PR #69, opened 2019-02-21
Script Dynatrace Request Attributes

alipatton10:
Added file applyRequestAttributeRules.sh which creates required request attributes in dynatrace. Currently needs to be manually executed but could be called from setupInfrastructure.sh


jetzlstorfer:
Thanks for the PR!
Could you please still make a distinction between
DT_TENANT_ID = id of the tenant, e.g., abc1234
DT_TENANT_URL = full URL including the ID and environemnt, e.g., abc1234.live.dynatrace.com
so we can use either variable on its own. in your PR they are set to the same value.
Thank you!

jetzlstorfer:
Hi @alipatton10 
we are closing this pull request at this moment since there are too many conflicts in the files and also, there are some values instead of placeholders in some of the yaml files.
Please update these files and open a new PR. Thanks!

PR #94, opened 2019-02-26
Adding instruction to create keptn namespace, if it doesn't already exist.

dirkwall:


dirkwall:
obsolete, wrong prerequisites on my end

PR #138, opened 2019-03-08
Cloud function for deleting obsolete cluster

agrimmer:


agrimmer:
Please check if we can use the prescribed license

PR #228, opened 2019-04-02
install pitometer service during keptn setup

bacherfl:


install/scripts/deployDynatrace.sh
```diff
@@ -33,3 +33,6 @@ echo "End applying auto tagging rules in Dynatrace "
 echo "--------------------------"
 
 ./createServiceEntry.sh $DT_TENANT_ID $DT_PAAS_TOKEN
+
```
AloisReitbauer: Does this support managed as well?

bacherfl: currently not, we need to adapt a few scripts to support managed

jetzlstorfer: right. I would suggest we still merge this PR and create a new one for managed support.

PR #232, opened 2019-04-03
install istio 1.1

bacherfl:


scripts/setupIstio.sh
```diff
@@ -3,8 +3,11 @@
 DT_TENANT_ID=$1
 DT_PAAS_TOKEN=$2
 
-kubectl apply -f ../manifests/istio/istio-crds.yml
-kubectl apply -f ../manifests/istio/istio-demo.yml
+for i in ../manifests/istio/crd*yaml; do kubectl apply -f $i; done
```
jetzlstorfer: I would prefer dedicated statements for the files to have better control on the installation

Issue #250, opened 2019-04-11
Smooth installation of keptn


dirkwall:
Installation of keptn follows the keptn walkthrough

johannes-b:
- [ ] consider managed environments (see: #255 )

robertjahn:
Please simply the steps needed for "keptn auth".  If it can automatically get the keptn token and endpoint this saves a step for the user to not look them up and pass them in.

From https://keptn.sh/docs/0.2.0/reference/cli/ 
![image](https://user-images.githubusercontent.com/24907089/57003364-2e70ff80-6b94-11e9-99c4-0c61155840d5.png)

PR #261, opened 2019-04-15
Clean-up install directory

johannes-b:


AloisReitbauer:
We still have the direct configuration dependency on GKE. Should this be put in a uniform?

johannes-b:
The configuration dependency on GKE is still there, but will become less dominant as we continue working on #250 (i.e., the installer will handle the connection to the cluster, which can be GKE, EKS, ...). 

But you brought up a question that needs clarification. From my understanding, the uniform defines the services that need to be installed and to which channel they subscribe. This comes after the installation of keptn:
1. `keptn install`
2. `keptn wear uniform` 

Thus, I don't see the cluster definition as part of the uniform. Or what do think?

dirkwall:
I agree with Johannes that uniforms define the services and their subscriptions to keptn channels, not the platform on which keptn is running. The platform needs to be addressed during the installation.

Issue #287, opened 2019-04-25
Implement error handling for all calls in all services


dirkwall:
- in some calls to external services there is no error handling implemented; we should fix that

PR #293, opened 2019-04-29
Error handling and improved logging

bacherfl:
Related to #287

### DIFF
--- a/core/control/src/svc/CredentialsService.ts
+++ b/core/control/src/svc/CredentialsService.ts
@@ -38,33 +38,11 @@ export class CredentialsService {
         }
       }
     } catch (e) {
+      Logger.error('', `Error retrieving API Token: ${e}`);
       token = '';
     }
 
     return token;
-  }
-
-  async getGithubCredentials(): Promise<KeptnGithubCredentials> {
-    const gitHubCredentials: KeptnGithubCredentials = {
-      user: '',
-      org: '',
-      token: '',
-    };
-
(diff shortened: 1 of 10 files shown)


core/control/src/svc/CredentialsService.ts
```diff
@@ -23,26 +22,23 @@ export class CredentialsService {
...
+
+      if (secret.body.items && secret.body.items.length > 0) {
+        const apiToken = secret.body.items.find(item => item.metadata.name === 'keptn-api-token');
+        if (apiToken && apiToken.data !== undefined) {
+          token = base64decode(apiToken.data['keptn-api-token']);
+        } else {
+          console.log('[keptn] The secret does not contain the proper information.');
+        }
       }
+    } catch (e) {
```
dirkwall: should we print log this error message some where, or is it clear which error happened? I'd make it more obvious and at least include the name of the error message, if not all details.

dirkwall: like in MessageService: Logger.error(e)

core/eventbroker/src/lib/log/Logger.ts
```diff
@@ -0,0 +1,43 @@
+export class Logger {
```
dirkwall: should we create a library that all core components include, instead of implementing the same class in different projects?

bacherfl: Absolutely, but I would suggest to create a separate story for that

Issue #292, opened 2019-04-26
The rollback of blue/green deployment following a pitometer 'fail' status fails to find the istio-virtual-service YAML file


robertjahn:
**Describe the bug**
The rollback of blue/green deployment following a pitometer 'fail' status fails to find the istio-virtual-service YAML file, thus stopping the "evaluation_done" pipeline

**To Reproduce**

Failure is because the YAML files for the keptn project are in "camelCase" and the jenkins job is looking for actual name.  In my scenario the service names is "front-end".   Project Yaml his ```istio-virtual-service-frontEnd.yaml``` but Jenkins job looking for ```istio-virtual-service-front-end.yaml```

1. run a build where the pitometer result is a fail.  Here is sample message.
```
{"keptnContext":"f5cddaf3-120b-4bd0-be4b-0e1b689ac704","logLevel":"INFO","message":"Sending event: {\"id\":\"e1fb220a-180a-46a3-af46-f0bbeb10e577\",\"specversion\":\"0.2\",\"time\":\"2019-04-26T12:58:57+00:00\",\"datacontenttype\":\"application/json\",\"shkeptncontext\":\"f5cddaf3-120b-4bd0-be4b-0e1b689ac704\",\"data\":{\"githuborg\":\"jahn-keptn-orders\",\"project\":\"orders-project\",\"deploymentstrategy\":\"blue_green_service\",\"stage\":\"staging\",\"service\":\"front-end\",\"image\":\"10.23.246.81:5000/orders-registry/front-end\",\"tag\":\"1-3\",\"evaluationpassed\":false},\"type\":\"sh.keptn.events.evaluation-done\"}","keptnService":"pitometer-service"}
```

2. the "evaluation_done" jenkins job runs.  It then attempts to update the manifest file and can't find it, thus failing the job.

```+ cd orders-project
+ git checkout staging
Branch 'staging' set up to track remote branch 'staging' from 'origin'.
Switched to a new branch 'staging'
[Pipeline] sh
+ cd orders-project
+ sed -i 's#weight: 0#weight: swap_100#' helm-chart/templates/istio-virtual-service-front-end.yaml
sed: helm-chart/templates/istio-virtual-service-front-end.yaml: No such file or directory
```

**Expected behavior**
Job will find the YAML file for the service.

**Screenshots**
![image](https://user-images.githubusercontent.com/24907089/56816226-17878180-6811-11e9-800b-3276be2daf19.png)

[...]

PR #302, opened 2019-05-02
Bugfix/292

jetzlstorfer:
Issue #292 adressed in the way that the CLI checks for valid project and service names. A valid project/service name is defined by the kubernetes naming conventions in combination with valid helm chart naming conventions: only lowercase letters and - are allowed.


dirkwall:
Can we please start providing a comment on what we've changed and why we've changed something when issuing a PR? I can look it up in the code changes, but a short summary wouldn't hurt to set the stage for the review and make it more transparent for everyone.

jetzlstorfer:
I've now updated the comment :)

dirkwall:
The version bump and the buildCLI script have nothing to do with this bugfix. Please separate those changes.

PR #344, opened 2019-05-13
Patch/272/improve messages

agrimmer:
Improves the help message of command `keptn install`


agrimmer:
I added descriptions for the log levels.

However, I did not add a description for the log file because the persistent store of the log file is independent of the specified log level (i.e. the log file always contains full details). 
Furthermore, Cobra (the used CLI framework) shows the default option. Therefore, I removed this output.

Issue #222, opened 2019-04-01
Install fails when AG is deployed


alipatton10:
When you have an AG installed the install fails as ISTIO creation doesn't seem to like having ports included in the endpoints. You receive the following error
Error from server: error when applying patch:
{"metadata":{"annotations":{"kubectl.kubernetes.io/last-applied-configuration":"{\"apiVersion\":\"networking.istio.io/v1alpha3\",\"kind\":\"ServiceEntry\",\"metadata\":{\"annotations\":{},\"name\":\"dynatrace-oneagent-hostnames\",\"namespace\":\"default\"},\"spec\":{\"hosts\":[\"pgr59512.live.dynatrace.com\",\"activegate.us-central1-a.c.acl-gsi.internal:9999\",\"10.128.0.37:9999\",\"sg-eu-west-1-18-202-8-23-prod23-ireland.live.ruxit.com\",\"sg-eu-west-1-52-31-108-81-prod23-ireland.live.ruxit.com\",\"sg-eu-west-1-34-241-182-103-prod23-ireland.live.ruxit.com\",\"sg-eu-west-1-34-241-58-226-prod23-ireland.live.ruxit.com\",\"sg-eu-west-1-52-214-86-21-prod23-ireland.live.ruxit.com\"],\"location\":\"MESH_EXTERNAL\",\"ports\":[{\"name\":\"HTTPS\",\"number\":443,\"protocol\":\"HTTPS\"}],\"resolution\":\"DNS\"}}\n"}},"spec":{"hosts":["pgr59512.live.dynatrace.com","activegate.us-central1-a.c.acl-gsi.internal:9999","10.128.0.37:9999","sg-eu-west-1-18-202-8-23-prod23-ireland.live.ruxit.com","sg-eu-west-1-52-31-108-81-prod23-ireland.live.ruxit.com","sg-eu-west-1-34-241-182-103-prod23-ireland.live.ruxit.com","sg-eu-west-1-34-241-58-226-prod23-ireland.live.ruxit.com","sg-eu-west-1-52-214-86-21-prod23-ireland.live.ruxit.com"]}}
to:
&{0xc420105080 0xc420180af0 default dynatrace-oneagent-hostnames ../manifests/gen/service_entries_oneagent.yml 0xc420d0c000 176401 false}
for: "../manifests/gen/service_entries_oneagent.yml": admission webhook "pilot.validation.istio.io" denied the request: configuration is invalid: 4 errors occurred:

[...]

PR #372, opened 2019-05-21
#222 Removed service entries for Dynatrace tenant

johannes-b:
Since Istio is configured to allow all outgoing traffic, it is not necessary to create service entries. Creating the service entries caused this issue.


johannes-b:
Valid point. I removed the createServiceEntry.sh as well as template files that were used by this script. 

(In addition, I removed the createCluster.sh file which is also obsolete and never used.)

Issue #326, opened 2019-05-07
As a keptn user I would like to use the CLI to send a new-artifact event


johannes-b:
**Is your feature request related to a problem? Please describe.**
This feature request simulates the web-hook to a container registry.  

**Describe the solution you'd like**
I would like to use the keptn cli similar to: `keptn send event new-artifact repo/image:tag` to send a new-artifact event to keptn, which triggers the delivery pipeline.

agrimmer:
**New CLI commands:**

For sending a new artifact event, the user can specify an image, e.g.
```
keptn send new-artifact --image=docker.io/keptn/keptn-authenticator:7e48486 --project=sockshop --service=carts
keptn send new-artifact --image=quay.io/keptn/keptn-authenticator:7e48486 --project=sockshop --service=carts
```



For sending arbitrary CloudEvents, the user should be able to pass a cloud event as json file, e.g.
`keptn send --cloudevent=ce.json`

agrimmer:
This issue can be closed after merging 
- https://github.com/keptn/github-service/pull/25
- https://github.com/keptn/keptn/pull/393
- https://github.com/keptn/keptn.github.io/pull/147

PR #375, opened 2019-05-23
#326 Send new artifact functionality

agrimmer:
- CLI: Implement `send` and `send new-artifact` commands
- Control: Implement endpoint `event` to receive external events and forward them to the eventbroker
- Fix Dockerfiles for kaniko
- Update releasenotes


core/control/src/ext-event/ExtEventController.ts
```diff
@@ -0,0 +1,76 @@
...
+        model: 'ExtEventRequestModel',
+        required: true,
+      },
+    },
+    responses: {
+      200: {
+      },
+      400: { description: 'Parameters fail' },
+    },
+    summary: '',
```
dirkwall: Please provide a meaningful summary and explanation of the error codes. We should also do that for all other swagger-ized components.

agrimmer: Thank you for pointing out. I added a meaningful summary.
I agree that we have to work on the error codes and their description. Now, we use in all endpoints the same description (i.e. _Parameters fail_) as it is used in the example of the library (i.e. https://github.com/olivierlsc/swagger-express-ts).

Issue #283, opened 2019-04-24
As a keptn user I would like to use something other than xip.io


akirasoft:
**Is your feature request related to a problem? Please describe.**
Specifically in the github-service as well as the installation scripts depend on xip.io heavily. Assumptions are made that the istio ingressgateway's loadbalancer will be an IP and hot a hostname. This results in the github service being unable to talk to the keptn event broker as well as creating invalid helm charts.

**Describe the solution you'd like**
Keptn should support IaaS providers that give hostnames for loadbalancers (like AWS) and make the vhost domain globally configurable. 

**Additional context**
Issue is also created via the github service:
https://github.com/keptn/github-service/issues/3

PR #379, opened 2019-05-24
#283 XIP replacement

bacherfl:
Instead of defaulting to using the ingressgateway's IP address and appending 'xip.io', we now check if the ingressgateway has a hostname assigned. If available, we will use that one.
The domain name of the ingressgateway is stored in a configmap `keptn-domain` in the keptn namespace. This configmap is used by the `github-service` (for creating the VirtualServices) and by the Jenkins pipelines (for running the tests). By using this config map, it will be possible to use any arbitrary domain name.


johannes-b:
@bacherfl, may I ask you to split this PR since the files you changed in the install folder won't live there anymore. 
In other words, the change you made in cli/cmd/install.go still goes into this repository, but the changes made in:
- install/manifests/keptn/keptn-domain-configmap.yaml
- install/manifests/knative/config-domain.yaml
- install/scripts/setupKeptn.sh 
- install/scripts/setupKnative.sh  

need to go the the keptn/installer repo.

PR #397, opened 2019-05-27
#326 Make docker.io prefix optional

agrimmer:
Check Docker image even when `docker.io/`-prefix is not specified.


cli/cmd/send_event_newArtifact_test.go
```diff
@@ -38,3 +38,58 @@ func TestNewArtifact(t *testing.T) {
...
+}
+
+func TestCheckImageAvailability(t *testing.T) {
+
+	validImgs := []DockerImg{DockerImg{"docker.io/keptnexamples/carts", "0.7.0"},
+		DockerImg{"docker.io/keptnexamples/carts:0.7.0", ""},
+		DockerImg{"keptnexamples/carts", "0.7.0"},
+		DockerImg{"keptnexamples/carts:0.7.0", ""},
+		DockerImg{"10.10.10.10:10/keptnexamples/carts", "0.7.5"},
+		DockerImg{"10.10.10.10:10/keptnexamples/carts:0.7.5", ""}}
```
dirkwall: Please add one more test for the docker image `httpd`, which is valid. No organization and no tag.

dirkwall: And also maybe one more where no tag is provided for keptnexamples/carts.

agrimmer: Done 👍

PR #700, opened 2019-08-22
#681 specification for cloud events for kernel

jetzlstorfer:


specification/cloudevents.md
```diff
@@ -288,6 +372,33 @@ The *start evaluation* event is sent when a test for a service is completed and
...
+  "shkeptnphaseid":"k5dg-565h-9o87",
+  "shkeptnphase":"RELEASE",
+  "shkeptnstepid":"a3ce-grdv-qwed",
+  "shkeptnstep":"TEST",
+  "data": {
+    "stage": "production",
+    "project": "sockshop",
+    "service": "carts",
+    "teststrategy":"real-user",
+    "startedat": "timestamp",
```
bacherfl: let's provide an example value here as well, to indicate that it should be in the same format as the `time` property of the cloud event, e.g. `2019-06-07T07:02:15.64489Z`

jetzlstorfer: thanks for the hint!

PR #935, opened 2019-10-07
ignore jetbrains .idea files

christian-kreuzberger-dtx:


christian-kreuzberger-dtx:
This PR can not be merged, as travis seems to be configured such that this does not cause and jobs to be created.
Screenshot from 2019-10-09 07-37-18

christian-kreuzberger-dtx:
Closed in favour of #945

PR #965, opened 2019-10-11
#5 Implementation of PKS support in Keptn CLI

johannes-b:


christian-kreuzberger-dtx:
Looks good to me. I assume this has been tested?

johannes-b:
Yes, I tested the install option on an PKS cluster (which is hosted on GCP).

Issue #971, opened 2019-10-15
keptn uninstall should hint that not everything is uninstalled


christian-kreuzberger-dtx:
**Describe the bug**
When calling `keptn uninstall`, not everything is cleaned up in the cluster.

According to the docs, istio, tiller and other services might still be there, see **https://keptn.sh/docs/0.5.0/reference/cli/#keptn-uninstall**.

While I would not expect the dynatrace service or any onboarded service to be removed by `keptn uninstall` (though this could also be questioned...), I do expect everything to be uninstalled that has been previously installed using `keptn install`, which includes istio and tiller.

This issue becomes especially apparent when the installation of `keptn install` fails.

**To Reproduce**
1. `keptn install [--platform=...]`
1. `keptn uninstall`
1. Verify the output of `kubectl get pods --all-namespaces`, e.g.:
```
NAMESPACE      NAME                                      READY   STATUS      RESTARTS   AGE
istio-system   istio-citadel-d5dc779f8-rngl8             1/1     Running     0          21m
istio-system   istio-cleanup-secrets-1.2.5-2gg87         0/1     Completed   0          21m
istio-system   istio-ingressgateway-65f756dddd-vkpbs     0/1     Running     0          21m
istio-system   istio-init-crd-10-hddrz                   0/1     Completed   0          22m
istio-system   istio-init-crd-11-9fz8w                   0/1     Completed   0          22m
istio-system   istio-init-crd-12-985mf                   0/1     Completed   0          22m
istio-system   istio-pilot-6cb78cd947-zrnxc              0/2     Pending     0          21m
istio-system   istio-security-post-install-1.2.5-sd7xc   0/1     Completed   0          21m
istio-system   istio-sidecar-injector-857549df44-29t57   1/1     Running     0          21m
kube-system    coredns-5644d7b6d9-jhbcp                  1/1     Running     0          34m
kube-system    coredns-5644d7b6d9-pj76f                  1/1     Running     0          34m
kube-system    etcd-minikube                             1/1     Running     0          33m
kube-system    kube-addon-manager-minikube               1/1     Running     0          33m
kube-system    kube-apiserver-minikube                   1/1     Running     0          33m
kube-system    kube-controller-manager-minikube          1/1     Running     0          33m
kube-system    kube-proxy-255rt                          1/1     Running     0          34m
kube-system    kube-scheduler-minikube                   1/1     Running     0          33m
kube-system    storage-provisioner                       1/1     Running     0          34m
```

**Expected behavior**
[...]

PR #972, opened 2019-10-15
#971 keptn uninstall should provide a hint about services/namespaces …

christian-kreuzberger-dtx:
This is a potential fix for #971 

Feel free to discuss the two proposed solutions in issue #971.

### DIFF
--- a/cli/cmd/uninstall.go
+++ b/cli/cmd/uninstall.go
@@ -61,9 +61,52 @@ var uninstallCmd = &cobra.Command{
 			}
 		}
 		logging.PrintLog("Successfully uninstalled keptn", logging.InfoLevel)
+		logging.PrintLog("Note: Please review the following namespaces and perform manual deletion if necessary:",
+			logging.InfoLevel)
+
+		namespaces, err := listAllNamespaces()
+
+		for _, namespace := range namespaces {
+			logging.PrintLog(" - " + namespace, logging.InfoLevel)
+			if namespace == "default" || namespace == "kube-public" {
+				// skip
+				logging.PrintLog("      Recommended Action: None (default namespace)", logging.InfoLevel)
+			} else if namespace == "kube-system" {
+				// we need to remove helm / tiller stuff
+				logging.PrintLog("      Recommended Action: Remove tiller/helm using", logging.InfoLevel)
+				logging.PrintLog("                          kubectl delete all -l app=helm -n kube-system", logging.InfoLevel)
+			} else if namespace == "istio-system" {
+				// istio is special, we will refer to the official uninstall docs
+				logging.PrintLog("      WARNING: Please consult the Istio Docs at https://istio.io/docs/setup/install/helm/#uninstall on how to remove istio completely from your cluster.", logging.InfoLevel)
+				logging.PrintLog("      Recommended action: kubectl delete namespace istio-system", logging.InfoLevel)
+			} else {
+				// just delete the namespace
+				logging.PrintLog(fmt.Sprintf("      WARNING: Please review this namespace in detail using 'kubectl get pods -n %s' before deleting it", namespace), logging.InfoLevel)
+				logging.PrintLog(fmt.Sprintf("      Recommended Action: kubectl delete namespace %s", namespace), logging.InfoLevel)
+			}
+		}
 
 		return nil
 	},
+}
+
+// returns a list of all namespaces
+func listAllNamespaces() ([]string, error) {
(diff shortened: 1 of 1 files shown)


christian-kreuzberger-dtx:
Tests currently fail. PR #975 will fix this.

cli/cmd/uninstall.go
```diff
@@ -61,6 +61,10 @@ var uninstallCmd = &cobra.Command{
 			}
 		}
 		logging.PrintLog("Successfully uninstalled keptn", logging.InfoLevel)
+		logging.PrintLog("Note: Some namespaces and services have not been uninstalled. Please inspect this yourself using", logging.InfoLevel)
```
agrimmer: This message sounds as if this was not intended.
Maybe this would be better: "Note that Keptn does not uninstall Istio, Tiller, and if installed Dynatrace."

cli/cmd/uninstall.go
```diff
@@ -61,6 +61,10 @@ var uninstallCmd = &cobra.Command{
...
+		logging.PrintLog("      kubectl get pods --all-namespaces", logging.InfoLevel)
```
agrimmer: Be more specific?
"You can delete those resources by 
`kubectl delete namespace istio-system`
`kubectl delete all -l app=helm -n kube-system`
`kubectl delete namespace dynatrace`

cli/cmd/uninstall.go
christian-kreuzberger-dtx: I wanted to be less specific about the services/namespaces that have not been uninstalled, as they might change in the future.

In that case I would go for `"Note that Keptn does not uninstall all services."`

cli/cmd/uninstall.go
christian-kreuzberger-dtx: Again, being less specific here, as at some point it might not be istio, but maybe linkerd, and it might not be helm but kustomize.

cli/cmd/uninstall.go
agrimmer: Being less specific makes the life of an unexperienced k8s user hard.

cli/cmd/uninstall.go
agrimmer: Same argument as above.

cli/cmd/uninstall.go
johannes-b: Even though I have approved the PR, I also have a thought about the comments. 
I think we should tell the user what remains installed (Istio, Tiller), as we are telling the user which components get installed during the Keptn install process.
When it comes to Dynatrace, there we should check whether the namespace is available or not.

christian-kreuzberger-dtx: Okay, I will rewrite the code and add some if statements.

christian-kreuzberger-dtx: I've rewritten the code to list all installed namespaces and to provide a recommended action (+ a warning, where I find it appropriate).

Here is an example output:

```
Would you like to uninstall keptn from this cluster? (y/n)
y
Starting to uninstall keptn
Successfully uninstalled keptn
Note: Please review the following namespaces and perform manual deletion if necessary:
 - default
      Recommended Action: None (default namespace)
 - istio-system
      WARNING: Please consult the Istio Docs at https://istio.io/docs/setup/install/helm/#uninstall on how to remove istio completely from your cluster.
      Recommended action: kubectl delete namespace istio-system
 - kube-public
      Recommended Action: None (default namespace)
 - kube-system
      Recommended Action: Remove tiller/helm using
                          kubectl delete all -l app=helm -n kube-system
 - sockshop-dev
      WARNING: Please review this namespace in detail using 'kubectl get pods -n sockshop-dev' before deleting it
      Recommended Action: kubectl delete namespace sockshop-dev
 - sockshop-production
      WARNING: Please review this namespace in detail using 'kubectl get pods -n sockshop-production' before deleting it
      Recommended Action: kubectl delete namespace sockshop-production
 - sockshop-staging
      WARNING: Please review this namespace in detail using 'kubectl get pods -n sockshop-staging' before deleting it
      Recommended Action: kubectl delete namespace sockshop-staging

```

If a dynatrace namespace exists, it would also print the following message:
```bash

```

cli/cmd/uninstall.go
christian-kreuzberger-dtx: I've rewritten the code, see above discussion (I'll close this one)

cli/cmd/uninstall.go
agrimmer: Thank you! Very, very detailed.
I would have omitted the namespaces where no action is required.

PR #984, opened 2019-10-17
Improve docker builds

christian-kreuzberger-dtx:
I improved the docker builds for the following services

* helm-service
* api-service
* configuration-service

by

* Adding .dockerignore files which mainly exclude the ./vendor subdirectory
* Installing `go-dep` before copying the source code

This way the docker build should be much faster locally when re-building (e.g., when using skaffold for in-cluster-debugging).

It is also possible that this will decrease build-time for docker image within travis-ci.

We can add this improvement to the other services too at a later point.


christian-kreuzberger-dtx:
This PR will most likely be made obsolute by PR #981 - so lets not merge this right now.

christian-kreuzberger-dtx:
Closing this as #981 got merged :)

PR #987, opened 2019-10-17
Trigger Installer Build

christian-kreuzberger-dtx:


christian-kreuzberger-dtx:
This PR was meant to trigger the installer build (which uploaded a new image to dockerhub):
Screenshot from 2019-10-18 07-21-30

Please merge so we can check if the installer within the develop branch is also built.

PR #996, opened 2019-10-21
Allow one distributor to listen to multiple messages/events

christian-kreuzberger-dtx:


christian-kreuzberger-dtx:
Closed for now, as the event-system will be touched in the next release.

Issue #1011, opened 2019-10-24
Refactor mongodb-datastore to use go-utils develop version 


johannes-b:
Refactoring of mongodb-datastore to use latest go-utils version on develop.

[[constraint]]
  name = "github.com/keptn/go-utils"
  branch = "develop"

PR #1016, opened 2019-10-24
#1011 Use latest version of go-utils for mongodb-datastore

johannes-b:
This PR adapts the mongodb-datastore service to support the latest go-utils version. 

Mainly, the logging mechanism has changed. 

When creating a logger (`logger := keptnutils.NewLogger("", "", serviceName)`), the first two parameters are empty, because the mongodb-datastore does not receive a keptn context or event ID; this is as designed.

### DIFF
--- a/mongodb-datastore/Gopkg.lock
+++ b/mongodb-datastore/Gopkg.lock
@@ -2,12 +2,12 @@
 
 
 [[projects]]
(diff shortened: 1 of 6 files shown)


christian-kreuzberger-dtx:
lgtm

PR #1017, opened 2019-10-25
Improved logging in API and bugfix in error message handling

johannes-b:
This PR has three topics:
- refactors logging in API
- improves Keptn CLI UX
- bug fix in error message handling for `get event` command

### DIFF
--- a/cli/cmd/create_project.go
+++ b/cli/cmd/create_project.go
@@ -47,7 +47,7 @@ Example:
 
 		if len(args) != 1 {
 			cmd.SilenceUsage = false
-			return errors.New("Requires PROJECTNAME")
+			return errors.New(`required argument "projectname" not set`)
 		}
 
 		if !utils.ValidateK8sName(args[0]) {
--- a/cli/cmd/create_project.go
+++ b/cli/cmd/create_project.go
@@ -47,7 +47,7 @@ Example:
 
 		if len(args) != 1 {
 			cmd.SilenceUsage = false
-			return errors.New(`required argument PROJECTNAME not set`)
+			return errors.New("required argument PROJECTNAME not set")
 		}
 
 		if !utils.ValidateK8sName(args[0]) {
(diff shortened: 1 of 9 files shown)


cli/cmd/delete_project.go
```diff
@@ -30,7 +30,7 @@ Example:
 
 		if len(args) != 1 {
 			cmd.SilenceUsage = false
-			return errors.New("Requires PROJECTNAME")
+			return errors.New(`required argument PROJECTNAME not set`)
```
christian-kreuzberger-dtx: Is it neccessary to use backticks/backquotes here?

cli/cmd/create_project.go
```diff
@@ -47,7 +47,7 @@ Example:
 
 		if len(args) != 1 {
 			cmd.SilenceUsage = false
-			return errors.New("Requires PROJECTNAME")
+			return errors.New(`required argument PROJECTNAME not set`)
```
christian-kreuzberger-dtx: Same question about backticks/backquotes here.

cli/cmd/onboard_service.go
```diff
@@ -39,7 +39,7 @@ Example:
 	Args: func(cmd *cobra.Command, args []string) error {
 		if len(args) != 1 {
 			cmd.SilenceUsage = false
-			return errors.New("Requires SERVICENAME")
+			return errors.New(`required argument SERVICENAME not set`)
```
christian-kreuzberger-dtx: Backticks?

cli/cmd/delete_project.go
johannes-b: No, it is not necessary here, but it does not make a difference.

cli/cmd/onboard_service.go
johannes-b: see above

cli/cmd/create_project.go
johannes-b: see above

PR #1040, opened 2019-11-04
Speed up develop build by fetching the latest image

christian-kreuzberger-dtx:
This heavily improves build time within the develop Branch for the docker images by caching the "builder" stage of the Dockerfile (which is also pushed to dockerhub).

For instance:
```
Step 2/10 : ARG version=develop
 ---> Using cache
 ---> 95302a3fd03c

Step 3/10 : WORKDIR /go/src/github.com/keptn/keptn/api
 ---> Using cache
 ---> 3e317e78b9fd

Step 4/10 : ENV GO111MODULE=on
 ---> Using cache
 ---> 9e59005e2bdf

Step 5/10 : COPY go.mod go.sum ./
 ---> Using cache
 ---> 421dfa20c29f

Step 6/10 : RUN go mod download
 ---> Using cache
 ---> 96ab95068ab1
```

This would make sure that we don't always have to re-download all the dependencies.


christian-kreuzberger-dtx:
Dismissing this as a full build still takes 40 minutes...

Issue #1045, opened 2019-11-05
Helm Service no longer adds image and tag to cloudnative deployment-finished event. Breaks Dynatrace Service


grabnerandi:
**Environment**
keptn0.5

* [x ] Dynatrace Integration
* [x ] Helm

**Describe the bug**
The Helm Service no longer adds tag & image to the cloud native event. Until 0.4 it was sent in the Data object. Image & Tag was used by other services such as the Dynatrace or the Notification Service. Right now the absence of this data breaks a key use case in Dynatrace Service. The Deployment Push Event failes with deploymentVersion not set which is a manadatory field

Dynatrace Service Code that requires Data.Tag to be set
![image](https://user-images.githubusercontent.com/675886/68209379-8bca8b00-ffd3-11e9-962f-e488a0759c4c.png)

PR #1060, opened 2019-11-06
#1045 send image and tag in deployment-finished event

bacherfl:
- Send image and tag in deployment-finished event
- undo downscaling when canary version is deployed again


christian-kreuzberger-dtx:
**Note** This is from the 0.5.1 release

LGTM

Issue #1078, opened 2019-11-08
JMeter/Wait-service needs to send Test-finished event with start/end time rather than startedat


christian-kreuzberger-dtx:
Right now jmeter-service sends a test-finished event which is manually constructed using a string map.

Task: Change jmeter-service such that the test-finished event from keptn/go-utils is used

johannes-b:
Hi Christian, 
the same change is necessary for the wait-service I figured out today. Please see: https://github.com/keptn/keptn/blob/develop/wait-service/main.go#L192

Can you address this change?

christian-kreuzberger-dtx:
Will do.

christian-kreuzberger-dtx:
We have a similar problem in wait-service for deployment finished that we need to tackle later: https://github.com/keptn/keptn/blob/9de712adb3c5a27c521e01dbd756952585bed41c/wait-service/main.go#L34-L39

PR #1080, opened 2019-11-08
#1078 adapt jmeter service to use test-finished and evaluation-done events

christian-kreuzberger-dtx:
Jmeter service was constructing events on its own, rather than using the events from go-utils.

I've fixed that, and also removed the "fallback code" I had to introduce in lighthouse-service because of the original behaviour of jmeter service.

Also, I've added skaffold.yaml and Dockerfile stuff to jmeter service for easier debugging.

### DIFF
--- a/jmeter-service/main.go
+++ b/jmeter-service/main.go
@@ -18,8 +18,11 @@ import (
 	"github.com/google/uuid"
 	"github.com/kelseyhightower/envconfig"
 	keptnutils "github.com/keptn/go-utils/pkg/utils"
+	keptnevents "github.com/keptn/go-utils/pkg/events"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+const eventbroker = "EVENTBROKER"
 
 type envConfig struct {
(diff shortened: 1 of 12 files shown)


jmeter-service/main.go
```diff
@@ -196,59 +199,47 @@ func getGatewayFromConfigmap() (string, error) {
...
+	// fill in timestamps
+	testFinishedData.Start = startedAt.Format(time.RFC3339)
+	testFinishedData.End = time.Now().Format(time.RFC3339)
 
 	event := cloudevents.Event{
 		Context: cloudevents.EventContextV02{
 			ID:          uuid.New().String(),
 			Time:        &types.Timestamp{Time: time.Now()},
-			Type:        "sh.keptn.events.tests-finished",
+			Type:        "sh.keptn.event.tests.finished",
```
agrimmer: Do we have this type in go-utils? If so, can/should we use it?

christian-kreuzberger-dtx: Thanks, good catch. Fixed.

Issue #1057, opened 2019-11-06
Remove pitometer, install Lighthouse


christian-kreuzberger-dtx:
As Lighthouse is ready, we need to replace Pitometer with Lighthouse.

- [x] Installer
- [ ] Use Cases (e.g., Auto-remmediation?)
- [ ] Docs (keptn.sh)
- [x] README (Keptn in version x.y.z requires...)
- [x] Remove pitometer from git repo / archive it?

Also, should we auto-install dynatrace-sli-service and/or prometheus-sli-service for this?

christian-kreuzberger-dtx:
Answere within daily standup on Nov 8th: We do not install any sli-provider. By default lighthouse will just send the "evaluation-done" event.

PR #1084, opened 2019-11-08
#1057 remove pitometer, install lighthouse

christian-kreuzberger-dtx:
- Pitometer is now thrown over board
- Installer adapted to now use lighthouse
- Removed unused (according to @bacherfl ) files for events (sub-cm*.yml)


christian-kreuzberger-dtx:
Installer failed as I've had the wrong name for distributors for lighthouse.

I'll fix this.

PR #1093, opened 2019-11-11
Patch/manifests

bacherfl:


installer/manifests/keptn/uniform-services.yaml
```diff
@@ -92,3 +92,61 @@ spec:
...
+            cpu: "50m"
+          limits:
+            memory: "128Mi"
+            cpu: "500m"
+        env:
+        - name: CONFIGURATION_SERVICE
+          value: 'http://configuration-service.keptn.svc.cluster.local:8080'
+        - name: EVENTBROKER
+          value: 'http://event-broker.keptn.svc.cluster.local/keptn'
+        - name: THROUGHPUT_QUERY
```
christian-kreuzberger-dtx: I think the QUERY env variables are not needed, are they?

Issue #1085, opened 2019-11-08
Travis Builds sometimes fail when downloading go library from go.opencensus.io


christian-kreuzberger-dtx:
This happened twice in a row now:
```
go: contrib.go.opencensus.io/exporter/ocagent@v0.4.12: unrecognized import path "contrib.go.opencensus.io/exporter/ocagent" (https fetch: Get https://contrib.go.opencensus.io/exporter/ocagent?go-get=1: read tcp 172.17.0.2:52632->216.239.32.21:443: read: connection reset by peer)
```

christian-kreuzberger-dtx:
This issue is appearing almost every day, and blocks our build pipeline.

Today it happened with openshift-route-service:
```
go: contrib.go.opencensus.io/exporter/ocagent@v0.4.12: unrecognized import path "contrib.go.opencensus.io/exporter/ocagent" (https fetch: Get https://contrib.go.opencensus.io/exporter/ocagent?go-get=1: read tcp 172.17.0.2:56480->216.239.32.21:443: read: connection reset by peer)
```

the dependency tree looks as follows:
```
# contrib.go.opencensus.io/exporter/ocagent
keptn/platform-support/openshift-route-service
github.com/keptn/go-utils/pkg/utils
github.com/Azure/go-autorest/autorest
github.com/Azure/go-autorest/tracing
contrib.go.opencensus.io/exporter/ocagent
```

`go-autorest` is used in go-utils only at this line:

```go

import (
        ...

	// Initialize all known client auth plugins.
	_ "github.com/Azure/go-autorest/autorest"
	_ "k8s.io/client-go/plugin/pkg/client/auth"
	"k8s.io/client-go/rest"
	"k8s.io/client-go/tools/clientcmd"
)
```

christian-kreuzberger-dtx:
This problem seems to have been a temporary issue. Closing it for now.

PR #1094, opened 2019-11-12
#1085 remove unused/wrong dependencies

christian-kreuzberger-dtx:


christian-kreuzberger-dtx:
Discarded for now as the dependency problem didn't appear again.

Issue #1116, opened 2019-11-18
Bridge: Use 24h time format


christian-kreuzberger-dtx:
At the moment bridge uses 12 hour time format without AM/PM.

This is confusing, as 15:35 would be displayed as 03:35 (without AM/PM).

I propose that we change this to 24h time format (for now).

christian-kreuzberger-dtx:
Fixed with 0.6.0.beta

PR #1117, opened 2019-11-18
#1116 bridge moment 24h time format

christian-kreuzberger-dtx:
Bridge used the wrong time format (12 hour)


christian-kreuzberger-dtx:
Discarding this as duplicate of #1118

PR #1203, opened 2019-12-05
Improve release process automation

christian-kreuzberger-dtx:
I removed the travis build for master branch, instead we will now focus on building releases in the actual release branch.
Instead of building some-image:0.6.0.beta2.latest, we will now build some-image:0.6.0.beta2.

Also, for the [version](version) file, we no longer rely on that (for the release branch). Instead, we use the current branch name using the following commands (bash):

**Travis before_install**
```console
# get current git branch name
- GIT_BRANCH_NAME=$(git branch | grep \* | cut -d ' ' -f2)
- BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
- echo "TRAVIS_BRANCH=$TRAVIS_BRANCH, BRANCH=$BRANCH"
```

**actual build of release**
```console
      # overwrite version for release branches based on the branch name
      - export VERSION=${BRANCH#"release-"}
```

Example:

```console
BRANCH=release-0.6.0.beta2
VERSION=${BRANCH#"release-"}
echo $VERSION
```
Output: ``0.6.0.beta2``

This also means that for every commit on the release branch we will automatically overwrite the tag on dockerhub (similar to what we did with the `.latest` tag). Same is true for the CLI.


.travis.yml
```diff
@@ -387,11 +392,13 @@ jobs:
     if: branch =~ ^release.*$ AND NOT type = pull_request
     os: osx
     script:
+      # overwrite version for release branches based on the branch name
+      - export VERSION=${BRANCH#"release-"}
       - echo "Build keptn cli"
       - cd ./cli
       - go test ./...
       - TAG="${VERSION}-${DATE}"
```
johannes-b: This environment variable (TAG) is not needed anymore. Or?

Issue #1104, opened 2019-11-13
Feature toggle remediation with Unleash


jetzlstorfer:
Extend the remediation service to be able to control feature toggles from the Unleash feature toggle framework in reponse to problem notifications.

PR #1205, opened 2019-12-06
#1104 Adding feature toggling as remediation actions

jetzlstorfer:
Adding the ability to toggle feature flags in the Unleash feature toggle framework as a remediation action.

Please do not merge yet.

### DIFF
--- a/bridge/server/lib/services/DatastoreService.js
+++ b/bridge/server/lib/services/DatastoreService.js
@@ -20,7 +20,7 @@ class DatastoreService {
 
     switch (mappedEvent.type) {
       case 'sh.keptn.event.configuration.change': mappedEvent.eventTypeHeadline = 'Configuration change'; break;
-      case 'sh.keptn.event.problem.open': mappedEvent.eventTypeHeadline = 'Problem'; break;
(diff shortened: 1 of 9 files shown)


bridge/server/lib/services/DatastoreService.js
```diff
@@ -35,6 +35,26 @@ class DatastoreService {
...
     }
 
+    //mappedEvent.data.project = "test";
+
+    if (event.source === 'dynatrace') {
+      var dttags = event.data.Tags.split(', ');
+      console.log(dttags);
+      var i;
+      for (i = 0; i < dttags.length; i++) {
+        if (dttags[i].indexOf("service:") == 0) {
```
christian-kreuzberger-dtx: Does this whole blocks till work with the new tags we are using?
Screenshot from 2019-12-10 12-35-17

christian-kreuzberger-dtx: This is no longer needed as it is handled by dynatrace-service.

christian-kreuzberger-dtx:
I have reworked the code for remediation service and for bridge.

PR #1235, opened 2019-12-12
Set tags of images back to latest

christian-kreuzberger-dtx:
Fixed tags for installer for develop branch


installer/deploy/installer.yaml
```diff
@@ -0,0 +1,21 @@
...
+kind: Job
+metadata:
+  name: installer
+  namespace: default
+spec:
+  template:
+    spec:
+      containers:
+        - name: keptn-installer
+          image: keptn/installer:5228f98
```
agrimmer: Should the installer also point to the latest image?

installer/deploy/installer.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: batch/v1
```
agrimmer: I know each service has a deploy folder. But do we need this for the installer? The same files are located in installer/manifests/installer.

christian-kreuzberger-dtx: Actually those two files in installer/deploy/ should not have made it to this PR; that was a mistake. I'm removing them.

agrimmer: Ok, thanks!

Issue #1234, opened 2019-12-12
Mongo-Datastore: New Endpoint for fetching root events


christian-kreuzberger-dtx:
At the moment root events are fetched in bridge using the following logic:

```javascript
  async getRoots() {
    const deploymentRoots = await this.getDeploymentRoots();
    const problemRoots = await this.getProblemRoots();
    const evaluationRoots = await this.getEvaluationRoots();
    let combinedRoots = deploymentRoots.concat(problemRoots);
    combinedRoots = combinedRoots.concat(evaluationRoots);
    combinedRoots.sort((a, b) => (a.timestamp < b.timestamp ? 1 : -1));
    return combinedRoots;
  }
```

whereas each type of root (deployment root, problem root,e valuation root) has its own logic on how to find out if something is a root.

This should be simplified in an endpoint within the mongo-datastore that provides all root events.

As @grabnerandi pointed out in #1230 a root event can be any cloudevent, and not just the one we filtered for previously.

Therefore this endpoint should provide all root events (as in: the first event based on the datetime and id).

This endpoint should also be filterable by:

* project
* type

Feel free to re-use the existing `/events` endpoint by that (e.g., by adding a new query parameter), but a new endpoint for this is also fine.

PR #1248, opened 2019-12-17
#1234 fetch root events from mongodb datastore

ermin-muratovic:
I added a new query parameter to the `/event` endpoint, which enables that only root events are fetched from the mongodb datastore using the [Collection.Distinct](https://godoc.org/go.mongodb.org/mongo-driver/mongo#Collection.Distinct) method.

Unfortunately, the Distinct method does not support paging, but we are now able to load the root events with a single request. Previously there have been three requests aggregated, see detailed description in #1234.


mongodb-datastore/handlers/events.go
```diff
@@ -114,64 +114,102 @@ func GetEvents(params event.GetEventsParams) (*event.GetEventsOKBody, error) {
...
+		}
 
-	cur, err := collection.Find(ctx, searchOptions, sortOptions)
-	if err != nil {
-		logger.Error(fmt.Sprintf("error finding elements in events collection: %v", err))
-	}
+		for _, value := range values {
+			var outputEvent interface{}
+
+			sortOptions := options.FindOne().SetSort(bson.D{{"time", -1}})
```
christian-kreuzberger-dtx: This might be wrong. I think setting the sort to `time` and `-1` means "decreasing", so you get the latest event. We need the earliest event.

Issue #1210, opened 2019-12-09
Accessing api.keptn.DOMAIN/v1/events without query params results in an error


christian-kreuzberger-dtx:
When accessing `api.keptn.DOMAIN/v1/events` without any parameters the Api Server prints the following error:

```
2019/12/08 09:37:13 Serving  at http://[::]:8080
{"timestamp":"2019-12-08T12:42:24.931740505Z","logLevel":"INFO","message":"API received a keptn event"}
{"timestamp":"2019-12-08T12:47:20.08956808Z","logLevel":"INFO","message":"API received a keptn event"}
{"timestamp":"2019-12-08T12:54:58.893848056Z","logLevel":"INFO","message":"API received a keptn event"}
2019/12/09 11:05:51 http: panic serving 10.8.0.33:55984: runtime error: invalid memory address or nil pointer dereference
goroutine 136 [running]:
net/http.(*conn).serve.func1(0xc000380280)
    /usr/local/go/src/net/http/server.go:1769 +0x139
panic(0x15db520, 0x2792df0)
    /usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/keptn/keptn/api/handlers.GetEventHandlerFunc(0xc00016eb00, 0x0, 0x0, 0xc000752080, 0xc0007b4901, 0xc00030a040)
    /go/src/github.com/keptn/keptn/api/handlers/event.go:71 +0x3b
github.com/keptn/keptn/api/restapi/operations/event.GetEventHandlerFunc.Handle(0x1864f08, 0xc00016eb00, 0x0, 0x0, 0xc000752080, 0x0, 0x0)
    /go/src/github.com/keptn/keptn/api/restapi/operations/event/get_event.go:21 +0x4e
github.com/keptn/keptn/api/restapi/operations/event.(*GetEvent).ServeHTTP(0xc000379ce0, 0x1a3a000, 0xc000718620, 0xc00016eb00)
    /go/src/github.com/keptn/keptn/api/restapi/operations/event/get_event.go:69 +0x287
github.com/go-openapi/runtime/middleware.NewOperationExecutor.func1(0x1a3a000, 0xc000718620, 0xc00016ea00)
    /go/pkg/mod/github.com/go-openapi/runtime@v0.19.4/middleware/operation.go:28 +0x75
net/http.HandlerFunc.ServeHTTP(0xc0002e8230, 0x1a3a000, 0xc000718620, 0xc00016ea00)
[...]

PR #1273, opened 2020-01-10
#1210 Set query parameters for /event endpoint as required

johannes-b:
Set the query parameters `keptnContext` and `type` for api endpoint GET `v1/event` since this endpoint is designed to get an event of a particular type from a certain keptnContext.

### DIFF
--- a/api/handlers/event.go
+++ b/api/handlers/event.go
@@ -68,16 +68,17 @@ func PostEventHandlerFunc(params event.PostEventParams, principal *models.Princi
 // GetEventHandlerFunc returns an event specified by keptnContext and eventType
 func GetEventHandlerFunc(params event.GetEventParams, principal *models.Principal) middleware.Responder {
 
-	logger := keptnutils.NewLogger(*params.KeptnContext, "", "api")
+	keptnContext := uuid.New().String()
+	logger := keptnutils.NewLogger(keptnContext, "", "api")
 
 	eventHandler := datastore.NewEventHandler(getDatastoreURL())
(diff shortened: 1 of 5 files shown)


api/swagger.yaml
```diff
@@ -52,17 +52,17 @@ paths:
...
       tags:
         - Event
-      summary: Get the latest event matching the parameters
+      summary: Get the latest event matching the required query parameters
       parameters:
         - name: keptnContext
           in: query
           type: string
-          required: false
+          required: true
```
agrimmer: Are you sure that **always** both, the `keptnContext` as well as the `type` are required fields?
Would it be a valid use case to only have the latest event of e.g. a specific `keptnContext`?

api/handlers/event.go
```diff
@@ -68,16 +68,17 @@ func PostEventHandlerFunc(params event.PostEventParams, principal *models.Princi
 // GetEventHandlerFunc returns an event specified by keptnContext and eventType
 func GetEventHandlerFunc(params event.GetEventParams, principal *models.Principal) middleware.Responder {
 
-	logger := keptnutils.NewLogger(*params.KeptnContext, "", "api")
+	keptnContext := uuid.New().String()
```
agrimmer: Do we really need a new `keptnContext` only for the logging? I do not think so.

api/swagger.yaml
johannes-b: This is intentional.

api/handlers/event.go
johannes-b: Reverted the change to use the provided KeptnContext for logging.

Issue #1208, opened 2019-12-06
Installer: Check if istio is already installed before installing


christian-kreuzberger-dtx:
As a first step: Don't install Keptn, and fail before doing anything if istio is already installed (aka: Don't break an existing installation).

For the future: Work with existing istio installations.

johannes-b:
Feedback from Stephan about this issue on Slack: https://keptn.slack.com/archives/CFH319TPE/p1576855611002700

PR #1274, opened 2020-01-10
#1208 Check availability of Istio installation

agrimmer:


installer/scripts/common/setupIstio.sh
```diff
@@ -1,23 +1,39 @@
 #!/bin/bash
 source ./common/utils.sh
 
-kubectl create namespace istio-system
+kubectl get ns istio-system
+ISTIO_AVAILABLE=$?
+if [[ "$ISTIO_AVAILABLE" == 0 ]] && [[ "$ISTIO_INSTALL_OPTION" == "Reuse" ]]; then
+    # An istio-version is already installed
+    print_info "Istio installation is reused but its compatibility is not checked"
+elif [[ "$ISTIO_AVAILABLE" == 0 ]] && ([[ "$ISTIO_INSTALL_OPTION" == "StopIfInstalled" ]] || [[ "$ISTIO_INSTALL_OPTION" == "" ]]); then
```
christian-kreuzberger-dtx: Please also handle 
```bash
[ "$ISTIO_INSTALL_OPTION" == "ISTIO_INSTALL_OPTION_PLACEHOLDER" ]]
```
to remain compatibility with older CLIs

(not sure if it goes into this if statement or into the else branch)

installer/scripts/common/setupIstio.sh
```diff
@@ -1,23 +1,39 @@
 #!/bin/bash
 source ./common/utils.sh
 
-kubectl create namespace istio-system
+kubectl get ns istio-system
+ISTIO_AVAILABLE=$?
+if [[ "$ISTIO_AVAILABLE" == 0 ]] && [[ "$ISTIO_INSTALL_OPTION" == "Reuse" ]]; then
+    # An istio-version is already installed
+    print_info "Istio installation is reused but its compatibility is not checked"
```
christian-kreuzberger-dtx: Would it make sense to perform some basic checks here, same as what we do after installation, e.g.:
```bash
 wait_for_crds "adapters.config.istio.io,attributemanifests.config.istio.io,authorizationpolicies.rbac.istio.io,clusterrbacconfigs.rbac.istio.io,destinationrules.networking.istio.io,envoyfilters.networking.istio.io,gateways.networking.istio.io,handlers.config.istio.io,httpapispecbindings.config.istio.io,httpapispecs.config.istio.io,instances.config.istio.io,meshpolicies.authentication.istio.io,policies.authentication.istio.io,quotaspecbindings.config.istio.io,quotaspecs.config.istio.io,rbacconfigs.rbac.istio.io,rules.config.istio.io,serviceentries.networking.istio.io,servicerolebindings.rbac.istio.io,serviceroles.rbac.istio.io,sidecars.networking.istio.io,templates.config.istio.io,virtualservices.networking.istio.io"
```

Please also state in the output that this option is not supported

installer/scripts/common/setupIstio.sh
```diff
@@ -1,23 +1,39 @@
 #!/bin/bash
 source ./common/utils.sh
 
-kubectl create namespace istio-system
+kubectl get ns istio-system
```
christian-kreuzberger-dtx: Is istio **always**  installed into `istio-system`, or is there a way to determine if istio is installed into another namespace?

installer/scripts/common/setupIstio.sh
agrimmer: Thanks, good point! Addressed.

installer/scripts/common/setupIstio.sh
agrimmer: I do not add this check as we do not know, which CRDs are actually used.

installer/scripts/common/setupIstio.sh
agrimmer: Istio can be installed in any namespace. However, in this first version, we only check for the `istio-system` namespace.

Issue #1225, opened 2019-12-10
Helm Service: WaitForDeploymentInNamespace blocks sending new artifacts


christian-kreuzberger-dtx:
affects 0.6.0.beta2:

I have a service **frontend** that failed to deploy before (for unknown reasons).
```console
keptn send event new-artifact --project=myproject --service=frontend --image=some-image/frontend --tag=1
```

If I send a new artifact for another service **backend**, helm-service fails.

```console
keptn send event new-artifact --project=myproject --service=backend --image=some-image/backend --tag=1
```

```
{"timestamp":"2019-12-10T14:47:55.94880231Z","logLevel":"ERROR","message":"Error when waiting for deployments in namespace myproject-staging: Deployment \"frontend\" exceeded its progress deadline"}
```

This will block until we have fixed the **frontend** service.

The underlying issue is this (occurs two times in the code):
```go
	if err := keptnutils.WaitForDeploymentsInNamespace(getInClusterConfig(), namespace); err != nil {
		return "", fmt.Errorf("Error when waiting for deployments in namespace %s: %s", namespace, err.Error())
	}
```

This waits for all deployments to finish in the given namespace (e.g., **myproject-staging**), but if one or more of the deployments within that namespace have an error, the current new-artifact deployment will fail (although it might have succeeded). 

## Why is this a problem?

Imagine you have the following service layout:

```
frontend <--> backend <--> database
```

and you execute the commands in that order:
```
keptn send event new-artifact --project=myproject --service=frontend --image=some-image/frontend --tag=1
keptn send event new-artifact --project=myproject --service=backend --image=some-image/backend --tag=1
keptn send event new-artifact --project=myproject --service=database --image=some-image/database --tag=1
```

the backend service tries to connect to the database when starting up, but fails to do so, as the database does not exist yet.
Therefore you have send the new-artifact event for database, but once you do that, you end up running in the error described above.
You now have a state where you can never bring up any service again within your project.

PR #1330, opened 2020-01-28
#1225 Only wait for deployments contained in Helm release

agrimmer:
Use Helm release to obtain deployments and services


christian-kreuzberger-dtx:
Tested with Keptn 0.6.0 and the new helm-service, seems to work.

Issue #1197, opened 2019-12-05
Quality Gates-Standalone version not working on OpenShift


bacherfl:
Using nginx as ingress seems to cause problems with openshift.

PR #1353, opened 2020-02-05
#1197 fixed QG-standalone installation on Openshift

bacherfl:


christian-kreuzberger-dtx:
Looks good, I`d like to try the installer with and without the quality-gates option on GKE to ensure that it still works.

Issue #1289, opened 2020-01-16
lighthouse-service: Retrieve SLIs even if tests fail


christian-kreuzberger-dtx:
As a follow-up of #542 

Lighthouse should retrieve SLIs even if tests failed before (if an slo.yaml is defined and test strategy is defined).

This allows users to potentially find out why the tests have failed.

PR #1354, opened 2020-02-05
#1289 evaluate test runs regardless of test execution result

bacherfl:
However, if the test run has failed, the result of the evaluation will always be set to `fail`


lighthouse-service/event_handler/evaluate_sli_handler.go
```diff
@@ -537,6 +532,45 @@ func (eh *EvaluateSLIHandler) getPreviousEvaluations(e *keptnevents.InternalGetS
 	return evaluationDoneEvents, nil
 }
 
+func (eh *EvaluateSLIHandler) getPreviousTestExecutionResult(e *keptnevents.InternalGetSLIDoneEventData, keptnContext string) (*keptnevents.TestsFinishedEventData, error) {
+	queryString := fmt.Sprintf("http://mongodb-datastore.keptn-datastore:8080/event?type=%s&project=%s&stage=%s&service=%s&keptnContext=%s&pageSize=%d",
```
christian-kreuzberger-dtx: Is there no environment variable for the URL (`mongodb-datastore.keptn-datastore...`) available that we can access?

bacherfl: there is now

Issue #1152, opened 2019-11-21
Display provided labels from start-evaluation event in bridge


dirkwall:
In #1149, the start-evaluation event is extended with labels. In #1150, the provided labels are stored with each evaluation result in the data store.
The provided labels should also be visible in the bridge.

PR #1356, opened 2020-02-05
#1152 Display start-evaluation labels

ermin-muratovic:


christian-kreuzberger-dtx:
are those labels displayed for every event type or just for start evaluation?

bridge/client/app/_interceptors/http-mock-interceptor.ts
```diff
@@ -0,0 +1,43 @@
...
+const evaluationLabelsMockData = {
+  "testid": "12345",
+  "buildnr": "build17",
+  "runby": "JohnDoe"
+};
+
+@Injectable({
+  providedIn: 'root'
+})
+export class HttpMockInterceptor implements HttpInterceptor {
```
christian-kreuzberger-dtx: Should this really be in this PR?

ermin-muratovic:
> are those labels displayed for every event type or just for start evaluation?

The labels are displayed for every event that has a `labels` property in the payload as a key-value pair.

bridge/client/app/_interceptors/http-mock-interceptor.ts
ermin-muratovic: Of course not. I removed the MockInterceptor.

Issue #1360, opened 2020-02-06
MongoDB for mongodb-datastore loses data when being shut down


christian-kreuzberger-dtx:
Happened on internal-demo-cluster with Keptn 0.6.0 when draining a node.

How to reproduce:

1. Install Keptn 
2. Onboard a project and send new artifacts
3. Open Keptn's Bridge

4. Stop the mongodb pod
```console
kubectl get pods -n keptn-datastore
kubectl -n keptn-datastore delete pod mongodb-758f7975c6-7ffnc
```

5. Verify the pod has been restarted:
```console
kubectl get pods -n keptn-datastore
```

6. Look at Keptn's Bridge - all events are gone

christian-kreuzberger-dtx:
The issue is that the mountpath for the volume is wrong:

https://github.com/keptn/keptn/blob/e789627d0e7d254689040d4cc25c90387c235927/installer/manifests/logging/mongodb-k8s/deployment.yaml#L35-L37

It should be `/var/lib/mongodb/data`

christian-kreuzberger-dtx:
I discovered that there are currently two major issues with our mongodb deployment.

1. `/data/db` is the wrong mountpath (though this seems to be the recommended way to do it with Mongodb in Kubernetes)
2. the persistent volume is everything but persistent, the data is cleaned every time we restart the pod.

christian-kreuzberger-dtx:
I fixed this in PR #1362 by doing the following three things:

1. The volume needs to be set to `/var/lib/mongodb/data`
2. The volume needs to have permissions for the user group set by using
```yaml
      securityContext:
        fsGroup: 184
```
3. The size of the volume needs to be increased from 100 MB to at least 1 GB

PR #1362, opened 2020-02-07
#1360 Set proper permission, path and size for mongodb volume

christian-kreuzberger-dtx:
The used mongodb image [centos/mongodb-36-centos7](https://hub.docker.com/r/centos/mongodb-36-centos7) requires proper permissions to be set on the `/var/lib/mongodb/data` volume, else it won't start properly:

```
      securityContext:
        fsGroup: 184
```

Also the size of the volume was set to 100 MB, which is way too less (a quick test shows that an empty MongoDB volume takes 350 MB already).

I also adapted the installer yaml files.


christian-kreuzberger-dtx:
I just tested this change with GKE and it works fine :) I can now delete the mongodb pod, it is afterwards restarted and the data is still there.

@bacherfl if you have time, can you please test this with OpenShift?

christian-kreuzberger-dtx:
I just added the changes in the yaml files for OpenShift as well.

@bacherfl can you try it out on OpenShift (or MiniShift) using
```console
keptn install --keptn-installer-image=keptn/installer:feature.1360.20200210.0859
```

bacherfl:
Just tested the changes on OpenShift. The installer works and the data stays persisted also when the mongodb pod is restarted -> LGTM

PR #1375, opened 2020-02-12
Revert "#1225 Only wait for deployments contained in Helm release"

christian-kreuzberger-dtx:
Reverts keptn/keptn#1330

I found an issue with the updated helm-service which we need to investigate. I'm reverting the change for now.


christian-kreuzberger-dtx:
Never mind, this was caused by #1376

PR #1382, opened 2020-02-16
Check compatibility based on Kubernetes cluster version before installing

### LINKED ISSUE #1326: Check compatibility based on Kubernetes cluster version before installing
**Story**: As a user I want to install Keptn only on a Kubernetes Cluster that is compatible with Keptn.

## Technical Details
When running `keptn install` the command (or the installer) should check if the cluster version is compatible with the current Keptn version (e.g., Keptn 0.6.0 is compatible with Kubernetes 1.13, 1.14 and 1.15, but not 1.16 AFAIK).

We could implement this check by executing
```console
kubectl version
```
and parsing the response of the server version
```
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.11-gke.14", GitCommit:"56d89863d1033f9668ddd6e1c1aea81cd846ef88", GitTreeState:"clean", BuildDate:"2019-11-07T19:12:22Z", GoVersion:"go1.12.11b4", Compiler:"gc", Platform:"linux/amd64"}
```

basically we need to check for `Major == "1"` and minor version `>= '13'` and `<= '15' (it's a string, so not sure whether we can easily check that with the `>=`/`<=` operators)

If we detect an unsupported Kubernetes version, the installer should exit with an appropriate error message.

## Definition of Done

When executing `keptn install ...`, the installer should not start execution if the cluster version is incompatible.
(Note: Whether this is done by a check in the CLI or in the installer needs to be discussed)

SVilgelm:
Fix #1326

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ before_install:
 - VERSION="unknownversion"
 - DATE="$(date +'%Y%m%d.%H%M')"
 - GIT_SHA="$(git rev-parse --short HEAD)"
+- KUBE_CONSTRAINTS=">= 1.13, <= 1.15"
 
 # store all changed files from this commit in files.txt (careful - travis commit range might fail)
 - git diff --name-only $TRAVIS_COMMIT_RANGE > files.txt || echo ""
@@ -200,9 +201,9 @@ jobs:
 
   # all commits (including external pull requests) should run basic tests for the CLI (if anything has changed)
   - stage: CLI tests
-    os: osx # ToDo: cli tests fail for os: linux 
+    os: osx # ToDo: cli tests fail for os: linux
     script:
-    - | 
+    - |
       if [[ $CHANGED_FILES == *"${CLI_FOLDER}"*  ]]; then
         echo "Testing keptn CLI on osx"
         cd ./cli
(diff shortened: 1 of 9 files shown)


christian-kreuzberger-dtx:
Thanks for the PR, looks good. I've commented on your questions in #1382.

Can you please also add a unit test for `CheckKubeServerVersion`? Thanks!

christian-kreuzberger-dtx:
We have also identified an issue with the unit tests being executed (hence the tests executed for your PR are failing right now). Can you please merge our current master branch (rebase works too, if you want)?

SVilgelm:
Sure, I will definitely add the unit tests and will rebase it. Thank you

SVilgelm:
@christian-kreuzberger-dtx I'm done with unit test, please review. I also adde the UTs for the IsKubectlAvailable function

christian-kreuzberger-dtx:
I just tested this with my Google Kubernetes Engine cluster and it seems to fail as you are expecting an integer for the version, but google is providing something a little less integer-like :man_facepalming: 

```console
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2", GitCommit:"59603c6e503c87169aea6106f57b9f242f64df89", GitTreeState:"clean", BuildDate:"2020-01-18T23:30:10Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.12-gke.25", GitCommit:"654de8cac69f1fc5db6f2de0b88d6d027bc15828", GitTreeState:"clean", BuildDate:"2020-01-14T06:01:20Z", GoVersion:"go1.12.11b4", Compiler:"gc", Platform:"linux/amd64"}
```

SVilgelm:
thanks, let me figure out what to do with it

christian-kreuzberger-dtx:
If the `hashicorp/go-version` package supports it, we could just use Strings for major and minor versions. Else we could just take the integer part of it.

SVilgelm:
@christian-kreuzberger-dtx unfortunately it doesn't support `1.13+`, there should be anything after the plus sing, like `1.13+foo76`

christian-kreuzberger-dtx:
Thanks for testing @SVilgelm . I guess we can just rewrite the regex to allow the minor and major version to contain an integer followed by an optional string, right?

SVilgelm:
I think we can allow anything after the number, but we should remove the `+` if the version ends with `+` only

SVilgelm:
@christian-kreuzberger-dtx here it is: https://github.com/keptn/keptn/pull/1382/commits/e1dbe64de835feb0c5f84862041e2e62fc4add42

Issue #1395, opened 2020-02-21
Configuration-service: Enhance mutex per project


christian-kreuzberger-dtx:
Right now configuration-service uses a global mutex. 

For performance reasons we should change to a mutex per project (based on the project-name).

PR #1400, opened 2020-02-25
#1395 lock mutex on project level

bacherfl:


configuration-service/handlers/project.go
```diff
@@ -160,8 +156,6 @@ func PutProjectProjectNameHandlerFunc(params project.PutProjectProjectNameParams
 
 // DeleteProjectProjectNameHandlerFunc deletes a project
 func DeleteProjectProjectNameHandlerFunc(params project.DeleteProjectProjectNameParams) middleware.Responder {
-	common.Lock()
```
christian-kreuzberger-dtx: Please add the lock here again, as deleting should not work while a lock is active.

configuration-service/handlers/project.go
```diff
@@ -160,8 +156,6 @@ func PutProjectProjectNameHandlerFunc(params project.PutProjectProjectNameParams
...
-	defer common.UnLock()
 	logger := utils.NewLogger("", "", "configuration-service")
```
agrimmer: I am not sure whether this lock can be removed

Issue #1278, opened 2020-01-14
Show last deployed image and URL to service in Bridge


christian-kreuzberger-dtx:
Within Keptn's Bridge the user should see the last successful deployed image for a service, as well as the URL to the service.

christian-kreuzberger-dtx:
This feature requires a proper user-story. Let's define what a user wants to see for each service in Bridge.

ermin-muratovic:
Could display the `deploymentURIPublic` from the last successful deployment in the stage tile on the left side?  See #1402 and #1404.
But I don't know if that is always the correct URL? Maybe there is a better way to determine the current deployed image and URL?

![image](https://user-images.githubusercontent.com/6098219/75449527-6a6c2100-596d-11ea-8eb6-56045f5e182b.png)

PR #1449, opened 2020-03-05
#1278 Show latest deployment in stage overview

ermin-muratovic:
The list of stages now shows the image from the latest `sh.keptn.events.deployment-finished` on that stage and is linked if `deploymentURIPublic` is provided.

image

### DIFF
--- a/bridge/client/_barista.scss
+++ b/bridge/client/_barista.scss
@@ -1,0 +1,6 @@
+@import '~@dynatrace/barista-components/style/main';
+@import '~@dynatrace/barista-components/style/index';
+
+.dt-button.dt-button-nested .dt-loading-spinner-svg {
+  stroke: $gray-400 !important;
+}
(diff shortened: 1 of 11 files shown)


johannes-b:
I like the enhancement. 
But I see no need for the information "2 Services" when I see the two tags. 
image

Besides, what does a tag show when just an onboarding but no deployment happened?

ermin-muratovic:
> I like the enhancement.
> But I see no need for the information "2 Services" when I see the two tags.
> image
> 
> Besides, what does a tag show when just an onboarding but no deployment happened?

If the service is onboarded, but now deployment happened yet, then there will be no tags, just the information "x Services". 

I think that the information "x Services" makes sense, although it is redundant, because it's like a title to the tags. But it could also be omitted, if you think that it's "to much".

christian-kreuzberger-dtx:
While I agree with @ermin-muratovic, I believe the information "x services" is not needed most of the time.
I vote for removing the "x services" information for now. Long term we should have an API endpoint that should return a list of services and whether an image has been deployed or not.

ermin-muratovic:
I removed the "x services" title and removed the link to _deploymentURIPublic_. Clicking on the service-tag now will bring you to the root event and scroll to the specific stage. There you can inspect that deployment and if you want you will find there also the link to _deploymentURIPublic_ with the _sh.keptn.events.deployment-finished_-Event

PR #1451, opened 2020-03-05
#1450 Better Support of Dynatrace Monitoring for Keptn core services

### LINKED ISSUE #1450: Monitor Keptn Installation with Dynatrace fails as our programs are statically linked with libc
Right now we can not monitor our go microservices of Keptn with Dynatrace. It shows us the following message:

Screenshot from 2020-03-05 10-26-17

This is happening as we statically link libc in our Dockerfiles:
https://github.com/keptn/keptn/blob/f4439d9e6f4a643584efe0ffd82b4e7a1346c1cc/gatekeeper-service/Dockerfile#L31

Removing the `CGO_ENABLED=0` would solve this problem, however with our multistage image this does not work.

christian-kreuzberger-dtx:
This fixes #1450 .

By switching to a dynamically linked library, Dynatrace is able to detect the Keptn core services written in Go and provide some more insights ;)

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -32,16 +32,12 @@ RUN sed -i "s/version: develop/${REPLACE}/g" /go/src/github.com/keptn/keptn/api/
 # Build the command inside the container.
 # (You may fetch or manage dependencies here, 
 # either manually or with a tool like "godep".)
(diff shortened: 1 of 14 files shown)


christian-kreuzberger-dtx:
I had to change some more things in order to be compatible with Dynatrace monitoring. 
In addition I tried debugging and found out that 128M is not enough for delve to run smoothly.

Please review again.

PR #1458, opened 2020-03-06
#1352 Show icons for certain events for better accessibility

### LINKED ISSUE #1352: Use icons for certain event types
We can use the barista icons (see https://barista.dynatrace.com/resources/icons/ )

1. If we highlight something with red or green, we should use an icon anyway to improve accessibility (just think about dyschromatopsia, in german "Rot-Grün Schwäche")

Examples where we should use icons in this case: failed tests, failed evaluations, problem events, failed stages. 

2. Map event types (e.g., root events) to icons.
Example:
* configuration-change
* problem open

All others can have a generic "log" or "info" icon

ermin-muratovic:
https://github.com/keptn/keptn/issues/1352#issuecomment-595781977


ermin-muratovic:
Latest changes indicate problem events more prominent, see Image below.

image

Issue #1316, opened 2020-01-22
Deep Links into Bridge


christian-kreuzberger-dtx:
Please provide the following deeplinks within the new bridge:

* project (already exists AFAIK)
* project/service (jumps to the project and opens the relevant service)
* keptncontextid (which should then automatically jump to the project/service of the event and open the traces based on the keptn context id)
* keptncontextid/trace (jump to the project/service, open the traces based on the keptn context id and scroll to specific trace)
* project/service/stage (same as project/service, but should automatically apply the filter for the stage which is still to be defined)

grabnerandi:
Thanks for putting this in. For the Quality Gate Use Case it would be interesting to generate a deep link into the bridge that shows the results of that particular quality gate run + the history of the previous runs for that same service on that stage. 
This would allow our users to integrate QGs with their CI/CD tools (Jenkins, GitLab, ...) and provide a direct link back to that quality gate visualization

christian-kreuzberger-dtx:
This feature requires that we expose Keptn's Bridge #1153 , which is currently not possible.

ermin-muratovic:
Implemented following deep links:

_'project/:projectName'_: opens project page of `projectName`
_'project/:projectName/:serviceName'_: opens project page of `projectName` and expands `serviceName`
_'project/:projectName/:serviceName/:contextId'_: opens project page of `projectName`, expands `serviceName` and selects root event with `contextId`
_'project/:projectName/:serviceName/:contextId/:eventId'_: opens project page of `projectName`, expands `serviceName`, selects root event with `contextId` and scrolls to event with `eventId`

![image](https://user-images.githubusercontent.com/6098219/76303644-5b1c9a00-62c2-11ea-91e3-9b97002cd49b.png)

grabnerandi:
That is awesome Ermin. Will this make it into 0.6.1? Or is this something we will see in a bridge update?

ermin-muratovic:
@grabnerandi we have already a feature freeze for 0.6.1, so it will be available in a new bridge update.

Added some new deep links for more convenience:

- `'trace/:shkeptncontext'`: will load that specific event and redirect to `'project/:projectName/:serviceName/:contextId'`
- `'trace/:shkeptncontext/:stage'`: will load that specific event and redirect to `'project/:projectName/:serviceName/:contextId/:eventId'` where _eventId_ is the id of the first event on the specific _stage_
[...]

PR #1463, opened 2020-03-09
#1316 Provide deep links to project/service/shkeptncontext

ermin-muratovic:
See #1316 for details


ermin-muratovic:
I used this deep link functionality also to integrate some smart interactions:
- clicking on a root element will scroll to the last event in that sequence
- clicking on a stage in a root element will scroll to the first event in that stage

robertjahn:
Big Thumbs up from me too.  Here is how I used it.

image

PR #1470, opened 2020-03-10
Fix bridge version not being applied in dockerfile

christian-kreuzberger-dtx:
The version of Keptn's Bridge needs to be available as an environment variable in the second stage.


christian-kreuzberger-dtx:
Works :)
Screenshot from 2020-03-10 16-32-01

PR #1472, opened 2020-03-10
#1465 Allow camel case service names if no Helm-chart is provided

### LINKED ISSUE #1465: Allow case-sensitive service names in "keptn create service"
In `keptn onboard service` and `keptn create service` we only allow service names, which only contain lower case letters. This is due to the fact that the k8s service name has to match Keptn service names. 
However, if we use quality gates only, we do not manage k8s service names and, hence, we should allow to use case-sensitive service names in `keptn create service`.

agrimmer:
Closes #1465


agrimmer:
@christian-kreuzberger-dtx May I also ask you for your feedback?

cli/cmd/create_service.go
```diff
@@ -40,9 +40,8 @@ Example:
 			return errors.New("required argument SERVICENAME not set")
 		}
 
-		if !keptnutils.ValidateKeptnEntityName(args[0]) {
-			errorMsg := "Service name contains upper case letter(s) or special character(s).\n"
-			return errors.New(errorMsg)
+		if !keptnutils.ValididateUnixDirectoryName(args[0]) {
+			return errors.New("Service name contains special character(s)")
```
christian-kreuzberger-dtx: Maybe specify that the service name contains unsupported characters and add some information (e.g., via a hyperlink?) about those.

helm-service/controller/onboarder.go
```diff
@@ -124,18 +124,18 @@ func (o *Onboarder) DoOnboard(ce cloudevents.Event, loggingDone chan bool) error
...
-	errorMsg := "Service name contains upper case letter(s) or special character(s).\n " +
-		"Keptn relies on the following conventions: " +
-		"start with a lower case letter, then lower case letters, numbers, and hyphens are allowed."
-
 	if event.HelmChart == "" {
 		// Case when only a service is created but not onboarded (i.e. no Helm chart is available)
-		if !keptnutils.ValidateKeptnEntityName(event.Service) {
-			return errors.New(errorMsg)
+		if !keptnutils.ValididateUnixDirectoryName(event.Service) {
+			return errors.New("Service name contains special character(s)")
```
christian-kreuzberger-dtx: Same as with the CLI, please add some more information and call it "unsupported characters"

cli/cmd/create_service.go
agrimmer: I improved the error message.

helm-service/controller/onboarder.go
agrimmer: I improved the error message.

PR #1548, opened 2020-03-24
Added basic authentication to Keptn's Bridge

christian-kreuzberger-dtx:
This provides an implementation and an example on how to enable Basic Authentication within Keptn's Bridge in the Readme file of Keptn's Bridge.

Using Kubernetes secrets and `optional: true`, I was able to add it to the Installer Manifest as well.

### DIFF
--- a/.idea/runConfigurations/keptn_bridge__Kube_ContDeploy_.xml
+++ b/.idea/runConfigurations/keptn_bridge__Kube_ContDeploy_.xml
@@ -1,0 +1,15 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="keptn/bridge [Kube ContDeploy]" type="google-container-tools-skaffold-run-config" factoryName="google-container-tools-skaffold-run-config-dev" show_console_on_std_err="false" show_console_on_std_out="false">
+    <option name="allowRunningInParallel" value="false" />
+    <option name="cleanupDeployments" value="true" />
+    <option name="envVariables" />
+    <option name="imageRepositoryOverride" />
+    <option name="kubeconfigFile" />
+    <option name="kubernetesContext" />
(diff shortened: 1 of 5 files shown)


.idea/runConfigurations/keptn_bridge__Kube_ContDeploy_.xml
```diff
@@ -0,0 +1,15 @@
+<component name="ProjectRunConfigurationManager">
```
agrimmer: Please add a .git ignore for .idea

bridge/README.md
```diff
@@ -14,6 +14,34 @@ To deploy the current version of the bridge in your Keptn Kubernetes cluster, us
...
+In the end, your container spec should look something like this:
+
+```yaml
+    spec:
+      containers:
+      - name: bridge
+        image: keptn/bridge2:0.6.1
+        imagePullPolicy: Always
+        env:
+          - name: BASIC_AUTH_USERNAME
```
agrimmer: I would propose to use k8s secrets.

.idea/runConfigurations/keptn_bridge__Kube_ContDeploy_.xml
christian-kreuzberger-dtx: This is a shared run configuration which was added on purpose, same as the other ones that are already in that folder: https://github.com/keptn/keptn/tree/master/.idea/runConfigurations

bridge/README.md
christian-kreuzberger-dtx: Agreed.

christian-kreuzberger-dtx: I'm using a secret now, and I've also added this to the installers manifest (with `optional: true`).

agrimmer: Cool, thanks

PR #1558, opened 2020-03-26
increased test coverage for lighthouse service

bacherfl:


christian-kreuzberger-dtx:
Please ignore the sonarcloud code duplication comment, but take a look at the code smells.

bacherfl:
> Please ignore the sonarcloud code duplication comment, but take a look at the code smells.

imho the code smell alerts are not necessarily applicable in that case - two of them address repeated strings that are part of unit test input. the other two concern the naming of the test functions, which are in line with the common guidelines regarding the naming of unit test functions.

Issue #1510, opened 2020-03-17
Refactor API-Service to only act as a proxy for other services


christian-kreuzberger-dtx:
At the moment our API-Service implements many things that are already implemented in the backend services such as the configuration-service.

In an ideal scenario we could switch to a more proxy-like approach, where the API-service only proxies certain endpoints of the backend services.

Important: Swagger-UI in API still needs to show the exposed endpoints, without having to duplicate code!

Please research this and create a prototype.

grabnerandi:
Important use case would be to retrieve the evaluation results that are currently shown in the heatmap in the bridge. If we provide a new API Endpoint that allows a user to retrieve the e.g: last X results for a service and stage then our partners / users can implement their own heatmap visualization in external tools such as Azure DevOps, Jenkins ...

agrimmer:
@grabnerandi This issue addresses no new end-points instead we would like to directly access the API from e.g. the `configuration-service`.

grabnerandi:
ok - @agrimmer -  shall I create a new issue then?

agrimmer:
Can you please formulate a mini KEP for this and make a PR? https://github.com/keptn/enhancement-proposals

agrimmer:
One possible solution would be to make the routing in the VirtualService and/or Ingress.
See https://github.com/keptn/keptn/blob/feature/1510/refactor-api-service/installer/manifests/keptn/keptn-api-virtualservice.yaml
for a possible implementation of the VS.

Pros:
- No new service for distributing the events
- Easy to configure and maintain

Drawbacks:
- No centralized authentication. Therefore, also other services like the `configuration-service` have to implement the authentication.

agrimmer:
Research Outcome:
- We introduced an NGINX as new k8s deployment and service. This NGINX allows to route the traffic and ensures that all requests are authenticated using the /auth endpoint of the `api-service`.
- The `api-service` now does not implement endpoints of the `configuration-service` anymore.
- The `configuration-service` is exposed to the public. Endpoints that are not intended to be used from the public (e.g. deleting a project) are marked and the description is accordingly adapted.

**Architecture for the full installation:**
![full](https://user-images.githubusercontent.com/10980861/78262547-03f49880-7501-11ea-9974-0f70bee617c4.png)

**Architecture for the quality gates installation:**
![qg](https://user-images.githubusercontent.com/10980861/78262560-09ea7980-7501-11ea-8720-edf8d3234766.png)

[...]

PR #1587, opened 2020-04-02
#1510 refactor api service

agrimmer:
Closes issue https://github.com/keptn/keptn/issues/1510 

This PR consists of:
- Remove configuration-service endpoints of api-service
- Add nginx for routing the traffic
- Refactoring of the installer


cli/cmd/configure_domain.go
```diff
@@ -35,7 +35,7 @@ var configureDomainParams *configureDomainCmdParams
 const installerPrefixURL = "https://raw.githubusercontent.com/keptn/keptn/"
 
 const apiVirtualServiceSuffix = "/installer/manifests/keptn/keptn-api-virtualservice.yaml"
-const apiIngressSuffix = "/installer/manifests/keptn/api-ingress.yaml"
+const apiIngressSuffix = "/installer/manifests/keptn/keptn-ingress.yaml"
```
christian-kreuzberger-dtx, line 38: You also removed `api-ingress`, but I guess `keptn-ingress` is correct as this will also be the ingress for Bridge?

cli/cmd/configure_domain.go
```diff
@@ -148,14 +148,6 @@ Example:
...
 			}
 
-			if err := keptnutils.RestartPodsWithSelector(false, "keptn", "run=api"); err != nil {
-				return err
-			}
-
-			if err := keptnutils.WaitForPodsWithSelector(false, "keptn", "run=api", 5, 5*time.Second); err != nil {
-				return err
-			}
-
```
christian-kreuzberger-dtx, line 158: Why do we not need to restart the pods anymore?

api/deploy/service.yaml
```diff
@@ -2,21 +2,21 @@
...
     metadata:
       labels:
-        run: api
+        run: api-service
     spec:
       containers:
-        - name: api
-          image: keptn/api:latest
+        - name: api-service
+          image: keptn/api-service:latest
```
christian-kreuzberger-dtx, line 19: Not sure if the image change was on purpose, maybe we don't even need to do that? If we however do, we should also update the build process in .travis.yml

api/skaffold.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: skaffold/v1beta13
 kind: Config
 build:
   artifacts:
-    - image: keptn/api
+    - image: keptn/api-service
```
christian-kreuzberger-dtx, line 5: Same here, let's make sure we actually want that image name

api/swagger-ui/index.html
```diff
@@ -49,12 +49,16 @@
         plugins: [
           SwaggerUIBundle.plugins.DownloadUrl
         ],
-        layout: "StandaloneLayout"
+        layout: "StandaloneLayout",
+        urls: [{
+          url: "/swagger-ui/swagger.yaml",
+          name: "api-service"
+        }, {url: "/configuration-service/swagger-ui/swagger.yaml", name: "configuration-service"}],
```
christian-kreuzberger-dtx, line 56: Not a major issue, but let's make this look cleaner by adding some newlines:
```javascript
        urls: [{
          url: "/swagger-ui/swagger.yaml",
          name: "api-service"
        }, {
          url: "/configuration-service/swagger-ui/swagger.yaml", 
          name: "configuration-service"
       }],
```

cli/cmd/configure_domain.go
```diff
@@ -274,7 +266,8 @@ func updateKeptnAPIIngress(path, domain string) error {
 	}
 
 	if err := file.Replace(keptnAPIIngress,
-		file.PlaceholderReplacement{PlaceholderValue: "domain.placeholder", DesiredValue: domain}); err != nil {
+		file.PlaceholderReplacement{PlaceholderValue: "domain.placeholder", DesiredValue: domain},
+		file.PlaceholderReplacement{PlaceholderValue: "ingress.placeholder", DesiredValue: "nginx"}); err != nil {
 		return err
```
christian-kreuzberger-dtx, line 271: Please add inline documentation about what the above 4 lines do (e.g., replace xyz in file foo.bar to provide something)

helm-service/deploy/service.yaml
```diff
@@ -31,7 +31,7 @@ spec:
         - name: EVENTBROKER
           value: 'http://event-broker.keptn.svc.cluster.local/keptn'
         - name: API
-          value: 'ws://api.keptn.svc.cluster.local'
+          value: 'ws://api-service.keptn.svc.cluster.local:8080'
```
christian-kreuzberger-dtx, line 34: I guess this change needs to be done for many other services as well :/

installer/manifests/keptn/core.yaml
```diff
@@ -1,32 +1,230 @@
...
+            proxy_pass         http://configuration-service:8080;
+            proxy_redirect     off;
+            proxy_set_header   Host $host;
+            proxy_http_version 1.1;
+            proxy_set_header X-Real-IP $remote_addr;
+            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+            proxy_set_header X-Forwarded-Proto $scheme;
+          }
+
+          location ~* /v1/project/[a-z0-9-]+/resource {
```
christian-kreuzberger-dtx, line 64: we should document that this is a legacy endpoint for v1, that will be removed in the future in favour of directly using configuration-service.

installer/manifests/keptn/core.yaml
```diff
@@ -1,32 +1,230 @@
...
+            - mountPath: /etc/nginx # mount api-nginx-config volume to /etc/nginx
+              readOnly: true
+              name: api-nginx-config
           resources:
             requests:
               memory: "64Mi"
               cpu: "50m"
             limits:
               memory: "128Mi"
               cpu: "500m"
```
christian-kreuzberger-dtx, line 176: we should double check the actual requirements for API gateway in terms of cpu and memory.

installer/manifests/keptn/keptn-ingress.yaml
```diff
@@ -0,0 +1,26 @@
...
+    kubernetes.io/ingress.class: ingress.placeholder
+  name: keptn-ingress
+  namespace: keptn
+spec:
+  tls:
+    - hosts:
+        - '*.keptn.domain.placeholder'
+      secretName: sslcerts
+  rules:
+    - host: '*.keptn.domain.placeholder'
```
christian-kreuzberger-dtx, line 15: Careful, this change is potentially also affecting the expose Keptn's Bridge task. I wonder if we can still write `host: 'api.keptn.domain.placeholder'` here?

installer/scripts/installKeptn.sh
```diff
@@ -29,33 +29,14 @@ kubectl apply -f ../manifests/keptn/namespace.yaml
...
-
 case $PLATFORM in
-  aks)    
-    echo "Installing Keptn on AKS"
-    ./common/install.sh
-    ;;
-  eks)
-    echo "Install Keptn on EKS"
+  aks|eks|gke|pks|kubernetes)
     ./common/install.sh
```
christian-kreuzberger-dtx, line 34: long overdue change :+1:

cli/cmd/configure_domain.go
agrimmer: Done

api/deploy/service.yaml
agrimmer: As discussed, let's stay with the image name `keptn/api`.

api/skaffold.yaml
agrimmer: As discussed, let's stay with the image name keptn/api.

api/swagger-ui/index.html
agrimmer: Done.

helm-service/deploy/service.yaml
agrimmer: Yes, in the dyantrace-service and prometheus-service.

installer/manifests/keptn/core.yaml
agrimmer: We reduced the limits.

installer/manifests/keptn/keptn-ingress.yaml
agrimmer: Agree, let's use `api.keptn` as host.

installer/manifests/keptn/core.yaml
agrimmer: Done.

cli/cmd/configure_domain.go
agrimmer: Done.

cli/cmd/configure_domain.go
agrimmer: This is not needed anymore.

installer/manifests/keptn/keptn-ingress.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: Ingress
+metadata:
+  annotations:
+    kubernetes.io/ingress.class: ingress.placeholder
+  name: keptn-ingress
+  namespace: keptn
+spec:
+  tls:
+    - hosts:
+        - '.keptn.domain.placeholder'
```
christian-kreuzberger-dtx, line 12: Just to be 100% sure on this:

This used to be
```yaml
spec:
  tls:
    - hosts:
        - domain.placeholder
      secretName: sslcerts
```

Now it starts with a `.` and it is limited to the `keptn` namespace.

agrimmer: Corrected. This was a c&p error.

PR #1588, opened 2020-04-02
Feature/1492/go utils refactoring

bacherfl:


api/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Use the offical Golang image to create a build artifact.
 # This is based on Debian and sets the GOPATH to /go.
-FROM golang:1.12.13-alpine as builder
+FROM golang:1.13-alpine as builder
```
christian-kreuzberger-dtx, line 3: Please be specific in the golang version, as Dynatrace oneagent does not support newer versions yet.
Screenshot from 2020-04-03 12-22-36
I think 1.13.7 should be fine.

helm-service/controller/configuration_changer.go
```diff
@@ -56,8 +56,13 @@ func (c *ConfigurationChanger) ChangeAndApplyConfiguration(ce cloudevents.Event,
 		return err
 	}
 
+	keptnHandler, err := keptnevents.NewKeptn(&ce, keptnevents.KeptnOpts{})
+	if err != nil {
+		c.logger.Error("Could not initialiue keptn handler: " + err.Error())
```
christian-kreuzberger-dtx, line 61: Typo *initialiue*

christian-kreuzberger-dtx:
Btw, when we merge this, we need to make sure that this is squash merged

api/Dockerfile
bacherfl: done

helm-service/controller/configuration_changer.go
bacherfl: done

Issue #1490, opened 2020-03-12
Refactor MongoDB Datastore to use one collection per project, write tests


christian-kreuzberger-dtx:
At the moment, MongoDB Datastore uses one collection for all projects. In terms of performance and to prepare for future features, it would be better to have one collection per project

We should also refactor the service and write tests for it.

PR #1609, opened 2020-04-07
#1490 store events in separate mongodb collection per project

bacherfl:


mongodb-datastore/handlers/events.go
```diff
@@ -94,11 +96,59 @@ func SaveEvent(event *models.KeptnContextExtendedCE) error {
...
+	// store mapping between context ID and project in separate table
+	/*
+		_, err = getProjectForContext(event.Shkeptncontext)
+		if err != nil {
+		}
+
+	*/
+	contextToProjectCollection := client.Database(mongoDBName).Collection(contextToProjectCollection)
+
+	_, err = contextToProjectCollection.InsertOne(ctx,
```
agrimmer, line 110: The `contextToProjectCollection` always contains one document for one `shkeptncontext`. 
If now a document with an `shkeptncontext` is already contained in the collection, this `InsertOne` will produce an error. 
Should we check if the collection `contextToProjectCollection` contains already the `shkeptncontext` and only insert it if it is not contained?

mongodb-datastore/swagger.yaml
```diff
@@ -102,62 +102,6 @@ paths:
           description: error
           schema:
             "$ref": "#/definitions/error"
-  /log:
```
agrimmer, line 105: I like it that we removed the /log endpoint as it is never used.

mongodb-datastore/handlers/events.go
bacherfl: I will add an additional check for that

Issue #1153, opened 2019-11-21
API: Expose Keptn Bridge via Istio Ingress Gateway OR nginx ingress depending on what is available


dirkwall:
It should be possible to expose the should on a public endpoint, e.g. `bridge.keptn.<domain>`, if the user wants to do that.
~This should be configurable in the installation process.~
This should be configurable at run time, e.g. using the CLI.

christian-kreuzberger-dtx:
`keptn expose bridge` should expose the bridge and print credentials (e.g., admin, password, URL)

grabnerandi:
Hi
I think its very important that we make it as easy as possible for users to access the bridge as the bridge is the "official face" and interface to Keptn. The current primary use case of Quality Gates also only makes sense if the users can access the bridge.
As I start I think it would be sufficient if we simply prompt the user for a secret that we auto-generate during installation. This is similar to what Argo or Jenkins are doing. We could either create a random key and print it out at the end of keptn install. When running on k8s this key could be the name of the pod that the bridge uses. When not running on k8s it can be a random generated string/key.
I suggest that our bridge prompts the user for this UI secret and if correct - sets a cookie that tells us that the user is authorized.
This would be a fast and easy way to prevent access to everyone in case the bridge is exposed to the public. As the project matures I agree we need to find more sophisticated options of authentication but at this time I think we are good with this approach

Andi

checkelmann:
Hi, 

we need at least a basic auth for the bridge within our Environment. A pre generated user/password would do the job. Or a command line parameter on keptn install. 

> Hi
> I think its very important that we make it as easy as possible for users to access the bridge as the bridge is the "official face" and interface to Keptn. The current primary use case of Quality Gates also only makes sense if the users can access the bridge.
> As I start I think it would be sufficient if we simply prompt the user for a secret that we auto-generate during installation. This is similar to what Argo or Jenkins are doing. We could either create a random key and print it out at the end of keptn install. When running on k8s this key could be the name of the pod that the bridge uses. When not running on k8s it can be a random generated string/key.
> I suggest that our bridge prompts the user for this UI secret and if correct - sets a cookie that tells us that the user is authorized.
> This would be a fast and easy way to prevent access to everyone in case the bridge is exposed to the public. As the project matures I agree we need to find more sophisticated options of authentication but at this time I think we are good with this approach
> 
> Andi

robertjahn:
[...]

PR #1620, opened 2020-04-08
#1153 Expose bridge

agrimmer:
Provide a new endpoint `configure/bridge/expose` in the API for exposing the bridge.


api/go.mod
```diff
@@ -16,7 +16,14 @@ require (
 	github.com/gorilla/websocket v1.4.1
 	github.com/jessevdk/go-flags v1.4.0
 	github.com/keptn/go-utils v0.6.1-compat
+	github.com/keptn/keptn/configuration-service v0.0.0-20200406090046-e6e7da3c4531
```
christian-kreuzberger-dtx, line 19: Was this added on purpose?

agrimmer: Thank you! No this was not intended.

api/handlers/configure.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	for _, ns := range nsList.Items {
+		if ns.Name == "istio-system" {
+			return istio, nil
+		} else if ns.Name == "ingress-nginx" {
+			return nginx, nil
+		}
+	}
+	return istio, errors.New("Cannot obtain type of ingress.")
```
christian-kreuzberger-dtx, line 64: Are you returning `istio` on purpose here? Or should this be nil?

christian-kreuzberger-dtx:
One more thing that just came up in my mind is whether this works with OpenShift or not.

api/handlers/configure.go
agrimmer: The type `Ingress` cannot be nil.

agrimmer:
Good point! I will implement `oc create route edge bridge --service=bridge --port=http --insecure-policy='None' -n keptn --hostname="bridge.keptn.$BASE_URL"` in the `openshift-route-service`.

agrimmer:
Now, this PR also supports exposing the Bridge using OC routes.

PR #1636, opened 2020-04-14
#1490 added db migration script

bacherfl:


upgrader/upgrade-061-070/mongodb-upgrade/main.go
```diff
@@ -0,0 +1,100 @@
...
+			fmt.Printf("Cannot migrate event because object does not have the expected structure.\n")
+			continue
+		}
+		project, ok := data["project"].(string)
+		if !ok || project == "" {
+			fmt.Printf("Cannot migrate event because object does not have the expected structure.\n")
+			continue
+		}
+
+		if doc["data"] == nil || project == "" || doc["shkeptncontext"].(string) == "" {
```
agrimmer, line 67: Is this check redundant?

upgrader/upgrade-061-070/mongodb-upgrade/main.go
```diff
@@ -0,0 +1,100 @@
...
+	"os"
+)
+
+const mongoDBURL = "mongodb://user:password@mongodb:27017/keptn"
+
+var client *mongo.Client
+
+var projectCollections map[string]*mongo.Collection
+
+func main() {
```
agrimmer, line 18: Can we please introduce arguments for passing the `mongoDBURL`? If provided use it, if not use the constant.
Reason: One customer used an external mongodb and potentially this upgrade script would also work for them.

Issue #1622, opened 2020-04-08
API: Provide metadata (git upstream URI, creation date) for each project  


christian-kreuzberger-dtx:
From [KEP 10](https://github.com/keptn/enhancement-proposals/pull/10).

After implementing #1621:

Provide metadata like git upstream URI, creation date, and whatever else we have available for the project.

Issue #1621, opened 2020-04-08
API: Query all stages and services for each project based on materialized views 


christian-kreuzberger-dtx:
From [KEP 10](https://github.com/keptn/enhancement-proposals/pull/10).

We need to implement an efficient way of providing all stages and services of a project, e.g., via a materialized view. There will also be other metadata available (this is imlemented in issue #1622)

Right now this is implemented in the configuration-service and it uses git checkout etc... to determine all services and stages, which is very inefficient (and error prone).

PR #1651, opened 2020-04-16
#1621 #1622 retrieve all projects/stages/services via materialized view and provide further information

bacherfl:


configuration-service/common/projects_materialized_view.go
```diff
@@ -0,0 +1,208 @@
...
+		if stg.StageName == stage {
+			stageIndex = idx
+			break
+		}
+	}
+	if stageIndex < 0 {
+		return nil
+	}
+
+	copy(prj.Stages[stageIndex:], prj.Stages[stageIndex+1:])
```
agrimmer, line 138: A short version for deleting an element of a slice would be
`prj.Stages = append(proj.Stages[:stageIndex], proj.Stages[stageIndex+1:]...)`

configuration-service/handlers/project.go
```diff
@@ -177,6 +178,9 @@ func DeleteProjectProjectNameHandlerFunc(params project.DeleteProjectProjectName
 		}
 	}
 
+	mv := common.GetProjectsMaterializedView()
+	mv.DeleteProject(params.ProjectName)
```
agrimmer, line 182: How can errors be handled? 
If the error is not handled, the materialized view and the state in the Git repo can potentially diverge.

Issue #1660, opened 2020-04-21
Upgrader support for API enhancements (API proxy)


johannes-b:
The upgrader from Keptn 0.6.1 to 0.6.2 has to upgrade the API based on the changes implemented in #1510 . 

* The existing api deployment has been renamed to api-service
* Virtual services and OpenShift routes have been changed
* Ingress gateway changes see below

# Definition of Done
- [ ] Upgrade servies
  - [ ] Update `api` to `api-service`
  - [ ] Apply `api-gateway-nginx` and `api-nginx-config`
  - [ ] Upgrade WS-URL of `shipyard-service`, `helm-service`
- [ ] Upgrade a full Keptn 0.6.1 installation to Keptn 0.6.2 (0.6.1 as well as 0.6.2 uses Istio for the Gateway)
  - [ ] Upgrade keptn-api-virtualservice
- [ ] Upgrade a quality-gates Keptn 0.6.1 installation to Keptn 0.6.2
  - [ ] Nginx Ingress based (GKE, EKS, AKS, Kubernetes): Apply `keptn-ingress` and delete `api-ingress`
  - [ ] Routes based (OpenShift): Correct routes
- [ ] Upgrade guide in Keptn docs (similar to https://keptn.sh/docs/0.6.0/installation/upgrade-keptn/#upgrade-from-0-6-0-to-0-6-1 )

PR #1717, opened 2020-04-29
#1660 upgrade Job for 0.6.1 -> 0.6.2

bacherfl:
The upgrader job is responsible for upgrading Keptn from version 0.6.1 to 0.6.2.
In addition to updating the service images, it configures the api ingress gateway as described in #1660

### DIFF
--- a/upgrader/upgrade-061-062/scripts/upgradeKeptn.sh
+++ b/upgrader/upgrade-061-062/scripts/upgradeKeptn.sh
@@ -13,12 +13,14 @@ manifests=(
   "https://raw.githubusercontent.com/keptn/keptn/release-$KEPTN_VERSION/installer/manifests/keptn/continuous-operations.yaml"
   "https://raw.githubusercontent.com/keptn/keptn/release-$KEPTN_VERSION/installer/manifests/keptn/keptn-api-virtualservice.yaml"
   "https://raw.githubusercontent.com/keptn/keptn/release-$KEPTN_VERSION/installer/manifests/keptn/keptn-ingress.yaml"
+  "https://raw.githubusercontent.com/keptn/keptn/release-$KEPTN_VERSION/installer/manifests/keptn/uniform-services-openshift.yaml"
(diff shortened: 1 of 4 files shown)


upgrader/upgrade-061-062/scripts/upgradeKeptn.sh
```diff
@@ -0,0 +1,131 @@
...
+
+  if [[ $? == '0' ]]; then
+      print_debug "Full installation detected. Upgrading CD and CO services"
+      kubectl apply -f https://raw.githubusercontent.com/keptn/keptn/release-$KEPTN_VERSION/installer/manifests/keptn/continuous-deployment.yaml
+      kubectl apply -f https://raw.githubusercontent.com/keptn/keptn/release-$KEPTN_VERSION/installer/manifests/keptn/continuous-operations.yaml
+      curl -k https://raw.githubusercontent.com/keptn/keptn/release-$KEPTN_VERSION/installer/manifests/keptn/keptn-api-virtualservice.yaml | \
+        sed 's~DOMAIN_PLACEHOLDER~'"$DOMAIN"'~' | kubectl apply -f -
+  else
+      print_debug "Quality gates installation detected. Upgrading Nginx ingress"
+      kubectl apply -f https://raw.githubusercontent.com/keptn/keptn/release-$KEPTN_VERSION/installer/manifests/keptn/keptn-ingress.yaml
```
agrimmer, line 69: I do not think that this should be applied in a Openshift installation as it does not contain a Keptn-ingress.

upgrader/upgrade-061-062/scripts/upgradeKeptn.sh
```diff
@@ -0,0 +1,131 @@
...
+
+      curl -k https://raw.githubusercontent.com/keptn/keptn/release-$KEPTN_VERSION/installer/manifests/keptn/keptn-ingress.yaml | \
```
agrimmer, line 71: Here we have to be careful because $DOMAIN can contain a port. Correct me if I am wrong.
But if so, the domain.placeholder must not contain a port.

PR #1720, opened 2020-04-29
#1715 Add tcp traffic shifting for blue green

### LINKED ISSUE #1715: Cluster-internal traffic does not use VirtualService in B/G deployments
Hi, 

I'm running a Django, Celery, Postgres, Redis stack but for this bug we just need to focus on Django and Postgres.
Django will mount a secret with the DATABASE_URL which it will use to connect to Postgres.

`DATABASE_URL: postgresql://x:y@postgres:5432/foo`

While onboarding postgres service, keptn automatically created a virtual service:

```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: postgres
  namespace: prod
spec:
  gateways:
  - public-gateway.istio-system
  - mesh
  hosts:
  - postgres.prod.domain.de
  - postgres
  http:
  - route:
    - destination:
        host: postgres-canary.prod.svc.cluster.local
    - destination:
        host: postgres-primary.prod.svc.cluster.local
      weight: 100
```

and the following services are getting created:

```
postgres           ClusterIP   10.100.174.169   <none>        5432/TCP
postgres-canary    ClusterIP   10.100.138.179   <none>        5432/TCP
postgres-primary   ClusterIP   10.100.164.191   <none>        5432/TCP
```

Now to the actual problem I encountered. You would except that the DATABASE_URL is still valid because the virtual service is changing the destination between the canary and primary service but this is **not** working from the pod perspective.

If the pod is now trying to connect to the postgres with the above connection string:
`DATABASE_URL: postgresql://x:y@postgres:5432/foo`
He'll end up at the service `postgres` instead of `postgres-primary` so the connection will fail because the pod is looking for the service name not the virtual service hosts.

You can overcome this, by changing the connection string to: 
`DATABASE_URL: postgresql://x:y@postgres-primary:5432/foo`
but this would mean, that the canary deployment is also using the primary database, which is not a good idea. 

I don't have a proposal to fix this issue, yet.

Kind regards,
Stephan

image

christian-kreuzberger-dtx:
This potentially fixes #1715, by adding a check for tcp traffic shifting in the virtualservice.

Needs to be cherry-picked for 0.6.2.

Helm-Service Image: `keptn/helm-service:patch.1715.20200429.1445`

### DIFF
--- a/helm-service/controller/mesh/istio.go
+++ b/helm-service/controller/mesh/istio.go
@@ -59,17 +59,38 @@ func (*IstioMesh) UpdateWeights(virtualService []byte, canaryWeight int32) ([]by
 		return nil, errors.New("Invalid canary weight")
 	}
 
-	for _, httpRoute := range vs.Spec.Http {
-		for _, dst := range httpRoute.Route {
+	if len(vs.Spec.Http) > 0 {
+		fmt.Println("Updating weight for HTTP Traffic")
+		for _, httpRoute := range vs.Spec.Http {
+			for _, dst := range httpRoute.Route {
 
-			if !strings.HasPrefix(dst.Destination.Host, vs.ObjectMeta.Name) {
-				return nil, fmt.Errorf("Cannot update VirutalService because host has unexpected name %s", dst.Destination.Host)
+				if !strings.HasPrefix(dst.Destination.Host, vs.ObjectMeta.Name) {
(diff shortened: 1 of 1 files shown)


christian-kreuzberger-dtx:
closing this, as the underlying issue #1715 can not be fixed with this.

PR #1727, opened 2020-04-30
Fix error due to long list of SLIs

### LINKED ISSUE #1641: HeatMap runs in JS error due to long list of SLIs
Environment: Keptn 0.6.1 + Bridge EAP: 20200402:1046

In this Keptn a list of 100+ SLIs is configured. 

**HeadMap Isssue:** 
For the first ~4 test evaluations, the HeatMap worked fine with this high number of SLIs. After a click on the HeatMap button the bridge runs in an error: `Highcharts error #12: www.highcharts.com/errors/12` 

The reason might be that the setting `plotOptions.series.turboThreshold` has a default of 1000 and in this case the default is exceeded.  

Problematic HeatMap (result): 
image

**Definition of Done:**  
- [ ] BugFix for the JS error caused by the default value:  `plotOptions.series.turboThreshold`

**SLI Breakdown Issue**
In the SLI breakdown, three columns are created for the list of SLIs. 

image

**Definition of Done:**  
- [x] Enforce the component to not create multiple columns, but to show the list of SLIs in one column.

ermin-muratovic:
Fixes #1641 

image

### DIFF
--- a/bridge/client/app/_components/ktb-sli-breakdown/ktb-sli-breakdown.component.html
+++ b/bridge/client/app/_components/ktb-sli-breakdown/ktb-sli-breakdown.component.html
@@ -1,16 +1,26 @@
 <p class="mb-0">SLI breakdown</p>
-<dt-key-value-list>
+<dt-key-value-list [columns]="1">
   <ng-container *ngFor="let result of _indicatorResultsFail">
     <dt-key-value-list-item>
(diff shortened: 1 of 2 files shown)


ermin-muratovic:
Also the threshold in the heatmap is disabled, so there is no limit for tiles in the heatmap.

image (2)

PR #1762, opened 2020-05-06
Improved time duration visualization

ermin-muratovic:
The time is now visualized in the format `x hours y minutes z seconds`. If hours or minutes is 0 they are omitted.

image


ermin-muratovic:
image

PR #1771, opened 2020-05-08
Use alias for Keptn Slack URL

agrimmer:


cli/cmd/generate_support_archive.go
```diff
@@ -151,7 +151,7 @@ keptn generate support-archive --dir=/some/directory`,
 
 		fmt.Println("The support archive is available here " + supportArchive)
 		fmt.Println("This support archive potentially contains sensitive data. Therefore, please first review it before distributing.")
-		fmt.Println("If you need help, please use the #help channel in the Keptn Slack workspace https://join.slack.com/t/keptn/shared_invite/enQtNTUxMTQ1MzgzMzUxLWMzNmM1NDc4MmE0MmQ0MDgwYzMzMDc4NjM5ODk0ZmFjNTE2YzlkMGE4NGU5MWUxODY1NTBjNjNmNmI1NWQ1NGY")
+		fmt.Println("If you need help, please use the #help channel in the Keptn Slack workspace http://slack.keptn.sh")
```
christian-kreuzberger-dtx, line 154: is it really `http://`? Can we make sure we have a valid SSL certificate for this domain?

jetzlstorfer: it is `http://` right now. for `https://` we need to use cloudfront in AWS. I don't have privileges to manage this.

Issue #1766, opened 2020-05-06
Extend automated integration tests for GKE to use various versions of Kubernetes: 1.14, 1.15, 1.16, 1.17


christian-kreuzberger-dtx:
Right now the integration test for GKE uses just Kubernetes version 1.14.

As recently described in #1529 and https://github.com/keptn/examples/issues/101, it is possible to run Keptn and the sockshop example on Kubernetes versions 1.16 and 1.17.

We should have a way to automatically verify this using our nightly integration/platform test.

christian-kreuzberger-dtx:
I've tried including various versions of GKE, but right now only 1.14 and 1.15 are supported (as they are available in the **regular** channel).

For all other versions, we have to wait for GKE to make them available to us.
See
* https://cloud.google.com/kubernetes-engine/docs/release-notes-regular
* https://cloud.google.com/kubernetes-engine/docs/release-notes-stable

However, I'll prepare our .travis.yml such that we can easily enable tests for 1.16 and 1.17 once they are available.

PR #1773, opened 2020-05-08
#1766 Enable GKE tests for 1.15, prepare of 1.16 and 1.17; enable tests for various minikube versions

christian-kreuzberger-dtx:
This PR will close issue #1766.

* Reworked the GKE tests for travis ci to allow more than one nightly cluster to be created. The name is now used based on the branch name and GKE version to be tested
* Added `GKE_VERSION` as an environment variable that allows matrix tests
* Also reworked the Minikube tests to use different Minikube versions as a matrix test
* Added a simple script for exposing the Keptn Bridge (for easier debugging/troubleshooting of failed integration tests). The username and password are stored as secret variables in travis-ci, and I have a copy of those values which I can hand out to anyone that needs them.

Please note: Matrix tests in travis are executed in parallel, so this change does not increase the overall test-time in travis-ci:

Screenshot from 2020-05-08 12-58-10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,6 +82,8 @@ before_install:
     export KEPTN_INSTALLER_IMAGE=keptn/installer:${BRANCH#"release-"}
     export KEPTN_CLI_VERSION=${BRANCH#"release-"}
   fi
+# generate a lower-case slug of the branch name
+- BRANCH_SLUG=$(echo ${BRANCH} | sed -r s/[^a-zA-Z0-9]+/-/g | sed -r s/^-+\|-+$//g | tr A-Z a-z)
 ### ATTENTION: please make sure installer is always the last in this list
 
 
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -491,3 +549,11 @@ jobs:
...
       - kubectl get services --all-namespaces
       - kubectl get ingress --all-namespaces
+
+  - <<: *minikubeStandaloneTest
+    env:
+      - MINIKUBE_VERSION=v1.3.1
+
+  - <<: *minikubeStandaloneTest
+    env:
+      - MINIKUBE_VERSION=v1.4.0
```
johannes-b, line 559: Am I right that:
- Minikube v1.3.1 runs K8s 1.15
- Minikube v.1.4.0 runs K8s 1.16
as shown here: https://github.com/kubernetes/minikube/blob/v1.4.0/CHANGELOG.md 

If yes, I will take care of updating the Minikube column of the Keptn/K8s support matrix: https://github.com/keptn/keptn.github.io/pull/539

christian-kreuzberger-dtx: Versions seem to be correct - though this only affects the control-plane installation (not delivery).

johannes-b: thanks for clarification!

Issue #1723, opened 2020-04-30
Travis-CI: Installation on Minishift fails sometimes (timing issue)?


christian-kreuzberger-dtx:
Sometimes the installation on Minishift via Travis-CI fails. Installation actually finishes (without an error):
```
[keptn|DEBUG] [2020-04-30 06:11:54] All pods are running in namespace keptn.
[keptn|INFO] [2020-04-30 06:11:54] Keptn wears Openshift uniform
[keptn|INFO] [2020-04-30 06:11:54] Installation of Keptn complete.
```

But the authentication at the end of the installation fails:
```
Trying to get auth-token and API endpoint from Kubernetes.
Received Keptn Domain: 127.0.0.1.nip.io
Starting to authenticate
Authentication was unsuccessful. Post https://api.keptn.127.0.0.1.nip.io/v1/auth: dial tcp 127.0.0.1:443: connect: connection refused
Retrying...
Authentication was unsuccessful. Post https://api.keptn.127.0.0.1.nip.io/v1/auth: dial tcp 127.0.0.1:443: connect: connection refused
Retrying...
Authentication was unsuccessful. Post https://api.keptn.127.0.0.1.nip.io/v1/auth: dial tcp 127.0.0.1:443: connect: connection refused
Retrying...
```

I'm guessing this is could be a timing issue, but we need to investigate this.

christian-kreuzberger-dtx:
After initial investigation, the problem does not appear when using a full installation.

christian-kreuzberger-dtx:
I was able to debug the problem on travis-ci using `oc`, maybe this output helps:

```
$> oc get route --all-namespaces -owide
NAMESPACE   NAME      HOST/PORT                    PATH      SERVICES            PORT      TERMINATION   WILDCARD
keptn       api       api.keptn.127.0.0.1.nip.io             api-gateway-nginx   http      edge/None     None
keptn       api2      api.keptn                              api-gateway-nginx   http      edge/None     None
```

```
$>
oc get endpoints --all-namespaces -owide
NAMESPACE                       NAME                          ENDPOINTS                                            AGE
default                         docker-registry               <none>                                               5m
default                         kubernetes                    10.20.1.202:8443                                     8m
default                         router                        <none>                                               5m
keptn-datastore                 mongodb                       172.17.0.11:27017                                    1m
keptn-datastore                 mongodb-datastore             172.17.0.12:8080                                     1m
[...]

PR #1783, opened 2020-05-13
#1723 added more debug output to minishift tests

christian-kreuzberger-dtx:
added some more debug output for minishift tests


christian-kreuzberger-dtx:
After some troubleshooting, I will dismiss this PR as it doesn't help us in fixing the random problem that occurs in #1723

Issue #1737, opened 2020-05-04
Keptn Core (gatekeeper): React on an approval.finished event to send configuration changed event for the current stage


johannes-b:
The implementation of the approval strategy is based on the event stream outlined here: #1561 
 
Functionality of the gatekeeper when receiving an approval.finished: 
 
1. Listen to `approval.triggered` and `approval.finished` event.
1. Check that image/tag from the approval.finished event match with the corresponding approval.triggered event. 
 1. If image/tag match and approval is valid, send a configuration.change event for the stage to trigger the approved deployment 

**Definition of Done:**  
- [x] Check that image/tag from the approval.finished event match with the corresponsing approval.triggered event. 
- [x] A configuration.change event is sent as response to a valid approval.finished 
- [x] Tests are provided to verify the implementation of the image/tag check

PR #1784, opened 2020-05-13
#1737 handle approval.finished events in gatekeeper

bacherfl:
This PR implements the behavior described in #1737.

When receiving an `approval.finished` event, the gatekeeper will retrieve the corresponding `approval.triggered` event from the configuration service. When the open approval has been fetched, the gatekeeper will check if the image and tag of then `approval.triggered` and `approval.finished` event match. If yes, the `configuration-change` event for the next stage is sent, otherwise the approval is discarded.


gatekeeper-service/pkg/handler/approval_triggered_event_handler.go
```diff
@@ -90,3 +98,47 @@ func (a *ApprovalTriggeredEventHandler) getApprovalFinishedEvent(inputEvent kept
...
+
+	newApproval := &approval{
+		EventID:      eventID,
+		Image:        image,
+		KeptnContext: keptnContext,
+		Tag:          tag,
+		Time:         time,
+	}
+
+	queryURL := fmt.Sprintf("%s://%s/v1/project/%s/stage/%s/service/%s/approval", configurationServiceEndpoint.Scheme, configurationServiceEndpoint.Host, project, stage, service)
```
christian-kreuzberger-dtx, line 116: Duplicate code? Can you use the `getApprovalsEndpoint` function here too?

Issue #1749, opened 2020-05-05
Keptn Core (CLI): Allow user to send an approval event to the provided stage and to approve a deployment using the CLI 


johannes-b:
By using the CLI, a Keptn user has the option to manually trigger a deployment in a stage that has an approval_strategy configured:  

1.) Get all open approvals that are open for a project/stage (optional service):  

```console 
keptn get event approval.triggered --project=PROJECTNAME --stage=STAGENAME 
``` 

TBD: Output of this command

2.) Send approval.finished event to Keptn for a project/stage specified by keptncontext: 
```console 
keptn send event approval-finished --project=PROJECTNAME --stage=STAGENAME --id=1234.5432  
``` 
 
- CLI sends sh.keptn.event.approval.finished event to the Keptn API 
- API (eventbroker) forwards the event to the gatekeeper 
- Gatekeeper makes sanity check of image and tag against sh.keptn.event.approval.triggered. 
- Gatekeeper sends configuration-changed event for this stage.  
- MongoDB updates materialized view 

 **Definition of Done:**  
- [x] Keptn CLI provides commands 
- [x] Keptn CLI sends sh.keptn.event.approval.finished 
- [x] MongoDB updates materialized view 
- [ ] Auto-gen CLI documentation on keptn.sh/docs 

**To be defined:**
- [x] API endpoints for fetching open approvals - see #1757 
- [ ] CLI command
- [ ] CLI output

PR #1798, opened 2020-05-15
#1791 expose mongodb datastore via API gateway

### LINKED ISSUE #1791: Expose /event endpoint from mongodb-datastore to the public Keptn API
Like the configuration-service exposes its API to the public Keptn API, the mongodb-datastore should do the same for its /event endpoint.

**Tasks:**

* The `/event` endpoint in the API has to be marked as deprecated. (Do not delete the implementation of the endpoint, as we don't want to introduce a breaking change.)
```
    get:
      tags:
        - Event
      deprecated: true
      summary: 'Deprecated endpoint - please use /mongodb-datastore/v1/event'
```
* Re-configure the mongodb-datastore swagger-ui, to be exposed on the public api-gateway.

* *Investigation:* Ist the `/logs` endpoint still needed? If not, it has to be removed.

**Definition of Done**
- [ ] The `/event` endpoint of the API is marked as deprecated: https://github.com/keptn/keptn/blob/master/api/swagger.yaml#L52
- [ ] The `/event` endpoint is available when selecting the mongodb-datastore in the drop-down of the Keptn swagger UI and can be used.
image
- [ ] Update of the keptn/docs: https://keptn.sh/docs/develop/reference/api/#explore-the-keptn-api
- [ ] The `/logs` endpoint is deleted (if not needed)

bacherfl:
This PR will make the mongodb-datastore available via the Keptn API.

This will allow for more flexibility in terms of retrieving events via the API, as needed in Issue #1749.

Deprecates the `GET v1/event` endpoint of the `api-service`, which only provided the possibility to retrieve a single event by its type and keptn context ID.

Closes #1791.

### DIFF
--- a/gatekeeper-service/deploy/service.yaml
+++ b/gatekeeper-service/deploy/service.yaml
@@ -20,8 +20,8 @@ spec:
         - containerPort: 8080
         resources:
           requests:
-            memory: "256Mi"
-            cpu: "500m"
+            memory: "32Mi"
+            cpu: "50m"
           limits:
             memory: "256Mi"
             cpu: "500m"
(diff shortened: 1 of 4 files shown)


gatekeeper-service/deploy/service.yaml
```diff
@@ -20,16 +20,18 @@ spec:
         - containerPort: 8080
         resources:
           requests:
-            memory: "32Mi"
-            cpu: "50m"
+            memory: "256Mi"
```
christian-kreuzberger-dtx, line 23: Do we really need to request such a high amount? Is it not enough to just increase the limit?

gatekeeper-service/deploy/service.yaml
```diff
@@ -43,4 +45,59 @@ spec:
...
+    kubectl.kubernetes.io/last-applied-configuration: |
+      {"apiVersion":"apps/v1","kind":"Deployment","metadata":{"annotations":{},"name":"gatekeeper-service-evaluation-done-distributor","namespace":"keptn"},"spec":{"replicas":1,"selector":{"matchLabels":{"run":"distributor"}},"template":{"metadata":{"labels":{"run":"distributor"}},"spec":{"containers":[{"env":[{"name":"PUBSUB_URL","value":"nats://keptn-nats-cluster"},{"name":"PUBSUB_TOPIC","value":"sh.keptn.events.evaluation-done"},{"name":"PUBSUB_RECIPIENT","value":"gatekeeper-service"}],"image":"keptn/distributor:latest","name":"distributor","ports":[{"containerPort":8080}],"resources":{"limits":{"cpu":"500m","memory":"128Mi"},"requests":{"cpu":"50m","memory":"32Mi"}}}]}}}}
+  creationTimestamp: "2020-05-08T07:50:00Z"
+  generation: 1
+  name: gatekeeper-service-evaluation-done-distributor
+  namespace: keptn
+  resourceVersion: "594181"
+  selfLink: /apis/extensions/v1beta1/namespaces/keptn/deployments/gatekeeper-service-evaluation-done-distributor
+  uid: 84dff9d8-9100-11ea-9b9a-42010a80002c
+spec:
```
christian-kreuzberger-dtx, line 65: did you add the deployment of the distributor on purpose? If so, please remove the metadata like uid, selfLink, resourceVersion, ...

gatekeeper-service/deploy/service.yaml
bacherfl: that change was unintentional and has been removed

gatekeeper-service/deploy/service.yaml
bacherfl: that change was unintentional and has been removed

PR #1803, opened 2020-05-15
#1802 adapt helm template commands to helm v3

### LINKED ISSUE #1802: Keptn full install on OpenShift fails due to Helm v3
In the current **master branch (build 2020-05-15)** the Keptn full installation on OpenShift fails with the following error:

```
[keptn|INFO] [2020-05-15 11:20:44] Installing Istio on OpenShift (this might take a while)
namespace/istio-system created
Error: unknown flag: --name
```

This seems to be the relevant line of code:
https://github.com/keptn/keptn/blob/5330ae0b1cac9c91883767edf7e4f6b25e0e61b2/installer/scripts/openshift/setupIstio.sh#L6

Initial troubleshooting (thanks to @bacherfl ) points to a change in Helm (from v2 to v3) when installing istio. For the non-openshift installation this is the correct code:
https://github.com/keptn/keptn/blob/5330ae0b1cac9c91883767edf7e4f6b25e0e61b2/installer/scripts/common/setupIstio.sh#L29-L43

# Definition of done
- [ ] Keptn full installation with Istio can be installed with helm v3 on OpenShift 3.11

bacherfl:
Closes #1802 
Due to the upgrade to Helm v3, the `--name` flag we previously used in the `helm template` commands, is not supported anymore. Instead, the name of the template is not a mandatory parameter, as described in the [Helm v3 docs](https://helm.sh/docs/helm/helm_template/)

### DIFF
--- a/installer/scripts/common/setupIstio.sh
+++ b/installer/scripts/common/setupIstio.sh
@@ -26,7 +26,7 @@ else
     print_info "Install Istio"
     kubectl create namespace istio-system
 
-    helm template ../manifests/istio/helm/istio-init --namespace istio-system | kubectl apply -f -
+    helm template istio-init ../manifests/istio/helm/istio-init --namespace istio-system | kubectl apply -f -
     verify_kubectl $? "Creating Istio resources failed"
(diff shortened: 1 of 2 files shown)


christian-kreuzberger-dtx:
Thanks, I'll test the installer image of this branch and will approve the PR once I confirmed that it works on both OpenShift and K8s

Issue #1777, opened 2020-05-11
Kubernetes 1.14 - 1.18 support: Validate master based on Keptn/K8s compatibility matrix for 0.7


johannes-b:
Keptn 0.7 targets to support K8s 1.14 - 1.18 (ignore Kubernetes 1.13 and lower; Kubernetes 1.19 is not out as of writing this issue; OpenShift 3.11 will stay as it is).

In this issue, **Keptn master** should be validated on the Keptn/K8s support matrix proposed here: https://github.com/keptn/keptn.github.io/pull/539

* For GKE, this is already done in: #1766 (and https://github.com/keptn/keptn/pull/1812 for GKE 1.16)
* For Keptn *Control Plane* on Minikube, this has also partially been handled via #1766 (see PR #1773 and PR #1782)

**Definition of Done**:
- [x] [Keptn CLI version check](https://github.com/keptn/keptn/blob/c67d4306adc5bcd38947f26a1cb88c5d8ff99167/cli/cmd/install.go#L220-L225) adapted based on compatible K8s versions; for both, control and execution plane (but ignore OpenShift)
- [x] Know limitations (or identified problems) are listed in the release notes for Keptn 0.7

christian-kreuzberger-dtx:
I'll do some initial checks on newer K8s versions using Minikube in #1782. I'll update the original description once I'm done

christian-kreuzberger-dtx:
Preparations have been partially done in 
* https://github.com/keptn/keptn/pull/1804
and
* https://github.com/keptn/examples/pull/105

christian-kreuzberger-dtx:
So far, compatibility with GKE has been tested using travis-ci for GKE 1.14 and 1.16 (newer versions are not available yet).
Also in travis-ci we are testing `keptn install --use-case=quality-gates` on Microk8s and Minikube in various versions between 1.14 and 1.18.

I still need to test the full installation on Kubernetes 1.17 and 1.18.

christian-kreuzberger-dtx:
Unfortunately I can not test the current master branch and the current installer. The following issue is blocking me: https://github.com/keptn/keptn/issues/1861

christian-kreuzberger-dtx:
Not sure if related to #1861 or not, but integration tests are currently failing: #1872

christian-kreuzberger-dtx:
Integration tests are working again. I'm testing K8s 1.17 and 1.18 manually right now (1.16 and below is automatically tested via GKE).

[...]

PR #1807, opened 2020-05-15
#1777 Validate Minikube against multiple Kubernetes versions

christian-kreuzberger-dtx:
Within this PR I've added Minikube tests for full installation.


christian-kreuzberger-dtx:
Unfortunately, the minikube full test does not work on travis-ci as we are running out of either cpu or memory.

Issue #1624, opened 2020-04-08
CLI: Implement get projects, services, stages and metadata


christian-kreuzberger-dtx:
From [KEP 10](https://github.com/keptn/enhancement-proposals/pull/10).

After #1621, #1622 and #1623 are done, we need to implement the same functionality within the CLI, but in a very generic way.

Examples:

* `keptn get projects` -> returns a list of projects
* `keptn get projects --project=sockshop` -> returns infos for the selected project
* `kepn get services` -> returns a list of services (for all projects)
* `keptn get services --project=sockshop` -> returns a list of services for project sockshop
* `keptn get stages` -> does not make sense, right?
* `keptn get stages --project=sockshop` -> returns a list of stages for the given project
* Every call should work with singular and plural (e.g., `keptn get project` vs. `keptn get projects`).

For every call, we just take the result from the API, and display the result (in a nice way). Do not create any "hard dependencies" in Go (e.g., models) for this, we want this to be as generic as custom resources in Kubernetes.

agrimmer:
In order to stay consistent within our CLI commands
`keptn get projects --project=sockshop` should be `keptn get project sockshop`

Right, `keptn get stages` makes no sense without any project.

robertjahn:
I suggest for the get project <project name> -- that this includes the upstream repo info or say NONE

Consider the verb "describe" like kubectl does `describe       Show details of a specific resource or group of resources`

johannes-b:
The following commands should return: 

**Project:**
* `keptn get projects`:
```
NAME           CREATION DATE                 
sockshop       2020-04-06T14:35:40.210Z
musicshop      2020-04-06T14:37:45.210Z
```

* `keptn get project sockshop`:
```
NAME           CREATION DATE                 
sockshop       2020-04-06T14:35:40.210Z
```

* `keptn get project sockshop --output=yaml`:
```
projects:
- creationDate: '1587455680742808213'
  projectName: sockshop
  stages:
  - services:
    - creationDate: '1587455682123389180'
      deployedImage: docker.io/keptnexamples/carts:0.10.1
      lastConfigurationChangedEvent:
        keptnContext: 01ad3740-53bb-47b6-bb7f-08c3707fa7f4
        time: '1587455897846488861'
      lastDeploymentFinishedEvent:
        keptnContext: 01ad3740-53bb-47b6-bb7f-08c3707fa7f4
        time: '1587455890460007826'
      lastEvaluationDoneEvent:
[...]

PR #1811, opened 2020-05-19
Feature/1624/cli enhancements get commands

checkelmann:
Resolves issue https://github.com/keptn/keptn/issues/1624


cli/cmd/get_project.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+type getProjectStruct struct {
+	outputFormat *string
+}
+
+var getProject getProjectStruct
+
+var getProjectCmd = &cobra.Command{
+	Use:   "project",
```
bacherfl, line 42: according to the Issue, the better approach would be to merge the command definitions for the singular and the plural into one command, using the `Aliases` property:

```
var getProjectCmd = &cobra.Command{
	Use:   "project",
        Aliases: []string{"projects"},
	Short: "Get details of a keptn project",
	Long:  `Get details of a given keptn project`,
	Example: `keptn get project sockshop
// ...
```

The same holds for the `get stage(s)` and `get service(s)` commands

bacherfl:
Thank you for the contribution! I have added one remark in the comments about merging the singular and plural versions of the commands

cli/cmd/get_project.go
checkelmann: Ok, will look into it and try to merge these two commands into a single file, but not sure how to handle required arguments like --project and the different help text on the different sub-commands. 

If you have an example how I can do that would be good.

checkelmann: Thank you @bacherfl, I've merged the commands.

checkelmann:
Hi,

I'm done with my changes. This could be reviewed now.
@bacherfl @johannes-b

PR #1812, opened 2020-05-19
#1777 enable GKE 1.16; configure domain to nip.io

christian-kreuzberger-dtx:
This PR adapts travis ci yaml to use GKE 1.16, as it is now available as a regular version. As we know that 1.15 is similar to 1.14, but 1.16 is not, I've disabled the tests for 1.15 to avoid creating too many clusters on GKE.

In addition, as xip.io is failing a lot of times now, I've switched it over to nip.io using a `keptn configure domain` command. This way, we do not only test `keptn configure bridge` but also `keptn configure domain`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -408,20 +408,19 @@ jobs:
       - KEPTN_INSTALLATION_TYPE=FULL
     <<: *gke_full # use template
 
-  - stage: Test GKE Full (--platform=gke)
-    if: branch = master AND type = cron # run for cron
-    env:
-      - GKE_VERSION=1.15
-      - KEPTN_INSTALLATION_TYPE=FULL
-    <<: *gke_full # use template
-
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -408,21 +408,20 @@ jobs:
       - KEPTN_INSTALLATION_TYPE=FULL
     <<: *gke_full # use template
 
-  - stage: Test GKE Full (--platform=gke)
-    if: branch = master AND type = cron # run for cron
-    env:
-      - GKE_VERSION=1.15
```
johannes-b, line 414: Just curious, did you disable 1.15 tests to save time/money or is there any other reason?

christian-kreuzberger-dtx: To save time, money and resources.

1.14 and 1.15 are very similar, so there is no benefit from testing both. 1.14 however is the oldest version we are still supporting.

Issue #1751, opened 2020-05-05
Upgrade path from 0.6.2 to 0.7.0 for manual approval use case


johannes-b:
* First version of Keptn upgrader for 0.7

* Research whether an upgrade path for the story-related changes/enhancements must be provided that is not covered by the regular upgrader (i.e., exchanging the gatekeeper-service) 

* The upgrader will not change the shipyard of an existing project. Consequently, documentation must be provided how-to do that. 

**Definition of Done:**  
- [ ] First version of upgrader for 0.7
- [ ] If needed, an upgrader job is provided for the missing upgrade functionality 
- [ ] Documenation on keptn.sh/docs about adding an approval_strategy to a stage (in /continuous_delivery/manual_approval page)

PR #1826, opened 2020-05-20
#1751 upgrade job

bacherfl:
This upgrader job will take care of upgrading keptn 0.6.2 to 0.7.0 and will perform the following actions:

- constructing the materialized view for the configuration service, based on existing project information
- migrating to helm 3
- updating the k8s manifests/image tags (NOTE: for the keptn-contrib services, we will likely have to update the applied versions as soon as we know which versions are going to be released)


upgrader/upgrade-062-070/mongodb-upgrade/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/keptn/keptn/upgrade-061-070/mongodb-upgrade
```
christian-kreuzberger-dtx, line 1: This still hast he wrong module name *upgrade-061-070* in it

upgrader/upgrade-062-070/scripts/upgradeKeptn.sh
```diff
@@ -0,0 +1,124 @@
...
+  helm3 2to3 convert $release
+  verify_install_step $? "Helm2-to3 release convertion failed"
+done
+
+yes y | helm3 2to3 cleanup --tiller-cleanup
+verify_install_step $? "Helm2-to3 cleanup failed"
+
+
+
+KEPTN_VERSION="release-0.7.0"
```
christian-kreuzberger-dtx, line 30: As `release-0.7.0` is not available yet, the upgrader is hard to test.

Any change we can make `KEPTN_VERSION` an environment variable for the manifest that we apply? We could then use it as follows in the bash script:

```console
KEPTN_VERSION=${KEPTN_VERSION:-"release-0.7.0"}
```

and add an environment variable to the manifest `KEPTN_VERSION=master`

PR #1832, opened 2020-05-20
Show stage details in environments screen

ermin-muratovic:
When clicking on a stage, you now see the deployed services with details like the public deployment URI and the deployment time.

Unfortunately at the moment we can't display more details about the stage in general, like the deployment strategy or the test strategy. This needs to be considered in a separate issue, once we can read those information via the api.

image


ermin-muratovic:
Unfortunately at the moment we can't display more details about the stage in general, like the deployment strategy or the test strategy. This needs to be considered in a separate issue, once we can read those information via the api.

PR #1833, opened 2020-05-20
Use public api in bridge

ermin-muratovic:


christian-kreuzberger-dtx:
Alright, we are ready to get this merged now :) 
@ermin-muratovic please fix the conflicts.

bridge/server/api/index.js
```diff
@@ -119,5 +54,23 @@ module.exports = (params) => {
     }
   });
 
+  router.all('*', async (req, res, next) => {
+    try {
+      const result = await axios({
+        method: req.method,
+        url: `${apiUrl}${req.url}`,
+        data: req.params,
```
christian-kreuzberger-dtx, line 62: are we sure we won't run into the same problem here as with the version check with CloudFront?

ermin-muratovic:
@christian-kreuzberger-dtx you are right. I added a check on `req.method` and omit the `data` parameter for "GET" requests.

PR #1874, opened 2020-06-04
#1752 added instructions to re-enable bridge access 

### LINKED ISSUE #1752: After keptn configure domain, an already exposed bridge is no longer accessible
Steps to reproduce:

1. Install Keptn 0.6.2 (full installation)
2. expose bridge using `keptn configure bridge --action=expose` - bridge is exposed at (e.g.) `bridge.keptn.1.2.3.4.xip.io`
3. change domain using `keptn configure domain 1.2.3.4.nip.io`
4. Try to access bridge using the new domain `https://bridge.keptn.1.2.3.4.nip.io` (does not work)

Interesting enough, the old domain (using xip.io) still works for Keptn's bridge.

# Some more data

```console
$ kk get vs
NAME     GATEWAYS                        HOSTS                                             AGE
api      [public-gateway.istio-system]   [api.keptn.1.2.3.4.nip.io api.keptn]   4m50s
bridge   [public-gateway.istio-system]   [bridge.keptn.1.2.3.4.xip.io]                131m
```

# Work-around

Right now, the user has to execute the following commands to make the setup work again:
```console
keptn configure bridge --action=lockdown
keptn configure bridge --action=expose
```

# Possible Fix

I guess one of the two following variants should work:

* `keptn configure domain` should output a warning, that the user needs to manually check and update all other virtual services (e.g., bridge, but also onboarded services)
* `keptn configure domain` needs to also update other virtual services, not just the API vs (which is handled via a `kubectl apply`). 

**Please note, I did not check if this also happens when using nginx-ingress (e.g., when using the quality-gates installation), but I guess nginx-ingress is handling this differently**

bacherfl:
Closes #1752 
After configuring the domain, the following output is appended:

```
NOTE: If you have exposed the Keptn's bridge via 'keptn configure bridge --action=expose', please execute the following commands to re-enable access:
keptn configure bridge --action=lockdown
keptn configure bridge --action=expose

NOTE: VirtualServices for services that have been onboarded previously have not been updated.
```

Note that updating the VirtualServices for onboarded services also requires updating the corresponding resources within the configuration service (GitOps repo for the service). If we would like to tackle this, I propose to create a follow up-issue for moving the execution of the `configure domain` command to the `api-service` (as we did with the `configure bridge` command). This would also remove the dependency of having a configured `kubectl` CLI available locally when executing this command

### DIFF
--- a/cli/cmd/configure_domain.go
+++ b/cli/cmd/configure_domain.go
@@ -194,6 +194,12 @@ Please find more information on https://keptn.sh/docs/develop/reference/troubles
 				}
 				logging.PrintLog("Successfully configured domain", logging.InfoLevel)
 			}
+			fmt.Println()
+			logging.PrintLog("NOTE: If you have exposed the Keptn's bridge via 'keptn configure bridge --action=expose', please execute the following commands to re-enable access:", logging.InfoLevel)
+			logging.PrintLog("keptn configure bridge --action=lockdown", logging.InfoLevel)
+			logging.PrintLog("keptn configure bridge --action=expose", logging.InfoLevel)
+			fmt.Println()
+			logging.PrintLog("NOTE: VirtualServices for services that have been onboarded previously have not been updated.", logging.InfoLevel)
 		}
 
 		return nil


christian-kreuzberger-dtx:
While I dislike that we have to go the approach of telling the user what to do (instead of actually doing it), I fully agree with what you say in the description:

> If we would like to tackle this, I propose to create a follow up-issue for moving the execution of the configure domain command to the api-service (as we did with the configure bridge command). This would also remove the dependency of having a configured kubectl CLI available locally when executing this command

I'll create a follow-up issue.

Issue #1486, opened 2020-03-12
Unit tests for Bridge


christian-kreuzberger-dtx:
Write some unit tests for Keptn's Bridge that test certain services and/or controller.

The Bridge is missing: 
- UI tests
- Integration tests

**Definition of Done:**
- [ ] A solid basis of UI tests is implemented
- [ ] A solid basis of Integration tests is implemented

PR #1882, opened 2020-06-05
#1486 Automate unit tests for Keptn Bridge

christian-kreuzberger-dtx:
This PR enhances travis-ci with the unit tests implemented in #1873 

In addition, I noticed that several import statements in the spec files were creating troubles, therefore I tried to reduce the imports to `AppModule` where possible.

@ermin-muratovic please take a look at my changes, and as discussed via Slack, adapt the imports if you believe that it is necessary (e.g., http loading spinner should not need the full appmodule)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,7 +128,25 @@ jobs:
   ##################################################################################
   # Jobs that are always executed
   ##################################################################################
-  - stage: Unit tests
+  - stage: Unit tests # Bridge
+    os: linux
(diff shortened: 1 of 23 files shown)


christian-kreuzberger-dtx:
@ermin-muratovic  Tests for `ktb-selectable-tile` seem to fail

christian-kreuzberger-dtx:
Thanks Ermin, this is great!

Issue #1862, opened 2020-06-02
Use k8s service account with a restricted set of permissions instead of default 


thschue:
**Is your feature request related to a problem? Please describe.**
Currently, keptn services use the default serviceaccount with a set of high privileges inside the k8s clusters

**Describe the solution you'd like**
Use service accounts with a minimum set of privileges 

**Additional context**
for example: 
- helm-service / verb=get, resource=configmap, resource-name=keptn-domain
- lighthouse-service / verb=create,update, resource=configmap

johannes-b:
Thanks for bringing this up! 

The problem is clear, but I have two questions: 
- Should we provide a set of pre-defined service accounts, which are then assigned to service? 
- How can we handle service accounts of (custom) keptn-services as available on keptn-contrib/sandbox? Do we just provide a recommendation for those?

johannes-b:
**Research Tasks:** 
- Research on which services require K8s API access.
- Research on RBAC for OpenShift.
- Research on NATS.

**Decision:**
- dynatrace-service, prometheus-service have their own role-binding

Issue #1770, opened 2020-05-07
Rename installer job to keptn-installer to avoid confusion


grabnerandi:
Right now the keptn installer job is called "installer". 
We should rename it to something that is more keptn specific, e.g.: "keptn-installer" to avoid confusion or to make it more clear what this job is really doing.

cavejay:
In poking my head around - Is this just as simple as updating [this line](https://github.com/keptn/keptn/blob/70acdf43a3e7280fda5e45d1f1288a9d022a0772/installer/manifests/installer/installer.yaml#L4) and then ensuring everything else continues to work?

johannes-b:
Hi, thanks for reaching out on that issue!

For clarification, the line (file) you are referencing is used to support the installation of Keptn 0.5.2. With Keptn 0.6, we moved the job into the install command of the CLI. This is also the place where the job need to be renamed. Please take a look at: 

- installerJob: https://github.com/keptn/keptn/blob/master/cli/cmd/install.go#L70
- deleteKeptnInstallerPod: https://github.com/keptn/keptn/blob/master/cli/cmd/uninstall.go#L130

**Tasks:**
- [ ] Renaming the installer job from `installer` to `keptn-installer` in the code referenced above.
- [ ] Adding the comment `## This installer job is deprecated and needed for Keptn 0.5 only. It will be removed in the future` to the [file](https://github.com/keptn/keptn/blob/70acdf43a3e7280fda5e45d1f1288a9d022a0772/installer/manifests/installer/installer.yaml).

Would you like to work on that?

agrimmer:
With my PR #2022, the Keptn installer is now distributed as Helm-chart and not as installer job anymore. Therefore, I am closing this issue.

PR #1894, opened 2020-06-08
#1862 Introduce RBAC for Keptn services

agrimmer:
This PR implements https://github.com/keptn/keptn/issues/1862. This PR also implements https://github.com/keptn/keptn/issues/1770.

The PR introduces dedicated k8s service accounts for the Keptn services. These service accounts implement a restricted, minimal set of permissions.

Open points: 
- [x] Do not assign cluster-admin rights to installer
- [x] Define tasks for upgrader
- [x] Formulate issues for keptn-contrib repositories how to define service account


platform-support/openshift-route-service/deploy/service.yaml
```diff
@@ -14,6 +14,7 @@ spec:
       labels:
         run: openshift-route-service
     spec:
+      serviceAccountName: keptn-default
```
bacherfl, line 17: are the service account privileges sufficient to create OpenShift routes?

agrimmer: Thank you. I added a rule.

PR #1912, opened 2020-06-15
#1830 do not include port in host for VirtualService and Ingress rules

### LINKED ISSUE #1830: keptn expose bridge broke for installations quality-gates and gateway nodeport
I installed keptn master branch using (edit: also confirmed for 0.6.2)

```console
keptn install --platform=kubernetes --gateway=NodePort --use-case=quality-gates
```
on Minikube

and tried to expose Keptn Bridge, which results in the following error:

```
$ keptn configure bridge --action=expose
Error: Could not complete command: {"code":500,"message":"Ingress.extensions \"keptn-ingress\" is invalid: spec.rules[2].host: Invalid value: \"bridge.keptn.192.168.99.100.xip.io:30716\": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')"}
```

# Analysis

The problem is that Keptn is trying to configure the nginx-ingress object such that the hostname contains a port (because of `keptn install ... --gateway=NodePort`).

This is the part of the code that is responsible for the problem:
https://github.com/keptn/keptn/blob/f7b0dd94baae466d2218881d73765a0afe71259a/api/handlers/configure.go#L302-L307

# Workaround

For now the workaround is to manually edit the ingress object and add Keptn Bridge manually.

```console
kubectl -n keptn edit ingress keptn-ingress
```

The spec of the ingress object should look as follows (where 10.200.0.2.xip.io is my keptn domain):
```console
spec:
  rules:
  - host: api.keptn.10.200.0.2.xip.io
    http:
      paths:
      - backend:
          serviceName: api-gateway-nginx
          servicePort: 80
  tls:
  - hosts:
    - '*.keptn.10.200.0.2.xip.io'
    secretName: sslcerts
```

The following rule needs to be added manually:
```yaml
  - host: bridge.keptn.10.200.0.2.xip.io
    http:
      paths:
      - backend:
          serviceName: bridge
          servicePort: 8080
```

This is what the end-result should look like:

```yaml
spec:
  rules:
  - host: api.keptn.10.200.0.2.xip.io
    http:
      paths:
      - backend:
          serviceName: api-gateway-nginx
          servicePort: 80
  - host: bridge.keptn.10.200.0.2.xip.io
    http:
      paths:
      - backend:
          serviceName: bridge
          servicePort: 8080
  tls:
  - hosts:
    - '*.keptn.10.200.0.2.xip.io'
    secretName: sslcerts
```

After this I was able to access my Keptn Bridge using https://bridge.keptn.10.200.0.2.xip.io:31741
(Please note: the prot 31741 might be different for your installation; please verify the port using `kubectl get cm -n keptn keptn-domain -ojsonpath={.data.app_domain}` )

bacherfl:
Closes #1830 

The api-service will now check if the keptn-domain ConfigMap contains a Port in the baseURL. To avoid generating invalid Ingress/VirtualService rules, the port is excluded from the host property of those rules.

### DIFF
--- a/api/handlers/configure.go
+++ b/api/handlers/configure.go
@@ -379,5 +379,10 @@ func removeBridgeFromIngress(ingress *networking.Ingress) {
 }
 
 func getHostForBridge(keptnDomain string) string {
+	// check if the domain contains a port. If yes, only the first part without the port will be used
+	split := strings.Split(keptnDomain, ":")
+	if len(split) > 1 {
+		keptnDomain = split[0]
+	}
 	return "bridge.keptn." + keptnDomain
 }
--- a/api/handlers/configure.go
+++ b/api/handlers/configure.go
@@ -381,8 +381,5 @@ func removeBridgeFromIngress(ingress *networking.Ingress) {
 func getHostForBridge(keptnDomain string) string {
 	// check if the domain contains a port. If yes, only the first part without the port will be used
 	split := strings.Split(keptnDomain, ":")
-	if len(split) > 1 {
-		keptnDomain = split[0]
-	}
-	return "bridge.keptn." + keptnDomain
+	return "bridge.keptn." + split[0]
 }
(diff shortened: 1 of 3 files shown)


api/handlers/configure.go
```diff
@@ -379,5 +379,7 @@ func removeBridgeFromIngress(ingress *networking.Ingress) {
 }
 
 func getHostForBridge(keptnDomain string) string {
-	return "bridge.keptn." + keptnDomain
+	// check if the domain contains a port. If yes, only the first part without the port will be used
+	split := strings.Split(keptnDomain, ":")
```
agrimmer, line 383: As far as I know, we have to differentiate between VirtualServices and Ingress:
- VirtualServices allow Ports
- Ingress does not allow Ports

agrimmer: Thank you for clarifying that also VirtualServices do not allow ports.

PR #1915, opened 2020-06-15
#1893 added username and password as required fields for CLI

### LINKED ISSUE #1893: Enforce username and password when exposing Keptn Bridge
When exposing the Keptn Bridge using `keptn configure bridge --action=expose`, the `username` is a required flag, `password` is optional.

 ```keptn configure bridge --action=expose --user=admin --password=asdf1234wxyz```

* If no `password` is provided, ask the user via stdin. 

* Executing the command a second time, overrides the secret (credentials). 

Additionally, move the step of manually creating a secret (see [here](https://keptn.sh/docs/0.6.0/reference/keptnsbridge/#enable-authentication)) to the API endpoint `/configure/bridge/expos` as implemented in: https://github.com/keptn/keptn/blob/374b4bc69748597c637d9b9ddf214460b394a2cb/api/handlers/configure.go#L93 

**Definition of Done**
- [ ] CLI command `configure bridge` is extended by the two flags `username` and `password` 
- [ ] Creating the secret is moved to API 
- [ ] Extend RBAC to just allow writing the secret

bacherfl:
Closes #1893 

The api-service now creates (or updates) a secret based on the credentials provided via the incoming request. When locking down the bridge, this secret will be deleted.

The CLI checks if a user and password have been provided if the bridge is being exposed.

Added RBAC rules:

```
  - apiGroups:
      - ""
    resources:
      - pods
    verbs:
      - get
      - delete
  - apiGroups:
      - ""
    resources:
      - secrets
    verbs:
      - create
  - apiGroups:
      - ""
    resources:
      - secrets
    verbs:
      - get
      - update
    resourceNames:
      - "bridge-credentials"
```


johannes-b:
The **Warning**: https://github.com/keptn/keptn/blob/5e942778db7f2e21eee36140106583b85bc4041f/cli/cmd/configure_bridge.go#L111 can be removed. Since basic auth is now enabled by default.

installer/manifests/keptn/rbac.yaml
```diff
@@ -78,6 +78,28 @@ metadata:
...
+    resources:
+      - secrets
+    verbs:
+      - create
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
```
agrimmer, line 99: Can the `create` be added here because this would allow us to limit the resources?

bacherfl: according to the k8s docs, `create` cannot be restricted by the `resourceName` attribute, because at the time of the creation the object is not available yet and hence gannot be verified (see https://github.com/kubernetes/kubernetes/issues/80295)

agrimmer: Thanks for the explanation.

PR #1956, opened 2020-06-25
close declined approvals

bacherfl:
This PR ensures that the gatekeeper service closes an open approval when it has been declined (previously, only approved approvals have been closed)


gatekeeper-service/pkg/handler/approval_finished_event_handler.go
```diff
@@ -93,6 +93,10 @@ func (a *ApprovalFinishedEventHandler) handleApprovalFinishedEvent(inputEvent ke
 		} else {
 			a.keptn.Logger.Info(fmt.Sprintf("Rejection for image %s for service %s of project %s and current stage %s received",
 				inputEvent.Image, inputEvent.Service, inputEvent.Project, inputEvent.Stage))
+			if err := closeOpenApproval(inputEvent, triggeredID); err != nil {
```
christian-kreuzberger-dtx, line 96: The code is duplicated from line 89-92 (if clause) to line 96-99 (else clause).

Any objections to move it below the if-else block so that it is executed in both cases? Or is there a case I am missing

PR #1959, opened 2020-06-26
Merging JMeter Extended Service into Keptn JMeter

grabnerandi:
The team has decided that we merge the JMeter Extended Service Implementation that was done in keptn-sandbox to become the implementation for the Keptn Core JMeter Service
The following changes contain that merge. Special Thanks goes to @bacherfl for helping with merging these code bases


bacherfl:
thank you for this contribution! I committed some minor changes in the dockerfile and deleted some files that are not needed anymore - the information in the release notes will be included in the release notes of the next Keptn version.

Issue #1794, opened 2020-05-14
Provide a “COPY JSON" button on the Bridge UI


robertjahn:
**Is your feature request related to a problem? Please describe.**
I often need to copy-and-paste the JSON from bridge and I need to carefully do this and can miss data.

**Describe the solution you'd like**
Perhaps a button on the popups or events details to just copy to clipboard.

**Describe alternatives you've considered**
None.  Manually do it.

**Additional context**

christian-kreuzberger-dtx:
This is the modal dialog: 
![Screenshot from 2020-05-18 08-47-02](https://user-images.githubusercontent.com/56065213/82182197-2b05fe80-98e4-11ea-9846-7c45a45f3194.png)

We could add a `copy to clipboard` button there.

sandroschmid:
@christian-kreuzberger-dtx @ermin-muratovic I would like to take care of this one

PR #1977, opened 2020-06-30
#1794: Provide a “COPY JSON" button on the Bridge UI

sandroschmid:
Issue #1794 

* Add a clipboard-utils service with a simple test
* Add a "copy"-button to the event payload dialog component (shown in the screenshot below)
* Format/cleanup code according to tslint-rules
* Remove unused method `getDuration`
* Remove redundant `OnInit`-implementation

image

### DIFF
--- a/bridge/client/app/app.module.ts
+++ b/bridge/client/app/app.module.ts
@@ -1,72 +1,74 @@
-import {NgModule} from '@angular/core';
-import {HttpClientModule} from '@angular/common/http';
-import {HTTP_INTERCEPTORS} from '@angular/common/http';
-import {BrowserModule} from '@angular/platform-browser';
-import {FlexLayoutModule} from "@angular/flex-layout";
-import {BrowserAnimationsModule} from '@angular/platform-browser/animations';
-import {MomentModule} from "ngx-moment";
-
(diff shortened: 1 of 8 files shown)


bridge/client/app/app.module.ts
```diff
@@ -1,72 +1,74 @@
-import {NgModule} from '@angular/core';
```
christian-kreuzberger-dtx, line 1: I guess this is the
> Format/cleanup code according to tslint-rules
stuff :+1:

It's a bit scary, as I think it will create conflicts when merging other PRs that are in progress. I'm pinging @ermin-muratovic on this, so he knows upfront and can later deal with the merge conflicts.

sandroschmid: true... it is also because I didn't switch of IntelliJ's automatic formatting. For the imports there are not tslint rules yet, so I could change my formatting settings without red lines reappearing 😄 

this PR is a lot smaller than Ermin's so it would be easier to revert here I guess

ermin-muratovic: Thank you for the cleanup. I merged it properly into the other PRs.

Issue #1989, opened 2020-07-02
Split project-board into smaller components


ermin-muratovic:
The project-board component has a lot of content in it and could be extracted into smaller components.

_Originally posted by @SaraDavilaMendez in https://github.com/keptn/keptn/pull/1877_

ermin-muratovic:
First of all the contents of the tabs should be separate view-components:
- #2939 Environments tab
- #2940 Services tab
- #2941 Sequences tab
- #2942 Integration tab

Additionally each tab has parts that can be extracted into components to make them easier maintainable and reusable:
- #2907 Stage timeline/selector component
- #2938 Task list with details
- #2943 Stage overview component
- #2944 Stage details component

Issue #1987, opened 2020-07-02
Use an HTTP-interceptor to add default headers and implement generic error handling


ermin-muratovic:
You might want to use an [HTTP-interceptor](https://angular.io/api/common/http/HttpInterceptor) to add default headers and implement generic error handling

_Originally posted by @sandroschmid in #1877_

PR #1993, opened 2020-07-02
Implement evaluation board to show Heatmap with deep link

ermin-muratovic:
The Bridge now provides a new view showing only the evaluation result as Heatmap/Chart.
With this view there are also to new deep links introduced:

`bridge.keptn.mykeptndomain/evaluation/{keptnContext}`: this shows all evaluations with the provided `keptnContext`

`bridge.keptn.mykeptndomain/evaluation/{keptnContext}/{stage}` or `bridge.keptn.mykeptndomain/evaluation/{keptnContext}/{eventId}`: this shows the evaluation with the provided `keptnContext `on the specified `stage `or with the specified `eventId`

image

### DIFF
--- a/bridge/client/app/evaluation-board/evaluation-board.component.html
+++ b/bridge/client/app/evaluation-board/evaluation-board.component.html
@@ -1,0 +1,48 @@
+<div class="container" *ngIf="error">
+  <dt-empty-state>
+    <dt-empty-state-item>
+      <dt-empty-state-item-img>
+        <img class="mt-2" src="/assets/keptn_logo.png" />
+      </dt-empty-state-item-img>
+      <ng-container [ngSwitch]="error">
+        <ng-container *ngSwitchCase="'contextError'">
+          <dt-empty-state-item-title class="mt-2" aria-level="2">Traces for <span class="italic" [textContent]="contextId"></span> not found</dt-empty-state-item-title>
+          <p>Sorry, we couldn't load any traces with that shkeptncontext. Check our <a href="https://keptn.sh/docs/quickstart/" target="_blank">Quick Start</a> instructions on how to <a href="https://keptn.sh/docs/0.6.0/manage/project/" target="_blank">setup a project</a> or go back to <a [routerLink]="['/dashboard']">your dashboard</a>.</p>
+        </ng-container>
+        <ng-container *ngSwitchDefault>
+          <dt-empty-state-item-title class="mt-2" aria-level="2">Some error occured</dt-empty-state-item-title>
+          <p>Sorry, some error occured. Check our <a href="https://keptn.sh/docs/quickstart/" target="_blank">Quick Start</a> instructions on how to <a href="https://keptn.sh/docs/0.6.0/manage/project/" target="_blank">setup a project</a> or go back to <a [routerLink]="['/dashboard']">your dashboard</a>.</p>
+        </ng-container>
+      </ng-container>
+    </dt-empty-state-item>
+  </dt-empty-state>
+</div>
+<div class="container" *ngIf="!error && !evaluations">
+  <div fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="15px">
+    <dt-loading-spinner></dt-loading-spinner>
+    <p>Loading ...</p>
+  </div>
+</div>
+<div class="evaluation-board pr-4 pl-4" fxLayout="column" *ngIf="evaluations">
+  <dt-info-group>
+    <dt-info-group-title>
+      <div fxFlex fxLayout="row" fxLayoutAlign="flex-start center" fxLayoutGap="15px">
+        <div fxLayout="row" fxLayoutAlign="start center">
+          <button dt-icon-button disabled variant="nested" *ngIf="root.isFaulty()"><dt-icon name="criticalevent"></dt-icon></button>
+          <p *ngIf="root.getShortImageName()" [textContent]="root.getShortImageName()"></p>
+          <p *ngIf="!root.getShortImageName()" [textContent]="root.getService()"></p>
+        </div>
+      </div>
+    </dt-info-group-title>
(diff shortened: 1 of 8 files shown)


bridge/client/app/evaluation-board/evaluation-board.component.ts
```diff
@@ -0,0 +1,74 @@
...
+    this._routeSubs = this.route.params.subscribe(params => {
+      if(params["shkeptncontext"]) {
+        this.contextId = params["shkeptncontext"];
+        this.apiService.getTraces(this.contextId)
+          .pipe(
+            map(response => response.body),
+            map(result => result.events||[]),
+            map(traces => traces.map(trace => Trace.fromJSON(trace)).sort((a, b) => new Date(a.time).getTime() - new Date(b.time).getTime()))
+          )
+          .subscribe((traces: Trace[]) => {
```
dominikmessner, line 46: Subscription is never cleaned up

bridge/client/app/evaluation-board/evaluation-board.component.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(EvaluationBoardComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create an instance', () => {
```
dominikmessner, line 29: We could speed up our pipeline by a few minutes by removing these default tests, you might want to remove them too.

christian-kreuzberger-dtx: I have no idea if there is value to those auto-generated "it should create an instance" tests. For this PR I would say it's fine to keep it as it is, but could we discuss this in https://github.com/keptn/keptn/issues/2000?

bridge/client/app/evaluation-board/evaluation-board.component.html
```diff
@@ -0,0 +1,48 @@
...
+        </ng-container>
+      </ng-container>
+    </dt-empty-state-item>
+  </dt-empty-state>
+</div>
+<div class="container" *ngIf="!error && !evaluations">
+  <div fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="15px">
+    <dt-loading-spinner></dt-loading-spinner>
+    <p>Loading ...</p>
+  </div>
```
sandroschmid, line 24: This probably could be useful as reusable component for other screens as well 🤔

bridge/client/app/evaluation-board/evaluation-board.component.scss
```diff
@@ -0,0 +1,6 @@
+@import "../../variables";
```
sandroschmid, line 1: you don't use any variables so you can remove the import

```suggestion

```

bridge/client/app/evaluation-board/evaluation-board.component.ts
sandroschmid: I've also mentioned this [here](https://github.com/keptn/keptn/pull/1877#discussion_r446008873) already, should be cleaned up in the entire bridge

bridge/client/app/evaluation-board/evaluation-board.component.ts
```diff
@@ -0,0 +1,74 @@
...
+            if(traces.length > 0) {
+              console.log("traces", traces);
```
sandroschmid, line 48: that was for local testing I guess?

```suggestion

```

bridge/client/app/evaluation-board/evaluation-board.component.ts
```diff
@@ -0,0 +1,74 @@
...
+              this.root = Root.fromJSON(traces[0]);
+              this.root.traces = traces;
+              this.evaluations = traces.filter(t => t.type == 'sh.keptn.events.evaluation-done' && (!params["eventselector"] || t.id == params["eventselector"] || t.data.stage == params["eventselector"])) ;
```
sandroschmid, line 51: As soon as #1877 is merged, you should use `EventTypes.EVALUATION_DONE`

bridge/client/app/evaluation-board/evaluation-board.component.ts
ermin-muratovic: I created an `unsubscribe$` Subject and use `.pipe(takeUntil(this.unsubscribe$))` before each subscribe.

bridge/client/app/evaluation-board/evaluation-board.component.html
ermin-muratovic: Yes, since the same template is also used in the `project-board`. There is already an issue to split the `project-board` into smaller components: #1989

bridge/client/app/evaluation-board/evaluation-board.component.ts
```diff
@@ -0,0 +1,71 @@
...
+            .subscribe((traces: Trace[]) => {
+              if(traces.length > 0) {
+                this.root = Root.fromJSON(traces[0]);
+                this.root.traces = traces;
+                this.evaluations = traces.filter(t => t.type == 'sh.keptn.events.evaluation-done' && (!params["eventselector"] || t.id == params["eventselector"] || t.data.stage == params["eventselector"])) ;
+              } else {
+                this.error = "contextError";
+              }
+            }, (err) => {
+              this.error = "error";
```
SaraDavilaMendez, line 52: This error messages do not help much about what the problem is.
For example: in the context error you could add what is the given context and in the second one you could add more information about what went wrong.

ermin-muratovic: Actually, the contextError says, that there couldn't be found any traces with that `contextId`.

https://github.com/keptn/keptn/blob/ba5bc699920625e73818b36fd3ac653aec186f32/bridge/client/app/evaluation-board/evaluation-board.component.html#L8-L11

And the "error"-case might only be in case of a error in the backend, and for this we have already a separate issue #1987

PR #1995, opened 2020-07-02
Show categories on xAxis in BarChart instead of timeline

ermin-muratovic:
The bar chart for evaluation results now shows categories on the xAxis instead of timeline

chart-labels-time

If the event has labels and a label with the key `buildId` is provided, that label will be used as category. Otherwise the time will be used as category.

chart-labels

### DIFF
--- a/bridge/client/app/_models/trace.ts
+++ b/bridge/client/app/_models/trace.ts
@@ -170,6 +170,15 @@ class Trace {
     return this.data.service;
   }
 
+  getChartLabel(): string {
+    let label;
+    if(this.data.labels)
+      label = this.data.labels.get("buildId");
+    if(!label)
+      label = this.time;
+    return label;
+  }
+
   static fromJSON(data: any) {
     return Object.assign(new this, data, { plainEvent: JSON.parse(JSON.stringify(data)) });
   }
--- a/bridge/client/app/_models/trace.ts
+++ b/bridge/client/app/_models/trace.ts
@@ -172,11 +172,6 @@ class Trace {
 
   getChartLabel(): string {
     return this.data.labels?.get("buildId") ?? this.time;
-    if(this.data.labels)
-      label = this.data.labels.get("buildId");
-    if(!label)
-      label = this.time;
-    return label;
   }
 
   static fromJSON(data: any) {
--- a/bridge/client/app/_models/trace.ts
+++ b/bridge/client/app/_models/trace.ts
@@ -171,7 +171,7 @@ class Trace {
   }
 
   getChartLabel(): string {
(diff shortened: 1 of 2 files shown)


bridge/client/app/_models/trace.ts
```diff
@@ -170,6 +170,15 @@ class Trace {
     return this.data.service;
   }
 
+  getChartLabel(): string {
+    let label;
```
dominikmessner, line 174: [minor] Could be written a bit shorter with Typescript 3.7
```suggestion
    return this.data.labels?.get("buildId") ?? this.time;
```

ermin-muratovic: It might happen, that there are labels defined, but without `buildId`. Is this also working in that case?

dominikmessner: If the result on the left side is `null` or `undefined`, it will fall back to `this.time`. If you additionally want to check for empty strings (falsy values) you can use `||` instead of `??`.

bridge/client/app/_models/trace.ts
```diff
@@ -170,6 +170,15 @@ class Trace {
     return this.data.service;
   }
 
+  getChartLabel(): string {
+    return this.data.labels?.get("buildId") ?? this.time;
+    if(this.data.labels)
+      label = this.data.labels.get("buildId");
+    if(!label)
+      label = this.time;
+    return label;
```
dominikmessner, line 179:
```suggestion
```
I'm sorry, this should also have been replaced, not just the single line, I still need to get used to Github. :smiley:

bridge/client/app/_models/trace.ts
```diff
@@ -170,6 +170,15 @@ class Trace {
     return this.data.service;
   }
 
+  getChartLabel(): string {
+    return this.data.labels?.get("buildId") ?? this.time;
+    if(this.data.labels)
```
sandroschmid, line 175: this is never called after the `return` in the line above, is it?

dominikmessner: Yes, I messed up the suggestion in https://github.com/keptn/keptn/pull/1995#discussion_r449120486. It should have replaced the whole method body but looks like I didn't select it correctly.

bridge/client/app/_models/trace.ts
ermin-muratovic: Of course, I removed the unused code.

bridge/client/app/_models/trace.ts
ermin-muratovic: Sounds greate, thx.

PR #2008, opened 2020-07-06
#1941 check ingress configuration

### LINKED ISSUE #1941: Keptn generate support-archive should have a separate check for ingress options
When troubleshooting Keptn installations, it is necessary to have some dedicated information about the following:

* is istio-system installed
* is public-gateway.istio-system (or what is defined in Issue #1986  ) available (which is hardwired in our [virtualservices](https://github.com/christian-kreuzberger-dtx/ck-sockshop/blob/8437064291eb556d25a416752a5649e49851762e/carts/helm/carts-generated/templates/carts-istio-virtualservice.yaml#L8for))
* List any services with type and internal/external IP in the istio-system namespace

* is any other ingress option available (e.g., ingress-classes)? (need to research if this is actually possible)
* What's the external and internal IP of the cluster node(s)
* In the whole cluster, is any service of type LoadBalancer available?

bacherfl:
Closes #1941 

This PR adds ingress configuration information to the generated support-archive

### DIFF
--- a/cli/cmd/generate_support_archive.go
+++ b/cli/cmd/generate_support_archive.go
@@ -214,7 +214,7 @@ func isConfiguredIngressGatewayAvailable() *errorableBoolResult {
 	if len(gwSplit) > 1 {
 		gatewayNamespace = gwSplit[1]
 	} else {
-		gatewayNamespace = "istio-system"
+		gatewayNamespace = "keptn"
 	}
 
 	res, err = exechelper.ExecuteCommand("kubectl", "get gateway -n "+gatewayNamespace+" "+gatewayName)
@@ -225,7 +225,7 @@ func isConfiguredIngressGatewayAvailable() *errorableBoolResult {
 }
 
 func writeIstioIngressGateways(dir string) {
-	writeErrorableStringResult(newErrorableStringResult(exechelper.ExecuteCommand("kubectl", "get gateway --all-namespaces -owide")),
+	writeErrorableStringResult(newErrorableStringResult(exechelper.ExecuteCommand("kubectl", "get gateway --all-namespaces -oyaml")),
 		filepath.Join(dir, "ingress-gateways.txt"))
 }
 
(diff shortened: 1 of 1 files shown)


cli/cmd/generate_support_archive.go
```diff
@@ -156,6 +195,75 @@ keptn generate support-archive --dir=/some/directory`,
...
+	}
+
+	var gatewayName, gatewayNamespace string
+	gwSplit := strings.Split(gatewayFullName, ".")
+
+	gatewayName = gwSplit[0]
+	if len(gwSplit) > 1 {
+		gatewayNamespace = gwSplit[1]
+	} else {
+		gatewayNamespace = "istio-system"
```
agrimmer, line 217: If no namespace is specified, the gateway is located in the `keptn` namespace.

cli/cmd/generate_support_archive.go
```diff
@@ -156,6 +195,75 @@ keptn generate support-archive --dir=/some/directory`,
...
+	}
+
+	res, err = exechelper.ExecuteCommand("kubectl", "get gateway -n "+gatewayNamespace+" "+gatewayName)
```
agrimmer, line 220: Can we here please add `-oyaml` or `-ojson` in order to get all details of the gateway?

cli/cmd/generate_support_archive.go
bacherfl: good catch, I have just corrected that

cli/cmd/generate_support_archive.go
bacherfl: Detailed information about all available gateways is provided in the output generated by 	https://github.com/keptn/keptn/blob/dc66510648c7f73071aff7a4630978c525677855/cli/cmd/generate_support_archive.go#L228

agrimmer: Thanks!

cli/cmd/generate_support_archive.go
agrimmer: Thanks!

PR #2009, opened 2020-07-06
#1986 added ingress gateway env var

### LINKED ISSUE #1986: Helm-Service: Make Gateway in generated VirtualServices configurable via Env-Variable
When generating the VirtualServices, the `helm-services` assumes that there is a gateway called `public-gateway` in the `istio-system` namespace available. This gateway is used when automatically generating VirtualService, e.g.:
```
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  creationTimestamp: null
  name: carts
spec:
  gateways:
  - public-gateway.istio-system
  - mesh
  hosts:
  - carts.sockshop-dev.demo.keptn.sh
  - carts
  http:
  - route:
    - destination:
        host: carts.sockshop-dev.svc.cluster.local
```

But by implementing [KEP 18](https://github.com/keptn/enhancement-proposals/pull/18/files), Keptn will not install Istio and no `public-gateway` anymore. Hence, we have to make the used gateway configurable.

**Tasks:**
- Introduce and use an environment for the used gateway.
- Check whether the gateway installed in an Istio demo-installation can be used as default for the gateway
- Describe in the Docs how to set the gateway

**Definition of Done:**
- The name and namespace of the used gateway can be configured via an environment variable.

bacherfl:
Closes #1986 

This PR makes the name of the ingress gateway configurable in the helm-service. To set the value for the gateway, the `ingress-config` ConfigMap is used. This ConfigMap now has the following format:

```
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: ingress-config
  namespace: keptn
data:
  ingress_hostname_suffix: ""
  ingress_protocol: ""
  ingress_port: ""
  istio_gateway: ""

```

### DIFF
--- a/helm-service/deploy/service.yaml
+++ b/helm-service/deploy/service.yaml
@@ -52,11 +52,11 @@ spec:
             configMapKeyRef:
               name: ingress-config
               key: ingress_port
-        - name: INGRESS_GATEWAY
+        - name: ISTIO_GATEWAY
           valueFrom:
             configMapKeyRef:
               name: ingress-config
-              key: ingress_gateway
+              key: istio_gateway
 ---
 apiVersion: v1
 kind: Service
--- a/helm-service/deploy/service.yaml
+++ b/helm-service/deploy/service.yaml
@@ -52,6 +52,11 @@ spec:
             configMapKeyRef:
               name: ingress-config
               key: ingress_port
+        - name: INGRESS_GATEWAY
+          valueFrom:
+            configMapKeyRef:
+              name: ingress-config
+              key: ingress_gateway
 ---
 apiVersion: v1
 kind: Service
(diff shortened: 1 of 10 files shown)


helm-service/deploy/service.yaml
```diff
@@ -52,6 +52,11 @@ spec:
             configMapKeyRef:
               name: ingress-config
               key: ingress_port
+        - name: INGRESS_GATEWAY
```
agrimmer, line 55: How much work would it be to rename `INGRESS_GATEWAY` into `ISTIO_GATEWAY`? Because this would follow the Istio naming https://istio.io/latest/docs/reference/config/networking/gateway/

PR #2026, opened 2020-07-09
Update UI look and feel

ermin-muratovic:
- The background color for the menu and dialog changed to gray.
- Removed the shadow from the header
- Failed root events are more visible (red border left and stage is marked red with background instead of border)
- Fixed the path to Fonts (uses now OpenSans correctly)

image

### DIFF
--- a/bridge/client/app/_components/ktb-root-events-list/ktb-root-events-list.component.scss
+++ b/bridge/client/app/_components/ktb-root-events-list/ktb-root-events-list.component.scss
@@ -1,26 +1,41 @@
 @import '../../../variables';
 @import '~@dynatrace/barista-components/core/src/style/variables';
 
-.dt-tag.error {
-  border-color: $error-color;
-}
-.dt-tag.warning {
-  border-color: $orange-400;
-}
+.ktb-root-events-list {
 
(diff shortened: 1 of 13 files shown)


bridge/client/app/_components/ktb-root-events-list/ktb-root-events-list.component.scss
```diff
@@ -1,26 +1,41 @@
...
+  .event-icon.warning {
+    fill: $orange-400;
+  }
+
+  .event-icon.success {
+    fill: $cta-color-active;
+  }
+
+  .ktb-selectable-tile.ktb-tile-error .ktb-selectable-tile-container {
+    border: 1px solid $gray-200;
```
sandroschmid, line 32: is this purposely `$gray-200` instead of `$gray-300` like in line 38? if they should have the same color you could use the same style for all three

ermin-muratovic: That is on purpose. Since I want only the left border to be red. By default the `.ktb-selectable-tile.ktb-tile-error` has a red border all around.

PR #2030, opened 2020-07-10
Run distributors as sidecar, correct URLs/namespaces

agrimmer:


agrimmer:
Implements https://github.com/keptn/keptn/issues/2031

PR #2048, opened 2020-07-14
#1965 adapted integration tests to changes made in KEP 18

### LINKED ISSUE #1965: Fix all integration tests that broke after KEP 18
Details of KEP 18: https://github.com/keptn/enhancement-proposals/pull/18/files

We are aware that KEP 18 will break many many integration tests.

After we have finished implementing KEP 18 and also documenting those steps, we need to fix the integration tests.

# Definition of Done

- [x] All integration tests work again
- [x] New integration test variants added or updated, such that API is exposed for GKE (using Loadbalancer), K3s (using built-in traefik) and minishift (using oc route)

bacherfl:
Closes #1965


test/utils/expose_keptn_bridge.sh
```diff
@@ -1,14 +1,16 @@
...
 echo "Exposing Keptn Bridge..."
+kubectl -n keptn delete secret bridge-credentials
 # expose Keptn's Bridge (for easier troubleshooting/debugging afterwards)
 kubectl -n keptn create secret generic bridge-credentials --from-literal="BASIC_AUTH_USERNAME=${NIGHTLY_BRIDGE_USERNAME}" --from-literal="BASIC_AUTH_PASSWORD=${NIGHTLY_BRIDGE_PASSWORD}"
-keptn configure bridge --action=expose
 # restart bridge pod to make use of the secret
 kubectl -n keptn delete pods --selector=run=bridge
 sleep 5
+kubectl patch svc bridge -n keptn -p '{"spec": {"type": "LoadBalancer"}}'
+sleep 10
```
agrimmer, line 11: In this test, we could re-use the already existing generated `bridge-credentials` secret. Furthermore, the service of the bridge does not require a LoadBalancer.

test/utils/expose_keptn_bridge.sh
```diff
@@ -1,14 +1,16 @@
 #!/bin/bash
```
agrimmer, line 1: Can we please rename this file in `check_bridge_availability.sh` or something similar?

test/assets/quality_gates_standalone_slo_step1.yaml
```diff
@@ -21,8 +21,8 @@ objectives:
   - sli: "throughput"
     pass:
       - criteria:
-          - "<=+20%"
-          - ">=-20%"
+          - "<=+100%"
```
agrimmer, line 24: Why have these criteria been increased?

test/test_install_gke.sh
```diff
@@ -2,42 +2,58 @@
...
 # Prepare creds.json file
-cd ./installer/scripts
+cd ./test/assets
 
 export CLN=$CLUSTER_NAME_NIGHTLY
 export CLZ=$CLOUDSDK_COMPUTE_ZONE	
 export PROJ=$PROJECT_NAME
 
-source ./gke/defineCredentialsHelper.sh
+source ../utils/gke/defineCredentialsHelper.sh
```
agrimmer, line 15: Actually, the creds.json file could be empty for platform `kubernetes`.

test/assets/quality_gates_standalone_slo_step1.yaml
bacherfl: they were too restrictive - sometimes the tests were failing simple because the throughput of a test run was a little bit (i.e. more than 20% higher) than in a previous test run - in my opinion, this should not lead to a failure because the goal of the test is only to verify that the mechanics of retrieving SLIs and evaluating them are working

PR #2058, opened 2020-07-16
Cleanup keptn-domain leftovers

agrimmer:


johannes-b:
Can you please provide an example with the `namespace` flag here: https://github.com/keptn/keptn/blob/dadc861c1be885eefa01a774a82e021bb8f3cf87/cli/cmd/generate_support_archive.go#L81

PR #2059, opened 2020-07-16
Add /api to endpoint for configure bridge

johannes-b:
Since we don't support the backward compatibility of the API, the endpoint for `keptn configure bridge` was broken.

### DIFF
--- a/cli/cmd/configure_bridge.go
+++ b/cli/cmd/configure_bridge.go
@@ -6,10 +6,11 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/keptn/keptn/cli/pkg/credentialmanager"
-	"github.com/spf13/cobra"
 	"io/ioutil"
 	"net/http"
+
+	"github.com/keptn/keptn/cli/pkg/credentialmanager"
+	"github.com/spf13/cobra"
 
 	keptnutils "github.com/keptn/go-utils/pkg/api/utils"
 )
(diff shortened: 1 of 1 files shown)


cli/cmd/configure_bridge.go
```diff
@@ -81,14 +80,14 @@ func retrieveBridgeCredentials(endpoint string, apiToken string) (*configureBrid
 
 	resp, err := client.Do(req)
 	if err != nil {
-		fmt.Println("Could not complete command: " + err.Error())
+		fmt.Println("Could not sent request: " + err.Error())
```
agrimmer, line 83: Correct wording: `send`.

cli/cmd/configure_bridge.go
```diff
@@ -117,20 +116,21 @@ func configureBridgeCredentials(endpoint string, apiToken string, configureBridg
...
 	}
+
 	req, err := http.NewRequest("POST", endpoint, bytes.NewReader([]byte(payload)))
 	req.Header.Add("x-token", apiToken)
 	req.Header.Add("content-type", "application/json")
 
 	resp, err := client.Do(req)
 	if err != nil {
-		fmt.Println("Could not complete command: " + err.Error())
+		fmt.Println("Could not sent request: " + err.Error())
```
agrimmer, line 126: Correct tense: `send`

cli/cmd/configure_bridge.go
johannes-b: fix it.

cli/cmd/configure_bridge.go
johannes-b: fix it.

PR #2070, opened 2020-07-18
#1944 Immediately return error if kube server version check error

### LINKED ISSUE #1944: CLI: kubectl version check incorrectly reports if no connection to cluster could be made
Source: https://keptn.slack.com/archives/CNRCGFU3U/p1592334110459600

In 0.6.2, the version check reports that the Kubernetes Version is invalid in the case that no connection to the Kubernetes server could be made:

```
$ kubectl get nodes
Unable to connect to the server: net/http: TLS handshake timeout 
```

However, when the user tries to install, they only see this:

```
$ keptn install ...
Used Installer version: docker.io/keptn/installer:0.6.2
Error: Keptn requires Kubernetes Server Version: >= 1.13, <= 1.15
```

---

The relevant part of the code is here:
https://github.com/keptn/keptn/blob/5e444816db7626d51ef588e874a23df1b387dafa/cli/cmd/install.go#L201-L207

https://github.com/keptn/keptn/blob/5e444816db7626d51ef588e874a23df1b387dafa/cli/pkg/kube/kube_utils.go#L43-L46

https://github.com/keptn/keptn/blob/5e444816db7626d51ef588e874a23df1b387dafa/cli/pkg/kube/kube_utils.go#L29-L32

An easy fix would be to check the error value of `CheckKubeServerVersion`. If it returns an error, we need to return that error (right now we create a new error object and only print the error when verbose logging is active).

imrenagi:
Resolve #1944

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -197,7 +197,7 @@ Please see https://kubernetes.io/docs/tasks/tools/install-kubectl/`)
 				if err := kube.CheckKubeServerVersion(KubeServerVersionConstraints); err != nil {
 					logging.PrintLog(err.Error(), logging.VerboseLevel)
 					logging.PrintLog("See https://keptn.sh/docs/"+keptnReleaseDocsURL+"/installation/k8s-support/ for details.", logging.VerboseLevel)
-					return errors.New(`Keptn requires Kubernetes server version: ` + KubeServerVersionConstraints)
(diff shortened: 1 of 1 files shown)


imrenagi:
hi @johannes-b @christian-kreuzberger-dtx can you please take a look to this PR? thanks!

agrimmer:
Can we please merge this PR after releasing Keptn 0.7.0?

imrenagi:
absolutely yes.

Issue #1655, opened 2020-04-20
API: Get Event returns HTTP 500 when event cant be found. Should be 404


grabnerandi:
When waiting for an evaluation-done event a client has to call GET /event with the keptnContext and evaluation-done as event type. When the event does not yet exists the Keptn API returns an HTTP 500 (Internal Server Error) instead of something that is more descriptive, e.g: 404 Resource Not Found.

HTTP 500 is definitely misleading as this is not an internal server error. It is really about an event not being present - so - a more suitable HTTP Status Code should be used, e.g: 404

imrenagi:
I will take this

johannes-b:
The /GET event endpoint still returns a 500 - Internal Server error, because the [GetEvent](https://github.com/keptn/keptn/blob/master/api/handlers/event.go#L116) function returns an error which leads to the 500 error. 

**Task:**
- Change the [getLatestEvent](https://github.com/keptn/go-utils/blob/master/pkg/lib/datastore.go#L75) to return a 404 when no event with the given keptnContext and type is available
- Catch a 404 in [GetEvent](https://github.com/keptn/keptn/blob/master/api/handlers/event.go#L116) and respond it.

**Definition of Done:**
- The /GET event endpoint returns 404 (resource not found) when the event from the given keptnContext and type is not available.

PR #2114, opened 2020-07-23
Return 404 when no event is found in /event endpoint

imrenagi:
Resolving #1655

Signed-off-by: Imre Nagi <imre.nagi2812@gmail.com>

### DIFF
--- a/api/handlers/event.go
+++ b/api/handlers/event.go
@@ -118,7 +118,7 @@ func GetEventHandlerFunc(params event.GetEventParams, principal *models.Principa
 	}
 
 	if cloudEvent == nil {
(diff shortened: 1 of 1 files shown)


imrenagi:
@bacherfl can you please check this as well? I found out that api docs says this `/event` endpoint is deprecated and moved to `/mongodb-datastore/v1/event`. Do i also have to update this endpoint?

agrimmer:
I will test the PR tomorrow.

agrimmer:
lgtm
But we need to leave the issue https://github.com/keptn/keptn/issues/1655 open because we also need to address this in the `mongodb-datastore`.

Issue #2012, opened 2020-07-07
Unsubscribe from all reading requests


ermin-muratovic:
Each subscription should also be unsubscribed.

For reading requests you could implement `OnDestroy`, add a `private readonly destroy$ = new Subject<void>();`, emit this in `ngOnDestroy` like so `this.destroy$.next();` and use `.pipe(takeUntil(this.destroy$))` everywhere where you subscribe (should be done for all services and components in order to prevent memory leaks)

_Originally posted by @sandroschmid in https://github.com/keptn/keptn/timeline_

Issue #2011, opened 2020-07-07
Create helper function (utility) for date comparison


ermin-muratovic:
the date comparison is needed numerous times, maybe you can use a helper function (but not important now I would say)

_Originally posted by @sandroschmid in https://github.com/keptn/keptn/pull/1877_

Issue #2000, opened 2020-07-03
Remove auto-generated Angular "it should create an instance" tests


christian-kreuzberger-dtx:
As mentioned by @dominikmessner in PR #1993, we could remove the auto generated "it should create an instance" tests:
https://github.com/keptn/keptn/pull/1993/files/8b3a047f610b7fec94f2806189f58453cf0efafa#r449132239
![Screenshot from 2020-07-03 08-02-53](https://user-images.githubusercontent.com/56065213/86436921-a3dde000-bd03-11ea-80ad-d56fffee4826.png)

As this affects not only the PR, but a whole lot of lines of code in our Angular app, I'd like this to be done in a separate issue. But before we go this way, can anyone provide some more details whether these auto generated tests have any value and how big the impact on removing them would be?

dominikmessner:
In our internal project, we removed all these tests and could speed up the test execution by 10%. I quickly had a look at a few Angular projects, the ones that I looked at don't have these default tests:
- [Angular](https://github.com/angular/angular/search?q=%22should+create+an+instance%22&unscoped_q=%22should+create+an+instance%22)
- [Material](https://github.com/angular/components/search?q=%22should+create+an+instance%22&unscoped_q=%22should+create+an+instance%22)
- [Barista](https://github.com/dynatrace-oss/barista/search?q=%22should+create+an+instance%22&unscoped_q=%22should+create+an+instance%22)

I think as soon as there is a proper test for a component, the instance creation will be tested implicitly anyways, that's why it's probably ok to remove them.

PR #2116, opened 2020-07-24
#2000 #2011 #2012 | Bridge code refactoring

ermin-muratovic:
- Unsubscribe from all reading requests
- Create helper function (utility) for date comparison
- Remove auto-generated Angular "it should create an instance" tests


bridge/client/app/dashboard/dashboard.component.ts
```diff
@@ -12,29 +13,31 @@ import {DataService} from "../_services/data.service";
 })
 export class DashboardComponent implements OnInit, OnDestroy {
 
+  private readonly unsubscribe$ = new Subject<void>();
+
   public projects$: Observable<Project[]>;
```
samuelfahrngruber, line 18: You could already set `projects$` here to `this.dataService.projects`.
by using the rxjs operator `catchError` you could then just return an empty array observable using `of([])` in erroneous cases.
Then, subscribing using the `| async` pipe in the template will always give you an empty array on error.
You won't have to manually subscribe in the component anymore, which means you don't need `unsubscribe$`.
However, just a suggestion, this works perfectly fine too :)

bridge/client/app/_utils/date.utils.ts
```diff
@@ -41,4 +42,8 @@ export default class DateUtil {
       sameElse : 'YYYY-MM-DD HH:mm'
     };
   }
+
+  static compareTime(a: Trace, b: Trace) {
```
samuelfahrngruber, line 46: [minor suggestion] do you think we should change the name to `compareTraceTimes` or add some documentation that it's clear that not actual timestamps are compared but Trace's `time` properties?

bridge/client/app/dashboard/dashboard.component.ts
samuelfahrngruber: generally prefer using `async pipe` above manual subscriptions when possible

ermin-muratovic: You are absolutely right, for this use case I can get rid of the manual subscription and provide an empty array as result in erroneous case. Fixed it.
I also log and show errors now in the http-error-interceptor and removed all error handling from the service.

bridge/client/app/_utils/date.utils.ts
ermin-muratovic: Yeah, makes sense to rename the function to `compareTraceTimes`. Fixed it.

bridge/client/app/_services/data.service.ts
```diff
@@ -153,7 +154,7 @@ export class DataService {
         map(roots => roots.map(root => Root.fromJSON(root)))
       )
       .subscribe((roots: Root[]) => {
-        service.roots = [...roots||[], ...service.roots||[]].sort((a, b) => new Date(b.time).getTime() - new Date(a.time).getTime());
+        service.roots = [...roots||[], ...service.roots||[]].sort((a, b) => DateUtil.compareTraceTimes(a, b));
```
sandroschmid, line 157:
```suggestion
        service.roots = [...roots||[], ...service.roots||[]].sort(DateUtil.compareTraceTimes);
```

could make this even shorter when using no wrapper function

bridge/client/app/_services/data.service.ts
```diff
@@ -209,7 +210,7 @@ export class DataService {
     if(root.traces.length > 0) {
       this._openApprovals.next(this._openApprovals.getValue().filter(approval => root.traces.indexOf(approval) < 0));
       if(root.traces[root.traces.length-1].type == EventTypes.APPROVAL_TRIGGERED)
-        this._openApprovals.next([...this._openApprovals.getValue(), root.traces[root.traces.length-1]].sort((a, b) => new Date(b.time).getTime() - new Date(a.time).getTime()));
+        this._openApprovals.next([...this._openApprovals.getValue(), root.traces[root.traces.length-1]].sort((a, b) => DateUtil.compareTraceTimes(a, b)));
```
sandroschmid, line 213:
```suggestion
        this._openApprovals.next([...this._openApprovals.getValue(), root.traces[root.traces.length-1]].sort(DateUtil.compareTraceTimes));
```

same here

bridge/client/app/_services/mock-data.service.ts
```diff
@@ -88,7 +89,7 @@ export class MockDataService {
   }
 
   public loadRoots(project: Project, service: Service) {
-    service.roots = [...this.mockRoots[project.projectName+"."+service.serviceName]||[], ...service.roots||[]].sort((a, b) => new Date(b.time).getTime() - new Date(a.time).getTime());
+    service.roots = [...this.mockRoots[project.projectName+"."+service.serviceName]||[], ...service.roots||[]].sort((a, b) => DateUtil.compareTraceTimes(a, b));
```
sandroschmid, line 92:
```suggestion
    service.roots = [...this.mockRoots[project.projectName+"."+service.serviceName]||[], ...service.roots||[]].sort(DateUtil.compareTraceTimes);
```

same here 🙂

bridge/client/app/dashboard/dashboard.component.ts
```diff
@@ -12,29 +13,24 @@ import {DataService} from "../_services/data.service";
 })
 export class DashboardComponent implements OnInit, OnDestroy {
 
+  private readonly unsubscribe$ = new Subject<void>();
```
sandroschmid, line 16:
```suggestion
```
not used to unsubscribe anything -> can be removed

bridge/client/app/dashboard/dashboard.component.ts
```diff
@@ -12,29 +13,24 @@ import {DataService} from "../_services/data.service";
...
+
   public projects$: Observable<Project[]>;
   public error: boolean = false;
```
sandroschmid, line 19: you never set this to true, therefore I guess it can be removed too?

bridge/client/app/dashboard/dashboard.component.ts
```diff
@@ -12,29 +13,24 @@ import {DataService} from "../_services/data.service";
...
-      this.error = false;
-    }, error => {
-      this.error = true;
-    });
   }
 
-  ngOnDestroy(): void {
-    this._projectsSubs.unsubscribe();
+  ngOnInit() {
   }
```
sandroschmid, line 26: empty implementation of `ngOnInit` can be removed

bridge/client/app/_components/ktb-http-loading-bar/ktb-http-loading-bar.component.ts
```diff
@@ -1,13 +1,17 @@
...
+import {Subject} from "rxjs";
+import {takeUntil} from "rxjs/operators";
 
 @Component({
   selector: 'ktb-http-loading-bar',
   templateUrl: './ktb-http-loading-bar.component.html',
   styleUrls: ['./ktb-http-loading-bar.component.scss']
 })
-export class KtbHttpLoadingBarComponent implements OnInit {
+export class KtbHttpLoadingBarComponent implements OnInit, OnDestroy {
```
SaraDavilaMendez, line 12: This class and KtbHttpLoadingSpinnerComponent have the loading flag, the show and hide funcitons and the subscribe to the httpState in common, may be extracting this parts to a general HttpLoadingState component could be an option to reduce duplicates.

bridge/client/app/_utils/search.utils.ts
```diff
@@ -0,0 +1,22 @@
...
+        m = k + 1;
+      } else if(cmp < 0) {
+        n = k - 1;
+      } else {
+        return k;
+      }
+    }
+    return -m - 1;
+  }
+}
```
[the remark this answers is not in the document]
ermin-muratovic: I changed the variable names to proper named names and added some comments and even fixed a bug. thx for your comment and suggestions

bridge/client/app/app-header/app-header.component.ts
```diff
@@ -33,37 +32,41 @@ export class AppHeaderComponent implements OnInit, OnDestroy {
...
+          } else {
+            versionInfo.keptnVersionInvalid = true;
+          }
+          if(semver.valid(versionInfo.bridgeVersion)) {
+            if(versionInfo.availableVersions.bridge)
+              this.doVersionCheck(versionInfo.bridgeVersion, versionInfo.availableVersions.bridge.stable, versionInfo.availableVersions.bridge.prerelease, "Keptn Bridge");
+          } else {
+            versionInfo.bridgeVersionInvalid = true;
+          }
         }
```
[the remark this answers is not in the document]
ermin-muratovic: yep, actually the check if the version is semver.valid could be done inside `doVersionCheck` and that function could return `false` in case it is invalid.

bridge/client/app/_components/ktb-http-loading-bar/ktb-http-loading-bar.component.ts
ermin-muratovic: Good point.
I actually removed the `KtbHttpLoadingSpinnerComponent` since it was never used. And actually I would implement content projection to `KtbHttpLoadingBarComponent` and then the developer could choose to either add a loading spinner, loading bar or show any other component while something is loading.

Issue #1735, opened 2020-05-04
keptn add resource flag checking


Dynatrace-Adam-Gardner:
**Suggestion:** Better parameter/flag checking of the `keptn add-resource` command.

Including a space in the flags results in a failed outcome (resource isn't added) but the output suggests it is successfully added.

Example (note the space between `--` and `resourceUri`)
```
keptn add-resource --project=website --service=front-end --stage=production --resource=remediations.yaml -- resourceUri=remediation.yaml
```
Output:
```
$ Adding resource remediations.yaml to service front-end in stage production in project website
$ Resource has been uploaded.
```

My `remediation-service` fails to find the file when executing:
```
{"timestamp":"...","logLevel":"ERROR","message":"Failed to get remediation.yaml file"}
{"level":"warn","ts":1588565664.9034595,"logger":"fallback","caller":"http/transport.go:337","msg":"got an error from receiver fn: %s","error":"{\"code\":404,\"message\":\"Service resource not found\"}\n"}
```

As soon as I fix the command (i.e., remove the space), it works perfectly.

https://github.com/keptn/keptn/blob/d118d835287ada4d154abb28961a17c0545e6401/cli/cmd/add_resource.go#L28

**Task:**
- Check the number of arguments to verify that the command can be successfully executed. 
- If `len(args) > 0` -> then abort this command
- Add this check here: https://github.com/keptn/keptn/blob/master/cli/cmd/add_resource.go similar to the implementation here: https://github.com/keptn/keptn/blob/master/cli/cmd/create_project.go#L61

**Definition of Done:**
- [ ] When executing the command  `keptn add-resource` with `-- resourceUri`, the user gets an info message as shown below:
```
keptn add-resource --project=website --service=front-end --stage=production --resource=remediations.yaml -- resourceUri=remediation.yaml
This command does not support any arguments
```

christian-kreuzberger-dtx:
Hi Adam,

when using `keptn add-resource`, the flag `--resource=remediations.yaml` is required, but `--resourceUri` is not (it only provides a name for the file to be used).

The way you have used the command is unfortunate, but I don't think it is something we can fix within the keptn CLI.

If anyone is still up for this, we will happily accept a PR.
We are using [spf13/cobra](https://github.com/spf13/cobra) for our command line programming in golang.

shouri007:
@johannes-b I'd like to take this up

johannes-b:
Hi @shouri007
Great, many thanks! 
[...]

PR #2122, opened 2020-07-26
Better parameter/flag checking of the keptn add-resource command.

shouri007:
For add-resource command,
If len(args) > 0 -> then abort this command.

Resolving [#1735 ](https://github.com/keptn/keptn/issues/1735l)


agrimmer:
@shouri007 Thank you for your contribution!
Nice solution!

PR #2183, opened 2020-08-06
Set height for chart and maxHeight for legend

### LINKED ISSUE #2150: Bridge shows unusable chart when having too many SLIs
Please see the following screenshot of my Bridge. When switching to the chart the height of the chart area is so small that it becomes unusable. It used to work in the beginning of this project but started showing this behavior a couple of runs ago

image

I am happy to give you access to my cluster for direct troubleshooting as I dont know how to best reproduce this

ermin-muratovic:
Fixes #2150

### DIFF
--- a/bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.ts
+++ b/bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.ts
@@ -76,7 +76,6 @@ export class KtbEvaluationDetailsComponent implements OnInit, OnDestroy {
         },
         min: 0,
         max: 100,
-        tickInterval: 10,
       },
       {
         title: null,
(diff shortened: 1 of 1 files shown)


ermin-muratovic:
Without a fix height for the chart and maxHeight for the legend, with a larger legend the chart area shrinks, see image below:

before

With a fix height for the chart and maxHeight for the legend, this issue is fixed, see below:

after

johannes-b:
* Can we please change the size of the legend, to see the full SLI names of the second line? 
With this height, they are cut off.  

* Can you also set the range of the y-axis from 0-100? The reason is that the lighthouse would never return a value above 100.

ermin-muratovic:
> * Can we please change the size of the legend, to see the full SLI names of the second line?
>   With this height, they are cut off.
> * Can you also set the range of the y-axis from 0-100? The reason is that the lighthouse would never return a value above 100.

Thanks for the feedback, you are right, the y-axis should be fixed from 0-100 and it makes sense to show 2-3 lines of SLI names in the legend, see new image below:

image

PR #2188, opened 2020-08-07
Don't filter the faulty event if deploymentstrategy is "direct"

### LINKED ISSUE #2133: Wrong version in environments overview when evaluation failed
I am using my simplenode app and have started extending my example into 3 stages: dev, staging, prod.

I've played around a bit with sending configuration-change events through both the Keptn CLI as well as using Jenkins where I accidentally sent configuration-change events into staging, therefore bypassing dev. 

This might be the reason for the wrong version information I am seeing in the bridge as you can see in the following screenshot:

image

Besides the wrong information for staging in the environment overview, I also want to make sure we are displaying the correct version information in the service view. 
* The "most relevant artifact" for me is the one that made it into production. 
* The other thing that is important to know is which artifact was the last that was sent to Keptn - whether it made it to prod or not. --> That is the "most recent artifact". 

You can also see this above in the screenshot.

ermin-muratovic:
Fixes #2133

### DIFF
--- a/bridge/client/app/project-board/project-board.component.ts
+++ b/bridge/client/app/project-board/project-board.component.ts
@@ -228,12 +228,12 @@ export class ProjectBoardComponent implements OnInit, OnDestroy {
     this.selectedStage = stage;
   }
 
-  countOpenApprovals(openApprovals: Trace[], stage: Stage, service?: Service) {
-    return openApprovals.filter(approval => approval == openApprovals.find(a => a.data.stage == stage.stageName && a.data.service == approval.data.service && (!service || a.data.service == service.serviceName))).length;
+  countOpenApprovals(openApprovals: Trace[], project: Project, stage: Stage, service?: Service) {
+    return this.getOpenApprovals(openApprovals, project, stage, service).length;
   }
 
-  getOpenApprovals(openApprovals: Trace[], stage: Stage, service: Service) {
-    return openApprovals.filter(approval => approval.data.stage == stage.stageName && approval.data.service == service.serviceName);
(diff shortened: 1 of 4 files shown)


bridge/client/app/project-board/project-board.component.ts
```diff
@@ -228,12 +228,12 @@ export class ProjectBoardComponent implements OnInit, OnDestroy {
...
-  countOpenApprovals(openApprovals: Trace[], stage: Stage, service?: Service) {
-    return openApprovals.filter(approval => approval == openApprovals.find(a => a.data.stage == stage.stageName && a.data.service == approval.data.service && (!service || a.data.service == service.serviceName))).length;
+  countOpenApprovals(openApprovals: Trace[], project: Project, stage: Stage, service?: Service) {
+    return this.getOpenApprovals(openApprovals, project, stage, service).length;
   }
 
-  getOpenApprovals(openApprovals: Trace[], stage: Stage, service: Service) {
-    return openApprovals.filter(approval => approval.data.stage == stage.stageName && approval.data.service == service.serviceName);
+  getOpenApprovals(openApprovals: Trace[], project: Project, stage: Stage, service?: Service) {
+    return openApprovals.filter(approval => approval.data.project == project.projectName && approval.data.stage == stage.stageName && (!service || approval.data.service == service.serviceName));
```
ermin-muratovic, line 236: This fixes a bug, that I figured out during testing. When you have the same service (same name) in different projects, and switch between those projects, it was showing you the open approvals also from the other project. Here I added also the filter on projectName, so you see only the approvals in this project, for this service in that stage.

johannes-b: Adding the project to uniquely identify a service (and its events) makes total sense.

Issue #1905, opened 2020-06-10
nx-ify Keptn Bridge


christian-kreuzberger-dtx:
Keptn Bridge currently has one app and one module. 

To improve mainability of our code, the Bridge should switch to a structure using an app-shell, preferably using [nx](https://nx.dev/).

Also paging @ddprrt and @lukasholzer on this.

warber:
will not be implemented

PR #2206, opened 2020-08-12
Provider better indication and workflow for artifacts waiting for approval

### LINKED ISSUE #2135: Allow user to provide approval also in service view
This is just a nice-to-have: I spend most of my time in the service view watching the events after I made a deployment.
It would be great to provide the end-user the ability to approve a deployment in that screen instead of having to navigate to the environment screen:

image

### LINKED ISSUE #2142: Provider better indication and workflow for artifacts waiting for approval
The Service view shows an artifact to be deployed into a stage even when it is waiting for approval. This is misleading and does not reflect the actual state. There is also not a good workflow for a user to get to approve or reject the artifact at that point. Either allow approval/reject in that view as described in #2135 or provide a link to the environment view.

The following screenshot should highlight the current implementation and how to make it better:

image

**Definition of Done**
- [ ] When an approval is required, add a blue border (same color as used in environment screen) to the stage that is waiting for approval
- [ ] When an approval is required, use a blue border (same color as used in environment screen) to highlight the event in the event stream 
- [ ] Added the accept/decline button to the **Approval.triggered** event - this closes: #2135

ermin-muratovic:
Resolves #2142 and resolves #2135 

image

### DIFF
--- a/bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.html
+++ b/bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.html
@@ -13,10 +13,9 @@
       </div>
     </dt-expandable-panel>
   </div>
-  <div class="mb-1 mt-1">
+  <div class="mb-1 mt-1" *ngIf="_showChart">
     <div fxLayout="row" class="pr-2">
-      <div fxFlex>
-      </div>
+      <div fxFlex></div>
       <div fxLayout="column" fxLayoutAlign="start end">
         <dt-button-group [(value)]="_comparisonView">
           <dt-button-group-item [value]="'heatmap'" (click)="_comparisonView = 'heatmap'">Heatmap</dt-button-group-item>
@@ -25,13 +24,13 @@
       </div>
     </div>
   </div>
-  <div (click)="$event.stopPropagation();" class="m-1 mr-2">
-    <dt-chart *ngIf="_comparisonView == 'heatmap'"
+  <div (click)="$event.stopPropagation();" class="m-0 mr-2">
+    <dt-chart *ngIf="_showChart && _comparisonView == 'heatmap'"
               [options]="_heatmapOptions"
               [series]="_heatmapSeries"
               (seriesVisibilityChange)="seriesVisibilityChanged($event)">
     </dt-chart>
-    <dt-chart *ngIf="_comparisonView == 'chart'"
+    <dt-chart *ngIf="_showChart && _comparisonView == 'chart'"
               [options]="_chartOptions"
               [series]="_chartSeries"
               (seriesVisibilityChange)="seriesVisibilityChanged($event)">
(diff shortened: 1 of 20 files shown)


bridge/client/app/_components/ktb-event-item/ktb-event-item.component.html
```diff
@@ -45,6 +45,13 @@ <h4 class="m-0 mt-1 mb-1" [class.error]="_event.isFaulty()" [textContent]="_even
       </dt-tag-list>
     </div>
   </div>
+  <div *ngIf="_event.isApproval() && project.getRootEvent(project.getService(_event.data.service), _event).isApproval()">
```
dominikmessner, line 48: Not necessary to change this in this PR, but in general function calls should be avoided in templates. Angular doesn't cache function results, so these functions are executed with every change detection cycle. This could degrade performance, depending on how often change detection is fired.
https://medium.com/showpad-engineering/why-you-should-never-use-function-calls-in-angular-template-expressions-e1a50f9c0496
https://stackblitz.com/edit/angular-hxmb6q

ermin-muratovic: Thanks @dominikmessner, actually I need to processes that once and store the reference in the event model. We have some refactoring tasks, see #1905, where I could address this issue, since I use this approach now already at several places.

ermin-muratovic:
As discussed with @johannes-b, I provide now the Score from the last evaluation next to the deployable artifact and on hover you see details from that evaluation.

image

bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.html
```diff
@@ -25,13 +24,13 @@
...
-  <div (click)="$event.stopPropagation();" class="m-1 mr-2">
-    <dt-chart *ngIf="_comparisonView == 'heatmap'"
+  <div (click)="$event.stopPropagation();" class="m-0 mr-2">
+    <dt-chart *ngIf="_showChart && _comparisonView == 'heatmap'"
               [options]="_heatmapOptions"
               [series]="_heatmapSeries"
               (seriesVisibilityChange)="seriesVisibilityChanged($event)">
     </dt-chart>
-    <dt-chart *ngIf="_comparisonView == 'chart'"
+    <dt-chart *ngIf="_showChart && _comparisonView == 'chart'"
```
sandroschmid, line 33: you could wrap the two charts in a ng-container and check for _showChart just onces but for such a simple condition I would consider a matter of personal taste

bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.ts
```diff
@@ -166,11 +167,24 @@ export class KtbEvaluationDetailsComponent implements OnInit, OnDestroy {
...
   }
 
+  @Input()
+  get showChart(): boolean {
+    return this._showChart;
+  }
+  set showChart(value: boolean) {
+    if (this._showChart !== value) {
+      this._showChart = value;
+      this._changeDetectorRef.markForCheck();
```
dominikmessner, line 177: A simple input might be sufficient here, it should do the same check internally.
```@Input() showChart: boolean;```

bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.html
ermin-muratovic: Yep, actually I would do it to with a ng-container, but since I have the *ngIf on each dt-chart anyway, I just thought I would integrate it there.

Issue #2195, opened 2020-08-11
Allow to set the namespace which is used for Keptn


agrimmer:
Currently, the CLI always installs Keptn in the namespace `keptn`.
Now, we would like to make the used namespace configurable, which allows to have multiple Keptn installations per cluster. 
Please note, that we do not support multiple Keptn installations within the same namespace.

**Tasks**
- Allow specifying the namespace in `keptn install` by introducing a flag `--namespace`. This should default to `keptn`.
- Allow specifying the namespace in `keptn uninstall` by introducing a flag `--namespace`. This should default to `keptn`.
- Allow specifying the namespace in `keptn generate support-archive` by introducing a flag `--namespace`. This should default to `keptn`.
- `/metadata` of the API also contains the namespace where Keptn is installed

In order to easily switch between Keptn installation, we also should investigate issue https://github.com/keptn/keptn/issues/1863

Furthermore, this issue solves https://github.com/keptn/keptn/issues/1287 because the name of a Keptn installation is then the name of the namespace.

**Definition of Done**
- [ ] Main tutorial (full-tour on DT/Prometheus and QG-only) work with a Keptn installation contained in a namespace other than `keptn`
- [ ] CLI docs are updated

TannerGabriel:
I will take this one!

TannerGabriel:
Information about the current status of this issue.

**Completed:**
- Allow specifying the namespace in `keptn install` by introducing a flag `--namespace`.
- Allow specifying the namespace in `keptn uninstall` by introducing a flag `--namespace`.
- Allow specifying the namespace in `keptn generate support-archive` by introducing a flag `--namespace`.
- Replaced hard coded namespace in API,  configuration service and lighthouse service.

**Still needs to be done:**
- Keptn contrib services need to be converted to Helm and feature dynamic namespaces instead of hard coded ones.
- The [Dynatrace SLI service](https://github.com/keptn-contrib/dynatrace-sli-service/blob/master/deploy/service.yaml#L72) still has a hard coded string for the `CONFIGURATION_SERVICE` that needs to be replaced.
- CLI documentation needs to be generated after the features are finished.

TannerGabriel:
#2214 #2216

agrimmer:
I've solved the "dynamice" namespace issues with these PRs
https://github.com/keptn-contrib/dynatrace-sli-service/pull/67
https://github.com/keptn-contrib/dynatrace-service/pull/175/files

johannes-b:
[...]

PR #2213, opened 2020-08-13
#2195 Define namespace for Keptn installation and other operations

TannerGabriel:
- Allow the user to specify a namespace for installing Keptn
- Replaced hard coded namespace variables with dynamic environment variables

### DIFF
--- a/configuration-service/common/git.go
+++ b/configuration-service/common/git.go
@@ -22,6 +22,8 @@ type GitCredentials struct {
 	RemoteURI string `json:"remoteURI,omitempty"`
 }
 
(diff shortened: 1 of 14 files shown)


configuration-service/common/git.go
```diff
@@ -22,6 +22,8 @@ type GitCredentials struct {
 	RemoteURI string `json:"remoteURI,omitempty"`
 }
 
+var namespace = os.Getenv("API_POD_NAMESPACE")
```
agrimmer, line 25: This should be `POD_NAMESPACE`

PR #2227, opened 2020-08-19
#2224 Change comparison filter to work on overall evaluation result

### LINKED ISSUE #2224: lighthouse-service: Change the comparison strategy to match the full quality gate result
Currently, the lighthouse-service applies the comparison strategy on an SLI basis. Therefore, the lighthouse-service checks SLI values considering the history of the same SLI. In other words, the lighthouse-service uses SLI values for the comparison regardless of the overall quality gate result.

**Example:**

Let’s assume the following evaluation history:
Screen Shot 2020-08-19 at 10 13 33

Let’s assume the following comparison strategy:
```
comparison:
  compare_with: "single_result"
  include_result_with_score: "pass"
  aggregate_function: avg
```

In the current implementation, a new evaluation would now take the response time of 50ms (this SLI was a pass) for the evaluation of the new response_time.

**With this issue, this behavior should be changed:**
- The comparison strategy `include_result_with_score` should target the overall quality gate result.
- If a new evaluation is performed, then only SLI values are considered which match the  `include_result_with_score` comparison strategy, i.e. the overall quality gate result decides whether SLI values are considered for the comparison or not.

**Definition of Done:**
- [ ] Behavior is implemented as described above
- [ ] Unit tests are available for this behavioral change
- [ ] Spec and docu are updated w.r.t. this behavioral change

bacherfl:
Closes #2224

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -591,7 +591,7 @@ jobs:
 
   - &K3sStandaloneTest
     stage: Test K3s Standalone (--platform=kubernetes)
-    if: branch = feature/2224/lighthouse-comparison #(branch = master or branch =~ ^release.*$) AND (type = cron OR type = push) # run for any change on master/release-* on push as well as cron
+    if: (branch = master or branch =~ ^release.*$) AND (type = cron OR type = push) # run for any change on master/release-* on push as well as cron
     os: linux
     env:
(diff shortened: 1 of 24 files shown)


agrimmer (approved):
lgtm

PR #2239, opened 2020-08-26
#2195 Do not use namespace a root context

johannes-b:
- Do not use namespace a root context for `keptn install`
- User message improvements for `keptn version`


agrimmer (approved):
lgtm

Issue #2149, opened 2020-07-30
Flaky integration tests: Integration tests fail (in unpredictable situations)


johannes-b:
**Situation 1:**
- Travis CI script cannot authenticate to gcloud: 

```
...
Please run:
  $ gcloud auth login
to obtain new credentials.
If you have already logged in with a different account:
    $ gcloud config set account ACCOUNT
to select an already authenticated account to use.
The command "gcloud auth activate-service-account --key-file ~/****" failed and exited with 1 during .
Your build has been stopped.
...
```

- [x] Problem is resolved. 

**Situation 2:**
- Bug in test/test_onboard_service.sh: 
  - Remove hard-coded project names from utility script
- [x] Problem is resolved in: #2164 

**Situation 3:**
- Suppress kubectl error message from builds
- [x] Problem is resolved in: #2167 

**Situation 4:**
- GKE 1.14 is not available anymore:
![image](https://user-images.githubusercontent.com/729071/91411213-4ddf7180-e848-11ea-8f5d-4d900d7bfd6c.png)
- [x] Issue is resolved in: #2242

**Situation 5:**
- unleash-service not deployed:
![image](https://user-images.githubusercontent.com/729071/91412592-4a4cea00-e84a-11ea-9561-c3e5be7f90c1.png)
- [x] Issue is resolved in: #2242

**Situation 6:**
- Quality gates on Minishift fail
- [x] Issue is resolved in: #2251

**Definition of Done:**
- [x] All situations as mentioned above are addressed and resolved. 
 
![image](https://user-images.githubusercontent.com/729071/91639574-ebd56680-ea17-11ea-87a1-d453f34138c0.png)

1 ... Cron Job on master
2 ... Master build

PR #2242, opened 2020-08-27
#2149 Fix problems in TravisCI for building/testing

johannes-b:
- GKE 1.14 is not available anymore - set version to 1.15
- Set namespace when deploying unleash-service


agrimmer (approved):
lgtm

Issue #2085, opened 2020-07-21
Show open remediation workflows in Environment screen


johannes-b:
When there is an ongoing remediation workflow: 

![problem open](https://user-images.githubusercontent.com/729071/88029051-c3fe0380-cb39-11ea-81df-e0a258c5b899.PNG)

* The environment screen displays an ongoing (i.e., sh.keptn.event.remediation.triggered) remediation workflow by a red icon in the stage tile 
* The counter shows how many remediation workflows are triggered.
* If a remediation workflow is finished (sh.keptn.event.remediation.finished) with `result: pass`, the counter is reduced by 1. 
* If now remediation is ongoing or no remediation workflow finished with `result: fail`, the icon is grey.  

(see stage: `prod-cluster-A`) 
![3_delivery_overview](https://user-images.githubusercontent.com/729071/88029203-f7409280-cb39-11ea-8a12-ec0c15c5e2d5.PNG)

When clicking on the stage:
* The legend on the top right corner shows the number of currently ongoing remediation workflows, e.g. `1 Problem(s) open` (The click functionality of this label will be addressed in #2087)
* The service tile shows the title of problem, e.g. `Out-of-memory problem detected`  
* Below is a list of the triggered actions: `sh.keptn.event.action.triggered`
* If there is a `sh.keptn.event.action.triggered` without the corresponding `sh.keptn.event.action.finished`, a loading icon displays that there is an action ongoing. 
* If there is an evaluation-result after the execution of an action, the result of the evaluation is shown, e.g, `61/100`

![image](https://user-images.githubusercontent.com/729071/89863253-069b8480-dbaa-11ea-9504-6ada5419c223.png)

**Definition of Done**
- [ ] Remediation workflow is shown in environment screen while it is running, `remediation.triggered` --> `remediation.finished`
- [ ] If remediation workflow finish successfully: `sh.keptn.event.remediation.finished` with `result: pass`, the environment screen does not show it. 
- [ ] If remediation workflow finish successfully: `sh.keptn.event.remediation.finished` with `result: fail`, the remediation workflow remains in the environment screen.

ermin-muratovic:
The environment screen now allows to easily see problems on a specific stage and remediation actions.
The stage tile shows the problem icon in red and a count how many services on this stage are "broken".
By clicking on the stage and expanding the details for a service you get the details about the problem and the remediation actions that are executed with the evaluation score after each remediation action. By hovering over the score, you can see the evaluation details in a popup.

[...]

PR #2243, opened 2020-08-27
#2085 Show remediation workflow in environment screen

ermin-muratovic:
The environment screen now allows to easily see problems on a specific stage and remediation actions.
The stage tile shows the problem icon in red and a count how many services on this stage are "broken".
By clicking on the stage and expanding the details for a service you get the details about the problem and the remediation actions that are executed with the evaluation score after each remediation action. By hovering over the score, you can see the evaluation details in a popup.

image


johannes-b (approved):
lgtm

PR #2245, opened 2020-08-28
Add messages if project or service does not exist

agrimmer:
This PR adds messages if a specified project or service does not exist.


johannes-b (approved):
lgtm

PR #2255, opened 2020-08-31
#2112 Fix missleading message "service not deployed yet" in quality gates use-case

### LINKED ISSUE #2112: Misleading message in bridge if no deployment was done but evaluations happened
Love the new bridge and the information it gives me when no deployment has yet been done on a service.
However, there are two things as you can see in this screenshot:

image

1: As I am using Keptn only for evaluation, the message "service not deployed yet" could be a bit misleading. It should say that - to get started send in some keptn send xxx events

2: If we want to let them know about delivery then we should fix the text. it should say "keptn send event new-artifact" and not "keptn send new-artifact"

**Definition of Done:**
- [x] Link text is changed to `keptn send event new-artifact` 
- [x] Don't show "Use keptn send event new-artifact ..." in case there is an open approval. This is shown in services and environments screen. 
- [x] Messages shows processed artifact for deployment.finished and evaluation.started (especially for evaluation.started, the image and tag must be provided; deployment.finished has it as default)
- [x] Document in keptn.sh/docs/bridge how to provide image and tag for evaluation use-case

ermin-muratovic:
Fixes #2112 

Now checking for deployment.finished or evaluation.done event and if `image` and `tag` are provided on the evaluation event, that will be displayed as `Last processed artifact`.

### DIFF
--- a/bridge/client/app/_models/project.ts
+++ b/bridge/client/app/_models/project.ts
@@ -31,14 +31,14 @@ export class Project {
     return this.stages.find(s => s.stageName == stageName);
   }
 
-  getLatestDeployment(service: Service, stage: Stage): Trace {
+  getLatestDeployment(service: Service, stage?: Stage): Trace {
     let currentService = this.getService(service.serviceName);
 
     if(currentService.roots)
       return currentService.roots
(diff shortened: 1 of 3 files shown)


johannes-b (approved):
lgtm

PR #2256, opened 2020-08-31
#2197 Provide API token and keptn auth command in user menu

### LINKED ISSUE #2197: Copy API Token and full keptn auth command to clipboard
As a Keptn user that has no access to the K8s cluster, I would like to use the Keptn bridge to get the API token and the full `keptn auth` command for the CLI. 

* The data is in the user menu
* The copy buttons copy the (1) API token or (2) full keptn auth command (containing the --endpoint) to the clipboard) 
* `keptn auth` refers to the documentation of the keptn auth command.  
* Extend the (internal) API of the bridge with an endpoint to retrieve the API Token

image

**Definition of Done**
- [x] New *Get started* section in the user menu
- [x] When clicking on a copy button, the data is in the clipboard
- [x] Pasting a `keptn auth` to the console automatically authenticates the CLI

ermin-muratovic:
Fixes #2197 
Provide API token and keptn auth command in user menu and allow to copy it easily to clipboard

image

### DIFF
--- a/bridge/client/app/_services/api.service.ts
+++ b/bridge/client/app/_services/api.service.ts
@@ -30,11 +30,10 @@ export class ApiService {
     this._baseUrl = `./api`;
   }
 
-  public getBridgeVersion(): Observable<any> {
+  public getKeptnInfo(): Observable<any> {
     let url = `${this._baseUrl}/`;
     return this.http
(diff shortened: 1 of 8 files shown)


johannes-b (approved):
lgtm

PR #2258, opened 2020-09-01
#2231 Mark info SLI correctly when empty pass/warning array is provided

### LINKED ISSUE #2231: Bridge: SLIs without warning or pass objections are "info only" and should not show up in red
SLIs that dont have any warning or pass criteria in the SLO definition are meant to be "informational only"
Right now it seems that the bridge shows the label of these SLIs in RED which I think is misleading. We should use a more neutral informational color such as BLUE

Here is a screenshot showing process_memory in RED
image

bacherfl:
Closes #2231 
Added a fix to check for empty pass/warning criteria arrays in addition to check for nil values. This avoids marking Info SLIs as being failed and sets the status to `info` instead.


johannes-b (approved):
lgtm

Issue #2087, opened 2020-07-21
Icons in stage tile and labels work as filter


johannes-b:
* The overview of `Problem open`, `Quality gates failed`, `Services are out-of-sync` (see stage details on the right) works as filter. When clicking on a label, the list of services is reduced. 

* The icons (problem, traffic-light, deployment) are clickable and activate a filter in the right panel. Thus, when clicking on the traffic-light, the `Quality gates failed` is selected.

![image](https://user-images.githubusercontent.com/729071/88032019-acc11500-cb3d-11ea-97ab-9c16d32dacd4.png)

**Definition of Done**
- [x] The labels `Problem open`, `Quality gates failed`, `Services are out-of-sync` on the details view of a stage work as filter
- [x] The icons activate a filter option

johannes-b:
**Minor UX improvements**

---

- [ ] **Singular/Plural** 
Currently, the label for "Services are out-of-sync" is inconsistent: 
![image](https://user-images.githubusercontent.com/729071/91799797-15fd7300-ec28-11ea-89ac-710cdf907d6b.png)
![image](https://user-images.githubusercontent.com/729071/91811222-7512b600-ec2f-11ea-9a08-ca0ded199691.png)


The label for "Services are out-of-sync" should work as follows:
- 0 Services are out-of-sync
- 1 Service is out-of-sync
- x Service is out-of-sync

- [x] **Deploy icon**
When there is an open approval, fill the "deploy" icon instead of using a border:
![image](https://user-images.githubusercontent.com/729071/91807161-768eaf00-ec2c-11ea-80ae-63925a872fee.png)
![image](https://user-images.githubusercontent.com/729071/91807811-a3db5d00-ec2c-11ea-9b72-2c32b4afaa1f.png)

- [ ] **Changing labels of events (to represent past tense)**
Source: https://github.com/keptn/keptn/blob/master/bridge/client/app/_models/event-labels.ts

Change event from/to
- Service create --> Service created
- Configuration change --> Configuration changed
- Start SLI retrieval --> SLI retrieval started
- Problem open --> Problem opened

PR #2259, opened 2020-09-01
#2087 Allow to filter services in environments screen by event type

ermin-muratovic:
Now the event types are displayed as toggle buttons with the details as labels.
By activating the toggle button, you will see only services with this event type.

image


johannes-b (approved):
lgtm

PR #2261, opened 2020-09-01
#2220 Added endpoint for forwarding events to Keptn API

### LINKED ISSUE #2220: Extend distributor to bridge traffic from Keptn service to Keptn API
**Tasks:**
- Implement an HTTP server with an endpoint /event, which accepts POST requests. This endpoint should accept an event as payload, add the x-token as header and forward the event to the Keptn API. Then, it should return the response from the Keptn API. This endpoint must only accept traffic from localhost.
- Keptn API URL as well as x-token can be configured via an ENV-variable.

**Advantages:**
- Execution plane services do not have to know and handle the x-token
- Execution plane services are decoupled from the Keptn API

**Definition of Done:**
- [ ] Keptn services can post an event to localhost:8080/event, which is forwarded to the Keptn API
- [ ] Distributor contains documentation on how to set the Keptn API URL and x-token

bacherfl:
Closes #2220 

The Endpoint can be configured using the env var `HTTP_EVENT_FORWARDING_ENDPOINT`
The API token can be configured using the env var `HTTP_EVENT_ENDPOINT_AUTH_TOKEN`

Further changes:
The env var for configuring the endpoint for polling events via HTTP has been renamed to `HTTP_EVENT_POLLING_ENDPOINT`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -254,7 +254,7 @@ jobs:
         cd ..
       - |
         cd "${DISTRIBUTOR_FOLDER}"
-        go test -race -v -coverprofile=coverage.txt -covermode=atomic ./...
+        go test -v -coverprofile=coverage.txt -covermode=atomic ./...
         cd ..
       - |
         cd "${EVENTBROKER_FOLDER}"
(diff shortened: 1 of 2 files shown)


agrimmer (approved):
lgtm

PR #2268, opened 2020-09-01
#2061 Make links in labels clickable

### LINKED ISSUE #2061: Make labels that are links clickable in the Bridge
In order to better integrate tools with Keptn, it is important to provide links between these tools. External tools can easily link to the Keptn Bridge as the bridge supports deep links. 

**Proposed feature:**
* To link back to a tool that integrates with Keptn, there should be the ability to add links to the Keptn events, e.g: when Dynatrace-SLI-Service pulls data from a Dashboard it should add that link to the SLI-done event. 
* If the bridge then makes this link clickable, it allows a direct tool integration via the Web UIs.

Right now the best way for tools that integrate with Keptn to pass additional context is through the collection of the labels in the CloudEvents. What I am asking for is extending the bridge so that labels that are actual links are displayed as links and are clickable.

Currently, the bridge only shows the text and also cuts the text which not even allows copy/paste of the URL. Here is an example:
image

**Definition of Done:**
- [ ] Labels where the value of the label starts with `http` or `https` only show the label name, but make it a hyperlink that opens that link that comes in the value.

ermin-muratovic:
When the value of a label is an URL, the value is linked and opens in a new tab.
Fixes #2061

### DIFF
--- a/bridge/client/app/_components/ktb-event-item/ktb-event-item.component.html
+++ b/bridge/client/app/_components/ktb-event-item/ktb-event-item.component.html
@@ -30,7 +30,7 @@
       <p class="bold nobreak m-0 mt-1 mb-1">Labels: </p>
       <dt-tag-list aria-label="affected-entities" (click)="$event.stopPropagation();">
         <dt-tag *ngFor="let tag of _event.data.ProblemDetails.tagsOfAffectedEntities">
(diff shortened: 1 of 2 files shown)


agrimmer (approved):
lgtm

PR #2279, opened 2020-09-02
Additional integration tests

bacherfl:
Added tests for:

- delete service
- install Keptn into custom namespace and go through Quality-gates standalone use-case


agrimmer (approved):
lgtm

PR #2324, opened 2020-09-11
Remove binary

agrimmer:


christian-kreuzberger-dtx (approved):
I guess @thschue owes us a beer now :)

LGTM

Issue #2325, opened 2020-09-11
Keptn 0.7: Swagger-ui uses https, but only http is available


christian-kreuzberger-dtx:
I've installed Keptn 0.7.0 (still need to verify this for 0.7.1) with Istio 1.6.2 on GKE 1.15.
I also exposed the api-gateway using istio ingress via http://1-2-3-4.nip.io

When trying to access the api via CLI everything works fine. When accessing it via swagger-ui, it tries to access the underlying API calls via HTTPS (instead of HTTP), which results in errors.

![Screenshot from 2020-09-11 10-54-12](https://user-images.githubusercontent.com/56065213/92897171-45c52b80-f41d-11ea-8aa9-9a599ccbcf45.png)

I'm guessing the part `https` is hardcoded for the base URL, and we need to make it dynamic.

FYI, Keptn Bridge seems to not have this issue and works fine.

christian-kreuzberger-dtx:
It is easily possible to just select `http` in a dropdown on top of the UI (though if you scroll down, you won't see it ... hence I ran into this issue):
![Screenshot from 2020-09-11 13-42-42](https://user-images.githubusercontent.com/56065213/92921634-b461b380-f434-11ea-90de-1ad2275bdf3d.png)

The problem here is that it defaults to HTTPS for whatever reason.
However, according to this PR https://github.com/swaggest/swgui/pull/7, swagger UI should
> Set current window location scheme if available

So I'll downgrade this bug to an enhancement for api/swagger-ui.

christian-kreuzberger-dtx:
Fixed for 0.7.2, backported for master
- #2330
- #2328

PR #2328, opened 2020-09-11
#2325 Remove anything scheme related to let swagger automatically determine the scheme

christian-kreuzberger-dtx:
Fixes #2328 
This PR drops `scheme` from swagger.yaml, as well as from the docker builds.

This allows swagger-ui to automatically determine the correct scheme by using the scheme that is used to access swagger-ui.


agrimmer (approved):
lgtm

PR #2335, opened 2020-09-14
#2307 include git commit id and repository in resource response

### LINKED ISSUE #2307: Configuration-Service: Include Git commit ID in Resource Responses
In the new CloudEvents that will be part of Keptn 0.8, we will reference the Version (i.e. Git commit ID) of the resource used by a service to perform its task. For example:

- The SLO file used by the lighthouse service
- The jmeter config file used by the jmeter-service

Execution services retrieve these files using the `GET v1/./.../resource` endpoint of the configuration service. The response from these endpoint should be extended to include the git commit, as well as the exact URL to the file within the upstream repo - if available. 

Research: 
- Can we generate a generic Git Url that points to the upstream repo?
- Tekton: https://vincent.demeester.fr/articles/tekton-usage.html#h:7da25768-794e-48c1-b92a-603e1c0572b2

Key-value map to store the values (in case the generic URL does not work):
- upstream URL (without token/secret)
- version
- branch
- ( (file) URI )

Tasks:
- Add the properties `version` and `upstreamURL` to the `Resource` response payload defined in the configuration service [API spec](https://github.com/keptn/keptn/blob/d62e52e487b5675f85f50e52809b8cf01790b46c/configuration-service/swagger.yaml#L352)
- Include those properties in all `GET, PUT, POST .../resource` responses
- Extend the corresponding struct in the [go-utils](https://github.com/keptn/go-utils/blob/010bbcf95e6f3735109995f1be44131cc323a74c/pkg/api/models/resource.go#L17) to include those additional fields

Definition of done:

- [ ] Configuration service updated
- [ ] go-utils updated

bacherfl:
Closes #2307 
Added the fields `UpstreamURL`, `Version` and `Branch` to responses for resource requests


configuration-service/handlers/project_resource.go
```diff
@@ -174,11 +174,15 @@ func GetProjectProjectNameResourceResourceURIHandlerFunc(params project_resource
...
-			ResourceURI:     &params.ResourceURI,
-			ResourceContent: resourceContent,
-		})
+	resource := &models.Resource{
+		ResourceURI:     &params.ResourceURI,
+		ResourceContent: resourceContent,
+		Branch:          "master",
+	}
+
+	common.AddResourceMetadata(params.ProjectName, err, resource)
```
agrimmer, line 183: Can we add those metadata also to the endpoints for PUT, DELETE, and POST?

agrimmer (approved):
lgtm

PR #2337, opened 2020-09-14
#2336 Allow http for git repo urls

### LINKED ISSUE #2336: Configuration-service: Allow non HTTPS connections (allow HTTP connections)
Right now, we assume `https://` for git ops repositories:

https://github.com/keptn/keptn/blob/61dd6d0f29962c0fb2efd1c5887157e4451d107d/configuration-service/common/git.go#L47-L53

This creates a problem when a user specifies `http://` as the upstream repository.

1. The log messages that is written contains the token (@bacherfl )
1. https is not always available

# Definition of Done 

- [ ] Within the `replace` statement, switch from `https://` to a more generic `://` replacement
- [ ] Hide the token in log output

bacherfl:
Closes #2336 
This PR allows to use an HTTP URL for a project upstream

### DIFF
--- a/configuration-service/common/git.go
+++ b/configuration-service/common/git.go
@@ -26,15 +26,15 @@ var namespace = os.Getenv("POD_NAMESPACE")
 
 // CloneRepo clones an upstream repository into a local folder "project" and returns
 // whether the Git repo is already initialized.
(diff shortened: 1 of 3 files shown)


agrimmer (approved):
lgtm

PR #2344, opened 2020-09-15
Fixed nil pointer access

bacherfl:
The changes made in PR #2338 introduced an unintended nil pointer access. This is fixed with this PR


agrimmer (approved):
lgtm

Issue #2295, opened 2020-09-07
CII Badge: Secret for generating WebSocket session token with 256-bit size


johannes-b:
**Session token for WebSocket communication:** `token`/`keptnContext`

- Struct: https://github.com/keptn/go-utils/blob/release-0.7.0/pkg/api/models/event_context.go#L17
- Generation:
  * `SECRET_TOKEN`: Is the `keptn-api-token` = 160-bit secret.
  * `token`: To generate this session token, the HS256 (Hash-based Message Authentication Code (HMAC) and SHA-256) is applied on the secret `SECRET_TOKEN`: https://github.com/keptn/keptn/blob/master/api/ws/client.go#L210
  * `keptnContext`: Is an UUID of the format: `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`
- Storing: Is not stored, but rather part of the server response and then used by the client to open a WebSocket communication. 

*Note on HS256:*
* Hash-based Message Authentication Code (HMAC) is an algorithm that combines a certain payload with a secret using a cryptographic hash function like SHA-256. The result is a code that can be used to verify a message only if both the generating and verifying parties know the secret. In other words, HMACs allow messages to be verified through shared secrets.

* The [JSON Web Algorithms RFC 7518](https://tools.ietf.org/html/rfc7518#page-7) states that a key of the same size as the hash output (for instance, 256 bits for "HS256") or larger MUST be used with the HS256 algorithm.

**Issue:**
- The used secret must have a size of 256-bit and it must not be the `keptn-api-token`. 

**Task:**
- A research task has to figure out whether it is possible to not use a persistently stored secret but rather create a random secret that is stored in memory and has a lifetime according to the lifetime of the corresponding WebSocket channel. If the WebSocket channel terminates, the secret has to expire. 
- The random secret MUST have a size of 256-bit. 

**Definition of Done:**
- [ ] Secret used for generating the WebSocket token is not the `keptn-api-token`. Instead, an individual token with size of 256-bit.

christian-kreuzberger-dtx:
Separating the websocket token generation from the api token is good, however we need to be aware that our websocket communication uses JWT in the background. The benefit is that multiple independent systems should be able to verify that the JWT is correct, therefore they need to know `SECRET_TOKEN` (but not necessarily the api token).

IMHO to make this whole thing more reliable, we should have a strong master token which we use 
* as the `SECRET_TOKEN` for generating websocket tokens,
[...]

PR #2345, opened 2020-09-15
#2295 Increase password and token entropy

christian-kreuzberger-dtx:
This PR increases the entropy of the api-token to >= 256 bit (Calculation: 62 possible characters (so almost 6 bit), 45 places -> 45 * 6 = 270 bit)

Also, I've increased the entropy of the mongo db admin password (I believe that's also one of those that can benefit)


agrimmer (approved):
lgtm

PR #2350, opened 2020-09-15
Allow to disable Bridge secret for basic auth

agrimmer:


bacherfl (approved):
lgtm

christian-kreuzberger-dtx (changes requested):
Hold on - disabling this secret also exposes the API Token as well as the REST API

christian-kreuzberger-dtx (approved):
lgtm

Issue #2354, opened 2020-09-15
Bridge UI breaks on first open approval event on stage


ermin-muratovic:
**Environment**

* Client OS (e.g., Linux, OSX, Windows): any
* Keptn Version (`keptn version`): 0.7.1

**Affected Component**
* [ ] Docs
* [ ] CLI
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [ ] Helm
* [ ] Istio
* [x] Bridge
* [ ] Gatekeeper
* [ ] Datastore
* [ ] REST API
* [ ] eventbroker / distributor
* [ ] jmeter

**Describe the bug**
UI breaks, when there is an open approval, but there is no image deployed on that stage yet. The bridge tries to retrieve the currently deployed artifact and calls `getShortImageName` on that artifact. If there is no artifact deployed yet, it is undefined, and it can't access `getShortImageName` of undefined.

**To Reproduce**
Create a new project, with manual approval configured for any stage. Send in a new artifact for the first time and open the event trace. The UI breaks.

**Expected behavior**
The event trace should show the new artifact and allow the user to send in the approval or decline the deployment. Currently deployed should show 'N/A', as there is no image deployed yet.

**Screenshots**

![image](https://user-images.githubusercontent.com/6098219/93205935-e8044c80-f758-11ea-8310-4795c1ddf20b.png)

![image](https://user-images.githubusercontent.com/6098219/93206043-108c4680-f759-11ea-9e95-4d5ca5b1e250.png)

PR #2355, opened 2020-09-15
#2354 use safe operator for access to getShortImageName()

ermin-muratovic:
use safe operator for access to getShortImageName() on currently deployed artifact and show 'N/A' if no image deployed yet


christian-kreuzberger-dtx (approved):
lgtm - please backport for 0.7.2

PR #2360, opened 2020-09-16
Remove credentials before storing event

bacherfl:
When a project is created, an internal event is sent by the API. This event also contains the git user and git token of the user, which should not be persisted. Therefore, this PR introduces a check for that specific event type in the mongodb-datastore, which makes sure that those properties are not persisted by removing them from the incoming payload


christian-kreuzberger-dtx (approved):
lgtm

Issue #2357, opened 2020-09-16
Bridge version is not displayed


johannes-b:
The Keptn Bridge in version 0.7.1, does not display its version (in the user menu):

![image](https://user-images.githubusercontent.com/729071/93302306-0cfccc00-f7fa-11ea-9631-b911f5130c0d.png)

**Definition of Done:**
- [ ] Bridge shows version as supported by 0.7.0

![image](https://user-images.githubusercontent.com/729071/93303001-1a668600-f7fb-11ea-9046-c39da4778bfd.png)

Issue #2320, opened 2020-09-10
Keptn Bridge: Disable the version check feature 


johannes-b:
The version check feature is not always relevant especially in the case where a Keptn user has no control over upgrading a Keptn installation. The same is true for the feature of copying a `Keptn API token` and `keptn auth command`.

![image](https://user-images.githubusercontent.com/729071/92734681-0c3ae480-f379-11ea-9000-021bd8c826c2.png)

**Tasks**
- Introduce an environment variable (flag) to enable/disable the `version check` as well as the `Keptn API token`
- Hide the `Keptn API token` and `keptn auth command` when disabled 
- Hide the `Version check` feature when disabled

PR #2361, opened 2020-09-16
#2320 Feature toggle for version check and api token info

ermin-muratovic:
1. The automatic version check feature can be disabled via setting the env `ENABLE_VERSION_CHECK` to `false`.

In this case the user won't to be asked to enable the version check and also the user menu on the top right won't show the toggle to enable the version check. The currently used version is still displayed, since this info is available in the bridge env, but the call to the `version.json` is omitted and therefore also the link to the data policy is not necessary.

2. The api token and keptn auth info fields can be hidden via setting the env `SHOW_API_TOKEN` to `false`

In this case the api token and keptn auth info fields won't be displayed and the api token also won't be provided to the bridge in the `GET /bridgeInfo` call.

This PR fixes also #2357 and shows the bridge version correctly again.


christian-kreuzberger-dtx:
@ermin-muratovic unit tests are failing
```
ERROR in client/app/_components/ktb-copy-to-clipboard/ktb-copy-to-clipboard.component.spec.ts:3:38 - error TS2305: Module '"./ktb-copy-to-clipboard.component"' has no exported member 'KtbExpandableTileHeader'.
```

ermin-muratovic:
> @ermin-muratovic unit tests are failing
> 
> ```
> ERROR in client/app/_components/ktb-copy-to-clipboard/ktb-copy-to-clipboard.component.spec.ts:3:38 - error TS2305: Module '"./ktb-copy-to-clipboard.component"' has no exported member 'KtbExpandableTileHeader'.
> ```

fixed it

johannes-b (approved):

Issue #2257, opened 2020-08-31
Hide the API token and keptn auth command per default


johannes-b:
(This is an enhancement of #2197)

As space is limited in the user menu, I see an enhancement of the "copy API token / keptn auth command" as depicted below:

![image](https://user-images.githubusercontent.com/729071/91737638-9d5ad000-ebaf-11ea-88a4-22edfc6eee60.png)

* Per default, the value of the *API token* / *keptn auth command* is hidden and the label "Keptn API token" or "keptn auth command" is shown. 
* By clicking on the "eye" button, the actual value is displayed.
* The copy button changes into an icon as shown. 

**TBD:**
- Do we lose the feature of the copy button that turns green when clicking?

PR #2367, opened 2020-09-17
#2257 Hide api token and keptn auth command

ermin-muratovic:
The api token and keptn auth command are now initially hidden and only a label is visible. By clicking on the toggle button, you can see the value, but the copy button still works in visible or hidden state.

image


johannes-b (approved):

johannes-b (approved):
lgtm

PR #2368, opened 2020-09-17
#2365 Send evaluation-done events with error information when service/stage/project could not be found

### LINKED ISSUE #2365: No sh.keptn.events.evaluation-done event when triggering an evaluation with an unknown project/service/stage
**Status:**
I have a Keptn setup as follows:
* project: *sampleproject*
* stage: *qualitygates*
* service: *sampleservice*

**Behaviour:**
When triggering an evaluation via `keptn send event start-evaluation --project=sampleproject--stage=qualitygates--service=sampleservice` with a typo in one of the three entities, then there is no sh.keptn.events.evaluation-done event available. 

**Tasks:**
- Cover the situations of unavailable project/stage/service: https://github.com/keptn/keptn/blob/master/lighthouse-service/event_handler/start_evaluation_handler.go#L47 (If entities are available but SLO is not there, continue with sending a `get-sli` event to trigger SLI retrieval.)
- When handling an `get-sli.done` event, respond with a `sh.keptn.events.evaluation-done` (result: Failed due to ...) in case of error states: https://github.com/keptn/keptn/blob/master/lighthouse-service/event_handler/evaluate_sli_handler.go#L46

**Definition of Done:**
- A typo in project/service: Event does not show up in bridge, but when executing a `keptn get event evaluation-done` the result  of the event says: `project/service xyz not found`
- A typo in stage: Event appears in bridge and it shows: `stage xyz not found`

bacherfl:
Closes #2365


lighthouse-service/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/cloudevents/sdk-go v0.10.0
 	github.com/ghodss/yaml v1.0.0
+	github.com/go-openapi/swag v0.19.5
```
christian-kreuzberger-dtx, line 8: I'm wondering why this dependency was added. Any idea?

bacherfl: That slipped through by copy&pasting a code snippet from another service. Has just been removed

christian-kreuzberger-dtx (approved):
lgtm

PR #2380, opened 2020-09-21
Verify helm templates before uploading them

christian-kreuzberger-dtx:
This PR verifies that the helm template that are being created are valid.

Hint: they are currently not valid.


agrimmer (approved):
lgtm

PR #2393, opened 2020-09-23
#2388 include event IDs of compared events

### LINKED ISSUE #2388: Lighthouse-service: include_result_with_score just works on SLO-level (and return compared evaluation results)
**Current behavior:**
The lighthouse-service filters out evaluation-results and objectives that failed (if `include_result_with_score: "pass_or_warning"` is set); see: https://github.com/keptn/keptn/blob/master/lighthouse-service/event_handler/evaluate_sli_handler.go#L371

This can lead to the situation that the quality gate opens even though the measured value is still not in an acceptable range. Please see this example: 

image

**Changing behavior:**
- The `include_result_with_score: "pass_or_warning"` parameter has no impact on the individual objectives from an SLO evaluation, but just on the overall evaluation result. 
- If an evaluation has a status pass/warning, then consider all objectives for comparison regardless of their status. 
- Explain this new behavior in keptn/spec and on keptn.sh/docs

_Example:_ 

image

**Additional task:**
- Extend the `sh.keptn.events.evaluation-done` event to contain the event-id of the evaluation results taken for comparison. 

**Definition of Done:**
- [ ] `include_result_with_score: "pass_or_warning"` works only on evaluation level, not on an objective or SLI level. 
- [ ] Event-id of compared evaluation results is part of the payload of the evalation.done event. 
- [ ] keptn/spec and keptn.sh/docs show this behavior

bacherfl:
Closes #2388
This PR adds a list of event IDs that were used for a comparison-based evaluation of SLOs.
Also, the PR removes the filtering logic on the SLI level and instead applies the `compare_results_with` filter on an event level.


lighthouse-service/go.sum
```diff
@@ -181,6 +186,10 @@ github.com/jstemmer/go-junit-report v0.0.0-20190106144839-af01ea7f8024/go.mod h1
 github.com/julienschmidt/httprouter v1.2.0/go.mod h1:SYymIcj16QtmaHHD7aYtjjsJG7VTCxuUUipMqKk8s4w=
 github.com/kelseyhightower/envconfig v1.4.0 h1:Im6hONhd3pLkfDFsbRgu68RDNkGF1r3dvMUtDTo2cv8=
 github.com/kelseyhightower/envconfig v1.4.0/go.mod h1:cccZRl6mQpaq41TPp5QxidR+Sa3axMbJDNb//FQX6Gg=
+github.com/keptn/go-utils v0.6.3-0.20200923070433-95a8b5159363 h1:p+33Y6OoBjdbEyeEYZZwINN4IO7SGZH71uNsEDTlN2I=
```
christian-kreuzberger-dtx, line 189: what's going on here, why are we adding go-utils in versino 0.6.3?

lighthouse-service/go.mod
```diff
@@ -10,9 +10,9 @@ require (
 	github.com/google/uuid v1.1.1
 	github.com/json-iterator/go v1.1.9 // indirect
 	github.com/kelseyhightower/envconfig v1.4.0
-	github.com/keptn/go-utils v0.7.1
+	github.com/keptn/go-utils v0.6.3-0.20200923082506-f715579724ec
```
christian-kreuzberger-dtx, line 13: What's going on here, why are we adding go-utils in version 0.6.3?

lighthouse-service/event_handler/common.go
```diff
@@ -8,7 +8,8 @@ import (
 
 	"github.com/ghodss/yaml"
 	utils "github.com/keptn/go-utils/pkg/api/utils"
-	keptn "github.com/keptn/go-utils/pkg/lib"
+	keptnv1 "github.com/keptn/go-utils/pkg/lib"
+	keptn "github.com/keptn/go-utils/pkg/lib/keptn"
```
christian-kreuzberger-dtx, line 12: something seems weird with those imports, what's going on here?

lighthouse-service/go.sum
bacherfl: this is the latest commit ID. Not sure why it starts with 0.6.3, but this is generated by go modules when targeting a specific commit

lighthouse-service/event_handler/common.go
bacherfl: the structure of the go-utils has changed

lighthouse-service/go.mod
bacherfl: this is the latest commit ID. Not sure why it starts with 0.6.3, but this is generated by go modules when targeting a specific commit

christian-kreuzberger-dtx (approved):
thanks for clarifying, LGTM

PR #2394, opened 2020-09-23
#2388 include event IDs of compared events

### LINKED ISSUE #2388: Lighthouse-service: include_result_with_score just works on SLO-level (and return compared evaluation results)

bacherfl:
Closes #2388
This PR adds a list of event IDs that were used for a comparison-based evaluation of SLOs.
Also, the PR removes the filtering logic on the SLI level and instead applies the compare_results_with filter on an event level.

### DIFF
--- a/lighthouse-service/event_handler/evaluate_sli_handler.go
+++ b/lighthouse-service/event_handler/evaluate_sli_handler.go
@@ -26,6 +26,7 @@ type datastoreResult struct {
 	PageSize    int    `json:"pageSize"`
 	Events      []struct {
 		Data interface{} `json:"data"`
+		ID   string      `json:"id"`
 	}
 }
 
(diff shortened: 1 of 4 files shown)


christian-kreuzberger-dtx (approved):
lgtm

PR #2398, opened 2020-09-23
Installer: Re-Enable uploading of index.yaml to google cloud

christian-kreuzberger-dtx:
I have disabled uploading index.yaml 2 weeks ago for the time being, so that we don't break our index.yaml in the process of creating the new release (also, to avoid accidental upgrades to 0.7.2).

With this PR I'm reverting this change.


agrimmer (approved):
lgtm

PR #2399, opened 2020-09-24
#2293 Point user to upgrade docs, fix bug in update project

### LINKED ISSUE #2293: Keptn CLI Update: Provide update instructions when new version is detected
The Keptn CLI does a status check and gives information in case there is a new version available. The following is an output of keptn version:

```
$ keptn version
CLI version: 0.7.0
keptn version 0.7.1 is available! Please visit https://keptn.sh for more information.
Daily version check is enabled. Keptn will collect statistical data and will notify about new versions and security patches for Keptn. Details can be found at https://keptn.sh/docs/0.7.x/reference/version_check      
To disable the daily version check, please execute:
keptn set config AutomaticVersionCheck false
```

It would be great to provide guidance on how to upgrade to the latest keptn CLI in that textual output as this will definitely increase adoption of new versions

agrimmer:
Closes #2293 

`keptn version` now outputs the following message, which contains the link to the upgrade instructions
```
Keptn CLI version 0.7.1 is available! Please visit https://keptn.sh/docs/0.7.1/operate/upgrade/ for more information.
To disable this notice, run: './keptn set config AutomaticVersionCheck false'

Keptn CLI version: 0.7.0

```

Please note that this requires that the alias in the docs is porperly set, e.g.
```
aliases:
  - /docs/0.7.0/operate/upgrade/
  - /docs/0.7.1/operate/upgrade/
```

### DIFF
--- a/cli/pkg/version/version_checker.go
+++ b/cli/pkg/version/version_checker.go
@@ -157,18 +157,24 @@ func (v *VersionChecker) CheckCLIVersion(cliVersion string, considerPrevCheck bo
 			}
 			msgPrinted := false
 			if newVersions.stable.newestCompatible != nil {
+				segments := newVersions.stable.newestCompatible.Segments()
+				majorMinorXVersion := fmt.Sprintf("%v.%v.x", segments[0], segments[1])
 				fmt.Printf(newCompatibleVersionMsg+"\n", newVersions.stable.newestCompatible.String(),
-					newVersions.stable.newestCompatible.String())
+					majorMinorXVersion)
 				msgPrinted = true
 			}
 			if newVersions.prerelease.newestCompatible != nil {
+				segments := newVersions.prerelease.newestCompatible.Segments()
+				majorMinorXVersion := fmt.Sprintf("%v.%v.x", segments[0], segments[1])
 				fmt.Printf(newCompatibleVersionMsg+"\n", newVersions.prerelease.newestCompatible.String(),
-					newVersions.prerelease.newestCompatible.String())
+					majorMinorXVersion)
 				msgPrinted = true
 			}
 			if newVersions.stable.newestIncompatible != nil {
+				segments := newVersions.stable.newestIncompatible.Segments()
+				majorMinorXVersion := fmt.Sprintf("%v.%v.x", segments[0], segments[1])
 				fmt.Printf(newIncompatibleVersionMsg+"\n", newVersions.stable.newestIncompatible.String(),
-					newVersions.stable.newestIncompatible.String())
+					majorMinorXVersion)
 				msgPrinted = true
 			}
 			if msgPrinted && considerPrevCheck {
--- a/cli/pkg/version/version_checker.go
+++ b/cli/pkg/version/version_checker.go
@@ -130,9 +130,9 @@ func (v *VersionChecker) getNewerCLIVersion(cliConfig config.CLIConfig, usedVers
 	return res, nil
 }
 
-const newCompatibleVersionMsg = `keptn version %s is available! Please visit https://keptn.sh for more information.`
-const newIncompatibleVersionMsg = `keptn version %s is available! Please note that this version might be incompatible with your Keptn cluster ` +
-	`version and requires to update the cluster too. Please visit https://keptn.sh for more information.`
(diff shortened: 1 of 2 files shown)


cli/pkg/version/version_checker.go
```diff
@@ -130,9 +130,9 @@ func (v *VersionChecker) getNewerCLIVersion(cliConfig config.CLIConfig, usedVers
 	return res, nil
 }
 
-const newCompatibleVersionMsg = `keptn version %s is available! Please visit https://keptn.sh for more information.`
-const newIncompatibleVersionMsg = `keptn version %s is available! Please note that this version might be incompatible with your Keptn cluster ` +
-	`version and requires to update the cluster too. Please visit https://keptn.sh for more information.`
+const newCompatibleVersionMsg = `Keptn CLI version %s is available! Please visit https://keptn.sh/docs/%s/operate/upgrade/ for more information.`
+const newIncompatibleVersionMsg = `Keptn CLI version %s is available! Please note that this version might be incompatible with your Keptn cluster ` +
+	`version and requires to update the cluster too. Please visit https://keptn.sh/docs/%s/operate/upgrade/ for more information.`
```
christian-kreuzberger-dtx, line 135: Regarding the URLs, this scheme might not work for all future releases, especially if we require aliases (we will definately forget to set theM).

Maybe we should rather point to a separate new page, e.g., https://keptn.sh/download, where we list all versions to download?

christian-kreuzberger-dtx: I've created an issue in the docs for this: https://github.com/keptn/keptn.github.io/issues/645

At the same time, I've done my homework and looked at other projects (namely helm and istio), and they don't have such a page. Instead, they always reference to the quickstart section in the docs.

So I'm wondering, should we also just reference our [Quickstart](https://keptn.sh/docs/quickstart/) page, and maybe adding a download link for Windows, MacOS and Linux in the appropriate subection?

johannes-b: For addressing the business value of this issue, which should provide guidance for a user to upgrade I see three options to point the user to: 
- (1) keptn.sh/docs
  - pro: most generic approach, the lowest risk to fail in the future
  - con: the user gets not guidance but just a download link
- (2) keptn.sh/docs/quickstart/
  - pro: the user gets a link to download the latest stable CLI, the quickstart page is stable with low risk to change in the future
  - con: the user does not get a hint of how to perform the upgrade
- (3) keptn.sh/docs/0.7.x/operate/upgrade/
  - pro: the user gets directed to the right spot where upgrade instructions are provided
  - con: when changing the structure of the docs > all CLIs fail

With option (1) and (2) we can run into the problem that someone is using CLI 0.7.0, downloads/installs 0.7.2, and performs an upgrade. We are working hard to make this scenario (upgrading from 0.7.0 to 0.7.2) possible, but I have my doubts that we can guarantee this for all upcoming releases. I think that when we have migration steps included, (e.g., you first need to upgrade to 0.7.1 before upgrading to 0.7.2.) we need to point the user to the **/operate/upgrade** page anyways. 

To reduce the risk of forgetting the alias, I would propose to ignore the patch version in the link, i.e., 0.7.2, 0.7.1 --> point to `0.7.x/operate/upgrade`

Final thought: The bridge is currently working the same way and we should keep both in sync: https://github.com/keptn/keptn/blob/fde3055c073b903b29f65977553c7e32b0c01dd7/bridge/client/app/app-header/app-header.component.ts#L71

**My Conclusion:**
When a new CLI version is available, lead the user to the instructions provided at: `keptn.sh/docs/0.7.x/operate/upgrade/`

christian-kreuzberger-dtx: I've udpated the source code accordingly, this is what it looks like now (I faked `currentVersion` to `0.6.0`):
```
Keptn CLI version: 0.6.0

Keptn CLI version 0.6.2 is available! Please visit https://keptn.sh/docs/0.6.x/operate/upgrade/ for more information.
Keptn CLI version 0.7.1 is available! Please note that this version might be incompatible with your Keptn cluster version and requires to update the cluster too. Please visit https://keptn.sh/docs/0.7.x/operate/upgrade/ for more information.
```

christian-kreuzberger-dtx (approved):
lgtm

PR #2403, opened 2020-09-24
#2387 API Endpoint for triggering evaluations

### LINKED ISSUE #2387: Introduce an endpoint in the api-service, which allows to trigger an evaluation
As a user, I would like to have a dedicated API endpoint that allows to start an evaluation. More precisely, currently, the `/event` endpoint can be used to start evaluations but this is too generic and, hence, complicated to use. Instead, as a user, I would like to have a dedicated endpoint, which allows triggering an evaluation and specifying the start, end, and timeframe as parameters similar to the CLI.

**Tasks:**
- Introduce a POST endpoint /project/{projectName}/stage/{stageName}/service/{serviceName}/evaluation/ with the following parameters:
    - start: The starting point from which to start the evaluation in UTC
    - end: The endpoint to which the evaluation data should be gathered in UTC (can not be used together with --timeframe)
    - timeframe: The time frame from which the evaluation data should be gathered (can not be used together with --end)
    All those parameters should be optional and there should be the same defaults as in the CLI (e.g. do the evaluation for the last 5 minutes)
- This endpoint returns the keptnContext and token (similar to `/event`)
- The api-service creates the appropriate CloudEvent and sends it to the event-broker (for 0.7.2) and to NATS for (0.8.0)
- Use this endpoint in the CLI command `keptn send event start-evaluation`
- The api-service has to check whether the project, service and stage exists.

**Note:** This endpoint does not return the evaluation data but only triggers the evaluation.

**Definition of Done:**
- A user can trigger an evaluation with `curl -X POST "HOST/api/v1/project/{projectName}/stage/{stageName}/service/{serviceName}/evaluation" -H "accept: application/json" -H "x-token: TOKEN"`

bacherfl:
Closes #2387
This PR introduces a new API endpoint that allows to trigger an evaluation using the following payload:

```
POST HOST/api/v1/project/{projectName}/stage/{stageName}/service/{serviceName}/evaluation

Payload:
{
    "from": "2006-01-02T15:04:05", // required
    "to": "2006-01-02T15:04:05", // can not be used in combination with 'timeframe'
    "timeframe": "5m", // can not be used in combination with ' to',
   "labels": {
        "test-id": "1234"
    }
}
```

This Endpoint also checks if the given project/stage/service do exist and give immediate feedback


christian-kreuzberger-dtx (approved):
lgtm

johannes-b:
Before merging, I would like to ask 2 questions:
- Do we run into a mismatch with the CLI? Because the CLI knows the flags: `--start` and `--end` and for this endpoint we call it `--from` and `--to`
- Does the endpoint evaluate the last 5 minutes when nothing is provided? (Like the CLI: https://github.com/keptn/keptn/blob/master/cli/cmd/send_event_evaluationStart.go#L155)

bacherfl:
> Before merging, I would like to ask 2 questions:
> 
> * Do we run into a mismatch with the CLI? Because the CLI knows the flags: `--start` and `--end` and for this endpoint we call it `--from` and `--to`
> * Does the endpoint evaluate the last 5 minutes when nothing is provided? (Like the CLI: https://github.com/keptn/keptn/blob/master/cli/cmd/send_event_evaluationStart.go#L155)

This has been taken into consideratino in the implementatino of the CLI

Issue #2305, opened 2020-09-07
Evaluation tile rework: from the overall result to the details


johannes-b:
**Status quo:**

Currently, the evaluation tile shows the HeatMap on the top followed by the overall result of the evaluation (horizontal bar). As a user, I would like to read the provided information from the overall view and then into the details.

![image](https://user-images.githubusercontent.com/729071/92399596-d6a0bc00-f12a-11ea-911b-d9ec4e282918.png)

**Proposal for restuctured view:**
![image](https://user-images.githubusercontent.com/729071/92399528-b7a22a00-f12a-11ea-9945-ca87f457d42d.png)

**Tasks:**
- [x] Move the Score (Bar) below the Heatmap
- [ ] ~~Display the limits for warning/failed in the bar~~ (Not supported by `<dt-bar-indicator>`)
- [x] Show the actual values for warning/failed (see example: 75 < 87.5 < 90) below the bar - see: #1661 
- [x] If the evaluation result is: "No evaluation performed ..." then hide: HeatMap/Chart and SLI Breakdown - see: #2134 -> implementing this feature has an impact on #2277
- [x] Show the comparison strategy as specified in the SLO.yaml
- [ ] ~~In the SLI breakdown, mark key-SLIs as bold~~ (Not provided in payload, should be provided by `lighthouse-service`)
- [x] When the HeatMap is selected, remove the headline `Heatmap` and `Evaluation results`

(Out-of-scope:)
- The SLI breakdown in a table format
- The indicators (and shadowing) in the HeatMap

PR #2406, opened 2020-09-24
#2305 Evaluation tile rework

ermin-muratovic:
- Moved the Score (Bar) below the Heatmap, showing the limits for warning/failed
- Showing the comparsion strategy as specified in the SLO.yaml
- Removed and rearranged some labels/infos

image


johannes-b:
If there is the result warning, can we change the "less" to a "less or equals".

image

should be: 50 <= 50 < 90

ermin-muratovic:
> If there is the result warning, can we change the "less" to a "less or equals".
> 
> image
> 
> should be: 50 <= 50 < 90

Yes, changed the "less" to a "less or equals". Also in case the evaluation is passed, it shows "less or equal" the pass limit.

johannes-b (approved):
lgtm

Issue #2300, opened 2020-09-07
Keptn CLI multiple contexts like KUBECONFIG


christian-kreuzberger-dtx:
Often users need to switch their kube context (e.g., `kubectl config use-context my-cluster-name`, or `
gcloud container clusters get-credentials ...` or `export KUBECONFIG=/etc/rancher/k3s/k3s.yaml`), but they end up still being connected to previous Keptn installation.

To overcome this, users need to manually authenticate the Keptn CLI with the correct Kubernetes cluster, as detailed here: https://keptn.sh/docs/0.7.x/reference/cli/#authenticate-keptn-cli

This could be as easy as `keptn auth --endpoint=$KEPTN_ENDPOINT --api-token=$KEPTN_API_TOKEN` on Linux (where as `KEPTN_ENDPOINT` and `KEPTN_API_TOKEN` can be retrieved via `kubectl` commands), but requires some manual work.

Another use-case that appears is that users wants to share their Keptn credentials (endpoint and api-token) with another user, without having to share access to the Kubernetes cluster. Therefore the user might want to be able to *export* the credentials, as well as re-use the credentials via an environment variable, e.g., 

```console
export KEPTNCONFIG=./keptn/workshop-credentials-1234
keptn send-event ...
```

# Goal

Implementation of this issue should accomplish the following: 

* make it easier for users to authenticate to a Keptn cluster by just configuring the Kubernetes context
* enable the user to switch between multiple Keptn installations (stored in config files locally)

# Definition of Done

- [ ] Introduce a config file containing the API Token and endpoint (e.g., using cobra) - similar to what `~/.keptn/.keptn` contains on Linux, but I'd prefer a format in JSON or YAML
- [ ] Introduce a command to export the current API Token and endpoint into a config file, e.g., `keptn auth --export`)
- [ ] CLI checks if `KEPTNCONFIG` is set, and uses the config file provided in this environment variable
- [ ] If `KEPTNCONFIG` is not set, the CLI should fall back to the existing mechanism (using credentialmanager for linux/darwin/windows)
- [ ] When executing `keptn auth` without any parameters and without `KEPTNCONFIG`, the CLI should try to do its best to try to authenticate with the Kubernetes cluster that `kubectl` is currently using. 
    - In case of multiple Keptn installations on the same cluster, the CLI should list all of them separated by ID/namespace/whatever, and provide a parameter to the `keptn auth` command to select one (e.g., `keptn auth --namespace=keptn-demo`)
[...]

PR #2417, opened 2020-09-27
Added support to export endpoint and api-token

ankitjain28may:
As the feature request in this issue #2300, I have added the support for exporting the current configuration i.e api-token and endpoint.

The configuration is exported to stdout due to undefined structure of config file based on multi-cluster support.

### DIFF
--- a/cli/cmd/auth.go
+++ b/cli/cmd/auth.go
@@ -16,9 +16,13 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var endPoint *string
-var apiToken *string
-var exportConfig *bool
+type authCmdParams struct {
+	endPoint     *string
+	apiToken     *string
+	exportConfig *bool
+}
+
+var authParams *authCmdParams
 var exportEndPoint url.URL
 var exportAPIToken string
 
@@ -35,30 +39,26 @@ More precisely, the Keptn CLI stores the endpoint and API token using *pass* in 
 	`,
 	Example:      `keptn auth --endpoint=https://api.keptn.MY.DOMAIN.COM --api-token=abcd-0123-wxyz-7890`,
 	SilenceUsage: true,
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		return verifyAuthParams(authParams)
+	},
 	RunE: func(cmd *cobra.Command, args []string) error {
 
 		var err error
-		if *exportConfig {
+		// User wants to print current auth credentials
+		if *authParams.exportConfig {
 			exportEndPoint, exportAPIToken, err = credentialmanager.NewCredentialManager().GetCreds()
 			if err != nil {
 				return err
 			}
-			fmt.Println("Endpoint = ", exportEndPoint.String())
-			fmt.Println("API Token = ", exportAPIToken)
+			fmt.Println("Endpoint: ", exportEndPoint.String())
+			fmt.Println("API Token: ", exportAPIToken)
 			return nil
-		}
-
-		if *endPoint == "" && *apiToken == "" {
-			return errors.New("required flag(s) \"api-token\", \"endpoint\" not set")
-		} else if *endPoint == "" {
-			return errors.New("required flag \"endpoint\" not set")
-		} else if *apiToken == "" {
-			return errors.New("required flag \"api-token\" not set")
 		}
 
 		logging.PrintLog("Starting to authenticate", logging.InfoLevel)
 
-		url, err := url.Parse(*endPoint)
+		url, err := url.Parse(*authParams.endPoint)
 		if err != nil {
 			logging.PrintLog("Error parsing Keptn API URL", logging.InfoLevel)
 			return err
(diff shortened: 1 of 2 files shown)


cli/cmd/auth.go
```diff
@@ -32,6 +36,26 @@ More precisely, the Keptn CLI stores the endpoint and API token using *pass* in
...
 	RunE: func(cmd *cobra.Command, args []string) error {
+
+		var err error
+		if *exportConfig {
+			exportEndPoint, exportAPIToken, err = credentialmanager.NewCredentialManager().GetCreds()
+			if err != nil {
+				return err
+			}
+			fmt.Println("Endpoint = ", exportEndPoint.String())
+			fmt.Println("API Token = ", exportAPIToken)
```
christian-kreuzberger-dtx, line 47: Please use colons instead of equals, and adapt the output as follows:
```
			fmt.Println("Endpoint: ", exportEndPoint.String())
			fmt.Println("API Token: ", exportAPIToken)
```

cli/cmd/auth.go
```diff
@@ -32,6 +36,26 @@ More precisely, the Keptn CLI stores the endpoint and API token using *pass* in
 	Example:      `keptn auth --endpoint=https://api.keptn.MY.DOMAIN.COM --api-token=abcd-0123-wxyz-7890`,
 	SilenceUsage: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
+
+		var err error
+		if *exportConfig {
```
christian-kreuzberger-dtx, line 41: Please add a comment here, e.g., 
```
// User wants to print current auth credentials
```

cli/cmd/auth.go
```diff
@@ -32,6 +36,26 @@ More precisely, the Keptn CLI stores the endpoint and API token using *pass* in
...
+		}
+
+		if *endPoint == "" && *apiToken == "" {
+			return errors.New("required flag(s) \"api-token\", \"endpoint\" not set")
+		} else if *endPoint == "" {
+			return errors.New("required flag \"endpoint\" not set")
+		} else if *apiToken == "" {
+			return errors.New("required flag \"api-token\" not set")
+		}
+
```
christian-kreuzberger-dtx, line 58: We needed a similar check in `keptn configure bridge [--output]`. 

https://github.com/keptn/keptn/blob/7d5fb6e6174a13c7fdfe45650266d18a1fab6780/cli/cmd/configure_bridge.go#L162-L165

Can you refactor the `keptn auth` command such that it also uses a `PreRun` check?

https://github.com/keptn/keptn/blob/7d5fb6e6174a13c7fdfe45650266d18a1fab6780/cli/cmd/configure_bridge.go#L43-L45

christian-kreuzberger-dtx:
FYI, tests are failing for `keptn auth` command: 
https://github.com/keptn/keptn/blob/7d5fb6e6174a13c7fdfe45650266d18a1fab6780/cli/cmd/auth_test.go#L17-L33

ankitjain28may:
@christian-kreuzberger-dtx Thanks for the review, will fix them :)

cli/cmd/auth.go
ankitjain28may: Sure, I thought to use `PreRun` check, will refactor

ankitjain28may:
@christian-kreuzberger-dtx I have made the changes, please review it

christian-kreuzberger-dtx (approved):
thanks, lgtm

cli/cmd/auth.go
```diff
@@ -87,11 +111,27 @@ More precisely, the Keptn CLI stores the endpoint and API token using *pass* in
...
+	if *authParams.exportConfig {
+		return nil
+	}
+
+	if authParams.endPoint == nil && authParams.apiToken == nil {
+		return errors.New("required flag(s) \"api-token\", \"endpoint\" not set")
+	} else if authParams.endPoint == nil {
+		return errors.New("required flag \"endpoint\" not set")
+	} else if authParams.apiToken == nil {
+		return errors.New("required flag \"api-token\" not set")
```
christian-kreuzberger-dtx, line 132: Actually, there is a little problem here, you're comparing to `nil`, but the default value (defined above) is `""`

christian-kreuzberger-dtx (changes requested):
Sorry, have to revert my LGTM, I found a small bug

cli/cmd/auth.go
ankitjain28may: Yes, that's because while mocking credentialstore, we are passing `""` for endpoint and api-token hence, my tests were getting failed before. I am checking `nil` so I can validate where the flag is passed or not.

ankitjain28may: @christian-kreuzberger-dtx Have made the changes, please check, Thanks :)

christian-kreuzberger-dtx: I only mentioned it becaise I ran into this problem when run `keptn auth` without params:

```
$ ./keptn auth
Starting to authenticate
Failed to resolve hostname 
Retrying...
```

And this is what I got with 0.7.1:
```
$ keptn auth
```

ankitjain28may: Yes, fixed this

christian-kreuzberger-dtx (approved):
Thanks, LGTM :)

Issue #2405, opened 2020-09-24
Add build and run targets in the Makefile


supra08:
Currently, the Makefile available in the `keptn` codebase has no targets, and building different components such as the bridge, CLI, etc is a manual task.

A Makefile containing commands for building all of these targets will come handy for quick setup and development. Besides building stuff, this Makefile can be used as a general-purpose task runner. To name a few:
* Format using `gofmt`
* Various tests such as e2e and unit
* Code lint
* Prepare-for-pr that would combine some of the above options to make the committed code ready for PR.
* Prepare for release.
This list is not exhaustive and a number of things can be added more. This can also handle better management of the actual `keptn` binary and the development binary.

supra08:
Also, I would like to be assigned and take this up.

christian-kreuzberger-dtx:
Thanks, assigned!

PR #2418, opened 2020-09-27
#2405 Add build and run targets in the Makefile 

supra08:
This PR adds functionality to the Makefile to speed up the local development process. Some of the features and newer additions in this needs discussion. The following commands have currently been added to the Makefile:
* `build-cli`: For building the CLI binary and saving it in a `./bin` directory.
* `start-bridge`: For starting the bridge server
* `fmt`
* `lint`
* `clean`
* `test-unit-cli`: For running unit tests on the CLI
* `prepare-for-pr`: To format, lint and check for commits to prepare it for PR.

Some points that are open for discussion are:
* Should there be commands to deploy the different services as well? This is because all of them are mostly `kubectl -f ...`
* How should the end-to-end tests be tackled in the Makefile? Since they are a set of scripts, what should be the best approach?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ UTILDIR := $(PROJECTROOT)/make-scripts/utils
 SPINNER := $(UTILDIR)/spinner.sh
 BUILDIR := $(PROJECTROOT)/make-scripts/build
 
-UNITTESTCOMMAND := $(shell cd cli/cmd/; go test -v)
+CREATEBIN := $(shell [ ! -d ./bin ] && mkdir bin)
 
 # Make is verbose in Linux. Make it silent.
 MAKEFLAGS += --silent
@@ -63,11 +63,12 @@ clean:
 ## Run unit tests on the CLI
 test-unit-cli:
 	@printf "⚙️ Running unit tests on the CLI\n" 
-	@$(UNITTESTCOMMAND)
+	@./make-scripts/run_cli_test.sh
 	@printf "👍 Done\n"
 
 ## Prepare code for PR
-prepare-for-pr: fmt lint test-unit-cli
+prepare-for-pr: fmt lint
+	@printf "❗️ Remember to run the tests"
 	@git diff-index --quiet HEAD -- ||\
 	(echo "-----------------" &&\
 	echo "NOTICE: There are some files that have not been committed." &&\
@@ -79,6 +80,17 @@ prepare-for-pr: fmt lint test-unit-cli
 	exit 0)
 
 help:
-	@printf "Keptn"
-	@printf "Help is coming soon..."
-	@printf ""
+	@echo "KEPTN"
+	@echo ""
+	@echo "* build-cli: Build the keptn cli and save it in bin/"
+	@echo "* start-bridge: Start the bridge server"
+	@echo "* install-helm: Install the helm binary in your local"
+	@echo "* install-golint: Install golint for linting the code"
+	@echo "* fmt: Formats the codebase"
+	@echo "* lint: Lints the codebase"
+	@echo "* clean: Cleans the build cache"
+	@echo "* test-unit-cli: Run unit tests on the Keptn CLI"
+	@echo "* prepare-for-pr: Makes the code ready for PR by formatting, linting and checking for uncommitted files"
+	@echo ""
+	@echo "Please visit https://keptn.sh for more information."
+	@echo "Get in touch with us via Slack: https://slack.keptn.sh"
--- a/Makefile
+++ b/Makefile
@@ -1,1 +1,84 @@
-target: ;
+PROJECTNAME := $(shell basename "$(PWD)")
+
+# Go related variables.
+PROJECTROOT := $(shell pwd)
+GOBIN := $(PROJECTROOT)/bin
+
+# Shell script related variables.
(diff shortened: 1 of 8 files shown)


christian-kreuzberger-dtx:
Before I review the code, here are my comments on the PR description:

>     * `build-cli`: For building the CLI binary and saving it in a `./bin` directory.
:+1:

>     * `start-bridge`: For starting the bridge server
IMHO not quite needed, but nice to have

>     * `fmt`
:+1:
 
>     * `lint`
:+1:

>     * `clean`
:+1:

>     * `test-unit-cli`: For running unit tests on the CLI
:+1:

>     * `prepare-for-pr`: To format, lint and check for commits to prepare it for PR.
:+1:

> Some points that are open for discussion are:
> 
>     * Should there be commands to deploy the different services as well? This is because all of them are mostly `kubectl -f ...`
Good catch. We try to make use `skaffold` for that, but that's service specific. For now, I don't think this should be included in the makefile.

>     * How should the end-to-end tests be tackled in the Makefile? Since they are a set of scripts, what should be the best approach?
Don't include those for now. Our end-to-end tests are way too unstructured. We will come up with a better concept in the future (looking at Helm, Kubernetes and co as an inspiration).

make-scripts/start_bridge.sh
```diff
@@ -0,0 +1,19 @@
...
+fi
+
+cd ./bridge/
+
+# install the dependencies
+npm install
+
+# accept the API URL and API TOKEN as the user inputs
+read -p 'Enter API URL: ' API_URL
+read -sp 'Enter API Token: ' API_TOKEN
```
christian-kreuzberger-dtx, line 16: FYI, the API_TOKEN can be left out and it will be automatically pulled via `kubectl` if not specified.

Makefile
```diff
@@ -1 +1,84 @@
-target: ;
\ No newline at end of file
+PROJECTNAME := $(shell basename "$(PWD)")
+
+# Go related variables.
+PROJECTROOT := $(shell pwd)
+GOBIN := $(PROJECTROOT)/bin
```
christian-kreuzberger-dtx, line 5: Can you make sure this directory always exists?
Else `make build-cli' will fail.

make-scripts/start_bridge.sh
```diff
@@ -0,0 +1,19 @@
...
+    exit
+fi
+
+cd ./bridge/
+
+# install the dependencies
+npm install
+
+# accept the API URL and API TOKEN as the user inputs
+read -p 'Enter API URL: ' API_URL
```
christian-kreuzberger-dtx, line 15: Can you make it such that if `API_URL` is already set, you just print it. Else, you use `read ...`

Makefile
```diff
@@ -1 +1,84 @@
...
+	echo "-----------------\n" &&\
+	git status &&\
+	echo "\n-----------------" &&\
+	echo "NOTICE: There are some files that have not been committed." &&\
+	echo "-----------------\n"  &&\
+	exit 0)
+
+help:
+	@printf "Keptn"
+	@printf "Help is coming soon..."
```
christian-kreuzberger-dtx, line 83: Maybe

```
Please visit https://keptn.sh" for more information. 
Get in touch with us via Slack: https://slack.keptn.sh
```

Makefile
supra08: yeah missed this. there should be a `mkdir bin` in the beginning.

make-scripts/start_bridge.sh
supra08: sure 👍 
Well is it a good idea to print the API_TOKEN? just displaying that it exists maybe.

Makefile
supra08: Yeah, I was thinking, it will contain the descriptions of all the commands and then this message at the end:
```
Please visit https://keptn.sh" for more information. 
Get in touch with us via Slack: https://slack.keptn.sh
```

make-scripts/start_bridge.sh
supra08: So whether to pull it or not depends on whether the `API_TOKEN`  env var exists or not, right?

christian-kreuzberger-dtx: The check is formulated in JavaScript:

https://github.com/keptn/keptn/blob/7d5fb6e6174a13c7fdfe45650266d18a1fab6780/bridge/server/app.js#L9-L23

I'm pretty sure javascript will evaluate an empty string as false, so it doesn't matter if the variable exists or not. So in the best case you do not need to handle this case in the code, just add a note here :)

make-scripts/start_bridge.sh
christian-kreuzberger-dtx: You're right, just print an info "Using API_TOKEN environment variable" without printing the actual value.

Makefile
christian-kreuzberger-dtx: :+1:

make-scripts/start_bridge.sh
supra08: sure!

supra08:
@christian-kreuzberger-dtx I've made the changes, please take a look.

christian-kreuzberger-dtx (approved):
LGTM

PR #2419, opened 2020-09-28
#2412 CLI: print keptn cluster version

### LINKED ISSUE #2412: Improve 'Keptn version' cli output with Keptn Api Version
# Improve 'keptn version' cli output 
When calling the 'keptn version' command on the cli it only prints out the client version and no information about the server that is authenticated to. Eventhough it states that automatic version is enabled, we should not assume that the server equals the client version if this is not printed out.

## Motivation
* Which use-cases does this KEP enable?
I believe is essential to print out the version of the client and server in technologies with fast development such as Keptn. Take as an example kubectl, docker, istioctl or alike.

## Explanation
As seen in the following output we can't tell the server version. I just upgraded the installation but I can't assume that the CLI version matches the server version.

```bash
keptn version

Keptn CLI version: 0.7.1

```
as an Example we can take a look at kubectl version command

```bash
kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:56:40Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.8", GitCommit:"9f2892aab98fe339f3bd70e3c470144299398ace", GitTreeState:"clean", BuildDate:"2020-08-13T16:04:18Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}

```
or docker

```
docker version
Client: Docker Engine - Community
 Version:           19.03.12
 API version:       1.40
 Go version:        go1.13.10
 Git commit:        48a66213fe
 Built:             Mon Jun 22 15:41:33 2020
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.12
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.10
  Git commit:       48a66213fe
  Built:            Mon Jun 22 15:49:27 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

anukul:
Closes #2412

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -115,7 +115,7 @@ func runDailyVersionCheck() {
 	vChecker := version.NewVersionChecker()
 	cliChecked, cliMsgPrinted = vChecker.CheckCLIVersion(Version, true)
 
-	keptnVersion, err := getInstalledKeptnVersion()
+	keptnVersion, err := getKeptnServerVersion()
 	if err != nil {
 		logging.PrintLog(err.Error(), logging.InfoLevel)
 	} else {
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -110,19 +110,20 @@ func (s *options) appendIfNotEmpty(newOption string) {
 }
 
 func runDailyVersionCheck() {
+	var cliMsgPrinted, cliChecked, keptnMsgPrinted, keptnChecked bool
+
 	vChecker := version.NewVersionChecker()
-	cliMsgPrinted, cliChecked := vChecker.CheckCLIVersion(Version, true)
+	cliChecked, cliMsgPrinted = vChecker.CheckCLIVersion(Version, true)
 
 	keptnVersion, err := getInstalledKeptnVersion()
 	if err != nil {
 		logging.PrintLog(err.Error(), logging.InfoLevel)
-		return
+	} else {
+		kvChecker := version.NewKeptnVersionChecker()
+		keptnChecked, keptnMsgPrinted = kvChecker.CheckKeptnVersion(Version, keptnVersion, true)
 	}
-	kvChecker := version.NewKeptnVersionChecker()
-	keptnMsgPrinted, keptnChecked := kvChecker.CheckKeptnVersion(Version, keptnVersion, true)
 
 	if cliMsgPrinted || keptnMsgPrinted {
-		fmt.Println(updateInfoMsg)
 		if len(os.Args) > 0 {
 			fmt.Printf(disableVersionCheckMsg+"\n", os.Args[0])
 		} else {
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -44,8 +44,7 @@ func Execute() {
 	currentLogLevel := logging.LogLevel
 	logging.LogLevel = logging.QuietLevel
 
-	vChecker := version.NewVersionChecker()
-	vChecker.CheckCLIVersion(Version, true)
+	runDailyVersionCheck()
 
 	// Set LogLevel back to previous state
 	logging.LogLevel = currentLogLevel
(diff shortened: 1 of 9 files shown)


cli/cmd/version.go
```diff
@@ -34,45 +34,110 @@ const versionCheckInfo = `Daily version check is %s.
...
 		}
-		if cliConfig.AutomaticVersionCheck {
-			fmt.Printf(versionCheckInfo, "enabled", "", "")
-			fmt.Printf(enableVersionCheckMsg, "disable", "false")
-		} else {
-			fmt.Printf(versionCheckInfo, "disabled", " not", " not")
-			fmt.Printf(enableVersionCheckMsg, "enable", "true")
+
+		// Keptn
+		keptnVersion, err := getInstalledKeptnVersion()
```
christian-kreuzberger-dtx, line 75: This is nice, I didn't know we have this function. For future reference: This function uses the `helm history` command to find out the latest Keptn helm release on the current cluster.

cli/cmd/version.go
```diff
@@ -34,45 +34,110 @@ const versionCheckInfo = `Daily version check is %s.
...
+		if err != nil {
+			logging.PrintLog(err.Error(), logging.InfoLevel)
+			return
+		}
+		fmt.Println("\nKeptn cluster version: " + keptnVersion)
```
christian-kreuzberger-dtx, line 80: Please be careful with the return statement and the if-else block here.

If fetching the Keptn Cluster version info failed (e.g., because the user does not have access to the K8s cluster), we can still continue the remaining code (except for printing the Keptn cluster version, obviously).



Right now the output would be like this:
```
Keptn CLI version: develop
Failed to check if Keptn release is available in namespace keptn: ${ERROR_MESSAGE}

```

But in my opinion this would be okay:
```
Keptn CLI version: develop
Keptn cluster version:  Failed to check if Keptn release is available in namespace keptn: ${ERROR_MESSAGE}


```

cli/cmd/root.go
```diff
@@ -109,3 +108,28 @@ func (s *options) appendIfNotEmpty(newOption string) {
...
+
+func runDailyVersionCheck() {
+	vChecker := version.NewVersionChecker()
+	cliMsgPrinted, cliChecked := vChecker.CheckCLIVersion(Version, true)
+
+	keptnVersion, err := getInstalledKeptnVersion()
+	if err != nil {
+		logging.PrintLog(err.Error(), logging.InfoLevel)
+		return
+	}
```
christian-kreuzberger-dtx, line 120: Same as with my comment down below: We don't necessarily need to `return` here, just keep in mind that we can not check for a new version if keptnVersion is nil or something like that ;)

christian-kreuzberger-dtx:
Thanks for the PR, I've left some comments.

anukul:
Thank you for reviewing. I have made the suggested changes, fixed merge conflicts and added tests.

anukul:
@christian-kreuzberger-dtx I changed this to use [`apiHandler.GetMetadata()`](https://github.com/keptn/go-utils/blob/32aece7691fd1f2e66f35ef4bf3d7e14ace4ccba/pkg/api/utils/apiUtils.go#L150), because I just noticed that [`getInstalledKeptnVersion()`](https://github.com/keptn/keptn/blob/master/cli/cmd/upgrade.go#L132) uses `upgradeParams.Namespace` while fetching the helm release. In the current state, this flag would always be hardcoded to the default value `keptn` for the version command.

Also, using `keptn auth` credentials here will allow us to display the correct version when we allow switching namespace & context later.

christian-kreuzberger-dtx (approved):
Thanks, very nice! LGTM

Issue #2389, opened 2020-09-22
Bridge: Heatmap highlights evaluation results that are used for comparison


johannes-b:
With the list of event-id provided by #2388, it is possible to highlight the evaluation results that were used for comparison. 

_Example:_

Assuming the following comparison strategy is configured: 
```
comparison:
  compare_with: "several_results"
  number_of_comparison_results: 3
  include_result_with_score: "pass"
  aggregate_function: "avg"
```

Then, the lighthouse-service returns a list of three event-IDs (the last 3 passed evaluations). 

In the heatmap, this information should be displayed as shown below. Please use a highlighting mechanism for the comparison values that is most appropriate and does not hurt the look-and-feel.

![image](https://user-images.githubusercontent.com/729071/93889482-878d8600-fce9-11ea-8141-d8748c82404b.png)

PR #2424, opened 2020-09-28
#2389 Highlight comparedEvents in heatmap

ermin-muratovic:
If the comparedEvents are provided, those events are highlighted in the heatmap, so that the user now can easily see the selected event and with which it was compared with.

image


christian-kreuzberger-dtx (approved):
LGTM - please back port to release-0.7.2

christian-kreuzberger-dtx:
actually, tests are failing, let's fix them first ;)

johannes-b (approved):

Issue #2131, opened 2020-07-27
Bridge: Show buildId also in Heatmap and not just in charts


grabnerandi:
Since Keptn 0.7 the bridge not only looks at the evaluation timestamp, but is also able to use a label called `buildId` to display in "Chart Mode" as the label on the x-axis. See image:

![image](https://user-images.githubusercontent.com/675886/88558341-a940f880-d02b-11ea-8320-4b8734a7ebcb.png)

**Idea:** This information should also be used in the HeatMap as it makes navigating through the data easier. I can envision to either have it as part of the score heatmap tile or as part of the header information above the table - see here:

![image](https://user-images.githubusercontent.com/675886/88558259-94fcfb80-d02b-11ea-8a53-6f14e2897428.png)

johannes-b:
Based on feedback from @ermin-muratovic, this information (the buildId) should be placed on the horizontal axis of the heatmap. (no tooltip or overlay in the heatmap) 

Like it is implemented for the chart:
![image](https://user-images.githubusercontent.com/729071/93893143-9aa25500-fced-11ea-8ab6-fa3bf7916577.png)

PR #2432, opened 2020-09-29
#2131 Show labels on xAxis in Heatmap

ermin-muratovic:
The Heatmap has now labels on the xAxis (either the timestamp or the `buildId` label if it is provided on the event payload is used)

image

### DIFF
--- a/bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.ts
+++ b/bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.ts
@@ -314,9 +314,9 @@ export class KtbEvaluationDetailsComponent implements OnInit, OnDestroy {
         let categories = series.data
           .sort((a, b) => moment(a.evaluationData.time).unix() - moment(b.evaluationData.time).unix())
           .map((item, index, items) => {
-            let uniqueIndex = items.filter(c => c.evaluationData.getHeatmapLabel() == item.evaluationData.getHeatmapLabel()).indexOf(item);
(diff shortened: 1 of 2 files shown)


christian-kreuzberger-dtx (approved):
LGTM - please backport to release-0.7.2

johannes-b:
When testing this feature on my Keptn, the Heatmap collapsed: 
image

ermin-muratovic:
@johannes-b you were right, in case that there are just a few (1-2) SLIs, the height was not calculated correctly. My last commit fixed that issue.

image

ermin-muratovic:
We have one more issue with the labels, because they have to be unique string values. In case there are multiple evaluations with the same `buildId` label, the heatmap won't show the results correctly.
Therefore I will have to find a solution, to make those labels unique. Maybe by adding a seq number at the end, in case there are duplicate values.

ermin-muratovic:
The label is now appended with a sequential number in case that there are duplicates in the current set of evaluations.

image

johannes-b:
I would actually start to add (1) to the first occurrence of a duplicate. 

image

Resulting in:
- build-16 (1)
- build-16 (2)
- build-16 (3)

@ermin-muratovic what do you think?

ermin-muratovic:
@johannes-b yeah, makes sense. adding now the seq. number from the first occurrence of a duplicate value, see screenshot below.

image

johannes-b (approved):
lgtm

johannes-b:
Please provide the backport to 0.6.2 and an update of the keptn.sh/docs.

PR #2446, opened 2020-09-30
#2445 Added endpoint in output while keptn auth

### LINKED ISSUE #2445: Improve CLI auth command by including endpoint in output
When authenticating the Keptn CLI to the Keptn cluster, I often use something like:

```
KEPTN_ENDPOINT=http://<ENDPOINT_OF_API_GATEWAY>/api
KEPTN_API_TOKEN=$(kubectl get secret keptn-api-token -n keptn -ojsonpath={.data.keptn-api-token} | base64 --decode)

keptn auth --endpoint=$KEPTN_ENDPOINT --api-token=$KEPTN_API_TOKEN
```
Output:
```
Starting to authenticate
Successfully authenticated
```

It would be great if the output after this command also includes the `$KEPTN_ENDPOINT`. I know it is easy to retrieve it by `echo $KEPTN_ENDPOINT` but it is an extra step that I would like to get rid of.

The output of `keptn status` would be the right solution in my mind:
```
$ keptn status
Starting to authenticate
Successfully authenticated
CLI is authenticated against the Keptn cluster http://91.xxx.1xx.182:31090/api
```

**Definition of Done:**
- [ ] `keptn auth` shows the endpoint (like the `keptn status` command)

ankitjain28may:
Closes #2445

### DIFF
--- a/cli/cmd/auth.go
+++ b/cli/cmd/auth.go
@@ -101,6 +101,7 @@ More precisely, the Keptn CLI stores the endpoint and API token using *pass* in 
 			}
 
 			logging.PrintLog("Successfully authenticated", logging.InfoLevel)
+			logging.PrintLog("CLI is authenticated against the Keptn cluster "+*authParams.endPoint, logging.InfoLevel)
 			return credentialmanager.NewCredentialManager().SetCreds(*url, *authParams.apiToken)
 		}
 
--- a/cli/cmd/auth.go
+++ b/cli/cmd/auth.go
@@ -100,8 +100,7 @@ More precisely, the Keptn CLI stores the endpoint and API token using *pass* in 
 				return fmt.Errorf("Authentication was unsuccessful - could not authenticate against the server.")
 			}
 
-			logging.PrintLog("Successfully authenticated", logging.InfoLevel)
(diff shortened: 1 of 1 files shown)


cli/cmd/auth.go
```diff
@@ -101,6 +101,7 @@ More precisely, the Keptn CLI stores the endpoint and API token using *pass* in
 			}
 
 			logging.PrintLog("Successfully authenticated", logging.InfoLevel)
+			logging.PrintLog("CLI is authenticated against the Keptn cluster "+*authParams.endPoint, logging.InfoLevel)
```
agrimmer, line 104: Thanks for your contribution!
Can we please combine those messages
`logging.PrintLog("Successfully authenticated against the Keptn cluster "+*authParams.endPoint, logging.InfoLevel)`

agrimmer (commented):
Thank you for your contribution! I have a small request.

cli/cmd/auth.go
ankitjain28may: @agrimmer Thanks for reviewing, have combined the messages :)

johannes-b:
We should file a backport to 0.7.2 since @jetzlstorfer requires it.

christian-kreuzberger-dtx:
I will handle the backport.

PR #2448, opened 2020-10-01
Disabled Microk8s 1.19 integration test

bacherfl:
The installation of Keptn via Helm currently fails on MicroK8s 1.19. There is already an issue on the microk8s repo: https://github.com/ubuntu/microk8s/issues/1605 
Until this is fixed we should disable the Integration tests for MicroK8s 1.19


christian-kreuzberger-dtx (approved):
LGTM

Issue #1863, opened 2020-06-02
Keptn support for multiple plans


pcjeffmac:
**Current situation:**
* If you install on EKS, Keptn install writes the authentication information to the  ~/.keptn/.keptn 
* If you then install on gcloud, you loose the authentication information to EKS. As a result, you would need to backup the ~/.keptn/.keptn manually, then revert or change this file for each cloud environment.

**Idea:**
Provide a way to switch from EKS to gcloud or any other vendor without loosing the authentication information.  

**Describe alternatives you've considered**
* I like the approach that kubectl offers by setting an environment variable KUBECONFIG, which can then point to a Keptn config file. So we could have KEPTNCONFIG which by default points to ~/.keptn/.keptn (that would already be the default for linux) and if the user chooses to overwrite it, it would look like this:
`KEPTNCONFIG=./my-cluster-keptn-config  | keptn send-event new-artifact ...`
* Creating a symlink to the necessary file with the relevant authentication information.

supra08:
I would like to take up this issue to gain some insight.
@christian-kreuzberger-dtx @johannes-b

christian-kreuzberger-dtx:
Reopening this, linking PR #2468

PR #2450, opened 2020-10-01
#1863 Add support for multiple plans

supra08:
In this PR, I've added support for multiple plans such as an EKS installation, GCloud installation, or a local installation.
In this, if the user wants to use a different set of API credentials, he can specify them by using an environment variable `KEPTNCONFIG` that points to the creds JSON file location.
The command looks like this:
```
KEPTNCONFIG=./keptn_test_creds.json ./keptn add-resource --project=keptn07project 
```
As mentioned before, the creds will be stored in a JSON file. The fields would be:
* `api_token`
* `endpoint`

A sample JSON file would look like this:
```
{
    "api_token": "blahblah",
    "endpoint": "http://localhost:3000/api"
}
```

### DIFF
--- a/cli/cmd/add_resource.go
+++ b/cli/cmd/add_resource.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"net/url"
 	"os"
 	"path"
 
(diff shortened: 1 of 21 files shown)


ankitjain28may:
Hi @supra08 This is great, just a little suggestion, it would be great if we incorporate `KEPTNCONFIG` file lookup logic in credentialmanager itself rather adding in all the files as its a single point from where every command fetches auth credentials.

supra08:
Yeah, I believe that's a better approach. I was working on the same in order to remove the code duplications.

christian-kreuzberger-dtx:
Thanks @ankitjain28may  and @supra08 , I was also going to comment on the duplications and whether we can centralize the approach. I'll take a look at the PR.

christian-kreuzberger-dtx (approved):
LGTM - this is awesome!

ankitjain28may:
@christian-kreuzberger-dtx @supra08 The PR still lacks the logic for the support of Mac and windows.

PR #2452, opened 2020-10-01
Fix npm vulnerabilities and update modules to latest compatible versions

ermin-muratovic:


christian-kreuzberger-dtx (approved):
LGTM - please also do the same for release-0.7.2 (though not sure if a backport works)

PR #2459, opened 2020-10-01
#2364 Support custom-values.yaml in keptn install

### LINKED ISSUE #2364: Support custom-values.yaml in keptn install
Currently `keptn install` can only be customized with different flags like:
--plattform
--use-case
--endpoint-serivice-type
etc.
which then are used as input for the helm client:

```go
values := map[string]interface{}{
		"continuous-delivery": map[string]interface{}{
			"enabled": installParams.UseCase == ContinuousDelivery,
			"openshift": map[string]interface{}{
				"enabled": *installParams.PlatformIdentifier == "openshift",
			},
		},
		"control-plane": map[string]interface{}{
			"enabled": true,
			"apiGatewayNginx": map[string]interface{}{
				"type": installParams.EndPointServiceType.String(),
			},
		},
	}

	if err := helm.NewHelper().UpgradeChart(keptnChart, keptnReleaseName, keptnNamespace, values); err != nil {
		logging.PrintLog("Could not complete Keptn installation: "+err.Error(), logging.InfoLevel)
		return err
	}
```

The problem is that underlying values for example `control-plane.prefixPath` can not be configured that way and one has to use keptn's helm charts for a custom deployment.

Example: PR #2172 

I would suggest to replace current flags related with the helm chart with one --values flag to accept a custom-values.yaml file. So you can use the keptn cli right away for new features and you would not depend on helm and there would be no additional development effort to bring new features to the cli.

If no values file is specified, fall back to keptn/installer/manifests/keptn/values.yaml.

Maybe add a custom-values-template.yaml with all customizable values to give a better overview on which parameters are experimental/stable. 

I tried to add a custom flag for the prefixPath to cobra, but helm had an issue with a missing value in the core.yaml file. Maybe a missing env variable.

**Additional context**

anukul:
Closes #2364

update: rebased on master


cli/cmd/install.go
```diff
@@ -259,3 +165,75 @@ func checkIstioInstallation() error {
...
+func isOpenShiftEnabled(values map[string]interface{}) (enabled bool) {
+	defer func() {
+		if recover() != nil {
+			// default chart value
+			enabled = false
+		}
+	}()
+	enabled = values["continuous-delivery"].(map[string]interface{})["openshift"].(map[string]interface{})["enabled"].(bool)
+	return
+}
```
anukul, line 228: This seems a bit hacky.
I chose it instead of hardcoding the complete values file (when not provided by user).
Since here we're only hardcoding the default values that we need, instead of the complete values file.

Please let me know if there is a better approach.

cli/cmd/install.go
```diff
@@ -181,6 +99,9 @@ func init() {
 
 	installParams.Namespace = installCmd.Flags().StringP("namespace", "n", "keptn",
 		"Specify the namespace where Keptn should be installed in (default keptn).")
+
+	installParams.ValuesFile = installCmd.Flags().StringP("values", "f", "",
+		"Specify values in a YAML file or a URL.")
```
anukul, line 104: If we want to make this more like helm, maybe we can allow the possibility of multiple values files, or additionally setting values from the command line using `--set` 

[source](https://github.com/helm/helm/blob/66034e403548417ac90d9769f5dcde2d354e0033/cmd/helm/flags.go#L41-L46)

cli/cmd/create_project.go
```diff
@@ -4,22 +4,26 @@ import (
...
+	keptncommon "github.com/keptn/go-utils/pkg/lib/keptn"
 	keptnutils "github.com/keptn/kubernetes-utils/pkg"
+
+	"github.com/keptn/keptn/cli/pkg/file"
+	"github.com/keptn/keptn/cli/pkg/websockethelper"
+
+	"github.com/asaskevich/govalidator"
 	"github.com/spf13/cobra"
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
```
anukul, line 26: See https://github.com/go-yaml/yaml/issues/139

christian-kreuzberger-dtx:
Thanks for the PR. Adding @agrimmer and @bacherfl for a review on this PR, as they know their way around the installer and helm ;)

cli/cmd/install.go
bacherfl: I think providing the option to pass values using `--set` is a good idea. @christian-kreuzberger-dtx @agrimmer what do you think?

christian-kreuzberger-dtx: Sounds good

anukul: Updated :)

cli/cmd/install.go
```diff
@@ -18,131 +18,66 @@ package cmd
...
 
 For more information, please follow the installation guide [Install Keptn](https://keptn.sh/docs/` + keptnReleaseDocsURL + `/operate/install/#install-keptn)
 `,
-	Example: `keptn install                                                        # install on Kubernetes
-
-keptn install --platform=openshift --use-case=continuous-delivery    # install continuous delivery on Openshift
-
-keptn install --platform=kubernetes --endpoint-service-type=NodePort # install on Kubernetes with gateway NodePort
-`,
+	Example: `keptn install --values=./my-values.yaml`,
```
bacherfl, line 50: @anukul can you please also add an example for setting values directly with `--set`?

jetzlstorfer: @anukul any updates on this?

PR #2460, opened 2020-10-02
#2430 Fixed broken link

### LINKED ISSUE #2430: Click on Keptn API > page not found
I installed a preview of Keptn 0.7.2 

When clicking on  `Keptn API`, I get forwarded to:
```
{"code":404,"message":"path / was not found"}
```

**Fix:**
- Remove / from the path

**Enhancement:**
- Can we please flip the links. 
  - Advantage: it then shows them in the same order like the copy/paste labels below. (goal: red, blue, red, blue arrows)
image

johannes-b:
Closes #2430

### DIFF
--- a/bridge/server/api/index.js
+++ b/bridge/server/api/index.js
@@ -30,7 +30,7 @@ module.exports = (params) => {
     try {
       const result = await axios({
         method: req.method,
-        url: `${apiUrl}/${req.url}`,
+        url: `${apiUrl}${req.url}`,
         headers: {
           'Content-Type': 'application/json'
         },
@@ -63,7 +63,7 @@ module.exports = (params) => {
     try {
       const result = await axios({
         method: req.method,
-        url: `${apiUrl}/${req.url}`,
+        url: `${apiUrl}${req.url}`,
         ...req.method!='GET' && { data: req.body },
         headers: {
           'x-token': apiToken,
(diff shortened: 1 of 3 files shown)


johannes-b:
The link works and I changed the order:

image

@ermin-muratovic Can I ask you for a review, please?

bridge/server/api/index.js
```diff
@@ -30,7 +30,7 @@ module.exports = (params) => {
     try {
       const result = await axios({
         method: req.method,
-        url: `${apiUrl}${req.url}`,
+        url: `${apiUrl}/${req.url}`,
```
johannes-b, line 33: @ermin-muratovic Is this change correct? (also see line 66)

ermin-muratovic: This `/` here is not necessary, because `req.url` starts with a `/` and therefore you would get a double `/` here.
And the `apiUrl` in the server-part actually has not changed. (same on line 66)

ermin-muratovic (changes requested):
Since the `apiUrl` in the server has not changed and usually is just the internal URL to the api service, that `/` there is not necessary.
The changes in the client are fine.

johannes-b:
Thanks for the feedback - I reverted the server code.

ermin-muratovic (approved):
lgtm

PR #2468, opened 2020-10-02
Added support for mac and windows as well

ankitjain28may:
Referring to #2450 which was missing the multiple plans using `KEPTNCONFIG` for mac and windows.

### DIFF
--- a/cli/pkg/credentialmanager/handler.go
+++ b/cli/pkg/credentialmanager/handler.go
@@ -71,6 +71,11 @@ func getCreds(h credentials.Helper) (url.URL, string, error) {
 	if MockAuthCreds {
 		return url.URL{}, "", nil
 	}
+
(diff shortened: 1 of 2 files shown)


christian-kreuzberger-dtx:
Thanks for bringing this up, I totally overlooked that in my review.

I'm wondering though, `handler.go` is just the generic implementation, we also have `handler_darwin.go` for MacOS, `handler_windows.go` for Windows.

ankitjain28may:
@christian-kreuzberger-dtx As both`handler_darwin.go` for MacOS, `handler_windows.go` for Windows are eventually calling the generic `handler.go` file which has the main logic so I added the `KEPTNCONFIG` lookup in that function.

christian-kreuzberger-dtx (approved):
LGTM

PR #2480, opened 2020-10-06
#2457 Show true number of compared events

### LINKED ISSUE #2457: Show the true count of compared evaluations in evaluation-tile
With PR https://github.com/keptn/keptn/pull/2406 we added this line to the evaluation details tile:

*Compared with last $x passed/... evaluations", whereas $x was taken from the respective SLO.yaml

https://github.com/keptn/keptn/blob/20482b9a6d8d4a17103d05da81f5d528b15fc64c/bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.html#L72

We should improve on that, as the CloudEvent now contains the evaluation-details cloudevent `comparedEvents` as a list, and we can use the length of that list.

bacherfl:
Closes #2457 

With this PR, the true number of compared evaluations is shown by using the length of the `comparedEvents` field in the evaluation details. Additionally, the `filter: objective` string has been removed.

**Before:**

bug_2457_before

**After (0 compared events):**

bug_2457_after_2

**One compared event:**

bug_2457_after_1


christian-kreuzberger-dtx (approved):
LGTM (please squash merge)

Issue #2444, opened 2020-09-30
Improve post-installation steps by including Keptn API endpoint IP


jetzlstorfer:
When installing Keptn with the option NodePort or LoadBalancer it would be useful to print the concrete `keptn auth` instructions after the installation. 

**Status quo:**
The instructions that are provided right now include a  `port-forward` which means that if the terminal is closed the Keptn CLI can not communicate to the Keptn installation anymore. 

*Current situation:*
```
$ keptn install --endpoint-service-type=NodePort
Helm Chart used for Keptn installation: https://storage.googleapis.com/keptn-installer/keptn-0.7.1.tgz

Please confirm that the provided cluster information is correct:
Cluster: keptn
Is this all correct? (y/n)
y
Installing Keptn ...
Start upgrading Helm Chart keptn in namespace keptn
Finished upgrading Helm Chart keptn in namespace keptn
Keptn has been successfully set up on your cluster.
---------------------------------------------------
* To quickly access Keptn, you can use a port-forward and then authenticate your Keptn CLI (in a Linux shell):
 - kubectl -n keptn port-forward service/api-gateway-nginx 8080:80
 - keptn auth --endpoint=http://localhost:8080/api --api-token=$(kubectl get secret keptn-api-token -n keptn -ojsonpath={.data.keptn-api-token} | base64 --decode)

* To expose Keptn on a public endpoint, please continue with the installation guidelines provided at:
 - https://keptn.sh/docs/0.7.x/operate/install#install-keptn
```

**Improvement:**
Since for NodePort and LoadBalancer the IPs can be retrieved via `kubectl` commands, it would be useful if this is done during the installation and the result is printed to the user at the end of the installation. 

*Desired solution (for NodePort and LoadBalancer installations):*
```
...
Finished upgrading Helm Chart keptn in namespace keptn
Keptn has been successfully set up on your cluster.
---------------------------------------------------
* To quickly access Keptn, you can use a port-forward and then authenticate your Keptn CLI (in a Linux shell):
 - keptn auth --endpoint=http://43.56.32.3:31090/api --api-token=xxx123
```

johannes-b:
Find the manual instructions (that should be automated by this improvement) here:
- NodePort: https://keptn.sh/docs/0.7.x/operate/install/#option-2-expose-keptn-via-a-nodeport
- LoadBalancer: https://keptn.sh/docs/0.7.x/operate/install/#option-1-expose-keptn-via-a-loadbalancer

[...]

PR #2486, opened 2020-10-06
#2444 Improved post installations steps for nodeport and loadbalancer service

ankitjain28may:
Closing #2444 

I have used `keptnutils.ExecuteCommand` for fetching details regarding endpoint/secrets which eventually uses `kubectl` binary, but later if we move out from using `kubectl` as a dependency for `keptn` we need to use kubeclient for fetching information regarding the same.


cli/cmd/install.go
```diff
@@ -259,3 +273,43 @@ func checkIstioInstallation() error {
...
+		}
+	} else if serviceType == "LoadBalancer" {
+		external, err := keptnutils.ExecuteCommand("kubectl", []string{"get", "svc", "api-gateway-nginx", "-n", keptnNamespace, "-o", "jsonpath='{.status.loadBalancer.ingress[0].ip}'"})
+		if err != nil {
+			return "", err
+		}
+		endpoint = strings.Trim(external, "'")
+		port = "80"
+	}
+	return "http://" + endpoint + ":" + port + "/api", nil
```
christian-kreuzberger-dtx, line 300: Please add some docs to this code (e.g., fetching external and internal node IP, getting port of api-gateway-nginx service)

cli/cmd/install.go
```diff
@@ -234,9 +235,22 @@ func doInstallation() error {
...
+		endpoint, err := getAPIEndpoint(keptnNamespace, installParams.EndPointServiceType.String())
+		apiToken, _ := getAPITokenFromSecret(keptnNamespace)
+		if err == nil {
+			fmt.Println("* To quickly access Keptn, you can authenticate your Keptn CLI (in a Linux shell):\n" +
+				" - keptn auth --endpoint=" + endpoint + " --api-token=" + apiToken + "\n")
+		} else {
+			errorFlag = true
+		}
+	}
+	if errorFlag {
```
christian-kreuzberger-dtx, line 249: What if `EndPointServiceType` is ClusterIP? We'll never see the message that's now hidden behind `if errorFlag`.

I recommend to turn this around by renaming  `errorFlag` into `showFallbackConnectMessage`, and set `showFallbackConnectMessage` to `false` when you have successfully fetched an endpoint and port.

cli/cmd/install.go
```diff
@@ -259,3 +273,43 @@ func checkIstioInstallation() error {
 
 	return nil
 }
+
+func getAPIEndpoint(keptnNamespace string, serviceType string) (string, error) {
+	var endpoint, port string
```
christian-kreuzberger-dtx, line 278: Regarding the `if`/`else if` statement: what if `serviceType` is something else? The function will always return
```
return "http://" + endpoint + ":" + port + "/api", nil
```

regardless of whether endpoint or port are set. I recommend returning string and error in this function, and in case that serviceType is unknown, return an error like "Unknown service-type %v".

christian-kreuzberger-dtx (changes requested):
Thanks for your PR, I've added some comments.

Also, do you think it's possible to add unit tests for `getAPIEndpoint` and `getAPITokenFromSecret` by mocking the `kubectl` calls? (paging @bacherfl on this)

cli/cmd/install.go
ankitjain28may: Thanks for raising this, my bad, I was thinking to implement this the other way around and ended up doing this. will fix this.

cli/cmd/install.go
ankitjain28may: Cool, as I was only calling this service for NodePort and Loadbalancer type, but I will return the error.

ankitjain28may:
@christian-kreuzberger-dtx I am not sure whether we can mock with `kubectl` binary but if we used `kubeclient` then its possible.

christian-kreuzberger-dtx:
Thanks for the update, I'll take a look

christian-kreuzberger-dtx:
Code looks good to me, I'll wait with my approval though.

The way this is working now displays the API URL and the Token, e.g.:
image

Before we merge this, one last question to @johannes-b and @jetzlstorfer (as you created the issue):
Do we REALLY REALLY want to show API URL and Token in the output of the `keptn install` command?

If you run `keptn install` on any system that prints log files to the public (e.g., Travis CI), anyone having access to those logs will get access to the Keptn installations (e.g., on Google Kubernetes Engine).

johannes-b:
So far, the enhancement looks good! 

However, @christian-kreuzberger-dtx brought up a very valid point because we would expose the IP address of our test cluster in our TavisCI logs. I took a look into our Travis scripts that install Keptn. They are: 
- keptn/test/test_install_gke.sh
- keptn/test/test_install_kubernetes_full.sh
- keptn/test/test_install_kubernetes_quality_gates.sh
- keptn/test/test_install_minishift_quality_gates.sh

What they all have in common is that the use `keptn install` with the `--creds` flag; without the `--creds` flag automation would not work since prompts are shown to the user. 

This brings me to the idea that we can do a fallback to the old output (as controlled by `showFallbackConnectMessage` in the code) whenever `--creds` is used. This would hide the IP when `keptn install` is part of an automation script.

@christian-kreuzberger-dtx what do you think?

ankitjain28may:
@johannes-b Just to add to it, if I am passing creds by environment variable then it will expose the api-token so I suggest let's don't show the api-token in the output.

johannes-b:
@ankitjain28may please forget my previous comment about re-using the `--creds` flag. This would confuse the Keptn user. 

Instead, I would propose to add a new flag to the `keptn install` command which is called: `--hide-sensitive-data`
When this flag is set, do **not** show the output below `Keptn has been successfully set up on your cluster`.

**Definition of Done:**
- [ ] Please add the flag `--hide-sensitive-data` to the `install` command to enable/disable the output
- [ ] Explain the flag with a comment in the code
- [ ] Please update the below test scripts to use this new flag:
   - keptn/test/test_install_gke.sh
   - keptn/test/test_install_kubernetes_full.sh
   - keptn/test/test_install_kubernetes_quality_gates.sh
   - keptn/test/test_install_minishift_quality_gates.sh

cli/cmd/install.go
```diff
@@ -233,10 +240,32 @@ func doInstallation() error {
...
+	if err != nil {
+		apiToken = "$(kubectl get secret keptn-api-token -n " + keptnNamespace + " -ojsonpath={.data.keptn-api-token} | base64 --decode)"
+	}
+
+	if installParams.EndPointServiceType.String() == "NodePort" || installParams.EndPointServiceType.String() == "LoadBalancer" {
+		endpoint, err := getAPIEndpoint(keptnNamespace, installParams.EndPointServiceType.String())
+		if err == nil {
+			showFallbackConnectMessage = false
+			fmt.Println("* To quickly access Keptn, you can authenticate your Keptn CLI (in a Linux shell):\n" +
+				" - keptn auth --endpoint=" + endpoint + " --api-token=" + apiToken + "\n")
```
johannes-b, line 261: In a previous commit, you used the command: 
`$(kubectl get secret keptn-api-token -n " + keptnNamespace + " -ojsonpath={.data.keptn-api-token} | base64 --decode)` instead of showing the `api-token`.

I really liked it since then we minimize the risk of leaking the API token. Can you please revert it to use: `$(kubectl get secret keptn-api-token -n " + keptnNamespace + " -ojsonpath={.data.keptn-api-token} | base64 --decode)`

cli/cmd/install.go
```diff
@@ -233,10 +240,32 @@ func doInstallation() error {
...
+		}
+	}
+	if showFallbackConnectMessage {
+		fmt.Println("* To quickly access Keptn, you can use a port-forward and then authenticate your Keptn CLI (in a Linux shell):\n" +
+			" - kubectl -n " + keptnNamespace + " port-forward service/api-gateway-nginx 8080:80\n" +
+			" - keptn auth --endpoint=http://localhost:8080/api --api-token=" + apiToken + "\n")
```
johannes-b, line 267: same as above.

cli/cmd/install.go
ankitjain28may: Sure, will make the changes, but then it won't be visible in any cases.

johannes-b: But that is okay. 
We actually don't want to show the API token, but rather help the user to run the `keptn auth` command.

ankitjain28may: Done

johannes-b (approved):
lgtm

PR #2492, opened 2020-10-07
#2483 Allow all monitoring types for configure monitoring cmd

### LINKED ISSUE #2483: Extend keptn configure monitoring to allow custom sli provider names
Keptn configure monitoring currently only supports dynatrace and prometheus as input values. If somebody creates a custom SLI provider the default installation step of creating the configmap for the SLI provier's project association has to be done directly via kubectl. This is not user friendly and requires access to the k8s cluster

I would like to have the CLI support this
```
keptn configure monitoring mysliprovider -p=myproject
```
This command should create the ConfigMap entry for myproject and specifyy mysliprovider as the provider to be used

Creating the configmap "manually"

**Additional context**
I stumbled across this when creating a custom SLI provider. I thought I can simply use keptn configure monitoring to create that necessary configmap

bacherfl:
Closes #2483 

Note: currently only the dynatrace-service and prometheus service report back using WebSockets. Therefore, when used with any other type, the command has to be called with the `--suppress-websocket` parameter

### DIFF
--- a/cli/cmd/configure_monitoring.go
+++ b/cli/cmd/configure_monitoring.go
@@ -28,17 +28,11 @@ type configureMonitoringCmdParams struct {
 
 var params *configureMonitoringCmdParams
 
-var allowedMonitoringTypes = []string{
-	"prometheus",
-	"dynatrace",
-}
-
 var monitoringCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


johannes-b (approved):
lgtm

PR #2497, opened 2020-10-08
Removing distributor from api-service in control-plane helm chart

ankitjain28may:
The distributor container in api-service deployment is not querying any Topic and `PUBSUB_RECIPIENT` is undefined/unset hence it is throwing "No pubsub recipient defined" error resulting in the Error state of the distributor container, which eventually puts the pod in CrashLoopBackOff. Hence removing the distributor container from api-service in control-plane helm chart


christian-kreuzberger-dtx (approved):
thanks, LGTM

Issue #2511, opened 2020-10-09
Keptn CLI: using create project with --shipyard pointing to an URL does not properly work


warber:
With PR #1648 the `--shipyard` flag was enhanced to make the distinction between a file path and a URL pointing to s shipyard file. However, when passing in a URL, the cli still tries to read the shipyard file from disk.

**Affected Component**

* [ ] CLI

`keptn create project <projectname> --shipyard=https://url.com/path/to/shipyard.yaml`

**Expected behavior**
The shipyard.yaml is downloaded from the location given by the `--shipyard` flag and used throughout the further proccess.

**Current behavior**
The cli tries to load the file from disk and fails

ankitjain28may:
Hi @warber Thanks for bringing this up. I cross-checked, it just silently fails without throwing any error. It would be great if you can attach more information like the logs and the keptn cli version.

Issue #1648, opened 2020-04-16
YAML input support for URIs


Dynatrace-Adam-Gardner:
It'd be a great feature to allow URI-based YAML files as input for commands that expect a `yaml` file.

For example `keptn create project sockshop --shipyard=https://example.com/shipyard.yaml`

Dynatrace-Adam-Gardner:
When evaluating this, also consider #1654

shouri007:
@johannes-b I'd like to take this up. Do you recommend any standard yaml parsing libraries for golang?
I found the following ones:
[go-yaml/yaml](https://github.com/go-yaml/yaml)
[go-gypsy](https://github.com/kylelemons/go-gypsy)
[goccy/go-yaml](https://github.com/goccy/go-yaml)

agrimmer:
@shouri007 great and thank you!

We already have [go-yaml/yaml](https://github.com/go-yaml/yaml) as a dependency in the CLI. Therefore, I would prefer to use this.

johannes-b:
Hello @shouri007, 
I'm just asking if you still want to work on this issue? 
(Every contribution is highly appreciated. :smiley: )

shouri007:
Hi @johannes-b , apologies for the delay. I'm still interested in working on this issue. Need some more time in ramping up on the whole infrastructure and go ahead with the implementation. Thanks!

shouri007:
@johannes-b Please feel free to assign this issue if anybody is interested in working on this. I currently don't have the bandwidth to attend to this. Apologies for the delay!

anukul:
Hi!
Just to confirm, `keptn create project --shipyard=<here>` is currently the only argument that accepts a YAML file, right?

johannes-b:
Hi @anukul! Yes, `keptn create project` is the only argument that takes a YAML which will then be processed.

PR #2513, opened 2020-10-10
#2511 Fixed the bug in creating project via http shipyard file

ankitjain28may:
The content that is fetched via HTTP client is never passed while creating the project and it ended up trying reading the file for the URL as well. Fixing #2511 
Feature requested in #1648

### DIFF
--- a/cli/cmd/create_project.go
+++ b/cli/cmd/create_project.go
@@ -27,11 +27,11 @@ import (
 )
 
 type createProjectCmdParams struct {
-	Shipyard  *string
-	GitUser   *string
-	GitToken  *string
-	RemoteURL *string
-	Content   string
+	Shipyard        *string
+	GitUser         *string
+	GitToken        *string
+	RemoteURL       *string
+	ShipyardContent string
 }
 
 var createProjectParams *createProjectCmdParams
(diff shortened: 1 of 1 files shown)


cli/cmd/create_project.go
```diff
@@ -190,13 +190,19 @@ func getYamlFromURL(arg string) (string, error) {
...
 func getYamlFromFile(arg string) (string, error) {
 	if _, err := os.Stat(keptnutils.ExpandTilde(arg)); os.IsNotExist(err) {
 		return "", fmt.Errorf("Cannot find file %s", arg)
 	}
+	fileContent, err := file.ReadFile(arg)
+	if err != nil {
+		return "", err
+	}
+	createProjectParams.Content = fileContent
 	return file.ReadFile(arg)
```
warber, line 206: Since you introduced reading the file in line 201, there is no need to read it again at this place.
```suggestion
	return fileContent, nil
```

cli/cmd/create_project.go
```diff
@@ -31,6 +31,7 @@ type createProjectCmdParams struct {
 	GitUser   *string
 	GitToken  *string
 	RemoteURL *string
+	Content   string
```
warber, line 34: For the sake of clarity, can you rename this field to e.g. "ShipyardContent"?

warber (changes requested):

cli/cmd/create_project.go
ankitjain28may: Yeah, my bad will remove this.

warber (approved):
lgtm

PR #2519, opened 2020-10-12
Handle case when KUBE_CONSTRAINTS is set to ""

### LINKED ISSUE #2504: Makefile: Build-CLI works, but the resulting binary is not
As a follow up of https://github.com/keptn/keptn/pull/2418

Our makefile contains a command to the build the cli, however this command builds a Keptn CLI binary that can not be used for installing Keptn, as it results in the following error:

```console
$ bin/keptn install --endpoint-service-type=ClusterIP
Helm Chart used for Keptn installation: https://storage.googleapis.com/keptn-installer/latest/keptn-0.1.0.tgz
Please confirm that the provided cluster information is correct: 
Cluster: default
Is this all correct? (y/n)
y
Using a file-based storage for the key because the password-store seems to be not set up.
Error: Failed to check kubernetes server version: Malformed constraint: 
```

The issue at hand is that the build-script adds uninitialized [KUBE_CONSTRAINTS](https://github.com/keptn/keptn/blob/05c86f6f887c5325c590524fea1a5d9597fd2efb/make-scripts/build/build-cli.sh#L14), and this leads to the CLI failing with above error message.

I believe the makefile should have a smart default for KUBE_CONSTRAINTS, e.g., ``>= 1.14, <= 1.19``. Alternatively we need to adapt the CLI code to not fail if the constraint is empty or nil (I think I'd prefer this variant).

warber:
Closes #2504 

`KubeServerVersionConstraints` is explicitely checked for empty string to fall back to default kube constraints.

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -11,8 +11,11 @@ var (
 	// Version information which is passed by ldflags
 	Version = "develop"
 
-	// KubeServerVersionConstraints the Kubernetes Cluster version's constraints is passed by ldflags
-	KubeServerVersionConstraints = ">= 1.14, <= 1.19"
+	// DefaultKubeServerVersionConstraints is used when no version is passed by ldflags
+	DefaultKubeServerVersionConstraints = ">= 1.14, <= 1.19"
+
+	//KubeServerVersionConstraints the Kubernetes Cluster version's constraints is passed by ldflags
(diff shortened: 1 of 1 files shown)


christian-kreuzberger-dtx (approved):
LGTM

PR #2524, opened 2020-10-12
#2517 Update Integrations page - master

### LINKED ISSUE #2517: Bridge Integrations Page: Update variables and timeframe such that API Call and CLI Call are consistent
Follow up of #2515:
In Keptn Bridge new Integration page, I propose to adapt the following things:

1. The arrows need to be clickable such that the content is expanded onclick (might already be fixed):
Screenshot from 2020-10-12 10-29-04

2. Adapt the start-time and timeframe for CLI and API Call to be the same, e.g., current date and time (in the correct format obviously) - this should also help the user to test this command, as it is very likely that the user has data for the current date+time combination

3. Make sure CLI and API Call use variables `${PROJECT}`, `${SERVICE}`, `${STAGE}`

Screenshot from 2020-10-12 10-30-52

ermin-muratovic:
Update variables and timeframe such that API Call and CLI Call are consistent
Fixes #2517

### DIFF
--- a/bridge/client/app/project-board/project-board.component.html
+++ b/bridge/client/app/project-board/project-board.component.html
@@ -279,7 +279,7 @@
       </ol>
       <p class="bold mt-2">Use Cases:</p>
       <ul class="list-style-none">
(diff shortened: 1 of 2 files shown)


christian-kreuzberger-dtx (approved):
LGTM

Issue #2505, opened 2020-10-09
Keptn CLI: commands are taking too long to return when no connection to a cluster can be established 


warber:
`keptn` cli commands are hanging for approx. 30 seconds in case the target cluster is not reachable.

```
$ ./keptn version

Keptn CLI version: develop
Keptn cluster version: Error occurred with response code 0 with message Get http://172.31.118.52:31090/api/v1/metadata: dial tcp 172.31.118.52:31090: i/o timeout
```

This also affects commands containing the `--help` flag.

The cli shall return and fail early, and connections to the cluster shall only be established as soon as it is necessary.

**Affected Component**

*  CLI

**To Reproduce**
Given an already configured keptn environment.
Change the target cluster IP address in `.keptn` config file to something not reachable.
Execute a keptn command (e.g.: `keptn --help`) and observer that the command only returns after approx. 30 seconds

**Expected behavior**
Commands that do not require a connection to the cluster may return immediately.
Fail as early as possible for other commands.

christian-kreuzberger-dtx:
This happened after merging #2419 (Print Keptn API version). Right now this is handled in the root command of the CLI: 
https://github.com/keptn/keptn/blob/05c86f6f887c5325c590524fea1a5d9597fd2efb/cli/cmd/root.go#L112-L136

I think to fix this, we should

a) not run the version check for the cluster connection in the root command at all (only when executing `keptn version`), and
b) introduce a timeout of 5 seconds (or so) for the HTTP connection checks.

pedroandrade:
HI @christian-kreuzberger-dtx, I'd like to pick this one, could you assign to me?

christian-kreuzberger-dtx:
Thanks! also paging @warber on this

agrimmer:
In order to get a daily version check, I think running this check in the root command is fine.

PR #2540, opened 2020-10-14
#2505 Remove version check for root cmds

pedroandrade:
Hey guys, this PR contains:
- Addi 5 seconds of a timeout to avoid hanging connection
- Remove version check for root calls

@christian-kreuzberger-dtx  @warber 

FIxing #2505

### DIFF
--- a/cli/cmd/check_endpoint_status.go
+++ b/cli/cmd/check_endpoint_status.go
@@ -9,9 +9,7 @@ import (
 	apiUtils "github.com/keptn/go-utils/pkg/api/utils"
 )
 
-const (
-	maxHttpTimeout = 5
-)
+var maxHTTPTimeout = 5 * time.Second
 
 var endPointErrorReasons = `
 Possible reasons:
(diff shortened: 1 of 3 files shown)


christian-kreuzberger-dtx:
Looks like unit tests are not working

warber:
@christian-kreuzberger-dtx @pedroandrade like so often, **global state is evil**! The unit test works in isolation but fails when executed together with all other tests, as some of them are fiddling around with the `checkEndPointStatusMock` variable.

Setting the variable to `false` before and resetting it to its old state after the test fixes the test.
This obviously is not a  fix for the actual problem - This would be: don't use global state variables at all...

I'll open create an issue for this.

pedroandrade:
@christian-kreuzberger-dtx and @warber yeah, you're right guys, global state variable is very bad, so for this PR I will set the checkEndPointStatusMock as false to make tests pass properly, but to avoid this behavior, I've got a proposal, and if you like, you can assign to me the new issue, so

it's just a draft:

```code
type Option func(*Options)
type Options struct {
	Timeout time.Durtion
}

func WithTimeout(timeout time.Durtion) Option {
	return func(o *Options) {
		o.Timeout = timeout
	}
}

Type KeptnRequester interface {
  HealthCheck(ctx context.Context, endpoint string) error
}

type KeptnRequest struct {
	Options       Options
}

func NewKeptnRequest(options ...Option) *NewKeptnRequest {
  //Setup default values
	opt := Options{
		Timeout: 5 * time.Second,
	}
  
	for _, o := range options {
		o(&opt)
	}

	return &NewKeptnRequest{Options: opt}
}

```
and.. about the tests, we just mock the interface via dependency injection and avoid the global state vars even for timeout.

what do you think?

bacherfl (approved):
lgtm

Issue #2485, opened 2020-10-06
No info or timeout while installing if there aren't renough resources available


cwtrsld:
**Environment**
- Client OS (e.g., Linux, OSX, Windows): WSL2, Ubuntu 20.04.1 LTS
- Keptn Version (```keptn version```): 0.7.1
- Kubernetes Cloud Provider (e.g., GKE, AKS): Azure Kubernetes Service, v1.16.13
- Kubernetes version (```kubectl version```): v1.19.2

**Affected Component**
- [ ] Docs
- [x] CLI
- [x] Kubernetes Integration
- [ ] Openshift Integration
- [ ] Helm
- [ ] Istio
- [ ] Bridge
- [ ] Gatekeeper
- [ ] Datastore
- [ ] REST API
- [ ] eventbroker / distributor
- [ ] jmeter

**Describe the bug**
If the (aks) cluster hasn't enough resources and you try to install keptn at some point the installation is stuck and nothing happens. The deployed pods are in a pending state and nothing happens anymore.

**Expected behavior**
Error message with information about a timeout or a long lasting pending state.

**Current behavior**
Stuck in message "Start upgrading Helm Chart keptn in namespace keptn message"

**Screenshots**
/

**Additional context**
I started the installation of keptn not knowing that there is only one node with small sizing present in the cluster. It was my fault but because of the missing timeout/info i've waited for some time.

cwtrsld:
The troubleshooting page (https://keptn.sh/docs/0.7.x/troubleshooting/) mentions this problem for EKS.

christian-kreuzberger-dtx:
Thanks for bringing this up.

I'm not sure if we can improve the installation procedure in this use-case, as we are relying on helm to return a value, but maybe we can configure a timeout for this.

For anyone wanting to take a look at this, the relevant lines of codes are here:
https://github.com/keptn/keptn/blob/40399ec765dff258b4946e59257b0bcfce6bbf39/cli/cmd/install.go#L230-L236
and
https://github.com/keptn/keptn/blob/40399ec765dff258b4946e59257b0bcfce6bbf39/cli/pkg/helm/helm_helper.go#L133-L181

Godtide:
I’d like to take this issue

Godtide:
[timeout](https://github.com/Godtide/keptn/tree/feature/2485/timeout)

christian-kreuzberger-dtx:
[...]

PR #2569, opened 2020-10-21
#2485 No info or timeout while installing if there aren't enough resources available

Godtide:


johannes-b:
Hi @Godtide, 
thanks for the PR!
I understand that a timeout makes sense there, but I think that 7sec is too short. 
Is there any special need to set the timeout to 7sec?

Godtide:
Not at all.
How many seconds do you suggest?

johannes-b:
Hi @Godtide,
I think my question was a little bit misleading and pointing in the wrong direction since we should not discuss the duration of the timeout, but rather a solution on getting more information on a failed Helm upgrade. 

In course of working on this issue, could you figure out whether the function: **UpgradeChart** see: https://github.com/keptn/keptn/blob/master/cli/pkg/helm/helm_helper.go#L133
provides more information when a Helm chart upgrade is not possible (maybe due to limited resources)?

Godtide:
https://github.com/keptn/keptn/blob/ca298963e5eac9bc9962a3347ad29751f4a25846/cli/pkg/helm/helm_helper.go#L164-L167
provides info on a failed upgrade

christian-kreuzberger-dtx:
Thanks for the contribution! However, as there is no progress in this PR I'm closing it now.

PR #2579, opened 2020-10-23
#2577 Update git upstream infos in materialized view

### LINKED ISSUE #2577: Wrong indicator for missing Git upstream
All my projects indicate that the Git upstream is missing. This is not true, since I created the Git upstream when I created the project. The original Keptn installation is from around Keptn 0.5 or 0.6 - maybe some service lost the information about the git upstream?
image

For clarification: the git upstream is still working as you can see here: https://github.com/keptn-public-demo (recent updates each day) 

Is there any way to fix this issue? Do I have to restart a service?

bacherfl:
Closes #2577


christian-kreuzberger-dtx (approved):
LGTM

Issue #2562, opened 2020-10-20
Distributor: Having a subscription topic does not have to be a requirement


bacherfl:
In previous versions (<0.8.x), the distributor's only responsibility was to receive events for a certain topic (defined by the env var `PUBSUB_TOPIC` and forward them to the recipient defined in the env var `PUBSUB_RECIPIENT`. Due to the changes made in #2220, the distributor now also allows services to forward events to the NATS Server/Keptn API, so the services do not have to send their events directly to the event broker (which has been removed in #2220).
Therefore, a `PUBSUB_TOPIC` does not necessarily has to be defined anymore when using the distributor as a sidecar for a service. E.g., when a service only needs the distributor to send events to the NATS Server/Keptn API, but does not need to subscribe to a topic - this will be the case for the api-service.
However, the distributor currently still requires a `PUBSUB_TOPIC` to be set, or otherwise it will shut down immediately. The implementation needs to be adapted so that it is also possible to use the distributor only as a forwarder for sending events to the NATS Server/Keptn API, in addition to subscribing to topics.

**Suggested Solution:** 

Currently, the listener for incoming CloudEvents for a certain topic coming from NATS, or the Keptn API is started in the `main` routine, while the HTTP endpoint for listening for events coming from the service using the distributor is started in a goroutine. This should be changed so that both listeners are started in separate goroutines as part of a `sync.WaitGroup`. This way, when the `main` routine waits for both listeners to finish, using the `Wait()` func of the `WaitingGroup`, the completion/termination of one listener does not cause the service to shut down.

**Definition of done**

- [ ] It's possible for services to use the distributor for sending events, without having to subscribe to a topic
- [ ] Subscribing to topics and using the distributor to send events can be used simultaneously 
- [ ] Re-add distributor to api-service, as api-service needs to be able to send events

Godtide:
I’d like to take this issue

johannes-b:
Hi @Godtide, 
great and thanks for taking it. I assigned it to you.

PR #2583, opened 2020-10-23
#2562 Distributor: Having a subscription topic does not have to be a requirement

Godtide:
…ement

### DIFF
--- a/distributor/cmd/main.go
+++ b/distributor/cmd/main.go
@@ -71,7 +71,16 @@ func main() {
 		fmt.Println("Failed to process env var: " + err.Error())
 		os.Exit(1)
 	}
-	go keptnapi.RunHealthEndpoint("10999")
+	//go keptnapi.RunHealthEndpoint("10999")
+
+	var wg sync.WaitGroup
+	   wg.Add(2)
+	   go connectionHTTP(connectionTypeHTTP, &wg)
+	   go connectionNATS(connectionTypeNATS, &wg)
+
+	   wg.Wait()
+
+
 	os.Exit(_main(os.Args[1:], env))
 }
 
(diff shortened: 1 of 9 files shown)


christian-kreuzberger-dtx:
Thanks Tide, @bacherfl will review this next week!

distributor/cmd/main.go
```diff
@@ -98,16 +107,38 @@ func _main(args []string, env envConfig) int {
...
+*/
 	return 0
 }
 
+  func connectionHTTP(connectionType string,  wg *sync.WaitGroup) {
+	  defer wg.Done()
+	  fmt.Printf("Type HTTP connection starting\n" )	
+	http.HandleFunc("/event", EventForwardHandler)
+	http.ListenAndServe("localhost:8081", nil)
+	createHTTPConnection()
```
bacherfl, line 119: I believe the `createHTTPConnection()` function will not be reached, since the `http.ListenAndServe()` function is blocking

distributor/cmd/main.go
```diff
@@ -83,9 +92,9 @@ func _main(args []string, env envConfig) int {
 	createEventForwardingEndpoint(env)
 
 	// initialize the http client
-	connectionType := strings.ToLower(os.Getenv("CONNECTION_TYPE"))
+	//connectionType := strings.ToLower(os.Getenv("CONNECTION_TYPE"))
```
bacherfl, line 95: the check for the connection type should not be removed, since we would like to have the possibility to subscribe to events either via NATS or via HTTP polling

distributor/cmd/main.go
```diff
@@ -98,16 +107,38 @@ func _main(args []string, env envConfig) int {
 	default:
 		createNATSClientConnection()
 	}
-
+*/
 	return 0
 }
 
+  func connectionHTTP(connectionType string,  wg *sync.WaitGroup) {
```
bacherfl, line 114: This function kind of mixes the initialisation of the HTTP event polling client, and the HTTP Server that is used to forward events to the Keptn API/NATS server - these two different responsibilities should be split up in their own go routines

bacherfl (approved):
thanks @Godtide for your contribution! I have added some changes, and now it's ready to be merged

Issue #2571, opened 2020-10-21
Re-add the version check into the root command


bacherfl:
In PR #2540 , the version check has been removed from the root command since it prevented commands that do not require a connection to the cluster the stall when no connection to the cluster could be established. The version check should be included again. However, commands that do not require a cluster connection should still work even if a connection to the cluster cannot be established.
In addition, commands like `keptn ... --help` should not run any version check at all. 

edit:  As mentioned in #2590, we also need to add a check if CLI version == API version, and print a warning if it doesn't.

christian-kreuzberger-dtx:
@johannes-b As mentioned in #2590, should we add a check if CLI version == API version, and print a warning if it doesn't?

johannes-b:
Yes, let's go this way.

PR #2586, opened 2020-10-27
#2505 Remove version check for root cmds

johannes-b:
- adding 5 seconds of a timeout to avoid hanging connection

* Remove version check for root calls
- keeping only on version command

* Add tests for http timeout
- add tests for check endpoint status function

* Does not mock the tests for checkEndpointStatus function
- forcing the checkEndPointStatusMock as true to avoid random tests

# To-be-discussed
- [ ] Providing a backport of https://github.com/keptn/keptn/pull/2540 does not work on release-0.7.3 since this branch does not know the function `checkEndPointStatus`. 
- [ ] When merging it, we must implement: https://github.com/keptn/keptn/issues/2571

### DIFF
--- a/cli/cmd/check_endpoint_status.go
+++ b/cli/cmd/check_endpoint_status.go
@@ -1,0 +1,47 @@
+package cmd
+
+import (
+	"context"
+	"crypto/tls"
+	"net/http"
+	"time"
+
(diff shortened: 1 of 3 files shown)


johannes-b:
Backport does not work. Closed issue.

PR #2591, opened 2020-10-27
Added docs on how to install the current master branch

christian-kreuzberger-dtx:
This PR  adds docs on how to install the current master branch on K8s.
During this task I also noticed that the developer docs need an update regarding k3s and `keptn install`


johannes-b (approved):
lgtm

PR #2593, opened 2020-10-28
#2589 Added timeout of 5 seconds to server version check

### LINKED ISSUE #2589: Keptn CLI: commands are taking too long to return when no connection to a cluster can be established
**On the release-0.7.3 branch:** `keptn` cli commands are hanging for approx. 30 seconds in case the target cluster is not reachable.

```
$ ./keptn version

Keptn CLI version: develop
Keptn cluster version: Error occurred with response code 0 with message Get http://172.31.118.52:31090/api/v1/metadata: dial tcp 172.31.118.52:31090: i/o timeout
```

This also affects commands containing the `--help` flag.

The cli shall return and fail early, and connections to the cluster shall only be established as soon as it is necessary.

**Affected Component**

*  CLI

**To Reproduce**
Given an already configured keptn environment.
Change the target cluster IP address in `.keptn` config file to something not reachable.
Execute a keptn command (e.g.: `keptn --help`) and observer that the command only returns after approx. 30 seconds

**Expected behavior**
Fail as early as possible for other commands.

**Task**
- Introduce a 5sec timeout in the function: `GetMetadata()` - https://github.com/keptn/keptn/blob/release-0.7.3/cli/cmd/version.go#L150

FYI: For the master branch, this feature is already implemented in: https://github.com/keptn/keptn/issues/2505

bacherfl:
Closes #2589

### DIFF
--- a/cli/cmd/version.go
+++ b/cli/cmd/version.go
@@ -17,6 +17,7 @@ package cmd
 import (
 	"errors"
 	"fmt"
+	"net/http"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


johannes-b (approved):
lgtm

PR #2594, opened 2020-10-28
#2530 Increase helm svc test coverage

### LINKED ISSUE #2530: Increase test coverage for helm-service
Due to the changes in https://github.com/keptn/keptn/pull/2502 the helm-services has seen many changes with uncertainty about their test coverage. 

**Task:**
- Increase the test coverage of the helm-service

warber:
fixes #2530 

This PR increases the test coverage of the current implementation of the helm service.
Further, "gomock" was introduced to be used as a mocking framework.

Note, that the release handler is not yet tested, as the current implementation needs to be verified to be correct.
Further tests will be implemented in a follow-up issue


bacherfl (approved):
lgtm

Issue #2548, opened 2020-10-15
Clean-up quality gate evaluation results in Bridge


ermin-muratovic:
Show an "Ignore for comparison"-Button on top of the evaluation details. When clicking the button, a confirmation dialog should be displayed, asking the user if he really wants to invalidate that evaluation (showing the date/time and if available also the `buildId` label). See #2449 

![image](https://user-images.githubusercontent.com/6098219/95983327-b945c880-0e21-11eb-8874-efbdf723a75c.png)

PR #2605, opened 2020-10-29
#2548 allow to invalidate evaluation results from bridge - backport 0.7.3

ermin-muratovic:
The evaluation tile now allows to invalidate evaluation results from the bridge by clicking on the "Ignore for comparison"-Button.

image

It will ask the user for confirmation and allow to provide a reason as well.

image


johannes-b (approved):
lgtm

Issue #2565, opened 2020-10-20
Bridge: Show all core use-cases on integrations page


christian-kreuzberger-dtx:
Next to the example of triggering a quality gate evaluation, Keptn is capable of two other important use-cases: "Deployment" and "Remediation". Accordingly, add these two examples to the list of *Use-Cases* on the integrations page.

**Wireframe**

![image](https://user-images.githubusercontent.com/729071/96591532-a2064f80-12e7-11eb-83e4-439e7c969d66.png)

**Tasks**
* helm-chart holds the use-case, the Bridge is used for. Forward this information to the Bridge via an environment variable.  
* Bridge: Display the additional use-cases depending on the environment variable:

Use-Cases:
- Keptn CLI: 
  - Trigger deployment with a new artifact: 
     ```
     keptn send event new-artifact --project=${PROJECT} --service=${SERVICE}--image=${IMAGE} --tag=${TAG}
     ```
  - Trigger remediation with a *dummy* problem event: `
    **Note: Linux/mac OS only**
    ```
    echo '{"type":"sh.keptn.event.problem.open","specversion":"0.2","source":"api","contenttype":"application\/json","data":{"State":"OPEN","ProblemID":"${PROBLEM_ID}","ProblemTitle":"${PROBLEM}","project":"${PROJECT}","stage":"${STAGE}","service":"${SERVICE}"}}' > dummy_problem.json
    keptn send event -f=dummy_problem.json
    ```

- Keptn API:
  - Trigger deployment with a new artifact: 
    ```
    curl -X POST "${KEPTN_API_ENDPOINT}/v1/event" \
    -H "accept: application/json; charset=utf-8" -H "x-token: ${KEPTN_API_TOKEN}" -H "Content-Type: application/json; charset=utf-8" \
    -d "{"type":"sh.keptn.event.configuration.change","specversion":"0.2","source":"api","contenttype":"application\/json","data":{"project":"${PROJECT}","stage":"${STAGE}","service":"${SERVICE}","valuesCanary":{"image":"${IMAGE}"}}}"
    ```
  - Trigger remediation with a *dummy* problem event:
    ```
    curl -X POST "${KEPTN_API_ENDPOINT}/v1/event" \
    -H "accept: application/json; charset=utf-8" -H "x-token: ${KEPTN_API_TOKEN}" -H "Content-Type: application/json; charset=utf-8" \
[...]

PR #2606, opened 2020-10-29
#2565 Show all core use-cases on integrations page

ermin-muratovic:
If keptn is not used as "quality gates only" installation, the bridge shows all core use-cases on integrations page.
The installation type is configured by providing `KEPTN_INSTALLATION_TYPE` environment variable, see _bridge/deploy/bridge.yaml_

image

### DIFF
--- a/bridge/deploy/bridge.yaml
+++ b/bridge/deploy/bridge.yaml
@@ -47,9 +47,9 @@ spec:
          - name: SHOW_API_TOKEN
            value: "true"
          - name: KEPTN_INSTALLATION_TYPE
-           value: "FULL"
+           value: "QUALITY_GATES,CONTINUOUS_OPERATIONS,CONTINUOUS_DELIVERY"
            Description: >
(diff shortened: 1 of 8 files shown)


bridge/deploy/bridge.yaml
```diff
@@ -46,6 +46,10 @@ spec:
            value: "true"
          - name: SHOW_API_TOKEN
            value: "true"
+         - name: KEPTN_INSTALLATION_TYPE
+           value: "QUALITY_GATES,CONTINUOUS_OPERATIONS,CONTINUOUS_DELIVERY"
```
christian-kreuzberger-dtx, line 50: Same as in the other PR, there is a mismatch between specifying `QUALITY_GATES` here, but expecting it to be `QUALITY_GATES_ONLY` in the checks above.

ermin-muratovic: I fixed that! Thanks.

christian-kreuzberger-dtx (approved):
LGTM

PR #2616, opened 2020-10-30
Set versions of keptn-contrib services based on tested keptn version

bacherfl:
This PR makes sure that the versions of the keptn-contrib services used in the integration tests are the ones that are compatible with the respective release


christian-kreuzberger-dtx (approved):
This painfully shows why we need Keptn Uniform @johannes-b 

LGTM

PR #2630, opened 2020-11-03
Helm service: ignore events that have been generated by itself

bacherfl:
Since the helm-service both listens to, as well as generates `service.delete.finished` events, it needs to ignore its own events.


warber (approved):
lgtm

PR #2634, opened 2020-11-04
#2533 Delivery Assistant Workflow in Keptn 0.8

### LINKED ISSUE #2533: gatekeeper-service becomes the approval-service for automatic approvals
The gatekeeper is currently responsible for the following tasks:

1. Task: It triggers the release in the current stage. More precisely, it sends out a `configuration.change` event with action promote/abort depending on the result of the evaluation result
2. Task: It checks whether an approval strategy is defined for the next stage. If so, it sends out an `approval.triggered` event
3. Task: It checks the result of the `approval.finished` result and depending on the result triggers a new deployment.

With the shipyard-controller, the tasks of the gatekeeper-service will get obsolete:
1. Task: The release is modeled as a dedicated step in the shipyard file. Therefore, the shipyard-controller will send out a release.triggered, which triggers the helm-service. Depending on the result in the event, the helm-service promotes or aborts the deployment.
2. Task: Also here the approval is modeled as a dedicated step in the shipyard file. Therefore, the shipyard-controller will send out an `approval.triggered` event.
3. Task: The subsequent task after the approval (i.e. the deployment) has to properly handle the result of the approval. That means, if a user has disapproved, the deployment must not happen

*Note:* It is important that every service in the execution-plane properly handles events which has a result=Fail !

**To-Do:**
* Use the collection of `open.triggered` events in the shipyard-controller to manage the open approvals. Remove the functionality in the configuration-service
* The gatekeeper-service subscribes to the event `sh.keptn.event.approval.triggered` and depending on the defined approval-strategy (i.e. only when the approval strategy is automatic), it sends an `approval.started` as well as an `approval.finished` event depending on the last result.
* The CLI uses the endpoint of the shipyard-controller in order to retrieve open approvals.

bacherfl:
Closes #2533


warber (approved):
lgtm

PR #2643, opened 2020-11-06
Removed unneeded env var from api-service distributor

bacherfl:


warber (approved):
good for now
created follow up issue: #2646

PR #2659, opened 2020-11-09
#2651 Update nginx version

### LINKED ISSUE #2651: Upgrade nginx unprivileged to latest version
See https://github.com/nginxinc/docker-nginx-unprivileged for latest version (and see [here](https://nginx.org/en/download.html) for a changelog).

We are using version 1.19.1-alpine of nginx-unprivileged
```
    repository: docker.io/nginxinc/nginx-unprivileged
    tag: 1.19.1-alpine
```

We should upgrade to the latest version to include all recent bugfixes etc...

bacherfl:
Closes #2651

### DIFF
--- a/installer/manifests/keptn/charts/control-plane/values.yaml
+++ b/installer/manifests/keptn/charts/control-plane/values.yaml
@@ -16,7 +16,7 @@ apiGatewayNginx:
   nodePort: 31090
   image:
     repository: docker.io/nginxinc/nginx-unprivileged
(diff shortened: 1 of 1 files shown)


johannes-b (approved):

christian-kreuzberger-dtx:
I think we will also need to update our documentation here: https://keptn.sh/docs/0.7.x/operate/advanced_install_options/ and https://keptn.sh/docs/0.8.x/operate/advanced_install_options/

PR #2680, opened 2020-11-12
Configure resource request and limit for jmeter-service

johannes-b:
In reviewing the resource quota of deployments, I recognized that the jmeter-service is missing the resource configuration.

All the other deployments have a resource config.

### DIFF
--- a/installer/manifests/keptn/charts/continuous-delivery/templates/continuous-deployment.yaml
+++ b/installer/manifests/keptn/charts/continuous-delivery/templates/continuous-deployment.yaml
@@ -283,9 +283,6 @@ spec:
             requests:
               memory: "64Mi"
               cpu: "50m"
-            limits:
-              memory: "128Mi"
-              cpu: "500m
           env:
           - name: CONFIGURATION_SERVICE
             value: 'http://configuration-service:8080'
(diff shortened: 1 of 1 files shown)


installer/manifests/keptn/charts/continuous-delivery/templates/continuous-deployment.yaml
```diff
@@ -278,7 +278,14 @@ spec:
...
           imagePullPolicy: IfNotPresent
           ports:
-          - containerPort: 8080
+            - containerPort: 8080
+          resources:
+            requests:
+              memory: "64Mi"
+              cpu: "50m"
+            limits:
+              memory: "128Mi"
```
christian-kreuzberger-dtx, line 287: In my monitoring jmeter-service shows up that it needs up to 1.5 GB memory depending on the use-case
image

christian-kreuzberger-dtx: FYI, a simple "new-artifact" on 0.7.2 leads to roughly 500 MB memory usage

johannes-b: Thanks for the insights. Well, then my resource limits are too (too) low. :) 

What is your suggestion on that?

johannes-b: As discussed, I will define the `requests` but not the `limits`.

christian-kreuzberger-dtx (approved):
LGTM

PR #2700, opened 2020-11-18
#2563 Removed livenessProbes from development manifests

### LINKED ISSUE #2536: Unable to do remote debugging of mongodb-datastore due to liveness-probe
Since the introduction of the livenessProbe for the mongodb-datstore and the distributor it is not possible to remotely debug the mongodb-datastore. The problem at hand is, that while sitting at a breakpoint inside the mongodb-datastore, the liveness-probe inside the cluster fails which causes the pod to be restarted by kubernetes. Hence, the debugging connection breaks.

**Affected Component**
* [ ] Datastore

Note, that other components might suffer from this issue too.

**To Reproduce**
1) Set a breakpoint in mongodb-datastore.
2) Deploy the mongodb-datastore with remote debugging.
3) While the execution stops at the breakpoint, observe that in kubernetes a "Liveness probe failed" event gets triggered and the debugging connection breaks.

**Expected behavior**
Pausing the execution of the process does not affect the lifecycle of the pod in the cluster.

**Current behavior**
The pod is restarted because liveness checks fail

bacherfl:
Closes #2536 

The livenessProbes have been removed from the individual service manifests that are used for developing. The manifests of the installer helm chart have been left untouched.

After removing the livenessProbe from the mongodb-datastore I was able to use the remote debugging feature again with no pod restarts

### DIFF
--- a/api/deploy/service.yaml
+++ b/api/deploy/service.yaml
@@ -37,12 +37,6 @@ spec:
             limits:
               memory: "256Mi"
               cpu: "500m"
-          livenessProbe:
-            httpGet:
-              path: /health
-              port: 10999
-            initialDelaySeconds: 5
-            periodSeconds: 5
           env:
             - name: PREFIX_PATH
               value: ""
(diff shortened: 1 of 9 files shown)


johannes-b (approved):
lgtm

PR #2705, opened 2020-11-19
#2704 Added -y option to keptn auth to automatically apply changed Kubernetes context

### LINKED ISSUE #2704: Integration Test stalls at the Keptn auth command
After installing Keptn during the integration tests, the auth command stalls because at this point, the ~/.keptn/config  file does not have the current context of the K8s cluster. After getting the current context from the KUBECONFIG, this is detected as a change (see [here](https://github.com/keptn/keptn/blob/b21f0b881cf9207e3eda57069772b7376df3d4e6/cli/pkg/credentialmanager/handler.go#L174), and a dialog asking the user wether the CLI should proceed with this context is shown: See [here](https://travis-ci.com/github/keptn/keptn/jobs/442408894#L3965)

Eventually, the build will be cancelled because no user input is provided during the integration tests.

**Proposed solutions:** 

1. Add a `-y` or `--yes` flag to the `keptn auth command` to automatically apply the new context

2. If the current Kubernetes context of the Keptn CLI is empty, automatically apply the detected context

bacherfl:
Closes #2704

Also, when no context has been set in the first place, it will automatically applied and the user will be informed

### DIFF
--- a/cli/cmd/add_resource.go
+++ b/cli/cmd/add_resource.go
@@ -50,7 +50,7 @@ keptn add-resource --project=rockshop --stage=production --service=shop --resour
 	SilenceUsage: true,
 	Args:         cobra.ExactArgs(0),
 	RunE: func(cmd *cobra.Command, args []string) error {
-		endPoint, apiToken, err := credentialmanager.NewCredentialManager().GetCreds(namespace)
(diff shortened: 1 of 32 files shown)


johannes-b (approved):
lgtm

PR #2714, opened 2020-11-23
#2571 Re-add the version check into the root command

### LINKED ISSUE #2571: Re-add the version check into the root command

### LINKED ISSUE #2590: Keptn CLI not backward compatible. Missing Keptn cluster version checks result in unclear error messages
Its possible to use a new Keptn CLI as the connected Keptn cluster, e.g: using CLI version 0.7.2 against a 0.7.1 Keptn Cluster.
This for instance leads to an unclear error message when executing ```keptn send event start-evaluation``` as the 0.7.2 CLI assumes the new /evaluation API endpoint which doesnt exist on a 0.7.1 installation. The error message simply returns that the /evaluation endpoint returned an error which is not very telling for the end user for troubleshooting.

I suggest that the CLI should do some version checks and either
1: warn the user that the CLI is newer or older then the keptn cluster and that an update on either side is suggested
2: is backward compatible, e.g: use the OLD API endpoints if the Keptn version is older

ermin-muratovic:
Closes #2571 

I re-added the version check into the root command again. And also added a check if CLI and cluster versions are equal, which fixes #2590 

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"github.com/keptn/keptn/cli/pkg/version"
 	"os"
+	"strings"
 
 	"github.com/keptn/keptn/cli/pkg/logging"
 	homedir "github.com/mitchellh/go-homedir"
@@ -44,7 +45,11 @@ func Execute() {
 	currentLogLevel := logging.LogLevel
 	logging.LogLevel = logging.QuietLevel
 
-	runVersionCheck()
+	containsHelp := strings.Contains(strings.Join(os.Args, " "), "help")
+
+	if !containsHelp {
+		runVersionCheck()
+	}
 
 	// Set LogLevel back to previous state
 	logging.LogLevel = currentLogLevel
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -2,14 +2,12 @@ package cmd
 
 import (
 	"fmt"
-	"github.com/keptn/keptn/cli/pkg/version"
-	"os"
-	"strings"
-
 	"github.com/keptn/keptn/cli/pkg/logging"
+	"github.com/keptn/keptn/cli/pkg/version"
 	homedir "github.com/mitchellh/go-homedir"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"os"
 )
 
 var cfgFile string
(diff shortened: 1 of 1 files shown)


cli/cmd/root.go
```diff
@@ -43,6 +44,8 @@ func Execute() {
 	currentLogLevel := logging.LogLevel
 	logging.LogLevel = logging.QuietLevel
 
+	runVersionCheck()
```
christian-kreuzberger-dtx, line 47: any chance to NOT run the version check if `--help` or similar is used?

christian-kreuzberger-dtx:
CLI Builds to test:
* Linux https://storage.googleapis.com/keptn-cli/feature-2571%2B20201123.1030/keptn-linux.zip
* OSX https://storage.googleapis.com/keptn-cli/feature-2571%2B20201123.1030/keptn-macOS.zip
* Windows https://storage.googleapis.com/keptn-cli/feature-2571%2B20201123.1030/keptn-windows.zip

cli/cmd/root.go
ermin-muratovic: I added the check, if `help` occurs in command arguments, the version check will not be executed.

warber (commented):
lgtm

warber (approved):

bacherfl (approved):
lgtm

cli/cmd/root.go
```diff
@@ -43,6 +45,12 @@ func Execute() {
 	currentLogLevel := logging.LogLevel
 	logging.LogLevel = logging.QuietLevel
 
+	containsHelp := strings.Contains(strings.Join(os.Args, " "), "help")
```
christian-kreuzberger-dtx, line 48: Will this also work when we use `-h` instead of `--help`?

Also, with the logic involved this check would look for any param that contains help, e.g., `keptn create service helper --project=carts` would also be caught.

I think we should check if any entry in `os.Args` has `--help` or `-h` in it.

christian-kreuzberger-dtx (changes requested):
Please address my comment.

Also, please rebase to master.

cli/cmd/root.go
ermin-muratovic: I changed it to check for `-h` or `--help` parameter and actually do it by iterating through all parameters instead of doing a `strings.Contains` check to prevent false detection of parameters only starting with `-h`.
Currently there are no to me known parameters starting with `-h`, but in future there might be some introduced.

johannes-b (approved):
lgtm

cli/cmd/root.go
christian-kreuzberger-dtx: Thanks, looks good!

christian-kreuzberger-dtx (approved):
LGTM

PR #2736, opened 2020-11-26
#2707 fixed encoding of URI for resources within subfolders

### LINKED ISSUE #2707: Fetching resources from folder via API not working
Talked to @sergiohinojosa regarding a bug, here are his words:

> I found a small bug that annoyed me this afternoon and was driving me crazy (thinking that something in the datastore was corrupt or the customer messed up the sli.yaml file). Downloading resources over the API is not possible if the ressource is behind a folder (meaning having a slash /). It always returns 404. I tried fetching the /dynatrace/sli.yamlin 0.7.2 and also on 0.7.3 with the same results. Also tried other ressources like the jmx file behind the jmeter folder with the same results. I was not succesfull to escape the slash or code it in a different way with curl or postman.

image

bacherfl:
Closes #2707 

Note that in order to access nested resources (e.g. `jmeter/load.jmx`) via the public API, the `%` character has to be escaped additionally:

Example: 
`http://<keptn-api>/api/configuration-service/v1/project/sockshop/stage/staging/service/carts/resource/jmeter%2Fload.jmx` => will NOT work. `%2F` needs to be replaced with `%252F`

`http://<keptn-api>/api/configuration-service/v1/project/sockshop/stage/staging/service/carts/resource/jmeter%252Fload.jmx` => will work


configuration-service/deploy/service.yaml
```diff
@@ -60,7 +60,7 @@ spec:
             memory: "256Mi"
             cpu: "500m"
           limits:
-            memory: "256Mi"
+            memory: "512Mi"
```
bacherfl, line 63: Not directly related to the Issue, but I increased this value because I kept getting OOM errors when trying to attach the debugger to the container.

johannes-b (approved):
lgtm

PR #2755, opened 2020-12-03
#2729 Moved PreRun checks into Run methods

### LINKED ISSUE #2729: CLI: Required Flags are not validated before PreRunE is called.
The CLI framework Cobra does not validate flags that are marked as required before the function `PreRunE` is called.
This is a known issue https://github.com/spf13/cobra/issues/655 and this behavior change has not been merged since 2018 https://github.com/spf13/cobra/pull/700.

However, the Keptn CLI assumes in some commands that the flags are already validated when `PreRunE` is called.

For example, `keptn onboard service` has `--project` and `--chart` as required flags. Furthermore, in the `PreRunE` function of the `onboard service` command checks are executed for validating the provided chart. 
If now a user executes `keptn onboard service carts` without specifying a chart, the user gets the error message `Error: Provided Helm chart does not exist`.
The correct error message would however be `Error: required flag(s) "project", "chart" not set`.

Tasks:
- [ ] Investigate all CLI commands whether PreRunE uses a required flag. If so and if possible, move the functionality of `PreRunE` into `RunE`

bacherfl:
Closes #2729 


cli/cmd/onboard_service.go
```diff
@@ -125,6 +98,36 @@ keptn onboard service SERVICENAME --project=PROJECTNAME --chart=HELM_CHART.tgz
 	},
 }
 
+func doOnboardServicePreRunChecks(args []string) error {
+	_, _, err := credentialmanager.NewCredentialManager(false).GetCreds(namespace)
+	if err != nil {
+		return errors.New(authErrorMsg)
+	}
```
christian-kreuzberger-dtx, line 105: To be honest, I wouldn't have expected the credential manager stuff in the `doOnboardServicePreRunChecks` function, but in `RunE` (as it is in `create_project`). But I'm fine if it stays in here too. What's your thought on this @bacherfl ?

bacherfl: True, that credential manager check didn't need to be there since the credential manager is checked again in the `RunE` method.

christian-kreuzberger-dtx: Oh yeah, I see it :) Thanks!

christian-kreuzberger-dtx (approved):
LGTM

Issue #2719, opened 2020-11-23
Remove hard-dependency of MacOS builds in travis-ci


christian-kreuzberger-dtx:
Due to the current limitations of Travis-CI https://github.com/keptn/keptn/issues/2715, we are forced to change some of our jobs.

One big issue is that MacOS is required to build the CLIs:

- https://github.com/keptn/keptn/blob/f89f31c5a5d95b80b4c933f9312b4598a1da92ee/.travis.yml#L324-L338
- https://github.com/keptn/keptn/blob/f89f31c5a5d95b80b4c933f9312b4598a1da92ee/.travis.yml#L355-L373
- https://github.com/keptn/keptn/blob/f89f31c5a5d95b80b4c933f9312b4598a1da92ee/.travis.yml#L456-L470

Unfortunately, we will need to keep building at least the MacOS binary on MacOS (see #2738), but we should move Windows and Linux builds to a Linux based job (either on Travis-CI or GitHub actions), as MacOS builds are expensive.

# Definition of Done

- [ ] CLI builds for Linux and Windows are running on Linux on GH Actions for each PR if any file in the `cli` folder has changed (does not matter which branch name; release, feature, patch, bugfix)
- [ ] CLI build for MacOS is still running on MacOS, but on GH Actions for each PR if any file in the `cli` folder has changed (does not matter which branch name; release, feature, patch, bugfix)
- [ ] Every push/commit on master branch should build the "latest" CLI
- [ ] Resulting binaries are still uploaded to Google Cloud

kadirselcuk:
I think this problem is exactly what I'm talking about, thank you.

PR #2795, opened 2020-12-11
#2719 Create GH Action for building the CLI

christian-kreuzberger-dtx:
This PR introduces a GitHub Action/Workflow that builds the CLI and uploads it as an artifact to the action itself (not to gcloud for now).

*Note*: The release step is not fully implemented yet. CLI  is currently uploaded to GH actions instead of google cloud, but I guess that's a good thing.

In addition, I've removed all scripts that were used to build the CLI via travis-ci previously.

# New Features

- Build CLI on the operating system it is meant for (windows, linux, macos)
- Upload CLI build artifacts via GH actions

# GH Actions

Looks as follows:
gh-action-step1-wip

## Build CLI on {Linux, Windows, MacOS}

This builds the cli on the target platform, and uploads an artifact to GitHub.

## Release Keptn

**Not implemented yet, but here is the idea**

image

This is only triggered for release branches (e.g., release-0.8.0).
It downloads the previously built artifact, creates a draft release, and uploads the CLI assets to the draft release.

The end result of this step is a draft release with up2date CLI binaries (packed as .tar.gz):
image

### DIFF
--- a/.github/workflows/cli-pipeline.yml
+++ b/.github/workflows/cli-pipeline.yml
@@ -1,0 +1,129 @@
+name: Keptn CLI Pipeline
+on:
+  # always execute CLI Pipeline when something is pushed to master or release-* branches
+  push:
+    branches:
+      - 'master'
+      - 'release-*'
+  # in addition, execute for pull requests to those branches if anything was changed in the CLI folder
+  pull_request:
+    branches:
+      - 'master'
+      - 'release-*'
+    paths:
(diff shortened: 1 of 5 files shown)


christian-kreuzberger-dtx:
Building on MacOS ic currently not working due to a Datacenter upgrade: https://github.com/actions/virtual-environments/issues/2247
Though we can still merge this.

christian-kreuzberger-dtx:
No longer needed, superseeded by #2817

PR #2798, opened 2020-12-14
Configuration Service: track last artifact in materialized view

### LINKED ISSUE #2692: Configuration Service: Keep track of last processed artifact in materialized view
In previous versions of Keptn (<0.8), we used the `sh.keptn.events.deployment-finished` events to keep track of the last deployed image of a service within the materialized view maintained by the configuration-service. Since we do not have this event type anymore, we need to adapt the logic in the configuration-service to use the newly introduced `deployment.finished` and their correlating `deployment.triggered` events (see [here](https://github.com/keptn/keptn/blob/908f3b23a3af3106ec592490d0a6657f7104a83b/configuration-service/common/projects_materialized_view.go#L289)). 

**Keep track of**
- Artifact
- Git commit
- Service URL

**Proposed solution:** 

When a `deployment.finished` event is received by the configuration service, check for a correlating `deployment.triggered` event - those two events can be linked by the `triggeredid`. When a correlating `deployment.triggered` event has been found, check if it contains the field `data.configurationChange.values.image`. If this property is present, the materialized view can be updated accordingly.

warber:
closes #2692 

Unfortunately, the generated swagger source code messes up the pull request. 
We should really discuss solutions for that. Ideally, these files are not pushed to Git.

The files with the relevant changes are:

* common/projects_materialized_view.go
* common/projects_materialized_view_test.go
* handlers/event.go
* swagger.yaml

### DIFF
--- a/configuration-service/cmd/configuration-service-server/main.go
+++ b/configuration-service/cmd/configuration-service-server/main.go
@@ -6,7 +6,7 @@ import (
 	"log"
 	"os"
 
-	loads "github.com/go-openapi/loads"
+	"github.com/go-openapi/loads"
 	flags "github.com/jessevdk/go-flags"
 	"github.com/keptn/keptn/configuration-service/restapi"
 	"github.com/keptn/keptn/configuration-service/restapi/operations"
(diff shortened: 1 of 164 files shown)


bacherfl (approved):
lgtm

PR #2812, opened 2020-12-15
#2759 mongodb performance investigation

### LINKED ISSUE #2759: Quering (root) events via mongodb-datastore is slow when there is many events in the db
In order to troubleshoot https://github.com/keptn/keptn/issues/2744 I have spammed my Keptn 0.7.3 with a lot of cloud-events:

1. start-evaluation cloud-events in project ck-keptn-qg (probably around 200 start-evaluation root events per service for 12 services = 2400 root events)
1. new-artifact cloud-events in project ck-sockshop (probably around 80 configuration-change root events, one service)

See https://pastebin.com/qVqSZFH9 for output of statistics-service.

This allowed me to discover a performance problem and even failed API Calls in mongodb-datastore, which will affect Keptn Bridge (and probably also in other parts of Keptn, but they are not too visible to the end-user at this point).

image (5)

Example for really slow calls (Response time > 10 seconds):  **<- this is where we need to improve heavily**
```
/event?root=true&pageSize=20&project=ck-keptn-qg&service=dynatrace-sli-service&fromTime=2020-12-04T07:21:15.000Z
/event?root=true&pageSize=20&project=ck-sockshop&service=carts-db
```

Examples for slower calls (around 1 second, but usually less):
```
/event?type=sh.keptn.events.tests-finished&project=ck-keptn-qg&stage=hardening&service=eventbroker-go&keptnContext=1c86c128-1703-400e-b5bb-9b2ae25d7efc&pageSize=1
/event/type/sh.keptn.events.evaluation-done?source=lighthouse-service&limit=1&excludeInvalidated=true&filter=data.project:ck-keptn-qg%20AND%20data.stage:hardening%20AND%20data.service:configuration-service%20AND%20data.result:pass
```

**Tasks:**
- [ ] Provide a script that allows to setup the environment (e.g., 20 projects, with 15 services each, + 100 evaluations) 
- [ ] Using a monitoring tool (Dynatrace or Prometheus), monitor the response time of the above-mentioned endpoints (+ note down the results) - please use this step to verify the claim that those endpoints are slow
- [ ] Create a (JMeter) Test that spams start-evaluation cloud-events; Use GET requests to query the above-mentioned endpoints
- [ ] Activate the MongoDB profiler to get more insights: https://docs.mongodb.com/manual/tutorial/manage-the-database-profiler/
- Identify the bottleneck that is causing a long response time. 

**Definition of Done**
- [ ] Update of Readme.md to list all collections used by the mongodb-datastore.
- [ ] **Document the results in the Pull Request**

--- 
**Out of scope:**
- Improve the above-mentioned endpoints (especially the root events one), either by adding indexes in MongoDB or by using other query methods
- Run the same JMeter Test again and verify that the results are better now
- Extend the endpoint to allow a list of services as a parameter when querying events. 
- Add an index for project, service, type.

bacherfl:
Closes #2759

**Performance Test Setup:**

 - Configured my Keptn installation to be monitored by Dynatrace, and created **keptn-selfmonitoring** project in my Keptn installation.
- Added each core service to the **keptn-selfmonitoring** project. Note: For the sake of this Issue, only the **mongodb-datastore** is needed, but having all services available for doing evaluations does not harm.

**Test protocol**:

1. Created a project
1. Create a service
1. Send 100 evaluation.triggered events
1. Send evaluation.triggered event for the **mongodb-datastore** - This triggers the **evaluation** task sequence which first performs a load tests using the jmeter service, and then evaluates the results using monitoring data from my DT tenant.
1. Send 10 evaluation.invalidated events
1. Trigger another test/evaluation for the **mongodb-datastore**
1. Repeat Steps 2 - 6 until I had 13 services with 1300 evaluations and 130 evaluation invalidations

Note: For this investigation, having one project was sufficient since the mongodb collections are separated for each project, meaning that potential scaling issues can only arise if the number of events for a particular project increases to a certain point. Also, to the best of my knowledge, we do not have a screen where requests for fetching events for multiple projects are sent in parallel in the bridge.

After executing this test, I obtained the following data from my DT tenant:

response-time

The load test script executed the following requests:

`/event/type/sh.keptn.event.evaluation.finished?invalidated=true&filter=data.project:project=project-0%20AND%20data.service:service-0%20AND%20data.result:pass`

`/event?project=project-1&service=service-1&root=true`

Jmeter configuration:

VUsers: 10
LoopCount: 500
ThinkTime: 250
AcceptedErrorRate: 0.1

**Interpretation:**
It clearly shows that an increased number of events leads to an increase of the response time. This is something that could be improved by using **indexes** for the mongodb collections. In my opinion, adding the following indexes could have a benefitial impact:

- Collection **project-rootEvents**: Add an index for `data.service`
- Collection **project**: Add indexes for `data.service` and `shkeptncontext`

On the bridge side, I observed that, when opening a project, all rootEvents, and their correlating event traces are loaded for each service simultaneously - this causes the spike in the response time (see the screenshot above). This could be improved by loading rootEvents and event traces on demand, i.e., when the user clicks on the service, or a particular root event.

### DIFF
--- a/test/assets/evaluation_triggered_event_template.json
+++ b/test/assets/evaluation_triggered_event_template.json
@@ -1,0 +1,14 @@
+{
+  "type": "sh.keptn.event.hardening.evaluation.triggered",
+  "specversion": "1.0",
+  "source": "travis-ci",
+  "contenttype": "application/json",
+  "data": {
+    "project": "$PROJECT",
+    "stage": "$STAGE",
+    "service": "$SERVICE",
+    "deployment": {
+      "deploymentURIsLocal": ["$SERVICE:8080"]
+    }
+  }
+}
--- a/test/assets/mongodb-datastore_evaluation_triggered_template.json
+++ b/test/assets/mongodb-datastore_evaluation_triggered_template.json
@@ -1,0 +1,19 @@
+{
+  "type": "sh.keptn.event.hardening.evaluation.triggered",
+  "specversion": "1.0",
+  "source": "travis-ci",
+  "contenttype": "application/json",
+  "data": {
+    "project": "$PROJECT",
+    "stage": "hardening",
+    "service": "$SERVICE",
+    "deployment": {
+      "deploymentURIsLocal": ["mongodb-datastore:8080"]
+    },
+    "labels": {
+      "nr_projects": "$project_nr",
+      "nr_services_per_project": "$service_nr",
+      "nr_evaluations_per_service": "$NR_NR_EVALUATIONS_PER_SERVICE",
+    }
+  }
+}
--- a/test/assets/mongodb-performance.jmx
+++ b/test/assets/mongodb-performance.jmx
@@ -1,0 +1,194 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jmeterTestPlan version="1.2" properties="4.0" jmeter="4.0 r1823414">
+  <hashTree>
+    <TestPlan guiclass="TestPlanGui" testclass="TestPlan" testname="Test Plan" enabled="true">
+      <stringProp name="TestPlan.comments"></stringProp>
+      <boolProp name="TestPlan.functional_mode">false</boolProp>
+      <boolProp name="TestPlan.serialize_threadgroups">false</boolProp>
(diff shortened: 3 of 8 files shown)


christian-kreuzberger-dtx (approved):
Thanks for the detailed report + screenshot, I love it!

As a follow-up task we can now create an automated load-test for evaluating Keptn with Keptn :)

PR #2818, opened 2020-12-16
#2816 Properly set result field of <task>.triggered event

### LINKED ISSUE #2816: Shipyard-controller does not set result field of next .triggered event
**Affected version**: [master](https://github.com/keptn/keptn/commit/67a99a016b9ea4cb48ee057cf813da894253bc6e)

**Problem description**:
After triggering an artifact-delivery sequence with an artifact that violates the quality gates for a stage, the `data.result` field of the `evaluation.finished` event was properly set to `fail`. However, the `release.triggered` event that was sent out next had the `data.result` field set to `pass`. This was most likely caused by the merging mechanism that merges all previous `.finished` events and includes their data in the following `.triggered` event: https://github.com/keptn/keptn/blob/67a99a016b9ea4cb48ee057cf813da894253bc6e/shipyard-controller/api/event.go#L784

**Proposed solution:**

We need to make sure that, after merging the `.finished` events, the `data.status` and `data.result` fields are set properly, meaning that they have the same value as the `.finished` event for the task that was just finished.

bacherfl:
Closes #2816

This PR also makes sure that the shipyard controller does not trigger a sequence in another stage if the last task.finished event does not a `data.result` field with its value set to `pass`

### DIFF
--- a/shipyard-controller/api/event.go
+++ b/shipyard-controller/api/event.go
@@ -573,7 +573,11 @@ func (sc *shipyardController) proceedTaskSequence(eventScope *keptnv2.EventData,
 			sc.logger.Error("Could not complete task sequence " + eventScope.Stage + "." + taskSequence.Name + " with KeptnContext " + event.Shkeptncontext)
 			return err
 		}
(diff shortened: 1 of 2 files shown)


warber (approved):
nice that we have it covered with a unit test ;)

Issue #2756, opened 2020-12-03
Extend JMeter to also download data files before executing a test


grabnerandi:
Currently the jmeter service only downloads the jmx file and executes it. 
This is limiting the use of JMeter because its not possible to reference any e.g: data files

We should exted jmeter.conf.yaml to reference additional files that should be downloaded into the same local directory in the jmeter pod so that the .jmx file can reference them without an issue

christian-kreuzberger-dtx:
> We should exted jmeter.conf.yaml to reference additional files that should be downloaded into the same local directory in the jmeter pod so that the .jmx file can reference them without an issue

Would it also work if jmeter-service downloads everything that's in the `jmeter` folder (including sub-folders)?

grabnerandi:
Yes. That would even be better. And - a feature that every service could benefit from. In the end we say that files should be uploaded to a specific folder in the config repo. If there would be an easy way to "clone" either the full repo or get that partial directory tree it would be great help and would also reduce the need to additoinally configure which files to download!

christian-kreuzberger-dtx:
@grabnerandi it looks like you are working on this? If so I would like to assign it to you :)

PR #2825, opened 2020-12-18
JMeter Dynatrace OSS and Data File Support

### LINKED ISSUE #2552: Extend JMeter Service to load JMeter Extensions such as Prometheus or Dynatrace Backend Listener
JMeter standalone is already really powerful but there are many extensions available in the open source community that bring additional value to JMeter. Examples are Backend Listener Plugins for [Prometheus](https://github.com/johrstrom/jmeter-prometheus-plugin) or [Dynatrace Backend Listener Plugin for JMeter](https://github.com/dynatrace-oss/jmeter-dynatrace-plugin).

After a discussing this with @christian-kreuzberger-dtx I want to propose two solutions to implement this
**Solution 1: Package useful extensions into Keptn JMeter Container Image**
The easiest solution would be to automatically package extensions with the container during build time and put them into the lib/ext subdirectory of the JMeter installation. The question here is which extensions should we include? I propose the two mention above!

**Solution 2: Dynamic Loading of Extensions**
Another option could be to allow our Keptn users to upload their extensions to the Keptn Project Configuration Repo. I would suggest to upload them on the master/main branch into a directory called jmeter/lib/ext. The Keptn JMeter Service can then pull all extensions before launching JMeter and copy them into the lib/ext directory. 
This solution would however require the Keptn Configuration API to provide an option to get the list of all files in a directory which is currently not possible. Without that capability its not possible to know which files have been uploaded

Open for feedback on the idea and my proposed options

grabnerandi:
This PR closes #2552 #2756 

*Dynatrace OSS Support*
Includes loading https://github.com/dynatrace-oss/jmeter-dynatrace-plugin/ which allows streaming JMEter data to Dynatrace

*Data File Support*
Allows downloading of ANY jmeter specific files such as other jmx or data files. Enables data driven test execution


jmeter-service/deploy/service.yaml
```diff
@@ -36,10 +36,22 @@ spec:
...
+            secretKeyRef:
+              name: dynatrace
+              key: DT_TENANT
         - name: EVENTBROKER
           value: 'http://event-broker/keptn'
+        - name: DEBUG_KEEP_TEMP_FILES
+          value: "true"
       - name: distributor
-        image: keptn/distributor:0.6.1
+        image: keptn/distributor:0.7.2
```
grabnerandi, line 54: update to 0.7.3

jmeter-service/deploy/service.yaml
```diff
@@ -26,7 +26,7 @@ spec:
     spec:
       containers:
       - name: jmeter-service
-        image: keptn/jmeter-service:latest
+        image: grabnerandi/jmeter-service:0.7.4
```
grabnerandi, line 29: reference keptn/jmeter-service:0.7.3-patch1

jmeter-service/common.go
```diff
@@ -52,14 +276,19 @@ func getJMeterConf(project string, stage string, service string, logger *keptn.L
...
 		}
 	} else {
-		resourceHandler := keptnapi.NewResourceHandler("configuration-service:8080")
-		keptnResourceContent, err := resourceHandler.GetServiceResource(project, stage, service, JMeterConfFilename)
+
+		logger.Info(fmt.Sprintf("Loading %s for %s.%s.%s", JMeterConfFilename, project, stage, service))
+
+		keptnResourceContent, err := GetKeptnResource(project, stage, service, JMeterConfFilename)
+
+		/* resourceHandler := keptnapi.NewResourceHandler(GetConfigurationServiceURL())
```
grabnerandi, line 284: remove comments

jmeter-service/deploy/service.yaml
```diff
@@ -36,10 +36,22 @@ spec:
...
+              key: DT_API_TOKEN
+        - name: DT_TENANT
+          valueFrom:
+            secretKeyRef:
+              name: dynatrace
+              key: DT_TENANT
         - name: EVENTBROKER
           value: 'http://event-broker/keptn'
+        - name: DEBUG_KEEP_TEMP_FILES
+          value: "true"
```
grabnerandi, line 52: set value to "false"

bacherfl (dismissed):
lgtm

jmeter-service/Dockerfile
```diff
@@ -56,7 +60,8 @@ RUN    apk update \
 	&& curl -L --silent ${JMETER_DOWNLOAD_URL} >  /tmp/dependencies/apache-jmeter-${JMETER_VERSION}.tgz  \
 	&& mkdir -p /opt  \
 	&& tar -xzf /tmp/dependencies/apache-jmeter-${JMETER_VERSION}.tgz -C /opt  \
-	&& rm -rf /tmp/dependencies
+	&& rm -rf /tmp/dependencies \
+	&& curl -L --silent ${DYNATRACE_EXTENSION_URL} > /opt/apache-jmeter-${JMETER_VERSION}/lib/ext/jmeter-dynatrace-plugin-${DYNATRACE_EXTENSION_VERSION}-SNAPSHOT.jar
```
christian-kreuzberger-dtx, line 64: I think this is okay for the time being, but long-term we **most not** download additional dependencies/plugins for JMeter and put them into the container. 
We will end up getting 100 of those dependencies and we will be responsible for ensuring that all of them are up2date, and if anything breaks it will also be our job to investigate.

jmeter-service/events/senddeploymentfinished.sh
```diff
@@ -0,0 +1,38 @@
...
+  VERSION="1.0"
+fi
+
+if [[ -z "$USER" ]]; then
+  USER="noname"
+fi
+
+if [[ -z "$PROJECT" || -z "$STAGE" || -z "$SERVICE" || -z "$URL" || -z "$TESTSTRATEGY" ]]; then
+  echo "Please specify project, stage, service, test strategy and URL. Optionally specify your version and user:"
+  echo "Usage: ./senddeployfinished.sh PROJECT STAGE SERVICE TESTSTRATEGY URL [VERSION] [USER]"
```
christian-kreuzberger-dtx, line 21: Script Filename is wrong, should be `senddeploymentfinished.sh`.

If you want this easier, you can use "$0" for the script filename in bash, e.g.:
```
#!/bin/bash

echo "use $0 --help for more info"
```

```
$ ./test.bash
use test.bash --help for more info
```

jmeter-service/jmeter/jmeter.conf.yaml
```diff
@@ -3,13 +3,25 @@ spec_version: '0.1.0'
 workloads:
   - teststrategy: performance
     vuser: 100
-    loopcount: 10
-    script: jmeter/load.jmx
+    loopcount: 500
+    script: jmeter/basicload.jmx
```
christian-kreuzberger-dtx, line 7: Will this break our end-to-end tutorials? We expect the filename to be load.jmx and basiccheck.jmx
```
keptn add-resource --project=sockshop --stage=dev --service=carts --resource=jmeter/basiccheck.jmx --resourceUri=jmeter/basiccheck.jmx
keptn add-resource --project=sockshop --stage=staging --service=carts --resource=jmeter/load.jmx --resourceUri=jmeter/load.jmx
```
In addition, the higher `loopcount` means that the load test is executed more often, therefore our tutorial will take longer?

jmeter-service/main.go
```diff
@@ -204,12 +204,16 @@ func runWorkload(serviceUrl *url.URL, testInfo *TestInfo, id string, workload *W
 		return true, nil
 	}
 
-	os.RemoveAll(workload.TestStrategy + "_" + testInfo.Service)
-	os.RemoveAll(workload.TestStrategy + "_" + testInfo.Service + "_result.tlf")
+	// the resultdirectory is unique as it contains context but also gives some human readable context such as teststrategy and service
+	// this will also be used for TSN parameter
+	resultDirectory := fmt.Sprintf("%s_%s_%s", workload.TestStrategy, testInfo.Service, testInfo.Context)
```
christian-kreuzberger-dtx, line 209: Thinking of troubleshooting with multiple projects, services, and stages, would it make sense to also use the project name and stage here?
E.g., 
```
fmt.Sprintf("%s_%s_%s_%s_%s", testInfo.Project, testInfo.Service, testInfo.Stage, workload.TestStrategy, testInfo.Context)
```

jmeter-service/Dockerfile
christian-kreuzberger-dtx: I have created a follow-up issue on that: https://github.com/keptn/keptn/issues/2832

jmeter-service/jmeterUtils.go
```diff
@@ -9,117 +9,59 @@ import (
...
-		logger.Error(err.Error())
-		return false, err
-	}
-	defer testScriptFile.Close()
+/**
+ * Returns additoinal JMeter Command Line Parameters including additional params passed to the JMeter script
+ */
+func addJMeterCommandLineArguments(testInfo *TestInfo, initialList []string) []string {
+	dtTenant := fmt.Sprintf("-JDT_TENANT=%s", os.Getenv("DT_TENANT"))
+	dtAPIToken := fmt.Sprintf("-JDT_API_TOKEN=%s", os.Getenv("DT_API_TOKEN"))
```
christian-kreuzberger-dtx, line 47: Not sure if this is ideal. In dynatrace-service we use `dynatrace.conf` to determine the secret that should be used. Here we refer to an environment variable based on a static secret.

jmeter-service/deploy/service.yaml
```diff
@@ -36,10 +36,22 @@ spec:
...
+        - name: DT_API_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: dynatrace
+              key: DT_API_TOKEN
+        - name: DT_TENANT
+          valueFrom:
+            secretKeyRef:
+              name: dynatrace
+              key: DT_TENANT
```
christian-kreuzberger-dtx, line 48: This will fail if the secret does not exist. Please make it optional by adding `optional: true` (see https://stackoverflow.com/a/48209289 for an example)

jmeter-service/Dockerfile
grabnerandi: agree - but - I was only able to add this plugin and so far I havent come across any other plugins that make sense. if anybody else wants to add plugins feel free to suggest

jmeter-service/jmeter/jmeter.conf.yaml
grabnerandi: this conf file - AFAIK - is not used but just shown as a reference. All our examples bring their own. But - I agree. I will revert it to what it was before

jmeter-service/jmeterUtils.go
grabnerandi: yes - because I didnt want to have a dependency to dynatrace.conf.yaml but simply allow additonal "secrets" to be passed to the JMEter script. I think this should to included into our general discussion of passing "secrets/endpoint configuraitons" to any type of keptn service

jmeter-service/main.go
grabnerandi: yeah - thats a valid point. I thought that test strategy and service would be unique. but - adding all these makes sense. just want to make sure we dont run itno any directory length restrictions

jmeter-service/events/senddeploymentfinished.sh
grabnerandi: Thanks for that hint. Added this to my bash script

jmeter-service/deploy/service.yaml
grabnerandi: Thanks. Added that to my service.yaml

christian-kreuzberger-dtx (dismissed):
LGTM

christian-kreuzberger-dtx (approved):
LGTM

bacherfl (approved):
lgtm

Issue #2824, opened 2020-12-18
Various improvements for GH actions


christian-kreuzberger-dtx:
This issue is a follow-up issue for PR #2817 . In addition, it acts as a meta-issue for making sure the GH Actions CI workflow is working as expected.

Plus:
* [x] Automatically create a dev (or latest) release which stores the CLI and Helm-Chart such that it can be directly downloaded
* [x] Automatically create a release draft for changes on release branches
* [x] Move certain "bash scripts" into a separate re-usable (for keptn-contrib/sandbox) GH action
* [x] Enable MacOS builds again once maintenance is complete
* [x] Upgrade next-semvers action to allow pre-releases https://github.com/WyriHaximus/github-action-next-semvers/releases/tag/v1.1.0

Issue #2811, opened 2020-12-15
Migrate integration tests from Travis-CI to GH action


christian-kreuzberger-dtx:
All integration tests that are currently running on Travis CI need to be migrated to GH Action.

Ideally, the scripts stay the same, but there might be some restrictions on GH action that require changes regarding K3s, Microk8s, Minikube, ...

For GKE they should work as they did before.

PR #2835, opened 2020-12-18
#2824 Replace helm-chart appVersion of latest with the actual version

christian-kreuzberger-dtx:

This PR replaces `appVersion: latest` with (e.g.) `appVersion: 0.7.4-dev-PR-2824` in the helm-chart that is being built and published as an artifact.

It's a requirement for #2811.

### DIFF
--- a/gh-actions-scripts/build_installer_helm_chart.sh
+++ b/gh-actions-scripts/build_installer_helm_chart.sh
@@ -13,6 +13,7 @@ helm dependency build ${BASE_PATH}/keptn/charts/control-plane
 
 # replace "appVersion: latest" with "appVersion: $VERSION" in all Chart.yaml files
 find -name Chart.yaml -exec sed -i -- "s/appVersion: latest/appVersion: ${VERSION}/g" {} \;
+find -name Chart.yaml -exec sed -i -- "s/version: latest/version: ${VERSION}/g" {} \;
 
 helm package ${BASE_PATH}/keptn --app-version $VERSION --version $VERSION
 if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


gh-actions-scripts/build_installer_helm_chart.sh
```diff
@@ -11,6 +11,9 @@ BASE_PATH=installer/manifests
 helm repo add nats https://nats-io.github.io/k8s/helm/charts/
 helm dependency build ${BASE_PATH}/keptn/charts/control-plane
 
+# replace "appVersion: latest" with "appVersion: $VERSION" in all Chart.yaml files
+find -name Chart.yaml -exec sed -i -- "s/appVersion: latest/appVersion: ${VERSION}/g" {} \;
```
agrimmer, line 15: Do we also need to properly set the `version`?

christian-kreuzberger-dtx: I can add that too, if wanted/needed. This would be the version of the chart itself, right?

christian-kreuzberger-dtx: Fixed.

agrimmer: Yes right. It is the version of the chart.

agrimmer (approved):
lgtm

gh-actions-scripts/build_installer_helm_chart.sh
agrimmer: Until now, the appVersion always matches the chart version.

PR #2862, opened 2020-12-23
Fixed Keptn Release Docs URL

christian-kreuzberger-dtx:

While generating docs for 0.8.x alpha I noticed that we need to reflect the version change in version.go within the CLI.

### DIFF
--- a/cli/cmd/version.go
+++ b/cli/cmd/version.go
@@ -35,15 +35,16 @@ var (
 	Version string
 )
 
+const keptnReleaseDocsURL = "0.8.x" // ToDo: Can we automate this?
+
 const versionCheckInfo = `Daily version check is %s. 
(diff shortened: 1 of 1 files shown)


bacherfl (approved):

agrimmer (approved):
lgtm

cli/cmd/version.go
```diff
@@ -35,15 +35,16 @@ var (
 	Version string
 )
 
+const keptnReleaseDocsURL = "0.8.x" // ToDo: Can we automate this?
```
agrimmer, line 38: Yes: The version is input for the CLI build. 
By convention, we always use `MAJOR.MINOR.x` as docs path.

christian-kreuzberger-dtx: Okay, but only if it's a semantic-version. I'll create an issue for this.

christian-kreuzberger-dtx: https://github.com/keptn/keptn/issues/2863

Issue #2880, opened 2021-01-07
Helm Services may create wrong deploymentURILocal as exposed port is not included


grabnerandi:
When Keptn deploys a service that is exposed via a different port other than http (80) - then the Helm Service will not correctly add the actual port information to the deploymentURILocal field that is sent as part of the Deployment Finished Event.

An example is the sockshop sample app and the catalog service which is exposed on port 8080.

The problem with this is that other services - such as JMeter - are taking the deploymentURILocal field and try to run tests against. Those tests will fail because JMeter assumes the service is internally exposed on port 80.

I think the issue is in workflow_engine_replace.go.getLocalDeploymentURI where no port information is retrieved when constructing that URL

**Environment**
* Keptn Version (`keptn version`): 0.7.3

**Affected Component**
* [ ] Docs
* [ ] CLI
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [x ] Helm
* [ ] Istio
* [ ] Bridge
* [ ] Gatekeeper
* [ ] Datastore
* [ ] REST API
* [ ] eventbroker / distributor
* [ x] jmeter

PR #2883, opened 2021-01-07
#2880 set port of local deployment URI based on K8s service

### LINKED ISSUE #2881: Helm Service is not correctly replacing keptn variables in generated helm charts
When onboarding a service the Helm service is generating additional helm charts for e.g: canary/primary deployments.

I have observed a weird behavior that leads to incorrectly generated deployment.yamls for the primary deployment.
Here is a snippet of my initial deployment yaml including several usages of the variable .Values.keptn.deployment:

```
spec:
  replicas: {{ .Values.replicaCount }}
  selector:
    matchLabels:
      app: simplenode
  template:
    metadata: 
      labels:
        app: simplenode
        app.kubernetes.io/name: {{ .Values.keptn.service }}
        app.kubernetes.io/part-of: "{{ .Values.keptn.project }}"
        app.kubernetes.io/managed-by: Keptn
        app.kubernetes.io/version: {{ (split ":" .Values.image)._1 | default "latest" }}
    spec:
      containers:
      - name: simplenode
        image: "{{ .Values.image }}"
        imagePullPolicy: IfNotPresent
        env:
        - name: DT_CUSTOM_PROP
          value: "keptn_project={{ .Values.keptn.project }} keptn_service={{ .Values.keptn.service }} keptn_stage={{ .Values.keptn.stage }} keptn_deployment={{ .Values.keptn.deployment }}"
        - name: KEPTN_PROJECT
          value: "{{ .Values.keptn.project }}"
        - name: KEPTN_STAGE
          value: "{{ .Values.keptn.stage }}"
        - name: KEPTN_SERVICE
          value: "{{ .Values.keptn.service }}"
        - name: KEPTN_DEPLOYMENT
          value: "{{ .Values.keptn.deployment }}"
``` 

When using this during `keptn onboard service` the resulting file simplenode-generated/templates/simplenode-primary-deployment.yaml looks like this where the deployment value is correctly replaced in DT_CUSTOM_PROP (shows as primary) but incorrectly in KEPTN_DEPLOYMENT (it is shown as canary)

```
spec:
  replicas: 1
  selector:
    matchLabels:
      app: simplenode-primary
  template:
    metadata:
      labels:
        app: simplenode-primary
        app.kubernetes.io/name: simplenode-primary
        app.kubernetes.io/part-of: student014
        app.kubernetes.io/version: 3.0.0
    spec:
      containers:
      - env:
        - name: DT_CUSTOM_PROP
          value: keptn_project=student014 keptn_service=simplenode keptn_stage=prod
            keptn_deployment=primary
        - name: KEPTN_PROJECT
          value: student014
        - name: KEPTN_STAGE
          value: prod
        - name: KEPTN_SERVICE
          value: simplenode
        - name: KEPTN_DEPLOYMENT
          value: canary
```

I am not sure how it is possible that the same placeholder is replaced with two different values. This currently leads to inconsistent information as this data is passed to the monitoring tool

**Environment**
* Kubernetes version (`kubectl version`): 01.17

**Affected Component**
* [ ] Docs
* [ ] CLI
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [x] Helm
* [ ] Istio
* [ ] Bridge
* [ ] Gatekeeper
* [ ] Datastore
* [ ] REST API
* [ ] jmeter

bacherfl:
Closes #2881 

### DIFF
--- a/helm-service/controller/deployment_handler.go
+++ b/helm-service/controller/deployment_handler.go
@@ -2,7 +2,6 @@ package controller
 
 import (
 	"fmt"
-
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	keptnevents "github.com/keptn/go-utils/pkg/lib"
 	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
@@ -10,6 +9,7 @@ import (
 	"github.com/keptn/keptn/helm-service/pkg/helm"
 	"github.com/keptn/keptn/helm-service/pkg/mesh"
 	"helm.sh/helm/v3/pkg/chart"
+	corev1 "k8s.io/api/core/v1"
 )
 
 // DeploymentHandler is a handler for doing the deployment and
@@ -154,6 +154,27 @@ func (h *DeploymentHandler) catchupGeneratedChartOnboarding(deploymentStrategy k
 	return h.onboarder.OnboardGeneratedChart(userChartManifest, event, deploymentStrategy)
 }
 
+func (h *DeploymentHandler) getPortOfDeployedService(e keptnv2.EventData) (string, error) {
+	userChartManifest, err := h.getHelmExecutor().GetManifest(helm.GetReleaseName(e.Project, e.Stage, e.Service, false),
+		e.Project+"-"+e.Stage)
+
+	if err != nil {
+		return "", err
+	}
+	services := helm.GetServices(userChartManifest)
+	if len(services) > 0 {
+		if len(services[0].Spec.Ports) > 0 {
+			for _, port := range services[0].Spec.Ports {
+				if port.Protocol == corev1.ProtocolTCP {
+					return fmt.Sprintf("%d", port.Port), nil
+				}
+			}
+		}
+
+	}
+	return "80", nil
+}
+
 func (h *DeploymentHandler) getStartedEventData(inEventData keptnv2.EventData) keptnv2.DeploymentStartedEventData {
 
 	inEventData.Status = keptnv2.StatusSucceeded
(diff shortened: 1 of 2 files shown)


helm-service/controller/deployment_handler.go
```diff
@@ -154,6 +154,27 @@ func (h *DeploymentHandler) catchupGeneratedChartOnboarding(deploymentStrategy k
...
+		if len(services[0].Spec.Ports) > 0 {
+			for _, port := range services[0].Spec.Ports {
+				if port.Protocol == corev1.ProtocolTCP {
+					return fmt.Sprintf("%d", port.Port), nil
+				}
+			}
+		}
+
+	}
+	return "80", nil
```
christian-kreuzberger-dtx, line 175: Can we add some logging here, stating that no services could be found and that we are defaulting to port 80?

helm-service/controller/deployment_handler.go
```diff
@@ -154,6 +154,27 @@ func (h *DeploymentHandler) catchupGeneratedChartOnboarding(deploymentStrategy k
...
+	}
+	services := helm.GetServices(userChartManifest)
+	if len(services) > 0 {
+		if len(services[0].Spec.Ports) > 0 {
+			for _, port := range services[0].Spec.Ports {
+				if port.Protocol == corev1.ProtocolTCP {
+					return fmt.Sprintf("%d", port.Port), nil
+				}
+			}
+		}
```
christian-kreuzberger-dtx, line 172: This will obviously be problematic for services that expose more than one port. 
Thinking of a service with multiple exposed ports, we could introduce a best practice to always take the lowest port that is exposed, what do you think @grabnerandi @bacherfl ?

helm-service/controller/deployment_handler.go
christian-kreuzberger-dtx: Actually, thinking about this: If no (valid) services could be found, would it make sense to raise an error?

helm-service/controller/deployment_handler.go
bacherfl: Agreed, I think taking the lowest port per default is a reasonable approach

helm-service/controller/deployment_handler.go
bacherfl: Yes, raising an error would be the right approach I think. Although in this case (i.e. when no port is defined), it will not reach that code anyway since the deployment will fail - K8s Services always require a port to be defined

christian-kreuzberger-dtx: Then let's raise an error here, just in case, so we can see if this happens.

helm-service/controller/deployment_handler.go
christian-kreuzberger-dtx: Okay, could you please adapt the implementation here to follow that approach?

bacherfl: Done - the implementation now follows that approach

helm-service/controller/deployment_handler.go
bacherfl: I have changed the logic to raise an error if a service has been defined, but no port could be found. Further, I changed the code to only generate a deploymentURI if a service has actually been defined in the helm chart.

christian-kreuzberger-dtx: That sounds nice, thanks!

christian-kreuzberger-dtx (approved):
LGTM, let's give this a try!

Issue #2646, opened 2020-11-06
Distributor: handle empty values of ENV variables more reliably


warber:
It was observed that the distributor container of the api-service pod crashes due an invalid value of the `PUPBSUB_TOPIC` environment variable (which was set to `''`).

The distributor crashes with:
`failed to subscribe to topic: nats: invalid subject`

However, the distributor does not crash, if the `PUBSUB_TOPIC` env var is explicitly set to `none` or if it is not present at all.

PR #2643 fixed the problem by removing the environment variable(s), however a more reliable solution should be to make the handling of the env var variables more robust.

christian-kreuzberger-dtx:
We should refine this issue (or dismiss it, if it's no longer needed).

johannes-b:
Before working on this issue, let's merge: https://github.com/keptn/keptn/pull/2884

johannes-b:
Has been fixed with: https://github.com/keptn/keptn/pull/2884

PR #2884, opened 2021-01-07
#2841 Use distributor as proxy for Keptn API

### LINKED ISSUE #2841: Helm-service and jmeter-service do not support https and x-token based communication with configuration endpoint
When running the helm-service and jmeter-service on the execution plane, they cannot communicate with the control-plane that exposes the configuration endpoint on `https` . The reason is that the used handlers - like ProjectHandler - have a fallback to `http`.  https://github.com/keptn/keptn/blob/04ba0b4b0bc41ea20ea4a4d5c9b88d17cf4f0cef/helm-service/main.go#L136

**As discussed**
The `distributor` should become the central component for communicating with the control-plane due to the fact that then just the distributor needs to know the access token. 

**Task**
- [ ] Enhance the distributor to communicate with the control plane via https:// and an x-token
- [ ] Distributor: Intercept calls to the configuration-service
- [ ] jmeter-service used `CONFIGURATION_SERVICE_URL` in code, but in Helm chart `CONFIGURATION_SERVICE` is defined --> change it to CONFIGURATION_SERVICE in code; https://github.com/keptn/keptn/blob/master/installer/manifests/keptn/charts/continuous-delivery/templates/continuous-deployment.yaml#L275
- [ ] In distributor support the possibility to skip TLS certificate validation: 
```
func main() {
    tr := &http.Transport{
        TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
    }
    client := &http.Client{Transport: tr}
    _, err := client.Get("https://golang.org/")
    if err != nil {
        fmt.Println(err)
    }
}
```

**Definition of Done:**
- Services in the execution plane can communicate with the control-plane (configuration API) via https and x-token

bacherfl:
Closes #2841

This will allow the helm-service and jmeter-service to run outside of the cluster in which Keptn is installed.
The distributor uses the following environment variables to configure access to the Keptn API:

`KEPTN_API_ENDPOINT`: The URL of the Keptn api, e.g. https://my-keptn.com/api
`KEPTN_API_TOKEN`: The Keptn API Token
`API_PROXY_PORT`: The port on which the distributor will listen for API requests, default = 8081

These variables will replace the previously used variables `HTTP_EVENT_POLLING_ENDPOINT` `HTTP_EVENT_ENDPOINT_AUTH_TOKEN` and `HTTP_EVENT_FORWARDING_ENDPOINT`

Using this new feature, execution plane services can send API requests to `localhost:<API_PROXY_PORT>` to communicate with the Keptn API (given that the distributor runs as a sidecar within the same pod). This means that the execution plane service itself does not have to know the API URL and the API Token.

The individual services within Keptn (e.g. datastore, configuration service) can be accessed based on the path of the request. The paths are mapped as follows:

```
"/mongodb-datastore":     "/api/mongodb-datastore"
"/datastore":             "/api/mongodb-datastore"
"/event-store":           "/api/mongodb-datastore"
"/configuration-service": "/api/configuration-service"
"/configuration":         "/api/configuration-service"
"/config":                "/api/configuration-service"
"/shipyard-controller":   "/api/shipyard-controller"
"/shipyard":              "/api/shipyard-controller"
```

Example: The endpoint `https://mykeptn.com/api/configuration-service/v1/project` can be called using `http://localhost:8081/configuration/v1/project`

**Example: Configuring the helm-service to use the API proxy**

To use the API proxy sidecar within the helm service, the env var pointing to the configuration-service and to the event endpoint  have to be set to the following values:

```
- name: CONFIGURATION_SERVICE
  value: 'http://localhost:8081/configuration'
- name: EVENTBROKER
  value: 'http://localhost:8081/event'
```

**NOTE:** In order for this feature to work completely, the following PR needs to be merged first: https://github.com/keptn/keptn/pull/2736


christian-kreuzberger-dtx:
@johannes-b I guess we also need docs for this (for creating integrations)?
@bacherfl does this have any impact on resource (memory/cpu) usage (you can measure it using Dynatrace)?

distributor/cmd/main.go
```diff
@@ -236,6 +366,14 @@ func forwardEventToAPI(event cloudevents.Event, apiEndpoint string) error {
 	return errors.New(string(body))
 }
 
+func getHTTPClient() *http.Client {
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: env.VerifySSL},
```
johannes-b, line 371: The naming of the environment variable VerifySSL is misleading because I would set it to `false` to disable it, but it determines whether to skip SSL verification.

christian-kreuzberger-dtx: We have set it to `HTTP_SSL_VERIFY` in dynatrace-service. Let's use the same naming here.

https://github.com/keptn-contrib/dynatrace-service/blob/81e8ff9dfef995e4cc42411e80e4353d59405fec/deploy/service.yaml#L126-L127

bacherfl:
> @bacherfl does this have any impact on resource (memory/cpu) usage (you can measure it using Dynatrace)?
I did some initial investigations on that (by triggering the artifact delivery task sequence a couple of times) and there was no noticeable increase in memory (~23MB) /CPU (~0.004 - 0.006%) when compared to other distributors without the API-proxy feature.

johannes-b (approved):
Last Friday, I successfully tested this PR.
From a usage point of view, this PR looks good to me. 

Before merging it, can we please provide an update of the documentation: https://keptn.sh/docs/0.8.x/integrations/custom_integration/#subscription-to-keptn-event

distributor/cmd/main.go
christian-kreuzberger-dtx: I think this was resolved in https://github.com/keptn/keptn/pull/2884/commits/31084d362f077430c956ae9fd98b063a691e533f

johannes-b:
Please close https://github.com/keptn/keptn/issues/2646 when this PR checks the environment variables.

johannes-b (approved):
great improvements!
lgtm

johannes-b (approved):
lgtm

Issue #2509, opened 2020-10-09
Create integration test that tests the Git upstream functionality for create and update project


christian-kreuzberger-dtx:
As a follow-up for #2493 we need an integration test that tries to

* create a project with upstream to GitHub (and maybe even other providers such as GitLab, Bitbucket?)
  * with existing branches and files
  * without any existing branches
* update a project (both, without having an upstream configured previously, and with having it configured previously)

I don't think we need to have a full test with onboard service etc..., just `keptn create project` should be sufficient.

christian-kreuzberger-dtx:
Looking at PR https://github.com/keptn/keptn/pull/2888, we should also create an integration test for `keptn update project`. I've added it to the issue description above.

christian-kreuzberger-dtx:
Next issue that could have been detected earlier with a proper integration test: https://github.com/keptn/keptn/issues/3334

thisthat:
This issue can be closed since at each integration test run, we create projects using gitea.

PR #2888, opened 2021-01-11
#2879 delete upstream repo credentials if there was an error

### LINKED ISSUE #2879: Failed update project with invalid upstream git still stores that upstream in keptns config repo
When calling `keptn update project PROJECT --git-user=user --git-token=token --git-remote-url=remote url` with invalid parameters, e.g: invalid URL or token then the CLI correctly returns an error like this

```
Error: Update project was unsuccessful. failed to set upstream: Error executing command git push --set-upstream http://keptn:********@git.student014/keptn/student014.git master: exit status 128
fatal: unable to access 'http://git.student014/keptn/student014.git/': Could not resolve host: git.student014
```

In the above case the URL was simply wrong.

While the error message correctly indicates that the update failed it seemed though that the upstream git information was stored in keptns config repo as the Keptns Bridge shows the invalid URL

image

**Environment**
* Client OS: Linux
* Kubernetes Cloud Provider (e.g., GKE, AKS): EKS
* Kubernetes version (`kubectl version`): 1.17

**Affected Component**

* [ ] Docs
* [x ] CLI
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [ ] Helm
* [ ] Istio
* [ ] Bridge
* [ ] Gatekeeper
* [ ] Datastore
* [ ] REST API
* [ ] jmeter

bacherfl:
Closes #2879 

In case there already was an upstream repository defined before updating the project with the wrong one, the previous credentials will be restored.

### DIFF
--- a/shipyard-controller/api/project.go
+++ b/shipyard-controller/api/project.go
@@ -315,7 +315,7 @@ func (pm *projectManager) updateProject(params *operations.CreateProjectParams) 
 	})
 
 	if errObj != nil {
-		msg := fmt.Sprintf("Could not update upstream repository of project %s: %s", *params.Name, errObj.Message)
+		msg := fmt.Sprintf("Could not update upstream repository of project %s: %s", *params.Name, *errObj.Message)
 
 		if delErr := pm.deleteUpstreamRepoCredentials(params); delErr != nil {
 			pm.logger.Error(fmt.Sprintf("Could not delete upstream repo credentials: %s", delErr.Error()))
--- a/shipyard-controller/api/project.go
+++ b/shipyard-controller/api/project.go
@@ -307,12 +307,21 @@ func (pm *projectManager) updateProject(params *operations.CreateProjectParams) 
 		}
 	}
 
-	pm.projectAPI.UpdateConfigurationServiceProject(keptnapimodels.Project{
+	_, errObj := pm.projectAPI.UpdateConfigurationServiceProject(keptnapimodels.Project{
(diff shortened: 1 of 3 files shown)


shipyard-controller/api/project.go
```diff
@@ -307,12 +307,21 @@ func (pm *projectManager) updateProject(params *operations.CreateProjectParams)
...
 		GitRemoteURI: params.GitRemoteURL,
 		GitToken:     params.GitToken,
 		GitUser:      params.GitUser,
 		ProjectName:  *params.Name,
 	})
+
+	if errObj != nil {
+		msg := fmt.Sprintf("Could not update upstream repository of project %s: %s", *params.Name, *errObj.Message)
+
+		if delErr := pm.deleteUpstreamRepoCredentials(params); delErr != nil {
```
christian-kreuzberger-dtx, line 320: If I'm not mistaken, this will only delete the upstream repo credentials, but not restore potentially previously correct credentials.

bacherfl: true. Although that raises to question if we do currently actually support the use case of updating the upstream repo multiple times? As far as I know we have not explicitly tested that use case yet, so maybe we need a separate issue for that

christian-kreuzberger-dtx: Agreed. I'll add it to #2509 (creating integration tests for the git upstream functionality).

christian-kreuzberger-dtx (approved):
LGTM, let's give it a try.

PR #2896, opened 2021-01-12
#2803 Check version of shipard file before starting task sequence

### LINKED ISSUE #2803: Shipyard-controller needs to checks whether the shipyard file is valid and has right version
When a `.triggered` event for a task sequence for a project is received, the shipyard controller initially retrieves the shipyard file for that project and checks whether it is valid. If not, the task sequence will not be executed, and a `.finished` event with `status = ‘errored’ ` and `result = ‘fail’` should be sent. The Message field of the event should contain information about how the shipyard differs from the current spec.

**Tasks:**
* When the shipyard file of a project is retrieved by the shipyard controller, check its validity
* If it is valid, proceed with executing the task sequence
* If it is not valid, send a .finished event with `data.status = ‘errored’` and `data.result = ‘fail’`, and data.message should contain information on how to fix/upgrade the shipyard

bacherfl:
Closes #2803

### DIFF
--- a/shipyard-controller/api/event.go
+++ b/shipyard-controller/api/event.go
@@ -519,8 +519,10 @@ func (sc *shipyardController) handleTriggeredEvent(event models.Event) error {
 		}, taskSequenceName)
 	}
 
+	// validate the shipyard version - only shipyard files following the '0.2.0' spec are supported by the shipyard controller
 	err = common.ValidateShipyardVersion(shipyard)
 	if err != nil {
+		// if the validation has not been successful: send a <task-sequence>.finished event with status=errored
 		sc.logger.Error("invalid shipyard version: " + err.Error())
 		return sc.sendTaskSequenceFinishedEvent(event.Shkeptncontext, &keptnv2.EventData{
 			Project: eventScope.Project,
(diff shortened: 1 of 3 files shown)


shipyard-controller/api/event.go
```diff
@@ -506,7 +506,31 @@ func (sc *shipyardController) handleTriggeredEvent(event models.Event) error {
...
+		sc.logger.Error("invalid shipyard version: " + err.Error())
+		return sc.sendTaskSequenceFinishedEvent(event.Shkeptncontext, &keptnv2.EventData{
+			Project: eventScope.Project,
+			Stage:   eventScope.Stage,
+			Service: eventScope.Service,
+			Labels:  eventScope.Labels,
+			Status:  keptnv2.StatusErrored,
+			Result:  keptnv2.ResultFailed,
+			Message: "Found shipyard.yaml with invalid version. Please upgrade the shipyard.yaml of the project using the Keptn CLI: 'keptn upgrade project " + eventScope.Project + " --shipyard'. '",
+		}, taskSequenceName)
```
christian-kreuzberger-dtx, line 533: Please add some comments to these code blocks

christian-kreuzberger-dtx (approved):
LGTM

PR #2900, opened 2021-01-13
#2804 Added shipyardVersion property to GET /project response

### LINKED ISSUE #2804: API endpoint for projects parses shipyard for each project and returns version or whether it is valid/invalid
The GET projects function in the configuration service (see https://github.com/keptn/keptn/blob/f22f425123ecc8d11a273c95336fa82ed3c79733/configuration-service/handlers/project.go#L26) should be adapted to also return information about the validity of the shipyard file. To avoid checking out the master branch of the project and retrieving the shipyard from the file system, we can use the `.Shipyard` property of the project object that has been stored in the materialized view. This field is updated each time a shipyard.yaml file has been uploaded to a project: https://github.com/keptn/keptn/blob/f22f425123ecc8d11a273c95336fa82ed3c79733/configuration-service/common/projects_materialized_view.go#L58 

**Remark:** The old shipyard has no dedicated version property. Consequently, return version 0.1.7 - which is the last released keptn/spec version. 

**Tasks:**
* Add a new field `ShipyardVersion` to the GET /project response payload (https://github.com/keptn/keptn/blob/f22f425123ecc8d11a273c95336fa82ed3c79733/configuration-service/swagger.yaml#L58) 
* When retrieving the projects, make sure this field is set for each project. If not, inspect the `.Shipyard` property that is already available for each project and set the `ShipyardVersion` property accordingly.
* Adapt the `UpdateShipyard` function (https://github.com/keptn/keptn/blob/f22f425123ecc8d11a273c95336fa82ed3c79733/configuration-service/common/projects_materialized_view.go#L58) of the materialized view to also set the `ShipyardVersion` property accordingly

bacherfl:
Closes #2804 

It turned out that up until now, the shipyard content was only persisted after updating a shipyard.yaml for a project, but not during the project creation. This has been fixed with this PR. 

### DIFF
--- a/configuration-service/common/projects_materialized_view.go
+++ b/configuration-service/common/projects_materialized_view.go
@@ -134,9 +134,11 @@ func (mv *projectsMaterializedView) GetProject(projectName string) (*models.Expa
 	if err != nil {
 		return nil, err
 	}
-	if err := setShipyardVersion(project); err != nil {
-		// log the error but continue
-		mv.Logger.Error(fmt.Sprintf("could not set shipyard version of project %s: %s", project.ProjectName, err.Error()))
+	if project != nil {
+		if err := setShipyardVersion(project); err != nil {
+			// log the error but continue
+			mv.Logger.Error(fmt.Sprintf("could not set shipyard version of project %s: %s", project.ProjectName, err.Error()))
+		}
 	}
 	return project, nil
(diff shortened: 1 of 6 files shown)


christian-kreuzberger-dtx (approved):
LGTM

Issue #2628, opened 2020-11-03
Refinement of the sequence tile 


johannes-b:
In the header of a sequence tile show: 
- <service> in <list of stages stage>
- Icon for the predefined sequence types: Deployment, Remediation, Evaluation
 - Requirement: Type of task sequence (add an optional property (values: deployment, remediation, evaluation) for a task sequence

As meta-information show: 
- Trigger
- Running for / From (is there a better approach to show the start time and duration?)
- Recent task / Last task 

**Behavior:**
- If the sequence is not finished, display a loading icon and add “started” to the name of the sequence
- If the sequence is finished, display the icon depending on the status: failed --> red, succeeded → green and add the status to the sequence name, e.g., Deployment *succeeded* 

![image](https://user-images.githubusercontent.com/729071/100993704-073f9880-3556-11eb-93b6-911efb3f1d21.png)

johannes-b:
I checked out the newest updates on the bridge and found an improvement that would make sense in course of this issue. @ermin-muratovic can you set the height of the list component to 100% to use the available space:

![image](https://user-images.githubusercontent.com/729071/106582874-c3144880-6544-11eb-861a-117e3b5d1397.png)

Issue #2627, opened 2020-11-03
Open details when selecting a sequence


johannes-b:
When clicking on a sequence, the details are displayed next to the sequence list. The sequence list is then shown in one stacked column, with the selected sequence on top. 

**Stage details:**
* Display stages in a timeline with stage name and start time of the task sequence in this stage.
* If the sequence is still active for a stage, show a loading icon 
* By default, the last stage of the sequence is selected, e.g., dev - staging - prod --> prod is selected. 

**Behavior:**
* When clicking on a stage, show meta-data, and tasks that belong to this sequence
* Meta-data about the stage:
  * Selected stage
  * Trigger
  * Git commit in case of a *Configuration change* 
* Tasks of this sequence:
  * Stacked representation of the tasks (youngest task on top)
  * The name of a task is extended by its status (started, succeeded, failed)
  * A representation of a task contains all events that belong to the task (deployment task: deployment.triggered, finished, started)
  * A task can be expanded containing details about the triggered Keptn-services and their progress/status

**Components:**
- Stage selector
- Breakdown of events

![image](https://user-images.githubusercontent.com/729071/100993577-df503500-3555-11eb-8345-ff2620e5e787.png)

johannes-b:
This issue has been broken down into: 
- #2907 
- #2938

johannes-b:
This issue can be closed since it is covered by others.

Issue #2625, opened 2020-11-03
Create sequence screen and load all triggers


johannes-b:
As a user, I click on the menu entry in the left panel that opens the sequence screen. Initially, this screen loads all triggers that triggered a sequence. 

From a technical perspective, this tab represents a new component:

![image](https://user-images.githubusercontent.com/729071/99404716-3f0ad700-28ec-11eb-86a9-5106e47e30f6.png)

*Note:* Introducing the icons as shown in the wireframe is out of the scope of this issue. 

**Tasks:**
- Add a new menu item: `Sequences`
- When clicking on the item, open the page with the list of all sequence triggers (filtering is out-of-scope since addressed in #2626)

PR #2902, opened 2021-01-13
#2625 Create sequence screen and load all triggers

ermin-muratovic:
Create sequence screen and load/display all triggers.

Since all triggers are actually "root" events, as shown in the service screen, the root events are polled regularly and distributed to the entities for the sequence and the service screen.

The `project-board` component will be reduced in to smaller components and is part of #1989 

The detail view in the `Sequence-Tab` is part of #2627 and the tiles will show more details which is part of #2628

image


bridge/client/app/_models/trace.ts
```diff
@@ -195,16 +198,19 @@ class Trace {
       if(this.isProblem() && this.isProblemResolvedOrClosed()) {
         this.label = EVENT_LABELS[EventTypes.PROBLEM_RESOLVED];
       } else if(this.isApprovalFinished()) {
-        this.label = EVENT_LABELS[EventTypes.APPROVAL_FINISHED][this.data.approval?.result] || this.type;
+        this.label = EVENT_LABELS[EventTypes.APPROVAL_FINISHED][this.data.approval?.result] || this.type.split(".").slice(-2)[0];
       } else {
-        this.label = EVENT_LABELS[this.type] || this.type;
+        this.label = EVENT_LABELS[this.type] || this.type.split(".").slice(-2)[0];
```
dominikmessner, line 203: [Suggestion] ```this.type.split(".").slice(-2)[0];``` could be extracted to a function so it's a bit easier to understand what it extracts (also in line 201)

dominikmessner (approved):

dominikmessner (approved):

dominikmessner (dismissed):

bridge/client/app/app.module.ts
```diff
@@ -36,10 +36,12 @@ import { DtTagModule } from '@dynatrace/barista-components/tag';
 import { DtTopBarNavigationModule } from "@dynatrace/barista-components/top-bar-navigation";
 import { DtCopyToClipboardModule } from "@dynatrace/barista-components/copy-to-clipboard";
 import { DtToggleButtonGroupModule } from "@dynatrace/barista-components/toggle-button-group";
+import {DtQuickFilterModule} from "@dynatrace/barista-components/experimental/quick-filter";
```
christian-kreuzberger-dtx, line 39: please align formatting of this import with the other imports (Note to myself: we need a linter :D )

bridge/client/app/_models/trace.ts
ermin-muratovic: Thanks. Extracted it to a function and actually changed also the returned value.

christian-kreuzberger-dtx:
Looks good, I'll give the final approval once the linter issue is fixed.

christian-kreuzberger-dtx (approved):
LGTM

bridge/client/app/app.module.ts
ermin-muratovic: Yes, we should activate a linter tool. I create a follow-up issue for this, see #2915

dominikmessner: Prettier is maybe also interesting and could make it a bit easier to introduce style rules to the existing code base (sorry, I'm a Prettier-fanboy :smile: )

PR #2906, opened 2021-01-14
#2905 fixed potential memleaks

### LINKED ISSUE #2905: Fix memory leak in shipyard-controller
When using the endpoint for retrieving all open `.triggered` events on the shipyard controller frequently, I noticed that the memory consumption was growing steadily, until the shipyard-controller pod ran into an OOM error and was restarted.

This is most likely caused by not calling the `cancel()` function of the context object that is being created when a connection to the mongodb is established:

https://github.com/keptn/keptn/blob/15a83681e5efe49f33adf736e0c9623dc836d6d9/shipyard-controller/db/mongodb_connection.go#L34

bacherfl:
Closes #2905

**Memory consumption before the fix:**
sc-memleak-before

**Memory consumption after the fix:**
sc-memleak-after

### DIFF
--- a/shipyard-controller/api/event.go
+++ b/shipyard-controller/api/event.go
@@ -156,20 +156,22 @@ type shipyardController struct {
 }
 
 func getShipyardControllerInstance() *shipyardController {
-	logger := keptncommon.NewLogger("", "", "shipyard-controller")
-	sc := &shipyardController{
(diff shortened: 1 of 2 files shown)


christian-kreuzberger-dtx (approved):
LGTM

PR #2917, opened 2021-01-15
#2625 Heatmap not showing up in Bridge

ermin-muratovic:
Since the `.triggered`, `.started` and `.finished` events of any type are now "merged" in the Bridge into one `Trace`, the heatmap was not showing up, since the results are provided with the `.finished` event, which now is not showing up as a separate tile anymore. 

This makes it easier to read the tasks that were executed, but also had to be considered for showing the evaluation results correctly.


christian-kreuzberger-dtx (approved):
LGTM

christian-kreuzberger-dtx:
image

PR #2954, opened 2021-01-21
#2870 improved logging in case of cluster API downtime

### LINKED ISSUE #2870: GET "/api/v1/metadata" returns null during K8s api downtime
With Keptn 0.8.0-alpha, accessing GET "/api/v1/metadata" returns null during K8s api downtime.

This is especially inconvenient as in the CLI we get the following message:
```
$ keptn-080-alpha status
* Warning: Your Keptn CLI version (0.8.0-alpha) and Keptn cluster version () don't match. This can lead to problems. Please make sure to use the same versions.Starting to authenticate
```

# Definition of Done
- [ ] Unit test available which tetest this edge case (the [fake](https://godoc.org/k8s.io/client-go/kubernetes/fake) package may be used to mock the kubernetes API)
- [ ] API returns a sensible error message (or even better a proper response) in case of Kubernetes Api not being reachable
- [ ] CLI needs a line-break after the message "Please make sure to use the same versions."

bacherfl:
Closes #2870

### DIFF
--- a/api/handlers/metadata.go
+++ b/api/handlers/metadata.go
@@ -23,46 +23,67 @@ import (
 // GetMetadataHandlerFunc returns metadata of the keptn installation
 func GetMetadataHandlerFunc(params metadata.MetadataParams, principal *models.Principal) middleware.Responder {
 
+	handler := newMetadataHandler()
+
+	return handler.getMetadata()
+}
+
+func newMetadataHandler() metadataHandler {
+	var clientSet kubernetes.Interface
 	logger := keptnutils.NewLogger("", "", "api")
-	logger.Info("API received a GET metadata event")
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		logger.Debug("Could not get InClusterConfig, will skip k8s-deployments.")
+	} else {
+		// creates the clientset
+		clientSet, err = kubernetes.NewForConfig(config)
+		if err != nil {
+			logger.Debug("Could not create kubernetes client, will skip k8s-deployments.")
+		}
+	}
+	return metadataHandler{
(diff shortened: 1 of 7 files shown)


api/handlers/metadata.go
```diff
@@ -23,46 +23,67 @@ import (
...
+	return handler.getMetadata()
+}
+
+func newMetadataHandler() metadataHandler {
+	var clientSet kubernetes.Interface
 	logger := keptnutils.NewLogger("", "", "api")
-	logger.Info("API received a GET metadata event")
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		logger.Debug("Could not get InClusterConfig, will skip k8s-deployments.")
```
christian-kreuzberger-dtx, line 36: Does it make sense that we log the error?

api/handlers/metadata.go
```diff
@@ -23,46 +23,67 @@ import (
...
+		clientSet, err = kubernetes.NewForConfig(config)
+		if err != nil {
+			logger.Debug("Could not create kubernetes client, will skip k8s-deployments.")
+		}
+	}
+	return metadataHandler{
+		k8sClient:       clientSet,
+		logger:          logger,
+		swaggerFilePath: "/swagger-ui/swagger.yaml",
+	}
```
christian-kreuzberger-dtx, line 48: What is returned here if `clientSet` is not set because of an error when retrieving the inClusterConfig?

api/handlers/metadata.go
bacherfl: absolutely, I have added the error message to the log output

christian-kreuzberger-dtx (approved):
LGTM

PR #2955, opened 2021-01-21
#2925 add indexes to mongodb collections

### LINKED ISSUE #2925: Improve mongodb datastore performance
Based on the outcome of the performance investigation reported in PR #2812, we can take some measures to improve the performance of the mongodb/mongodb-datastore to avoid performance issues in case the number of events within a project has grown to a certain point.

On the API-side, the following enhancements have shown a promising impact:

 - Collection *project-rootEvents*: Add an index for data.service
 - Collection *project*: Add indexes for data.service and shkeptncontext

On the client side, we need to avoid sending too many requests at once - i.e., by only loading root events for a service when the user has clicked on that service

bacherfl:
Closes #2925

So far the results seem promising. I executed the same test as described in #2812, and the response time pretty much stays constant at a median of 100ms, and 95th percentile of 200ms:

In addition to the indexes mentioned in #2925, adding an index for the property `type` in the collections containing a project's events proved to be beneficial (e.g. for loading the evaluation heatmap which queries a specific event type)

Screenshot 2021-01-22 at 08 05 56

### DIFF
--- a/mongodb-datastore/handlers/events.go
+++ b/mongodb-datastore/handlers/events.go
@@ -314,6 +314,7 @@ func dropProjectEvents(logger *keptncommon.Logger, event *models.KeptnContextExt
 
 	logger.Debug(fmt.Sprintf("Delete all root events of project %s", projectName))
 	createdIndexes[rootEventsCollection.Name()+"-data.service"] = false
+	createdIndexes[rootEventsCollection.Name()+"-time"] = false
 	err = rootEventsCollection.Drop(ctx)
 	if err != nil {
 		err := fmt.Errorf("failed to drop collection %s: %v", rootEventsCollection.Name(), err)
@@ -514,6 +515,13 @@ func findInDB(collectionName string, pageSize int64, nextPageKeyStr *string, onl
 			ctx,
 			collection,
 			"type",
+			logger,
+		)
+	} else {
+		ensureIndexExistsOnCollection(
+			ctx,
+			collection,
+			"time",
 			logger,
 		)
 	}
--- a/mongodb-datastore/handlers/events.go
+++ b/mongodb-datastore/handlers/events.go
@@ -303,6 +303,7 @@ func dropProjectEvents(logger *keptncommon.Logger, event *models.KeptnContextExt
 	logger.Debug(fmt.Sprintf("Delete all events of project %s", projectName))
 
 	createdIndexes[projectCollection.Name()+"-shkeptncontext"] = false
+	createdIndexes[projectCollection.Name()+"-type"] = false
 	createdIndexes[projectCollection.Name()+"-data.service"] = false
 	err := projectCollection.Drop(ctx)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mongodb-datastore/handlers/events.go
```diff
@@ -178,6 +173,31 @@ func insertEvent(logger *keptncommon.Logger, event *models.KeptnContextExtendedC
...
+		Keys: bson.M{
+			indexName: 1,
+		},
+	}
+	// CreateOne() is idempotent - this operation checks if the index exists and only creates a new one when not available
+	_, err := collection.Indexes().CreateOne(ctx, indexDefinition)
+	if err != nil {
+		// log the error, but continue anyway - index is not required for the query to work
+		logger.Debug("could not create index for " + collection.Name() + ": " + err.Error())
+	}
```
christian-kreuzberger-dtx, line 195: Can we improve this kind of error handling a little bit? Is there a way to check if the error is somethingl ike `already exists`?

If not, does it make sense to rename `createdIndexes` to `skipCreatingIndex`?

mongodb-datastore/handlers/events.go
```diff
@@ -275,39 +295,46 @@ func dropProjectEvents(logger *keptncommon.Logger, event *models.KeptnContextExt
...
+	rootEventsCollection := client.Database(mongoDBName).Collection(projectName + rootEventCollectionSuffix)
+	invalidatedEventsCollection := client.Database(mongoDBName).Collection(getInvalidatedCollectionName(projectName))
+
+	logger.Debug(fmt.Sprintf("Delete all events of project %s", projectName))
 
-	logger.Debug(fmt.Sprintf("Delete all events of project %s", collectionName))
-	err := collection.Drop(ctx)
+	createdIndexes[projectCollection.Name()+"-shkeptncontext"] = false
+	createdIndexes[projectCollection.Name()+"-type"] = false
+	createdIndexes[projectCollection.Name()+"-data.service"] = false
```
christian-kreuzberger-dtx, line 307: The code above, as well as the code a couple lines below
```
createdIndexes[rootEventsCollection.Name()+"-data.service"] = false
createdIndexes[rootEventsCollection.Name()+"-time"] = false
```

is painful to maintain - as in: We will forget about it in the future :)

If anything, at the end of `dropProjectEvents` we could have an assert or something to check if `createdIndexes` contains anything that starts with the project name (maybe the separator `-` can be replaced by something that's not allowed for project names?).

mongodb-datastore/handlers/events.go
bacherfl: I just checked - in case the index does already exist, the function does not return an error. renaming `createdIndexes` to `skipCreatingIndex` makes sense, as that will make the purpose of this var easier to understand

mongodb-datastore/handlers/events.go
bacherfl: good idea - using a separator that is not allowed within a keptn project name would do the trick. I will adjust this accordingly

bacherfl: did it a little differently - I added arrays containing the names of each index to be created for each collection. When deleting the collections for a project, I iterate over them and make sure the correlating values in the `skipCreateIndex` map are set to `false`. Also, when creating the indexes the service will use those arrays, which makes the code cleaner

christian-kreuzberger-dtx (approved):
LGTM

PR #2963, opened 2021-01-22
#2824 Report Docker Build to PR

christian-kreuzberger-dtx:

This PR adds a feature to our CI build where we post a comment with the updated docker images of the PR.

The comment is updated every time a new commit is built (so the previous comment gets deleted).

If no Docker Images have been updated, it looks like this:
image

And if an image is built, it looks like this:
image


johannes-b (approved):
nice! lgtm

PR #2964, opened 2021-01-22
#2914 check project name and stage names during project creation

### LINKED ISSUE #2914: API allows creating projects with special characters while CLI does not - leads to issues
Check the following screenshot. This project was created by accident by using the Keptn API. the CLI would never allow this as it does a name sanity check. We should enforce the same checks on the API because this project can now no longer be deleted

image

**Environment**

bacherfl:
Closes #2914 


christian-kreuzberger-dtx (approved):
LGTM

PR #2966, opened 2021-01-22
Introduce user-managed deployment strategy

warber:


helm-service/controller/onboarder.go
```diff
@@ -248,14 +99,18 @@ func (o *onboarder) OnboardGeneratedChart(helmManifest string, event keptnv2.Eve
...
-	} else {
+	} else if strategy == keptnevents.Direct {
 		o.getKeptnHandler().Logger.Debug(fmt.Sprintf("For service %s in stage %s with deployment strategy %s, a mesh chart is generated",
 			event.Service, event.Stage, strategy.String()))
 		generatedChart, err = o.chartGenerator.GenerateMeshChart(helmManifest, event.Project, event.Stage, event.Service)
 		if err != nil {
 			o.getKeptnHandler().Logger.Error("Error when generating the managed chart: " + err.Error())
 			return nil, err
 		}
+	} else {
```
christian-kreuzberger-dtx, line 110: `else` is always very implicit. Please add a comment for which strategies the following code is executed.

helm-service/main.go
```diff
@@ -79,69 +79,81 @@ func gotEvent(ctx context.Context, event cloudevents.Event) error {
...
+		onboardHandler := createOnboardHandler(url, keptnHandler, mesh)
+		onboardHandler.HandleEvent(event)
 	} else if event.Type() == keptnv2.GetTriggeredEventType(keptnv2.ActionTaskName) {
 		actionHandler := createActionTriggeredHandler(url, keptnHandler)
-		go actionHandler.HandleEvent(event)
+		actionHandler.HandleEvent(event)
 	} else if event.Type() == keptnv2.GetFinishedEventType(keptnv2.ServiceDeleteTaskName) {
 		deleteHandler := createDeleteHandler(url, keptnHandler)
-		go deleteHandler.HandleEvent(event)
+		deleteHandler.HandleEvent(event)
```
christian-kreuzberger-dtx, line 94: Why is `go `  removed in these calls? Is the multi-threading aspect of helm-service handled elsewhere?

warber: it is handled in the the cloudevents sdk

helm-service/controller/onboarder.go
warber: Actually, you cannot know. Its everything except bluegreen and direct ;) I would not mention this in a comment.

christian-kreuzberger-dtx: True, could be anything, but it's implemented with `user-managed` in mind. How about
```
// No need to generate further charts for other strategies, e.g., user-managed
```

agrimmer (dismissed):
LGTM

agrimmer:
This PR has been implemented in https://github.com/keptn/keptn/pull/3218

PR #2968, opened 2021-01-22
Improve integration tests for new-artifact

christian-kreuzberger-dtx:

During log analysis of our integration tests I noticed two things:

1. Our "wait timeouts" are way too high, leading to integration tests that are running for ages
2. While we waited for `carts-db` to be deployed, we never exited

This PR tries to fix both.


johannes-b (approved):
lgtm

Issue #2912, opened 2021-01-15
Keptn 0.8.0-alpha CLI crashes for auth after upgrade from 0.7.3


GustavAT:
Upgrading the CLI from 0.7.3 to 0.8.0-alpha as described in the [Docs](https://keptn.sh/docs/0.8.x/operate/upgrade/) results in the CLI crashing when patching namespaces or authenticating again.

**Environment**

* Client OS (e.g., Linux, OSX, Windows): Ubuntu 20.04
* Keptn Version (`keptn version`): 0.8.0-alpha
* Kubernetes Cloud Provider (e.g., GKE, AKS): GKE
* Kubernetes version (`kubectl version`): v1.16.15-gke.4300

**Affected Component**

* [ ] Docs
* [X] CLI
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [ ] Helm
* [ ] Istio
* [ ] Bridge
* [ ] Gatekeeper
* [ ] Datastore
* [ ] REST API
* [ ] eventbroker / distributor
* [ ] jmeter

**Describe the bug**

After upgrading the CLI as described in the [Docs](https://keptn.sh/docs/0.8.x/operate/upgrade/), commands such as `keptn auth` or `keptn upgrade --patch-namespace` crash the CLI.

**To Reproduce**

Have the keptn CLI 0.7.3 installed.

Download the 0.8.0-alpha release (`keptn-0.8.0-alpha-linux-amd64.tar.gz
`)  from [GitHub](https://github.com/keptn/keptn/releases/tag/0.8.0-alpha) and extract it.

Replace old keptn installation:
`mv keptn-0.8.0-alpha-linux-amd64 /usr/local/bin/keptn`
Set execute permissions
`chmod +x /usr/local/bin/keptn`

`> keptn version` Shows the expected version and tells us that the CLI needs to be authenticated against the cluster
```bash
Keptn CLI version: 0.8.0-alpha
Keptn cluster version: This command requires to be authenticated. See "keptn auth" for details
```

`> keptn auth` crashes
```bash
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/keptn/keptn/cli/cmd.smartKeptnCLIAuth(0xc00003e0c5, 0xc00083dd58, 0xc00083dd58, 0x13de9a4)
	/home/runner/work/keptn/keptn/cli/cmd/auth.go:250 +0x617
[...]

PR #2733, opened 2020-11-25
Keptn Smart Auth: Autofetch keptn endpoint and api-token

ankitjain28may:
Signed-off-by: ankitjain28may <ankitjain28may77@gmail.com>

PR #2971, opened 2021-01-24
Fixed the patch-namespace flag and also added the error message to run auth  command without upgrade

ankitjain28may:
The new flag `--patch-namespace` was introduced in the PR #2733 but the calling function was later being removed somehow because of which the CLI was throwing error/crashing as the issue raised in #2912 

This has been fixed and also added the error message while running `keptn auth` without patching the namespace.

Signed-off-by: ankitjain28may <ankitjain28may77@gmail.com>


christian-kreuzberger-dtx (approved):
LGTM

Issue #2976, opened 2021-01-25
Integration Tests: use newer istio version


christian-kreuzberger-dtx:
In our current integration test setup we are using Istio 1.6.14:
https://github.com/keptn/keptn/blob/354f662c4b56cd65daac92affccdf502abf6d721/test/utils/install_istio.sh#L6

As already stated in https://github.com/keptn/tutorials/issues/118, Istio 1.6 is no longer supported, and we should upgrade to a newer version.

For our integration tests, I believe we can go to the latest 1.8 (currently istio 1.8.2) release.

christian-kreuzberger-dtx:
Fixed in PR #2979

Issue #2973, opened 2021-01-25
GKE: Enable Shielded GKE Nodes for Integration Tests and in Docs


christian-kreuzberger-dtx:
![image](https://user-images.githubusercontent.com/56065213/105685080-fb3cdb00-5ef5-11eb-860a-4f28d7d5df78.png)

This is a security feature which will be enabled by Default on GKE soon(ish) (once Kubernetes 1.18 is used).
We should also enable this in our integration tests, and document this in our GKE installation guide
* in quickstart https://keptn.sh/docs/quickstart/
* and in the 0.8.x docs https://keptn.sh/docs/0.8.x/operate/install/
*

christian-kreuzberger-dtx:
Fixed in PR https://github.com/keptn/keptn/pull/2979 and in docs https://github.com/keptn/keptn.github.io/pull/712

PR #2979, opened 2021-01-25
#2976 #2973 Upgrade istio, enable GKE shielded nodes

christian-kreuzberger-dtx:

This PR addresses the first parts of #2976 and #2973:

* Upgrade Istio to 1.8.2 in integration tests
* Use GKE Shielded Nodes in integration tests
* In addition, I used the time to remove the istio addon from our GKE installation (we don't want to be forced to use the old version of istio)

We still need to adapt our docs.


Kirdock (approved):
LGTM

Issue #2908, opened 2021-01-14
keptn get project: display shipyard version


christian-kreuzberger-dtx:
Now that #2804 has been implemented and merged, we could show the shipyard version when executing
```
$ keptn get projects
NAME		CREATION DATE                 SHIPYARD VERSION
sockshop	2020-12-23T16:04:23+01:00     spec.keptn.sh/0.1.7
unleash		2020-12-28T16:33:08+01:00     spec.keptn.sh/0.1.7
sockshop-dt	2020-12-29T10:11:46+01:00     spec.keptn.sh/0.2.0
```

christian-kreuzberger-dtx:
The PR has already been merged, is this issue done?

johannes-b:
I double-checked the code and can confirm that it has been implemented.

PR #3000, opened 2021-01-26
#2899 added padding to tabular output & added shipyard version to get…

### LINKED ISSUE #2899: Keptn tabular cli output breaks automation with too long project, stage or service names
In some of my automation scripts I use the Keptn CLI to e.g: iterate over projects or services.
The problem though is that parsing the CLI output breaks as there is no space between columns when names get too long. See the following output

image

This for instance breaks the following script that wants to just get the project names but ends up merging project name + creation date:
image

Automation can obviously be done through the API as well - but - I think this should be an easy fix which enables automation through the CLI as well

**Environment**

warber:
fixes #2899 and #2908

### DIFF
--- a/cli/cmd/get_projects.go
+++ b/cli/cmd/get_projects.go
@@ -99,9 +99,9 @@ keptn get project sockshop -output=json  # Returns project details in JSON forma
 		if !mocking {
 
 			w := new(tabwriter.Writer)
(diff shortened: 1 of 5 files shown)


christian-kreuzberger-dtx (approved):
LGTM

Issue #2948, opened 2021-01-19
helm-service: Check length of release name


agrimmer:
The Helm release name is limited to 53 characters. See e.g. https://github.com/helm/helm/issues/6006#:~:text=Jump%20to%20bottom-,Why%20Helm%20release%20name%20is%20a%20DNS,max%20length%20of%2053%20characters%3F

If now the release name is too long, the `helm-service` cannot successfully execute the upgrade. Furthermore, Helm does not give a good error message here (I do not have any at hand)

**Implementation details:**
The `helm-service` uses the following release name patterns:
- [NAMESPACE]-[SERVICE-NAME]
- [NAMESPACE]-[SERVICE-NAME]-generated

With these release names, we can easily run into the 53 character limit.

A possible solution that would lower the risk of running into the limit would be to remove the namespace from the release name. Please let's use this issue to discuss this.

**Tasks:**
- Add a check for the length of the release name

christian-kreuzberger-dtx:
> A possible solution that would lower the risk of running into the limit would be to remove the namespace from the release name.

1. We would run into an issue with duplicate service names in different projects/namespaces, right? (Or can we distinguish helm-releases by namespace)
2. If we change the naming, do we need to migrate old service names (before Keptn 0.8.x)?


As a counter-proposal, we could keep the same naming format, but automatically limit the service-name to, e.g.:
```
max_len_service_name=53 - len(NAMESPACE) - 1 - len("generated") - 1
SERVICE_NAME=SERVICE_NAME[0:max_len_service_name]
```

In fact, we could even save some strings if we change `generated` to `gen` (though I believe our service names sometimes are called primary/canary, right?).

> Add a check for the length of the release name
We should definately add a check.

agrimmer:
Thanks for your input!

A Helm release lives in a namespace. Hence, there should be no collisions.

But, the migration will not be easy and needs in-depth research.

Furthermore, using only a substring of the Keptn service name could lead to collisions.

christian-kreuzberger-dtx:
As a quick-fix I would recommend that we limit `SERVICE_NAME` as I posted it above.
This won't break anything, but fix the problem for now.

We should furthermore make sure that this gets documented.

johannes-b:
Please add the length check to the shipyard-controller.

cpanato:
hello, this is open for contribution?

christian-kreuzberger-dtx:
[...]

PR #3005, opened 2021-01-26
shipyard-controller: check service name lenght to not be bigger than 53 chars

cpanato:
Add validation to check if the service name is > 53 chars

- add tests
- organize the import to follow good practice, can revert if you don't like it, also I can apply to the other parts as well 
- refactor the existing test file to use testify, if you don't like I can revert, if you like it I can open a PR to make it for the other tests 😄 

Fixes: https://github.com/keptn/keptn/issues/2948


shipyard-controller/api/service.go
```diff
@@ -130,11 +136,17 @@ func validateCreateServiceParams(params *operations.CreateServiceParams) error {
...
+
 	if !keptncommon.ValididateUnixDirectoryName(*params.ServiceName) {
 		return errors.New("Service name contains special character(s). " +
 			"The service name has to be a valid Unix directory name. For details see " +
 			"https://www.cyberciti.biz/faq/linuxunix-rules-for-naming-file-and-directory-names/")
 	}
+
+	if len(*params.ServiceName) > serviceNameMaxLen {
+		return errors.New("Service name need to be less than 53 characters")
+	}
```
christian-kreuzberger-dtx, line 148: This check is a little bit too soft :)
Not sure if we have described this properly in #2948, but what we need is that the service name is short enough, such that the generated helm chart release name is less than 53 characters.
The helm chart release name is: `${PROJECTNAME}-${STAGENAME}-${SERVICENAME}-generated`, therefore this string needs to be less than 53 characters (e.g., `len(ServiceName) < 53 - len(ProjectName) - len(StageName) - len("generated") - 2`). 

In addition, we would like to verify this for all stages, not just the current stage (see [Keptn's Shipyard File](https://github.com/keptn/spec/blob/0.2.0-alpha/shipyard.md#stage)).

As a referene, you can take a look at the `createService` function in this file. It has the projectName and it also fetches all stage names via
```
stages, err := sm.stagesAPI.GetAllStages(projectName)
```

christian-kreuzberger-dtx (changes requested):
Hi and thanks for the PR!
I've left a comment regarding the check for the length of the service name.

shipyard-controller/api/service.go
```diff
@@ -3,14 +3,20 @@ package api
...
+	"net/http"
+
 	"github.com/gin-gonic/gin"
 	"github.com/google/uuid"
+
 	keptncommon "github.com/keptn/go-utils/pkg/lib/keptn"
 	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
 	"github.com/keptn/keptn/shipyard-controller/common"
 	"github.com/keptn/keptn/shipyard-controller/models"
 	"github.com/keptn/keptn/shipyard-controller/operations"
```
christian-kreuzberger-dtx, line 15: Imports look much cleaner this way, thanks! I think we should autoformat it via our IDE anyway...

shipyard-controller/api/service_test.go
```diff
@@ -1,17 +1,65 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
 	keptnapimodels "github.com/keptn/go-utils/pkg/api/models"
 	keptnapi "github.com/keptn/go-utils/pkg/api/utils"
 	keptncommon "github.com/keptn/go-utils/pkg/lib/keptn"
 	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
 	"github.com/keptn/keptn/shipyard-controller/models"
 	"github.com/keptn/keptn/shipyard-controller/operations"
```
christian-kreuzberger-dtx, line 14: Same here, imports look much cleaner this way, thanks.

shipyard-controller/api/service.go
cpanato: thanks for your review @christian-kreuzberger-dtx 
I did not realize the stage and project. I did a small change, please let me know if that is good otherwise let's discuss the best place to add the validation.

cpanato: humm, looks like new changes in the file, will rebase and check the conflicts

shipyard-controller/handler/servicemanager.go
```diff
@@ -39,6 +45,11 @@ func (sm *serviceManager) createService(projectName string, params *operations.C
 	}
 
 	for _, stage := range stages {
+		sm.logger.Info(fmt.Sprintf("Validating service %s", *params.ServiceName))
+		if err := validateServiceName(projectName, stage.StageName, *params.ServiceName); err != nil {
+			return sm.logAndReturnError(fmt.Sprintf("could not create service %s in stage %s of project %s: %s", *params.ServiceName, stage.StageName, projectName, err.Error()))
+		}
+
```
christian-kreuzberger-dtx, line 52: The validation itself looks okay, but I think it's in the wrong place.

In the worst case you will create services in two stages (e.g., `dev`, `staging`), but in the third stage (`production`) you hit the length and it will return.

I know the for-loop is not optimal in itself, but to avoid introducing new side-effects, let's move your check in a separate for loop above the existing once.

```
	for _, stage := range stages {
		sm.logger.Info(fmt.Sprintf("Validating service %s", *params.ServiceName))
		if err := validateServiceName(projectName, stage.StageName, *params.ServiceName); err != nil {
			return sm.logAndReturnError(fmt.Sprintf("could not create service %s in stage %s of project %s: %s", *params.ServiceName, stage.StageName, projectName, err.Error()))
		}
	}

	for _, stage := range stages {
		...
	}
```

cpanato: @christian-kreuzberger-dtx thanks that makes sense, I thought on that but was unsure. thanks again for your review
PTAL

christian-kreuzberger-dtx (approved):
Looks good to me! @warber can you please take a look at the tests?

cpanato:
@christian-kreuzberger-dtx would be welcome if  I open a PR to update the tests to use testify/require?

christian-kreuzberger-dtx:
I think you'd have to sync with @warber because there are some bigger changes planned regarding tests. If you're on the Keptn Slack, I can get the two of you in a chat.

cpanato:
@christian-kreuzberger-dtx Im not in that slack, but would love to work togheter

christian-kreuzberger-dtx:
https://slack.keptn.sh if you want ;) 
For now, I recommend that you write a GH issue detailing your plan and maybe also telling us which files would be affected. From there we can have the discussion.

Issue #1512, opened 2020-03-17
Automatically create/update a github release/tag as a draft when pushing something to a release-branch


christian-kreuzberger-dtx:
To lighten our release process I propose the following Github Action:

* Within the releasing workflow, a draft release should be automatically created when pushing something to a release-* branch. If a release is already published/exists, the workflow/gh action needs to exit (if we ever re-release something, we will need to some manual steps anyway).

* Automatically create the release-notes from the [releasenotes](releasenotes/) folder and add them to the draft (only as long as the release is in *draft* state).

* Determine cpu/memory requests and limits per K8s deployment (after a fresh keptn install) and report it to the Github release (e.g., as a markdown file)

* Report integration test results to the release (e.g., as a markdown file)

* Automatically add CLI binaries and helm-chart to the release (and only update them as long as the release is in *draft* state)
   Note: Please take care of correct naming, see [existing releases](https://github.com/keptn/keptn/releases) as examples

**Note**: I did some research, and updating a *draft* release is not always straight forward (creating a *draft* release is). It might be possible that we do not automatically update a *draft* release, but just create it on demand using an action.

PR #3010, opened 2021-01-27
#2824 #1512 Automate creating draft release

christian-kreuzberger-dtx:

Implements #1512.

This PR introduces a job in the integration-test workflow that publishes a draft release. In addition, it captures the Kubernetes resource limits/requests for cpu/memory for OpenShift and for GKE (indicated by `COLLECT_RESOURCE_LIMITS` in the test matrix).

Noteworthy changes:
* keptn install: distinguish between the full installation (`--use-case=continuous-delivery`) and just control-plane - this should slightly speed up our integration tests, in addition it was needed to get proper resource consumption reports
* if a release with the same VERSION exists -> do not overwrite existing published releases -> abort workflow
* If a draft releasewith the same VERSION exists -> delete the draft release -> continue (this allows us to get a clean draft release)
* collecting resources limits etc.. is done in https://github.com/keptn/keptn/blob/feature/2824/release_drafter/test/utils/k8s_collect_resources.sh - it's not perfect, but this is as good as an automation can get (there is an open issue on Kubernetes that would help us a lot https://github.com/kubernetes/kubernetes/issues/55046 )

Eventually, a draft release is created, containing:
* release specific release notes (if available, else releasenotes_develop.md)
* integration test report
* docker build report
* some CI debug variables

I have removed the input variable that specifies that a release needs to be published. Instead, we always publish a draft release (unless the release already exists).

---

Result:
https://github.com/keptn/keptn/releases/tag/untagged-fc941536ed41752b7983
image


johannes-b (dismissed):
The end result LGTM, but I would propose to ask someone else for validating the code.

.github/workflows/integration_tests.yml
```diff
@@ -269,15 +286,21 @@ jobs:
           source test/utils.sh
           echo "Verifying that services and namespaces have been created"
 
-          # verify the deployments within the keptn namespace
+          # verify the deployments within the keptn namespace (for keptn control plane)
```
bacherfl, line 289: please also add the statistics-service to the checks

christian-kreuzberger-dtx: done. I also rebased from master branch, because statistics-service was not part of this when I started.

bacherfl (approved):
lgtm

Issue #3009, opened 2021-01-27
Mongodb-datastore: Refactoring for better testability


bacherfl:
Currently, the mongodb datastore is quite hard to test, mainly because dependencies such as the mongodb client are declared as global `vars` (https://github.com/keptn/keptn/blob/b424443440126eec46cb2919355a8f469b87ed7a/mongodb-datastore/handlers/events.go#L30) and initialized within the implementation of the endpoint handlers.
This should be refactored to have a struct for the API endpoint handlers to which we can inject dependencies such as the mongodb client. These dependencies then can be easily mocked during unit tests.

Also, it would make sense to migrate to gin-gonic since the approach of generating the boilerplate code from the swagger definition of the service proved to be cumbersome (PRs got way to big even with small changes to the API).

PR #3011, opened 2021-01-27
#2949 provide backwards compatibility for evaluation-done event queries

### LINKED ISSUE #2949: Mongodb datastore: support backwards compatibility for evaluation-done events used in Keptn < 0.8
Since the [evaluation-done](https://github.com/keptn/spec/blob/0.1.7/cloudevents.md#evaluation-done) events used in previous Keptn versions are an important part of the quality gates use case, we need to provide the possibility to still support them after we have switched to the new `evaluation.finished` events used in spec v0.2.0 (e.g. they might still be useful for comparison-based evaluations).

To still be able to access those events, we need to adapt the [GetEvents()](https://github.com/keptn/keptn/blob/1b0e37d93486defc4a634f61449ea98e30655b25/mongodb-datastore/handlers/events.go#L349) and [GetEventsByType](https://github.com/keptn/keptn/blob/1b0e37d93486defc4a634f61449ea98e30655b25/mongodb-datastore/handlers/events.go#L627) functions in the mongodb datastore in the following way:

If the provided event type is set to `sh.keptn.event.evaluation.finished` we will also need to search for events of the type `sh.keptn.events.evaluation-done`, transform their structure to be compatible with the current spec of the `sh.keptn.event.evaluation.finished`, and add them to the result set returned by the API

bacherfl:
Closes #2949 

Note: in general, the mongodb-datastore service is hard to unit test and needs a bigger refactoring - I have created an Issue for that: https://github.com/keptn/keptn/issues/3009


johannes-b (dismissed):
lgtm

johannes-b (approved):
lgtm

PR #3014, opened 2021-01-27
Changed DT entity tag filters for qg integration test

bacherfl:
This should solve the problem of not receiving any data from the DT metrics API in our qg integration tests


christian-kreuzberger-dtx (approved):
LGTM

PR #3022, opened 2021-01-28
#2988 added timeout to helm install

### LINKED ISSUE #2988: keptn install hangs in case of ImagePullBackOff
From current master branch:

When I execute `keptn install` with a helm chart that references a docker image that does not exist, the installation triggered by `keptn install` hangs (and will never complete).

# Definition of Done

- [ ] Installation should abort after a certain amount of time (e.g., 10 minutes) with an error message
- [ ] If possible, can we check for ImagePullBackOff errors based on the helm chart and notify the user via the cli?

bacherfl:
Closes #2988 

Added a timeout of 10minutes to the helm upgrade command - QQ: would it make sense to make this configurable as a parameter to the install command? 

In case of a timeout, added a hint to check the status of the installation via `kubectl get pods -n <keptn-namespace>`. In issue #2988 it was suggested to specifically check for an `ImagePullBackOff` in the deployments, but given that a timeout might have numerous reasons, I was thinking a more general hint might be better suited in this case.

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -244,8 +244,8 @@ func doInstallation() error {
 	}
 
 	if err := helm.NewHelper().UpgradeChart(keptnChart, keptnReleaseName, keptnNamespace, values); err != nil {
-		logging.PrintLog("Could not complete Keptn installation: "+err.Error(), logging.InfoLevel)
-		return err
+		msg := fmt.Sprintf("Could not complete Keptn installation: %s \nFor troubleshooting, please check the status of the keptn deployment by executing the following command: \n\nkubectl get pods -n %s\n", err.Error(), keptnNamespace)
+		return errors.New(msg)
 	}
 
 	logging.PrintLog("Keptn has been successfully set up on your cluster.", logging.InfoLevel)
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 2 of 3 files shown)


cli/cmd/install.go
```diff
@@ -244,8 +244,8 @@ func doInstallation() error {
 	}
 
 	if err := helm.NewHelper().UpgradeChart(keptnChart, keptnReleaseName, keptnNamespace, values); err != nil {
-		logging.PrintLog("Could not complete Keptn installation: "+err.Error(), logging.InfoLevel)
```
bacherfl, line 247: Note: removed the `logging.Printlog` since the error message was printed twice otherwise.

cli/cmd/upgrade.go
```diff
@@ -207,8 +208,8 @@ func doUpgrade() error {
 	}
 
 	if err := helm.NewHelper().UpgradeChart(keptnUpgradeChart, keptnReleaseName, keptnNamespace, nil); err != nil {
-		logging.PrintLog("Could not complete Keptn upgrade: "+err.Error(), logging.InfoLevel)
```
bacherfl, line 210: Note: removed the `logging.Printlog` since the error message was printed twice otherwise.

christian-kreuzberger-dtx:
> Added a timeout of 10minutes to the helm upgrade command - QQ: would it make sense to make this configurable as a parameter to the install command?
IMO we could make it configurable, but I don't see the need of it right now.

> In case of a timeout, added a hint to check the status of the installation via kubectl get pods -n <keptn-namespace>. In issue #2988 it was suggested to specifically check for an ImagePullBackOff in the deployments, but given that a timeout might have numerous reasons, I was thinking a more general hint might be better suited in this case.

Sounds good.

christian-kreuzberger-dtx (approved):
I like it, LGTM

Issue #2938, opened 2021-01-19
Show task details in sequence details


ermin-muratovic:
When clicking on a sequence, the details are displayed next to the sequence list. 
Each step/task in the sequence may contain different details, which should be display here. Also the status (success/failed) should be easily visible and the error message should be displayed properly.

**Component behavior:**
* Display steps/task which have been executed in the selected stage.
* If the step/task is still active, show a loading icon 
* If the step/task is finished, show all details that are provided in the payload in an expandable tile
  - All triggered services and the status of the finished event
  - Show error message if status failed
  - Show the result if status is successful

![100993577-df503500-3555-11eb-8345-ff2620e5e787-seqdetails](https://user-images.githubusercontent.com/6098219/105020872-2633b480-5a48-11eb-86ea-0a83963cf576.png)

christian-kreuzberger-dtx:
On the right side it says "2 services triggered", I'm wondering if we should call it "triggered", or if we just say "2 services have reacted" or just "2 services".

Reasoning (please correct me if I am wrong here):
In Keptn 0.8.x, we don't actually trigger multiple services, instead it is possible that multiple services return a `.started` and `.status.changed` and `.finished` event.

johannes-b:
Is part of: #2672

PR #3030, opened 2021-01-28
#2938 Show task details in sequence details

ermin-muratovic:
Show sequence tasks filtered by stage and show progress details of each task in tiles.

sequence-screen-details-2021-01-28


ermin-muratovic:
Show sequence tasks filtered by stage and show progress details of each task in tiles.

image

When expanding a task, you see the details (started, finished, message) and the payload of each step.

image

Kirdock:
The result of sequence tasks should be taken from `.finished` instead of `.triggered` traces

Kirdock (changes requested):
There are some issues
- The selected task is not reset after changing the project. The last selected task remains.
- If a task detail has labels, the expand-icon is offset.
image
- It is not possible to send manual approvals. The detail is empty.
image
- If a sequence does not have stages, an image or service name, it's incorrectly displayed
image

Kirdock (approved):
LGTM

Issue #2936, opened 2021-01-19
Upgrade CLI and Keptn-services to latest Go release (e.g., go 1.16)


christian-kreuzberger-dtx:
At the moment we are using Go 1.13, which is quite old (released in September 2019).

To stay up2date with the latest and greatest things as well as security fixes that Go is providing, we should upgrade to ~1.15 (latest release as of writing this issue) or even~ **1.16** (scheduled for Feb 2021).

In addition, [GH actions will also switch to Go 1.15 in Feb. 2021 as their default](https://github.com/actions/virtual-environments/issues/2447).

*Side note*: Go 1.16 will have support for ARM based Macs (M1, Apple silicon), so we can compile Keptn CLI natively!

# Blockers / Restrictions

We don't want to lose Dynatrace monitoring capabilities, so we need to make sure that a recently released version of OneAgent also supports the Go release we want to use, see https://www.dynatrace.com/support/help/shortlink/go#version-matrix for details
![image](https://user-images.githubusercontent.com/56065213/105004612-83be0600-5a34-11eb-8175-b5c96df8e631.png)
(Screenshot as of Jan 19th, 2021)

# Definition of Done

- [x] Before starting, make sure the desired Go Version is supported by a recent OneAgent release

- [x] CI (GH Actions) Scripts etc... are upgraded to use Go 1.15/1.16
- [x] All Docker Images are upgraded to use Go 1.15/1.16
- [x] go.mod files are upgraded to use Go 1.15/1.16
- [x] go-utils CI is upgraded to use Go 1.15/1.16
- [x] go-utils go.mod file is upgraded to use Go 1.15/1.16
- [x] kubernetes-utils CI is upgraded to use Go 1.15/1.16
- [x] kubernetes-utils go.mod file is upgraded to use Go 1.15/1.16
- [ ] CLI builds for apple silicon added in github workflows ci.yaml

warber:
@christian-kreuzberger-dtx @johannes-b 

> Side note: Go 1.16 will have support for ARM based Macs, so there is an added benefit of waiting until 1.16 is released and switching then.

[Go 1.16](https://blog.golang.org/go1.16) is now released.

christian-kreuzberger-dtx:
Thanks, I've updated the issue description. Unfortunately, OneAgent does not support 1.16 yet (but I'm sure it will soon).

christian-kreuzberger-dtx:
Latest version of oneagent supports Go 1.16.2

[...]

PR #3031, opened 2021-01-29
#2824 Cache go modules, node modules, and other things for faster CI builds

christian-kreuzberger-dtx:

* Added Caching for CI and integration-test workflows (this should speed up those workflows by a bit)
* Refactored test/utils/download-*.sh scripts such that they make use of the download/cache folder (separated downloads from the actual install scripts)
* Removed travis-scripts/download_gcloud.sh and created a better script in test/utils/
* Added a check whether internet connection is working (to verify #2936


johannes-b (approved):
* Adding the caching capabilities looks good.
* Like it to have separate download/install scripts 

lgtm

PR #3033, opened 2021-01-29
#2929 Improve CodeCoverage reporting

### LINKED ISSUE #2929: Fix coverage reporting
Our code coverage reporting on PRs often shows wrong numbers.

In addition, coverage reporting on GitHub README page seems broken:
image

Our coverage should be way higher, but it seems that it always dips down to 22% for no reason: https://codecov.io/gh/keptn/keptn

image

# Definition of Done
- [ ] Coverage Reporting on PR is correct 
- [ ] Coverage Reportingo n master is correct

christian-kreuzberger-dtx:

This PR tries to fix #2929 by doing the following:

* Added codecov reporting to CLI tests
* Added codecov call to every unit test (instead of a single call for all of them)
* Added a flag (`-F flagname`) when running codecov to split tests and codecov by directory
* Added `-c` flag to cleanup after sending the report (to avoid having a mix of reports)
* Added `coverage.txt` to .gitignore
* Added the `carryforward` flag for all directories, such that it *should* carry over the coverage from the last build (e.g., master branch) - this is recommended for monorepos, but it remains to be seen if it also works as expected.

Whether those changes have the desired effect or not can not be seen within this PR, but after it has been merged to master and builds are running for a week or so.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -290,66 +290,88 @@ jobs:
 
       - name: Test api
         if: (needs.prepare_ci_run.outputs.BUILD_EVERYTHING == 'true') || (needs.prepare_ci_run.outputs.BUILD_API == 'true')
-        run: go test -coverprofile=coverage.txt -covermode=atomic -v ./handlers/... ./utils/...
+        run: |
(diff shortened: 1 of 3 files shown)


johannes-b (approved):
lgtm

PR #3039, opened 2021-02-01
#2877 add: doc output to obtain endpoint & apiToken

### LINKED ISSUE #2877: Extend keptn auth with more doc output to explain how to obtain token & endpoint
This could be an easy win for user experience. When doing a keptn auth the current output says 
```
```

Would be better to reflect some of the documentation or at least point to the doc where we explain how to get token and endpoint
Here is a version that I think would help:
```
$ keptn auth
keptn auth requires api-token and endpoint.

For more information on how to obtain token and endpoint to go https://keptn.sh/docs/0.7.x/reference/cli/#authenticate-keptn-cli
```

DarthBenro008:
- Fixes #2877 

### Preview
image


cli/cmd/auth.go
```diff
@@ -180,6 +181,10 @@ func verifyAuthParams(authParams *authCmdParams) error {
 				return fmt.Errorf("Error in fetching the api-token\n" + err.Error() + "\nCLI is not authenticated")
 			}
 		}
+		namespace, err = smartKeptnCLIAuth()
+		if err != nil {
+			return err
+		}
```
DarthBenro008, line 187: The smartKeptnCLIAuth() check is done after the other checks are completed as proposed in https://github.com/keptn/keptn/issues/2877#issuecomment-770708493

DarthBenro008:
Closing this draft in favour of #3159

PR #3046, opened 2021-02-02
#2560 fix(cli): add better keptn upgrade message

### LINKED ISSUE #2560: Better instructions on how to download new version of CLI to do a keptn upgrade
As the Keptn team is constantly putting out new cool features and with that also an updated version of the CLI which is needed to upgrade to the latest Keptn I suggest some user experience improvements in that process.

#1: keptn version should give me the command to update my CLI
Currently when doing keptn version I get this
```console
Keptn CLI version: 0.7.1

keptn version 0.7.2 is available! Please visit https://keptn.sh for more information.
```

I suggest to provide an easy - here is how you upgrade your CLI command. So - instead of just saying "please visit the keptn website" - I suggest to also add the following:
```
To update to the latest CLI run this command: curl -sL https://get.keptn.sh | sudo -E bash
```
Another option would be to point the user to a site that shows the user how to download the latest CLI depending on the target OS!

#2: Do a version check with keptn upgrade
Right now when doing a keptn upgrade and the CLI version is not on the latest available keptn version we get the following error message which is not that descriptive to the end user on what to do next:
```console
$ keptn upgrade
Error: No upgrade path exists from Keptn version 0.7.1 to 0.7.1
```

I suggest that keptn upgrade should also highlight that there is a new keptn version available but first the user has to upgrade to the latest CLI in order to perform that updgrade

DarthBenro008:
- Fixes #2560 

### Preview
keptn_version
keptn_upgrade

### DIFF
--- a/cli/pkg/version/version_checker.go
+++ b/cli/pkg/version/version_checker.go
@@ -129,7 +129,9 @@ func (v *VersionChecker) getNewerCLIVersion(usedVersionString string) (available
 	return res, nil
 }
 
-const newCompatibleVersionMsg = `* Keptn CLI version %s is available! Please visit https://keptn.sh/docs/%s/operate/upgrade/ for more information.`
+const newCompatibleVersionMsg = `* Keptn CLI version %s is available! 
+ - To update to the latest CLI run this command: curl -sL https://get.keptn.sh | bash
+ - For more information, please visit https://keptn.sh/docs/%s/operate/upgrade/`
 const newIncompatibleVersionMsg = `* Keptn CLI version %s is available! Please note that this version might be incompatible with your Keptn cluster ` +
 	`version and requires to update the cluster too. Please visit https://keptn.sh/docs/%s/operate/upgrade/ for more information.`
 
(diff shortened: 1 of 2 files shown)


cli/pkg/version/version_checker.go
```diff
@@ -129,7 +129,9 @@ func (v *VersionChecker) getNewerCLIVersion(usedVersionString string) (available
 	return res, nil
 }
 
-const newCompatibleVersionMsg = `* Keptn CLI version %s is available! Please visit https://keptn.sh/docs/%s/operate/upgrade/ for more information.`
+const newCompatibleVersionMsg = `* Keptn CLI version %s is available! 
+ - To update to the latest CLI run this command: curl -sL https://get.keptn.sh | sudo -E bash
```
christian-kreuzberger-dtx, line 133: Please use the `curl ...` command without `sudo -E`, as shown in our [Quickstart](https://keptn.sh/docs/quickstart/#2-1-install-keptn-cli).

DarthBenro008: @christian-kreuzberger-dtx my bad, I presumed sudo would be required to mv the binary to /usr/local/bin, I have fixed it!

christian-kreuzberger-dtx: It is, but we moved it inside the get.keptn.sh script ;) thanks for the fix

christian-kreuzberger-dtx (approved):
LGTM

Issue #2297, opened 2020-09-07
Multi-architecture support for CLI (32 bit, ARM, ...)


christian-kreuzberger-dtx:
Right now our release on GitHub builds and contains the following binaries for the CLI:

![Screenshot from 2020-09-07 13-33-01](https://user-images.githubusercontent.com/56065213/92383471-ae0ac900-f10e-11ea-8cbf-9c1dbcf6855e.png)

As far as I can tell, we are building each of these for Linux, MacOS and Windows on 64 Bit for AMD/Intel:

* [darwin amd64](https://github.com/keptn/keptn/blob/0e49d8d74c29736329448535ca65e1c673306c73/travis-scripts/build_cli.sh#L10-L11)
* [linux amd64](https://github.com/keptn/keptn/blob/0e49d8d74c29736329448535ca65e1c673306c73/travis-scripts/build_cli.sh#L36-L37)
* [windows amd64](https://github.com/keptn/keptn/blob/0e49d8d74c29736329448535ca65e1c673306c73/travis-scripts/build_cli.sh#L62-L63)

Knowing that Apple is switching to ARM CPUs, we will end up losing users that can not use our binaries (unless MacOS has some magic in it to emulate AMD64 binaries, though I wouldn't to rely on that).
In addition, looking at other projects they provide multiple architectures and platforms for their binaries, e.g., [helm](https://github.com/helm/helm/releases):
![Screenshot from 2020-09-07 13-37-04](https://user-images.githubusercontent.com/56065213/92383798-3e490e00-f10f-11ea-9e94-dc0be40ed9b9.png)

I'm not sure if we need to support all of these, though if we touch this aspect or our build process anyway, why not? Let's discuss...

# To be discussed

## ARM support (Linux & MacOS)

Though compiling for ARM would make a lot of sense, as this would allow us to run the CLI on Raspberry PIs (with Linux) and later on MacBooks/etc... (once they release the first Mac with ARM CPUs), so adding those makes a lot of sense to me.

## 32 Bit

I don't have a strong opinion on 32 bit support (I do not see a reason to add it nor to not add it).

## PowerPC

I have no idea whether this is needed at all

## MIPS

Same as with PowerPC, I have no idea whether this is needed at all.

# Definition of Done

- [x] Have an agreement of which architectures should be supported
- [ ] Potentially add binaries for additional architectures to older releases (e.g., 0.7.0, 0.7.1)
[...]

PR #3051, opened 2021-02-02
#2297 Added cli builds for various architectures, rework makefile usage for cli

christian-kreuzberger-dtx:

Implements #2297

Now that we are getting closer to the release of Keptn 0.8.0, we need to
* re-enable macos builds (it's still slow unfortunately, therefore I'll keep unit tests disabled for macos)
* adding more build architectures (e.g., to support raspberry pis, macos with apple silicon, ...)
* move windows builds to linux and cross-compile (this apparently works... it's just macos that needs some extra care)
* Simplified build-script for cli builds in the CI workflow
* Make use of `make` for building and testing the cli
* Removed outdated travis-scripts folder


.github/workflows/CI.yml
```diff
@@ -425,7 +425,7 @@ jobs:
     strategy:
       matrix:
         go-version: [ 1.13.x ]
-        platform: [ ubuntu-20.04 ] #[ ubuntu-20.04, macOS-11.0 ] # , windows-2019
+        platform: [ ubuntu-20.04 ] #, macOS-11.0, windows-2019 ]
```
warber, line 428: is windows-2019 disabled on purpose?

christian-kreuzberger-dtx: Yes, as it was before this PR, I only restructured the list. 
Tests are currently failing on Windows, and on MacOS we don't have enough runners on GitHub, so the workflow often just stalls/times out.

bacherfl (approved):
lgtm

Issue #2475, opened 2020-10-05
Show key SLI in bold


johannes-b:
**Improvement:**
SLIs that are marked as [key_sli](https://github.com/keptn/spec/blob/master/service_level_objective.md#objectives):`true` has to be highlighted by bold in the Heatmap and SLI breakdown.

**Task:**
- Enhance the lighthouse-service to send the information (`key_sli:[true or false]) in the `sh.keptn.events.evaluation-done payload: data.evaluationdetails.indicatorResults[].value` 
- Use the info from the `key_sli` property to set the font to bold in the Heatmap
- Use the info from the `key_sli` property to set the font to bold in the SLI Breakdown

**Definition of Done:**
- [ ] Update of keptn/spec (for additional property key_sli), see here: https://github.com/keptn/spec/blob/master/cloudevents.md#evaluation-done
- [ ] Representation in Keptn Bridge - `repsonse_time_p95` is a key SLI (_Note_: The Heatmap does not show all SLIs as depicted by the SLI breakdown) 
![image](https://user-images.githubusercontent.com/729071/95076244-d12a9780-0711-11eb-96cb-6705f41071ff.png)

Godtide:
I’d like to take up this issue

Kirdock:
Adjusting the style in the heatmap seems not possible.
KeySli for SLI-breakdown is fixed via #2935 
KeySli in cloudevents is fixed via #3340

Issue #2478, opened 2020-10-05
SLI breakdown in table


johannes-b:
**Status quo**
The SLI breakdown of a quality gate result is shown as a list. Each entry of this list has the SLI name on the left and the value/criteria on the right. Criteria that are not met are strikethrough with a font color in red:
![image](https://user-images.githubusercontent.com/729071/95090447-96caf580-0725-11eb-89ce-c6d3667c9c59.png)

**Improvement:** 
Change this list to a table that allows sorting for the different columns. The columns should be: 

- Name
- Value
- pass Criteria
- warning Criteria
- Result 
- Score (default column for sorting) 

**Tasks**
- Get familiar with the table component from: https://barista.dynatrace.com/components/table
- Show SLIs in a table
- If the objective of an SLI resulted in a warning, show the score in orange; if failed --> red.
- By default sort the table based on the score
 
**Definition of Done**
![image](https://user-images.githubusercontent.com/729071/95090592-bc57ff00-0725-11eb-843a-295ce219063b.png)

pahofmann:
I'd like to work on this.

johannes-b:
Hi @pahofmann, this would be really great if you can work on that issue! Thanks! 👍 
I assigned it to you.

If you have a first preview of the table you would like to discuss, please ping me or @ermin-muratovic.

pahofmann:
I finally had time to start working on this and have a first version with the table:
![image](https://user-images.githubusercontent.com/5699104/98277792-0b4cba80-1f98-11eb-822d-85f711684b7b.png)

However I ran into a problem with the criteria. In the mock-up it's separated by pass and warning.
We do have that information in the SLO file, but not in the targets that a returned for display in the SLI table.

```javascript
0:
  name: "response_time_p90"
  score: 0.5
  status: "warning"
  targets: Array(3)
    0: {criteria: "<=5000", targetValue: 5000, violated: false}
    1: {criteria: "<3000", targetValue: 3000, violated: true}
    2: {criteria: "<=+200%", targetValue: 1448.7341558353953, violated: true}
  value: 4210 
```

Without that information we couldn't tell if the third (relative) criteria is a warning or pass criteria in the example above .

Would it be feasible to add the information if the criteria is pass/warning to the target?

johannes-b:
Hi @pahofmann
wow, that looks already great!

[...]

PR #3053, opened 2021-02-02
#2478 SLI evaluation breakdown as barista table (#3023)

johannes-b:
Definition of Done:

- [x]  Table refreshes when clicking on the heatmap
- [x] Fixed size of columns (name 55%, result 10%, value 10%, score 10%, criteria 15%), at least the Name needs more space
- [x]  Column Score shows the weight but should be translated into points between 0 and 100 In the above example, score 1 --> 33.3 points, score 2 --> 66.6 points
- [x]  Split of criteria in warning/pass (currently just the pass criteria are shown)
   - [x]  show warning criteria
- [x] Mark criteria as red (or with an icon as an indicator) when criteria failed
- [x]  Sorting the rows according to the header


Kirdock:
Also fixes #2475 (the part that is possible)
image

cli/internal/cespec/data.go
```diff
@@ -191,7 +192,7 @@ var evaluationFinishedEventData = keptnv2.EvaluationFinishedEventData{
 				Success: true,
 				Message: "a message",
 			},
-			Targets: []*keptnv2.SLITarget{&keptnv2.SLITarget{
+			PassTargets: []*keptnv2.SLITarget{&keptnv2.SLITarget{
```
bacherfl, line 195: please also add an example for the `WarningTargets` property

bacherfl (changes requested):

johannes-b:
A minor change: Can you please align the column "Result" to the right. (to have it the same way like "Value", "pass Criteria", "warning Criteria", and "Score")  

image

johannes-b:
- [ ] The Score column needs to be fixed. Please see my example below, where I have two metrics: response_time (weight = 1), error_rate (weight=2), response_time resulted in a warning (this means that we get half of the points here). In the end, I would like to see: 
- response_time: Score: 16.5
- error_rate Score: 66

In total, this has to sum up to the 83.

Can you please get this resolved? Other than that, the table is a great feature!

image

ermin-muratovic (dismissed):
lgtm

ermin-muratovic (approved):
lgtm

PR #3152, opened 2021-02-03
Slow down dependabot

christian-kreuzberger-dtx:

Dependabot is creating roughly 100 PRs as we have some dependencies that we could update... This PR tries to lower the amount of PRs and disables rebase strategy for now (to lower the strain on our CI).


Kirdock (approved):
LGTM

Issue #3150, opened 2021-02-03
Format the Go imports


cpanato:
**Is your feature request related to a problem? Please describe.**
n/a

**Describe the solution you'd like**
When working on this PR https://github.com/keptn/keptn/pull/3005 notice the go import was not well formated and not following a good practice.

My proposal is to go over all files and format that to organize the imports and make it clear

the possible format can be

```
import (
  <go std libs>
   ...

  <other libs>
   ...

  <project/org libs>
   ...
)
```

for example 

```
import (
	"errors"
	"fmt"

	"github.com/google/uuid"

	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
	"github.com/keptn/keptn/shipyard-controller/common"
	"github.com/keptn/keptn/shipyard-controller/operations"
)
```

from https://github.com/keptn/keptn/blob/master/shipyard-controller/handler/servicemanager.go

I also volunteer myself to work on this

/cc @christian-kreuzberger-dtx 

**Describe alternatives you've considered**
n/a
**Additional context**
n/a

christian-kreuzberger-dtx:
This should really be automated, maybe via our GoLang IDE.

The main problem I see is that we will get 100s of merge conflicts. I think we could create PRs for each sub-directory, to avoid having huge merge conflicts.

cpanato:
yes, agree!
I will start the work :)
/assign

christian-kreuzberger-dtx:
For now, please avoid the following directories

* helm-service
* cli
* api-service
* shipyard-controller

christian-kreuzberger-dtx:
also
* configuration-service
* remediation-service

cpanato:
TODO:
- [ ] api
- [ ] distributor
- [ ] gatekeeper-service
- [ ] jmeter-service
- [ ] lighthouse-service
- [ ] mongodb-datastore
- [ ] platform-support
- [ ] statistics-service

Out of scope for now:
- [ ] helm-service
- [ ] cli
- [ ] api-service
- [ ] shipyard-controller
- [ ] configuration-service
- [ ] remediation-service

cpanato:
the defined work is done so will close this Issue 🎉

christian-kreuzberger-dtx:
thanks a lot!

PR #3157, opened 2021-02-03
api: format go import

cpanato:
Part of https://github.com/keptn/keptn/issues/3150

/cc @christian-kreuzberger-dtx


christian-kreuzberger-dtx (approved):
LGTM

api/restapi/configure_keptn.go
```diff
@@ -5,27 +5,24 @@ package restapi
...
 	"net/http"
 	"os"
 	"strings"
 
-	keptnapi "github.com/keptn/go-utils/pkg/api/utils"
-
-	"github.com/keptn/keptn/api/handlers"
-
 	"github.com/go-openapi/errors"
+	openapierrors "github.com/go-openapi/errors"
```
cpanato, line 15: not sure about this file, the header says not to be overwritten, but also this is not generated.
I can drop this is needed

christian-kreuzberger-dtx: I think it's generated, but whatever :D

PR #3159, opened 2021-02-04
#2877 cli: better keptn auth messages

### LINKED ISSUE #2877: Extend keptn auth with more doc output to explain how to obtain token & endpoint

DarthBenro008:
- Fixes #2877 

### Preview

keptn_auth_endpoint_message

### DIFF
--- a/cli/cmd/auth.go
+++ b/cli/cmd/auth.go
@@ -148,6 +148,13 @@ func init() {
 }
 
 func verifyAuthParams(authParams *authCmdParams) error {
+
+	const parametersRequiredMessage = "keptn auth requires api-token and endpoint \n\n" +
+		"For more information on how to obtain token and endpoint to go https://keptn.sh/docs/%s/reference/cli/#authenticate-keptn-cli \n\n" +
+		"Alternatively,To quickly access Keptn, you can use a port-forward and then authenticate your Keptn CLI: \n" +
+		"- kubectl -n keptn port-forward service/api-gateway-nginx 8080:80 \n" +
+		"- keptn auth --endpoint=http://localhost:8080/api --api-token=$(kubectl get secret keptn-api-token -n keptn -ojsonpath={.data.keptn-api-token} | base64 --decode)"
+
 	var err error
 	if *authParams.exportConfig {
 		return nil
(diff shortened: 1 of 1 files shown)


cli/cmd/auth.go
```diff
@@ -148,6 +148,13 @@ func init() {
 }
 
 func verifyAuthParams(authParams *authCmdParams) error {
+
+	const parametersRequiredMessage = "keptn auth requires api-token and endpoint \n\n" +
+		"For more information on how to obtain token and endpoint to go https://keptn.sh/docs/%s/reference/cli/#authenticate-keptn-cli \n\n" +
+		"Alternatively,To quickly access Keptn, you can use a port-forward and then authenticate your Keptn CLI: \n" +
```
christian-kreuzberger-dtx, line 154: I think there is a space missing between *Alternatively,* and *To*. Also, *To* should be lowercased *to*

cli/cmd/auth.go
```diff
@@ -148,6 +148,13 @@ func init() {
...
+		"- kubectl -n keptn port-forward service/api-gateway-nginx 8080:80 \n" +
```
christian-kreuzberger-dtx, line 155: Is it possible to make the namespace a variable here?

cli/cmd/auth.go
```diff
@@ -229,7 +249,7 @@ func smartKeptnCLIAuth() (string, error) {
 	}
 
 	if len(keptnInstallations) > 1 {
-		fmt.Println("We have found multiple Keptn Installation, Please select the one from the list below to continue")
+		fmt.Println("We have found multiple Keptn Installations, Please select the one from the list below to continue")
```
christian-kreuzberger-dtx, line 252: please change to
> There are multiple Keptn installations on your Kubernetes cluster, please select the correct one from the list below to continue

christian-kreuzberger-dtx (changes requested):
Thanks for the PR. Please adress my comments :)

cli/cmd/auth.go
DarthBenro008: By making the namespace variable, my understanding it that we want fetch and show the namespace that keptn is installed in (passed in args of `keptn install`)?
If that is the case, we will have to write a separate function that detects the installation namespace.

But in my opinion, whosoever follows the tutorial, they will probably have the default installation namespace as `keptn`.

Do let me know your thoughts!

cli/cmd/auth.go
DarthBenro008: Fixed in 42c037f666185b728f27f042d7e3faca704c5ec5 !

cli/cmd/auth.go
christian-kreuzberger-dtx: I agree that most people will follow the tutorial where you install it in the `keptn` namespace, nevertheless, I think the variable `namespace` is available (see line `namespace, err = smartKeptnCLIAuth()`), so we could make use of that (and in case it is empty or an has occured when fetching it, it could default to `keptn`).

So my proposal would be:
```
namespace := "keptn"
namespace, err = smartKeptnCLIAuth()
...
return fmt.Errorf(err.Error()+parametersRequiredMessage, keptnReleaseDocsURL, namespace)
```

DarthBenro008:
Hey @christian-kreuzberger-dtx ,

I have tested and added variable namespacing the message. Apparently we the fallback to default keptn namespace is already handled and we do not need to define it again (tested)

Here is the message when keptn is installed in a custom namespace called `testnamespace`:

namespace_variable

christian-kreuzberger-dtx (approved):
LGTM

PR #3165, opened 2021-02-04
jmeter-service: format go import

cpanato:
Part of https://github.com/keptn/keptn/issues/3150

/cc @christian-kreuzberger-dtx


christian-kreuzberger-dtx (approved):
LGTM

PR #3169, opened 2021-02-04
#2628 Sequence tile refinement

ermin-muratovic:
image


Kirdock (changes requested):
If a sequence does not have a stage, image or service name, it is incorrectly displayed
image

Kirdock (approved):
LGTM

Kirdock (approved):
LGTM

PR #3171, opened 2021-02-05
Remove payload of received response from distributor logs

bacherfl:
The log output removed in this PR was only used for debugging and was not intended to stay here


christian-kreuzberger-dtx (approved):
LGTM

PR #3197, opened 2021-02-08
Added a workflow to automatically create a release branch

christian-kreuzberger-dtx:

This PR adds a workflow to automate the creation of a release branch (and copying of release notes).
In addition, the workflow checks if the release name adheres to semver :) 

FYI: This workflow is not very specific to keptn/keptn itself, it can be copied to our keptn-contrib/sandbox services aswell. Might be worth to extract some functionality as another gh-action.


bacherfl (approved):
lgtm

Issue #2990, opened 2021-01-26
Keptn CLI: `keptn onboard service` becomes `keptn create service + add resources`


johannes-b:
With the change that the helm-services will no longer upload the Helm charts of a service, this functionality must be moved to the CLI. 

Since the `keptn onboard service` command is opinionated in the way this command limits a Keptn service to be deployed with a Helm chart, this command will become deprecated and replaced by the simpler `keptn create service` followed by `keptn add-resource`.

`keptn onboard service carts --project=sockshop --chart=./carts`

```
Output:

This command is deprecated and is replaced by: 

1.) keptn create service carts --project=sockshop
2.) keptn add-resource --project=sockshop --stage=all --service=carts  --resource=./carts --resourceUri=helm/carts

Please execute these commands sequentially. 
```

**Tasks:**
* Change the `keptn onboard service` command to print the sub-commands: `keptn create service + keptn add-resource` and a note showing that this command has been deprecated
* (!) Provide the CLI commands that do the onboarding (create + add-resource) in a way that copy-paste is possible
* (!) When executing `keptn onboard` return exit-code != 0
* Mark the command as depreciated - to see it on: https://keptn.sh/docs/0.8.x/reference/cli/commands/keptn_onboard_service/
* Update of integration tests: change `keptn onboard` to `keptn create svc` + `keptn add-resource`
* Update of documentation: https://keptn.sh/docs/0.8.x/manage/service/#onboard-a-service

**Note:**
- To ensure feature parity with Keptn 0.7.3, issue #2989 must be implemented. --> Consider merging the change of this issue with #2989. 

**TBD:**
To upload the helm chart, I see two options: 
1. `keptn add-resource` supports `--stage=all` which adds the resource to all stages
2. `keptn add-resource --project=sockshop --service=carts  --resource=./carts --resourceUri=helm/carts` - without `stage`

**Definition of Done:**
- [ ] `keptn onboard` has been replaced
- [ ] Documentation (https://keptn.sh/docs/0.8.x/manage/service/#onboard-a-service) is up-to-date

christian-kreuzberger-dtx:
I think we should not call it `--stage=all`, as this means we can never create a stage called `all` (which is not obvious to the user).

Not having the argument in `--stage` provided seems to be suboptimal, because (IMO) this should push resources to the master/main branch.

How about new argument called `--allStages` or `--all-stages` (similar to `kubectl get pods --all-namespaces`).

johannes-b:
[...]

Issue #3001, opened 2021-01-26
Hide implementation details in the API


agrimmer:
The goal of this issue is to hide implementation details (e.g. the name of the services providing the API) in the API.

**Tasks:**
- Move endpoints of shipyard-controller into path `/api/controlPlane/v1`
- Move endpoints of `/api/configuration-service/v1` into path `/api/resources/v1`
- Move endpoints of `/api/mongodb-datastore` into path `/api/eventStore/v1`
- Remove NGINX backward-compatible rules

johannes-b:
Just one remark, should we call the `api/eventStore/v1` endpoint just `api/events/v1` to have it aligned with  `/api/resources/v1` ?

johannes-b:
We agreed on: 
~`/api/controlplane/v1`~ --> `/api/v1/` 
`/api/events/v1`
`/api/resources/v1`

johannes-b:
Task: 
- [ ] Rename endpoint of `shipyard-controller`  to `/api/controlPlane/v1`

--> We agreed on non-touching the other endpoints.

PR #3201, opened 2021-02-08
#2990 deprecate onboard service, add --all-stages to add-resource

christian-kreuzberger-dtx:

As discussed in #2990, we will leave keptn onboard service as it is (but we will add a deprecation notice to it).

In addition, I implemented the following tasks:
* Add flag --all-stages to `add-resource` (+ unit tests)
* Improved unit-tests for add-resource with a setup and teardown function


cli/cmd/add_resource.go
```diff
@@ -86,20 +88,66 @@ keptn add-resource --project=rockshop --stage=production --service=shop --resour
...
 
 		if !mocking {
-			_, errorObj := resourceHandler.CreateResources(*addResourceCmdParams.Project, *addResourceCmdParams.Stage, *addResourceCmdParams.Service, resources)
-			if errorObj != nil {
-				return errors.New("Resource " + *addResourceCmdParams.Resource + " could not be uploaded: " + *errorObj.Message)
+			if addResourceCmdParams.AllStages != nil && *addResourceCmdParams.AllStages {
+				// Upload to all stages
+
+				// get stages
+				stagesHandler := apiutils.NewAuthenticatedStageHandler(endPoint.String(), apiToken, "x-token", nil, endPoint.Scheme)
```
bacherfl, line 128: please append `/shipyard-controller` to the endpoint, i.e.:

```golang
stagesHandler := apiutils.NewAuthenticatedStageHandler(endPoint.String()+"/shipyard-controller", apiToken, "x-token", nil, endPoint.Scheme)
```

bacherfl (changes requested):

cli/cmd/add_resource.go
christian-kreuzberger-dtx: I've changed it, but I believe we should make sure this is reflected in go-utils rather than in the code that uses it.

bacherfl: thanks, but I agree with your point - this will be done in https://github.com/keptn/keptn/issues/3001

bacherfl (approved):

christian-kreuzberger-dtx:
When compiling locally and trying it out with the latest master branch, it fails
``
./cli add-resource --project=sockshop --service=carts --all-stages --resource=go.mod
```

Error: Failed to retrieve stages for project : path /shipyard-controller/configuration-service/v1/project/stage was not found

**Review by bacherfl — CHANGES_REQUESTED** [review]
as discussed reverting the approval because we found some build errors

File: cli/cmd/add_resource.go [hunk 2]
```diff
@@ -77,29 +77,70 @@ keptn add-resource --project=rockshop --stage=production --service=shop --resour
...
+		}
+
+		// When setting --all-stages, project and service needs to be set
+		if (addResourceCmdParams.AllStages != nil && *addResourceCmdParams.AllStages) && (
+			(addResourceCmdParams.Service == nil || *addResourceCmdParams.Service == "") || (addResourceCmdParams.Project == nil || *addResourceCmdParams.Project == "")) {
+			return errors.New("--service and --project need to be supplied when using --all-stages")
+		}
+
+		// Handle different cases of adding resource to a projects default branch, stage branch, and/or service sub-directory
+		if (addResourceCmdParams.Service != nil && *addResourceCmdParams.Service != "") && (addResourceCmdParams.AllStages != nil && *addResourceCmdParams.AllStages) {
```
warber, line 101: provide helper functions which help checking the conditions... something like:
```
func assertFlagsSet(el... interface{}) bool {
	for e := range el {
		if !assertFlagSet(e) {
			return false
		}
	}
	return true
}


func assertFlagSet(s interface{}) bool {
	return &s != nil && s != ""
}
...
```

usage:

```
if (!assertFlagsSet(addResourceCmdParams.Stage,addResourceCmdParams.AllStages)) {
return errors.New("Cannot use --stage and --all-stages at the same time")
}
```

cli/cmd/add_resource.go
```diff
@@ -77,29 +77,57 @@ keptn add-resource --project=rockshop --stage=production --service=shop --resour
...
 			return fmt.Errorf("Resource %s could not be uploaded: %s"+endPointErrorReasons,
 				*addResourceCmdParams.Resource, endPointErr)
 		}
 
-		if (addResourceCmdParams.Service != nil && *addResourceCmdParams.Service != "") && (addResourceCmdParams.Stage != nil && *addResourceCmdParams.Stage != "") {
+		// Handle different cases of adding resource to a projects default branch, stage branch, and/or service sub-directory
+		if isStringFlagSet(addResourceCmdParams.Service) && isBoolFlagSet(addResourceCmdParams.AllStages) {
+			// add to all stages
+			logging.PrintLog("Adding resource "+*addResourceCmdParams.Resource+" to all stages of project "+*addResourceCmdParams.Project, logging.InfoLevel)
+		} else if isStringFlagSet(addResourceCmdParams.Service) && isStringFlagSet(addResourceCmdParams.Stage) {
```
warber, line 91: you can use `areStringFlagSet()`here

cli/cmd/add_resource.go
```diff
@@ -77,29 +77,57 @@ keptn add-resource --project=rockshop --stage=production --service=shop --resour
...
+			// add to service and stage
 			logging.PrintLog("Adding resource "+*addResourceCmdParams.Resource+" to service "+*addResourceCmdParams.Service+" in stage "+*addResourceCmdParams.Stage+" in project "+*addResourceCmdParams.Project, logging.InfoLevel)
-		} else if (addResourceCmdParams.Service == nil || *addResourceCmdParams.Service == "") && (addResourceCmdParams.Stage != nil && *addResourceCmdParams.Stage != "") {
+		} else if !isStringFlagSet(addResourceCmdParams.Service) && isStringFlagSet(addResourceCmdParams.Stage) {
+			// service is empty, add to stage
 			logging.PrintLog("Adding resource "+*addResourceCmdParams.Resource+" to stage "+*addResourceCmdParams.Stage+" in project "+*addResourceCmdParams.Project, logging.InfoLevel)
-		} else if (addResourceCmdParams.Service == nil || *addResourceCmdParams.Service == "") && (addResourceCmdParams.Stage == nil || *addResourceCmdParams.Stage == "") {
+		} else if !isStringFlagSet(addResourceCmdParams.Service) && !isStringFlagSet(addResourceCmdParams.Stage) {
```
warber, line 97: you can use `!areStringFlagSet()` here

cli/cmd/add_resource.go
bacherfl: good catch ;) i have adapted this line

cli/cmd/add_resource.go
bacherfl: this would alter the logic of this check - currently, the condition is fulfilled only if both flags are not set - `!areStringFlagsSet()` would return `true` if any of those flags is not set

warber (approved):

bacherfl (approved):

PR #3202, opened 2021-02-08
Fixed codecoverage for distributor

christian-kreuzberger-dtx:

This PR fixes the wrongly displayed partial codecoverage for distributor in PRs:
image

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -40,7 +40,7 @@ flags:
     carryforward: true # see https://docs.codecov.io/docs/carryforward-flags
   configuration-service:
     carryforward: true # see https://docs.codecov.io/docs/carryforward-flags
-  distributor:
+  distributor-service:
     carryforward: true # see https://docs.codecov.io/docs/carryforward-flags
   gatekeeper-service:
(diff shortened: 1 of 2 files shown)


codecov.yml
```diff
@@ -40,7 +40,7 @@ flags:
     carryforward: true # see https://docs.codecov.io/docs/carryforward-flags
   configuration-service:
     carryforward: true # see https://docs.codecov.io/docs/carryforward-flags
-  distributor:
+  distributor-service:
```
bacherfl, line 43: Should we instead rename the `distributor-service` to `distributor` [here](https://github.com/keptn/keptn/blob/31d73ea209b2211d50f5500aa4bce5d2ae4388af/.github/workflows/CI.yml#L338)? In all other places, e.g. the K8s deployments, it's referred to as `distributor`, so naming it differently here might cause confusion

christian-kreuzberger-dtx: good point, I'll also check for other occurances of distributor "service"

bacherfl (approved):
lgtm

PR #3232, opened 2021-02-10
#1924 implement getting resource list for a service

### LINKED ISSUE #1924: configuration-service: Implement GET ​/project​/{projectName}​/stage​/{stageName}​/service​/{serviceName}​/resource
**Problem statement:**
We have received extension requests for, e.g., jmeter-service, where a user wants the jmeter-service to have access to additional uploaded resources such as data files. The approach would be to let users upload resources to a particular folder, e.g: `/jmeter`, and then have the jmeter-service download ALL resources in its subfolder before executing the test. The problem right now is that there is no way for a Keptn-service to say: _Give me all uploaded resources_. That's why we would need a new API call that can list resources in a particular project/stage/service and optionally specify a resourceURI pattern, e.g: /jmeter

**Proposed feature:**
I would like to propose a call similar to this: `GET /project/{projectName}/stage/{stageName}/resource/{resourceURI}?retrieveList=true`

This would extend the existing endpoint and returns those resources matching the resourceURI. This call should then also work for the other GET resource calls such as the one on project and stage level

Alternatively, we would need to specify another file, e.g: jmeter.conf where we list all data files that should be downloaded. While this is doable it means additional configuration efforts

daemon1024:
Closes #1924

### DIFF
--- a/configuration-service/handlers/service_resource.go
+++ b/configuration-service/handlers/service_resource.go
@@ -24,7 +24,28 @@ import (
 // GetProjectProjectNameStageStageNameServiceServiceNameResourceHandlerFunc get list of resources for the service
 func GetProjectProjectNameStageStageNameServiceServiceNameResourceHandlerFunc(
 	params service_resource.GetProjectProjectNameStageStageNameServiceServiceNameResourceParams) middleware.Responder {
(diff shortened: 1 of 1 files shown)


bacherfl (approved):
lgtm

PR #3233, opened 2021-02-10
#2764 Fix inferring urls

### LINKED ISSUE #2764: helm-service: Support for `deployment_strategy: user_managed`
The helm-service currently supports two deployment strategies:
- `direct`: Here the chart of the user is installed and additionally a VirtualService as well as a DestinationRule are generated in the `generated-chart` and also installed.
- `blue_green_service`: Here the chart of the user is installed as `canary`-version and additionally a second Deployment, Service, VirtualService, DestinationRules are generated in the `generated-chart` and also installed.

In order to disable the automatic resource generation (i.e. the generation of the `generated-chart`) as well as to overcome the dependency of having Istio installed on the cluster, this issue proposes to introduce a new deployment strategy named `user_managed`. 
In case the user chooses `user_managed` as deployment strategy, no `generated-chart` should be generated and installed.

Additionally, when the user selects `user_managed` as deployment strategy, there should not be any requirements for the used chart (i.e. the Helm chart had to consist of exactly 1 Deployment and 1 Service). 
The `helm-service` does not need this requirement for deployment strategy `user_managed` because it does not need to generate a `generated-chart`.

**Supported use cases:**
- The `helm-service` allows Keptn services that contain multiple Deployments and Services within one Helm-chart. The Helm-chart can even contain e.g. an Argo Rollout which can be used for B/G deployments or Canary releases.

**Restrictions:**
- In case the chart contains multiple Deployments or Services, the deployment strategies `direct` and `blue_green_service` are not supported.
- As the `helm-service` does not generate any Istio configuration, it cannot infer `deploymentURIsLocal` and `deploymentURIsPublic` which are used for e.g. testing. Those URLs can however be set by the user in the domain event.

**Tasks:**
- `onboard service` command in the CLI: Relax the check for the Helm-chart and only print a warning in case the onboarded chart cannot be used for deployment strategy `direct` or `blue_green_service`. Here we cannot decide which deployment strategy will be used.
- `helm-service` file `onboarder.go`: Remove the check whether the k8s service name matches the Keptn service name. In the generated VirtualServices, we use the name of the provided K8s service anyway.
- `helm-service` file `deployment_handler.go`:  Introduce the new deployment strategy `user_managed`
- `helm-service`: Introduce a check for Helm-chart requirements when `deployment_strategy` is equals to `direct` or `blue_green_service`. Therefore, move the `helm_validator.go` functionalities into the `kubernetes-utils` so that it can be used by the CLI as well as the `helm-service`.
- Update and extend docs

agrimmer:
Closes #2764 
This PR fixes the inferring of the URL as follows: In case a user uses deployment strategy `direct` or `blue_green_service`, the `DeploymentURIsLocal` and `DeploymentURIsPublic` are always inferred regardless they are already present in the incoming event.

### DIFF
--- a/helm-service/controller/deployment_handler.go
+++ b/helm-service/controller/deployment_handler.go
@@ -3,6 +3,8 @@ package controller
 import (
 	"errors"
 	"fmt"
+	"math"
+
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	keptnevents "github.com/keptn/go-utils/pkg/lib"
 	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
@@ -11,7 +13,6 @@ import (
 	"github.com/keptn/keptn/helm-service/pkg/mesh"
 	"helm.sh/helm/v3/pkg/chart"
 	corev1 "k8s.io/api/core/v1"
-	"math"
 )
 
 // DeploymentHandler is a handler for doing the deployment and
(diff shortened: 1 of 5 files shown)


warber (commented):
you added logic to the [GenerateDuplicateChart](https://github.com/keptn/keptn/blob/cdafdb49bbd344c3c0c23464a7e51fed1ec1251d/helm-service/pkg/helm/generated_chart_handler.go#L46) function which is not covered by unit tests. (see codecov complain). Well, this function currently does not have a unit test at all but it is easy to test. Can you add test(s) for it?

agrimmer:
@warber Thanks for the hint! I added respective tests.

warber (approved):

PR #3235, opened 2021-02-10
Remove duplicate / in URL

agrimmer:
GetStageResource was called with a URL containing two slashes, e.g. `/project/{projectName}/stage/{stageName}/resource//{resourceURI}`


warber (commented):
Approved, but we should definitely aim to do a propper refactoring and increase test coverage for this service!

johannes-b (approved):
lgtm

PR #3265, opened 2021-02-17
#3043 make sure the message property of a .triggered event is not propagated

### LINKED ISSUE #3043: Shipyard controller: remove 'data.message' property from previous .finished event before sending next .triggered event
Currently, the shipyard controller passes the `data.message` property of the previous .finished event on to the subsequent .triggered event. Since the service executing the subsequent task does not need to know about this property, we should remove it before sending the .triggered event (https://github.com/keptn/keptn/blob/35e0e2c989e3359a8f9abd99f1389d3a03a0ffeb/shipyard-controller/handler/shipyardcontroller.go#L621).

bacherfl:
Closes #3043


johannes-b (approved):
lgtm

PR #3268, opened 2021-02-17
updated cli upgrade_project command

warber:

### DIFF
--- a/cli/cmd/upgrade_project_test.go
+++ b/cli/cmd/upgrade_project_test.go
@@ -152,24 +152,6 @@ func Test_transformShipyard(t *testing.T) {
 									},
 								},
 								{
-									Name: "rollback",
-									TriggeredOn: []keptnv2.Trigger{
-										{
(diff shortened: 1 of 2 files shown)


johannes-b (approved):
lgtm

cli/cmd/upgrade_project_test.go
```diff
@@ -155,7 +155,7 @@ func Test_transformShipyard(t *testing.T) {
 									Name: "rollback",
 									TriggeredOn: []keptnv2.Trigger{
```
bacherfl, line 156: I think we need to check wether the `deployment` task in the `delivery` sequence uses a `direct` or `blue_green_service` deployment strategy, and only add the `rollback` sequence if the `blue_green_service` strategy is being used

warber: well, this absolutely makes sense, sir :)

bacherfl (approved):

PR #3318, opened 2021-02-22
Smart Helm values merger

agrimmer:


christian-kreuzberger-dtx:
I think this PR shows how the merger should work.

Now that we have this feature, is there a chance we can get rid of the [replica_count_manipulator.go](https://github.com/keptn/keptn/blob/master/helm-service/pkg/configurationchanger/replica_count_manipulator.go)?

cli/internal/textutils.go
```diff
@@ -0,0 +1,77 @@
...
+	"fmt"
+	"strings"
+)
+
+type node interface {
+	String() string
+}
+
+func newInnerNode() innerNode {
+	return innerNode{childs: make(map[string]node)}
```
jsoref, line 13: Generally you want `children`

jsoref (changes requested):
Please add the other words the bot suggested into `expect.txt`, if you're on Linux, it has a command you can run to do it for you.

warber:
closes #3341

cli/internal/textutils.go
```diff
@@ -0,0 +1,90 @@
...
+func newInnerNode() *innerNode {
+	return &innerNode{
+		sortedKeys: []string{},
+		childs:     make(map[string]node),
+	}
+}
+
+type innerNode struct {
+	sortedKeys []string
+	childs     map[string]node
```
bacherfl, line 21: can we rename this to `children`?

bacherfl (approved):

Issue #3325, opened 2021-02-22
Integration Tests are failing (Minishift, self healing)


christian-kreuzberger-dtx:
Note: Until #3166 is implemented, I'll post this manually!

Integration tests have failed.

* Date Time: 2021-02-22 04:00
* Link to run: https://github.com/keptn/keptn/actions/runs/587804865
* Branch: master
* Commit:   af49f6d  
* Last committer: @Kirdock 

| Target Platform        |  Keptn Install    |  Keptn Auth  | Linking Stages   |   Quality Gates Standalone  |  Self Healing  |  Delivery Assistant  | Continuous Delivery |
| ---------------------- | ----------------- | ------------ | ---------------- | --------------------------- | -------------- | -------------------- | ------------------- |
| kubernetes/GKE 1.15 | success | success | success | success | success | success | success |
| kubernetes/GKE 1.17 | success | success | success | success | success | success | success |
| kubernetes/k3s-on-GHA v1.16.15+k3s1 | success | success | success | success | success | skipped | skipped |
| kubernetes/k3s-on-GHA v1.19.5+k3s1 | success | success | success | success | success | skipped | skipped |
| openshift/minishift-on-GHA 3.11 | success | success | success | skipped | **failure** | skipped | skipped |

Seems to be an error in openshift/minishift self-healing test: https://github.com/keptn/keptn/runs/1948759970?check_suite_focus=true#step:40:31

```
Error: [2021-02-22 03:13:41] ERROR: Checking .source, expected 'remediation-service', got 'null' ❌
Error: Process completed with exit code 1.
```

PR #3327, opened 2021-02-22
#3325 added logs of remediation service to self healing integration test

bacherfl:

Better log output in case of failing tests for self healing use-case


christian-kreuzberger-dtx (approved):
LGTM

PR #3336, opened 2021-02-22
Show evaluation score for approval-item

ermin-muratovic:
The score was not displayed, because the evaluation event was not found.
Now it should load the evaluation for approval-item correctly, taking the `.finished` event from the evaluation task:

image


Kirdock (approved):
LGTM

PR #3343, opened 2021-02-22
#3335 update shipyard content of project when receiving event

### LINKED ISSUE #3335: shipyard-controller api returns shipyard version 0.1.7, although 0.2.0 is used
Keptn-Version: 0.8.0-dev.202102191239 (last nightly)

For no apparent reason, my sockshop project is listed as "0.1.7" spec project in Keptn Bridge:
image

This information is returned by the API call

GET /api/controlPlane/v1/project?disableUpstreamSync=true&pageSize=50

Shipyard.yaml looks as follows:
```yaml
apiVersion: "spec.keptn.sh/0.2.0"
kind: "Shipyard"
metadata:
  name: "shipyard-sockshop"
spec:
  stages:
...
```

This would be the relevant code:
https://github.com/keptn/keptn/blob/1a0e88b4fb48e9ee076d9acce0b01cecd7aea774/shipyard-controller/db/projects_materialized_view.go#L84-L101

bacherfl:
Closes #3335 

This PR makes sure to regularly update the shipyard content of a project in the materialized view (whenever an event for that project is received). This makes sure to achieve eventual consistency, between the information stored in the db and the actual state of the shipyard in the git repo.

### DIFF
--- a/shipyard-controller/handler/shipyardcontroller.go
+++ b/shipyard-controller/handler/shipyardcontroller.go
@@ -103,22 +103,6 @@ func (sc *shipyardController) HandleIncomingEvent(event models.Event) error {
 				event.Triggeredid,
 			); err != nil {
 				sc.logger.Error(fmt.Sprintf("could not update event for project %s: %s", eventData.Project, err.Error()))
-				return
-			}
-
-			// also update the shipyard in the materialized view because it might have been changed in the meantime
-			shipyard, err := common.GetShipyard(eventData)
-			if err != nil {
-				sc.logger.Error(fmt.Sprintf("could not retrieve shipyard file of project %s: %s", eventData.Project, err.Error()))
-				return
-			}
-			shipyardContent, err := yaml.Marshal(shipyard)
-			if err != nil {
-				sc.logger.Error(fmt.Sprintf("could not encode shipyard file of project %s: %s", eventData.Project, err.Error()))
-				return
-			}
-			if err := sc.eventsDbOperations.UpdateShipyard(eventData.Project, string(shipyardContent)); err != nil {
-				sc.logger.Error(fmt.Sprintf("could not update shipyard content of project %s: %s", eventData.Project, err.Error()))
 			}
 		}()
 	}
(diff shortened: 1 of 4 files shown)


shipyard-controller/handler/shipyardcontroller.go
```diff
@@ -102,6 +103,22 @@ func (sc *shipyardController) HandleIncomingEvent(event models.Event) error {
 				event.Triggeredid,
 			); err != nil {
 				sc.logger.Error(fmt.Sprintf("could not update event for project %s: %s", eventData.Project, err.Error()))
+				return
+			}
```
warber, line 107: any chance to get the codecov warning(s) fixed ?

shipyard-controller/handler/shipyardcontroller.go
```diff
@@ -102,6 +103,22 @@ func (sc *shipyardController) HandleIncomingEvent(event models.Event) error {
...
+				sc.logger.Error(fmt.Sprintf("could not retrieve shipyard file of project %s: %s", eventData.Project, err.Error()))
+				return
+			}
+			shipyardContent, err := yaml.Marshal(shipyard)
+			if err != nil {
+				sc.logger.Error(fmt.Sprintf("could not encode shipyard file of project %s: %s", eventData.Project, err.Error()))
+				return
+			}
+			if err := sc.eventsDbOperations.UpdateShipyard(eventData.Project, string(shipyardContent)); err != nil {
+				sc.logger.Error(fmt.Sprintf("could not update shipyard content of project %s: %s", eventData.Project, err.Error()))
```
christian-kreuzberger-dtx, line 121: I think having this in `HandleIncomingEvent` is suboptimal, as we don't need to retrieve the shipyard for every event we process (in fact, we might not even want that in the future).

Looking at shipyardcontroller.go, I think a better place would be to have it in `handleTriggeredEvent`, where we already retrieve the shipyard.

christian-kreuzberger-dtx (changes requested):

shipyard-controller/handler/shipyardcontroller.go
bacherfl: agreed, have moved it to `handleTriggeredEvent`

christian-kreuzberger-dtx (approved):
LGTM

PR #3367, opened 2021-02-24
#3332 Rename send event to trigger commands

### LINKED ISSUE #3332: Bridge: Replace occurances of old "send event" with the new "trigger" functionality
In Keptn Bridge (last nightly) we still refer to `keptn send event`, e.g.:

image

image

These should be replaced with the latest `keptn trigger evaluation` and `keptn trigger delivery` commands.

ermin-muratovic:
Fixes #3332


Kirdock (approved):
LGTM

PR #3372, opened 2021-02-24
fix self healing integration test

warber:
use `.data.status` instead of `.data.action.status` in jq check


bacherfl (approved):
lgtm

Issue #3170, opened 2021-02-04
Incorporate shellcheck into CI pipeline


unacceptable:
Overview
---
[shellcheck](https://github.com/koalaman/shellcheck) is a command-line tool to ensure the quality of shell scripts.

Why this is important
---
It is extremely easy to make mistakes in bash, and extremely hard to write unit tests for shell scripts.  shellcheck is a stop-gap to help ensure that common issues are avoided in shell scripts.

christian-kreuzberger-dtx:
Thanks for the suggestion.

If we include it, I'd like to do it as a GH action, e.g., https://github.com/marketplace/actions/shellcheck

Also, are you willing to take this task?

unacceptable:
@christian-kreuzberger-dtx,

Yeah, I can incorporate it as a GitHub action, but I am not a fan of that one in the marketplace.  I would probably do something like this: 

```
name: shellcheck

on: [push]

jobs:
  lint_shellscripts:
    runs-on: [ubuntu-latest]
    container:
      image: koalaman/shellcheck-alpine:latest
      options: --entrypoint /bin/sh
    steps:
      - uses: actions/checkout@v1
      - run: |
            #!/usr/bin/env sh
            set -ex
            shellcheck --version
            shellcheck *.sh
```

I haven't done too much with GitHub actions since I mostly use GitLab, but I am sure I can figure it out (and maybe simplify the above example).

Do you all use `latest` for your GitHub action container tags or do you all want to pin versions?  I have no strong feelings about either way.

Respectfully,
Robert J.

christian-kreuzberger-dtx:
We try to use official GH actions from marketplace as much as possible. If there is a security vulnerability in it, then we'll get notified. We can't just run an external docker image.
It doesn't have to be https://github.com/marketplace/actions/shellcheck, can be any other one as long as it's used by others and actively maintained.

Alternatively, we'd like to have full control of what is happening, and install the tools ourselves, e.g.:
https://github.com/keptn/keptn/blob/master/.github/workflows/deprecated-version-check.yml#L13-L15

unacceptable:
@christian-kreuzberger-dtx,

Sorry, I missed your message before.  I threw a comment on the thread in the PR [here](https://github.com/keptn/keptn/pull/3394#discussion_r583339820).

Please let me know what you think.

Respectfully,
Robert J.

unacceptable:
@christian-kreuzberger-dtx I just use the Marketplace action as you wanted.

PR #3394, opened 2021-02-25
#3170 - Implementing shellcheck for scripts

unacceptable:


unacceptable:
It is my understanding that this would not be able to go out until after 0.8.0.

Also, there are a ton of changes here so please be super strict in reviewing this PR.  I would hate to be the reason that 0.8.x becomes a bad release.

Thanks for being awesome!

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,29 @@
+name: shellcheck
+
+on: [push, pull_request]
+
+jobs:
+  lint_shellscripts:
+    runs-on: [ubuntu-latest]
+    container:
+      image: koalaman/shellcheck-alpine:v0.7.1
+      options: --entrypoint /bin/sh
```
christian-kreuzberger-dtx, line 10: Already asked this on the issue #3170 - what's the problem with using the actual GH action?

unacceptable: Hey sorry, I missed that comment in the ticket.  I can try to get that GitHub action working if that's a deal-breaker for you all. In previous versions of that marketplace action, I wasn't able to get it to run on another, which is why I chose to use the maintainer of shellcheck's (Koalaman's) container.

In my mind using Koalaman's docker image is a more elegant solution than the [logic outlined](https://github.com/ludeeus/action-shellcheck/blob/master/action.yaml#L59-L66) it the GitHub action (which downloads and installs Koalaman's binaries from GitHub).

As for the security of the image, it is [FROM scratch](https://github.com/koalaman/shellcheck/blob/master/Dockerfile.multi-arch#L22-L26) which would arguable make it much more secure than that Ludeeus' GitHub Action.

At the end of the day, I guess this container preference just arose from building CI steps with containers from running Jenkins on K8s with dynamically rendered pods and GitLab CI.

christian-kreuzberger-dtx: Hi!

I agree that the fact that it is using "FROM SCRATCH" and the Dockerfile is nicely readable makes the Docker Image itself more secure. 
Unfortunately, there is no way of verifying that the Docker Image on DockerHub is actually being build from the Dockerfile provided in the repository. Even if it is possible to verify that **today**, the image might change tomorrow. Or next week. or ...
On DockerHub, you can always delete or overwrite a tag. I'm not saying that Koalaman is going to do that, but it's entirely possible that someone will try to do that (or try to hack an account on DockerHub with that purpose), hence it's better to avoid referring to (non official) images on DockerHub.

FYI, [Ludeeus' shellcheck GitHub Action](https://github.com/ludeeus/action-shellcheck) is a [composite action](https://github.com/ludeeus/action-shellcheck/blob/f0d446b80e91cc0a7e8bec41ada75e2e73fd2218/action.yaml#L40) which does not require a Docker Image to be shipped with it.

In the end, using that action does not make it more secure, but it makes it traceable (plus GH is having an eye on it as it's on marketplace).

Anyway, an alternative way that neither uses the Docker Image on DockerHub nor the GH action is to write the installation/download of the tool in bash, similar to what we did with [pluto and helm](https://github.com/keptn/keptn/blob/master/.github/workflows/deprecated-version-check.yml#L11-L15). 

(Btw, I'm not saying that this is the wrong way of doing it. I used to do a lot of stuff with GitLab CI in the past and we used external Docker Images all the time. Though we had to get permission to use an image, the image was audited and then pulled to our internal registry **once**.)

unacceptable: Alright, I just tested that action from the market place and it seems to be pretty interchangeable.

christian-kreuzberger-dtx:
Many thanks!
The shellcheck action looks fine, thanks for changing this.

We will have to check all the shell scripts for compatibility, which we will do once 0.8.0 is out. Thanks for your patience.

unacceptable:
@johannes-b, @christian-kreuzberger-dtx - I just pulled upstream changed for 0.8.0 and linted the new shell script changes.

christian-kreuzberger-dtx (approved):
LGTM

PR #3404, opened 2021-03-01
#3382 Fix duplicate tasks

### LINKED ISSUE #3382: Bridge: Duplicated Tasks in sequence screen
After triggering a delivery from the CLI I've noticed that in the bridge the **test** task of a sequence appeared twice:

duplicated-test

The ID of both events is identical, so it's clearly a duplicate which might be caused by a flaw in the event polling logic in the bridge. Further, the duplicated task does not turn green/red and keeps showing a rotating icon.

**Workaround:** Pressing **F5** :smile_cat:

ermin-muratovic:
Fixes #3382 

If last event in response is after responseTime, the Bridge considers the `event.time` as fromTime for next GET call.

I also changed the labeling of "Problem events" to "open" and "resolved", instead of "started" and "succeeded".


christian-kreuzberger-dtx:
I tested the image, outcome looks good to me (and it fixes #3405)

Kirdock (approved):
LGTM

Issue #3411, opened 2021-03-01
lighthouse-service needs to send previous payloads (e.g., "deployment") in "get-sli.triggered"


christian-kreuzberger-dtx:
Currently, a `get-sli.triggered` event is sent by lighthouse-service and looks as follows:
```
{
  "data": {
    "get-sli": {
      "end": "2021-03-01T13:54:42Z",
      "indicators": [
        "response_time_p95"
      ],
      "sliProvider": "dynatrace",
      "start": "2021-03-01T13:46:04Z"
    },
    "labels": {
      "DtCreds": "dynatrace"
    },
    "project": "mockshop",
    "service": "carts",
    "stage": "staging"
  },
  "id": "8d20b418-9417-471f-9380-097648489b99",
  "source": "lighthouse-service",
  "specversion": "1.0",
  "time": "2021-03-01T13:54:42.392Z",
  "type": "sh.keptn.event.get-sli.triggered",
  "shkeptncontext": "45b972d0-b522-4fd0-97cb-0722b3fa2d1c"
}
```

Unfortunately, it does not contain `deployment` or anything else that's being provided in `evaluation.triggered`, e.g. `deployment`, `test`,  `configuration-change`. 
This leads to a problem with our SLI providers that rely on that information, e.g., https://github.com/keptn-contrib/dynatrace-sli-service/issues/154

# Definition of Done

- [ ] Lighthouse-service needs to mimick the behaviour of shipyard-controller when sending a get-sli.triggered cloudevent, and include relevant payloads from previous cloudevents

PR #3415, opened 2021-03-02
#3411: introduced GetSLITriggeredEventDataExtended to fix event payload

warber:


lighthouse-service/event_handler/start_evaluation_handler.go
```diff
@@ -168,13 +168,29 @@ func (eh *StartEvaluationHandler) sendInternalGetSLIEvent(shkeptncontext string,
 		},
 	}
 
+	deployment := ""
+	if e.Deployment.DeploymentNames != nil && len(e.Deployment.DeploymentNames) > 0 {
```
christian-kreuzberger-dtx, line 172: Please also add a `e.Deployment != nil`

warber: not necessary, as it is a "non-nillable" type.

christian-kreuzberger-dtx (approved):
LGTM

Issue #3456, opened 2021-03-08
GET services from a stage endpoint requires stage but contains service in path


johannes-b:
**Bug:** Mismatch between path parameter (`service`) and actual parameter (`stage`). 

![image](https://user-images.githubusercontent.com/729071/110286563-6e516b00-7fe5-11eb-90be-0a4b932c30e0.png)

**Fix:**
* The path for the endpoint must be: `/project/{project}/stage/{stage}`

christian-kreuzberger-dtx:
This seems to be a copy-paste error. I've fixed it (locally) and it seems there are *two* endpoints now:
![image](https://user-images.githubusercontent.com/56065213/110301078-f476ad00-7ff7-11eb-8a4a-26cd5bdcfa2a.png)

PR #3460, opened 2021-03-08
Updated swagger docs of shipyard-controller

christian-kreuzberger-dtx:
While working on https://github.com/keptn/keptn/issues/3456 I noticed that shipyard-controllers swagger docs are not updated according to what they should contain. 

This PR manually updates swagger docs.
Also, I've updated README to explain development using skaffold.


bacherfl (approved):
lgtm

PR #3464, opened 2021-03-08
#3214 integration test for remote execution plane

### LINKED ISSUE #3214: Integration Test: Run jmeter and helm in execution plane (independent namespace)
Depends on https://github.com/keptn/keptn/issues/3253

Currently, our integration tests (see https://github.com/keptn/keptn/blob/master/.github/workflows/integration_tests.yml) run the continuous-delivery tests only for GKE, and it's being done by using `keptn install -use-case=continuous-delivery` (which is configured by setting `RUN_CONTINUOUS_DELIVERY_TEST`).

To properly test whether jmeter-service and helm-service are running in execution plane (separated from keptn control-plane), we need to create an integration test.
To avoid over-complicating it, we will use the same Kubernetes Cluster (e.g., on GKE), but work in a different namespace.

# Tasks

* Extend [Test-Matrix](https://github.com/keptn/keptn/blob/59dbf4ce9208345c7723d2f6c74710b0b3a27cfe/.github/workflows/integration_tests.yml#L26-L66) to indicate that a test is run with jmeter/helm running in a separate execution plane
* Install JMeter Service + Helm Service using `helm install` or `helm upgrade` (helm-charts will be available soon) in Execution plane (e.g., on a separate namespace  `keptn-exec-plane`)
* Make sure Distributor is properly setup to communicate with Keptn API via Istio (and not internally via svc.cluster.local and not via NATS)
* Run/make use of the [existing continuous-delivery integration test](https://github.com/keptn/keptn/blob/59dbf4ce9208345c7723d2f6c74710b0b3a27cfe/.github/workflows/integration_tests.yml#L430-L442)
* If needed, extend the [test report](https://github.com/keptn/keptn/blob/59dbf4ce9208345c7723d2f6c74710b0b3a27cfe/.github/workflows/integration_tests.yml#L523-L530)
* Cleanup: while we are calling [`keptn uninstall`](https://github.com/keptn/keptn/blob/59dbf4ce9208345c7723d2f6c74710b0b3a27cfe/.github/workflows/integration_tests.yml#L481-L483), we obviously also should clean up execution-plane services

**Note**: Try to not touch/modify existing integration tests (too much).

# Definition of Done

- [ ] Integration Test for Continuous-Delivery on Execution Plane is running and reported in the test report

bacherfl:
Closes #3214

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -438,7 +438,7 @@ jobs:
 
       - name: Test Continuous Delivery (with sockshop)
         id: test_continuous_delivery
-        if: env.RUN_CONTINUOUS_DELIVERY_TEST == 'true' # run only if variable is set
+        if: (env.RUN_CONTINUOUS_DELIVERY_TEST == 'true') && env.REMOTE_EXECUTION_PLANE != 'true'
         timeout-minutes: 45
         continue-on-error: true
         env:
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -383,6 +383,22 @@ jobs:
         timeout-minutes: 1
         run: keptn status
 
+      - name: Install Remote Execution Plane
+        id: install_remote_execution_plane
+        if: env.REMOTE_EXECUTION_PLANE == 'true'
+        timeout-minutes: 5
+        run: |
+          kubectl scale deployment/helm-service -n "${KEPTN_NAMESPACE}" --replicas=0
+          kubectl scale deployment/jmeter-service -n "${KEPTN_NAMESPACE}" --replicas=0
+
+          KEPTN_API_HOSTNAME=$(echo "${KEPTN_ENDPOINT}" | awk -F[/] '{print $3}')
+
+          helm install helm-service http://0.0.0.0:8000/"${HELM_SERVICE_HELM_CHART_NAME}" -n keptn-helm-service --set remoteControlPlane.enabled=true --set remoteControlPlane.api.protocol=http --set remoteControlPlane.api.hostname="${KEPTN_API_HOSTNAME}" --set remoteControlPlane.api.token="${KEPTN_API_TOKEN}" --create-namespace
+          helm install jmeter-service http://0.0.0.0:8000/"${JMETER_SERVICE_HELM_CHART_NAME}" -n keptn-jmeter-service --set remoteControlPlane.enabled=true --set remoteControlPlane.api.protocol=http --set remoteControlPlane.api.hostname="${KEPTN_API_HOSTNAME}" --set remoteControlPlane.api.token="${KEPTN_API_TOKEN}" --create-namespace
+
+          helm test jmeter-service -n keptn-jmeter-service
+          helm test helm-service -n keptn-helm-service
+
       - name: Test Linking Stages
         id: test_linking_stages
         timeout-minutes: 5
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -388,6 +388,7 @@ jobs:
         if: env.REMOTE_EXECUTION_PLANE == 'true'
         timeout-minutes: 5
         env:
+          KEPTN_ENDPOINT: ${{ steps.authenticate_keptn_cli.outputs.KEPTN_ENDPOINT }}
           HELM_SERVICE_HELM_CHART_NAME: ${{ steps.extract_helm_chart_name.outputs.HELM_SERVICE_HELM_CHART_NAME }}
           JMETER_SERVICE_HELM_CHART_NAME: ${{ steps.extract_helm_chart_name.outputs.JMETER_SERVICE_HELM_CHART_NAME }}
         run: |
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -388,6 +388,7 @@ jobs:
         if: env.REMOTE_EXECUTION_PLANE == 'true'
         timeout-minutes: 5
         run: |
+          KEPTN_API_TOKEN=$(kubectl get secret keptn-api-token -n "$KEPTN_NAMESPACE" -o jsonpath='{.data.keptn-api-token}' | base64 --decode)
           kubectl scale deployment/helm-service -n "${KEPTN_NAMESPACE}" --replicas=0
           kubectl scale deployment/jmeter-service -n "${KEPTN_NAMESPACE}" --replicas=0
 
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -436,9 +436,7 @@ jobs:
         timeout-minutes: 45
         continue-on-error: true
         env:
-          VERSION: ${{ env.VERSION }}
           KEPTN_ENDPOINT: ${{ steps.authenticate_keptn_cli.outputs.KEPTN_ENDPOINT }}
-          REMOTE_EXECUTION_PLANE: ${{ matrix.REMOTE_EXECUTION_PLANE }}
           PROJECT: "sockshop"
           SERVICE: "carts"
         run: |
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -558,7 +558,7 @@ jobs:
           TEST_REPORT_FILENAME: test-report-${{ github.run_id }}-${{ matrix.PLATFORM }}-${{ matrix.CLOUD_PROVIDER}}-${{ matrix.VERSION }}.txt
         run: |
           echo "write test report to $TEST_REPORT_FILENAME"
-          echo "| ${{ matrix.PLATFORM }}/${{ matrix.CLOUD_PROVIDER}} ${{ matrix.VERSION }} | ${{ steps.keptn_install.outcome }} | ${{ steps.authenticate_keptn_cli.outcome }} | ${{ steps.test_linking_stages.outcome }} | ${{ steps.test_quality_gates.outcome }} | ${{ steps.test_quality_gates_backwards_compatibility.outcome }} | ${{ steps.test_self_healing.outcome }} | ${{ steps.test_delivery_assistant.outcome }} | ${{ steps.test_continuous_delivery.outcome }} |" > $TEST_REPORT_FILENAME
+          echo "| ${{ matrix.PLATFORM }}/${{ matrix.CLOUD_PROVIDER}} ${{ matrix.VERSION }} | ${{ steps.keptn_install.outcome }} | ${{ steps.authenticate_keptn_cli.outcome }} | ${{ steps.test_linking_stages.outcome }} | ${{ steps.test_quality_gates.outcome }} | ${{ steps.test_quality_gates_backwards_compatibility.outcome }} | ${{ steps.test_self_healing.outcome }} | ${{ steps.test_delivery_assistant.outcome }} | ${{ steps.test_continuous_delivery.outcome }} | ${{ steps.test_continuous_delivery_remote_exec.outcome }} |" > $TEST_REPORT_FILENAME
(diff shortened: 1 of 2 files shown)


test/test_onboard_service.sh
```diff
@@ -4,9 +4,24 @@
...
 
+echo "Using remote execution plane services: ${REMOTE_EXECUTION_PLANE}"
+
+# check if REMOTE_EXECUTION_PLANE is set to true. If yes, scale down the helm-service and jmeter in the keptn namespace and install the services via their helm charts
+if [[ "${REMOTE_EXECUTION_PLANE}" == "true" ]]; then
+  kubectl scale deployment/helm-service -n ${KEPTN_NAMESPACE} --replicas=0
+  kubectl scale deployment/jmeter-service -n ${KEPTN_NAMESPACE} --replicas=0
+
+  KEPTN_API_HOSTNAME=$(kubectl -n ${KEPTN_NAMESPACE} get service api-gateway-nginx -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
+  KEPTN_VERSION_TAG=$(kubectl get deployment -n ${KEPTN_NAMESPACE} helm-service -ogo-template='{{index .metadata.labels "app.kubernetes.io/version"}}')
```
jsoref, line 20: Fwiw, you might be able to add
```
 -o
```
to your `patterns.txt` file instead of adding `ogo` to your `expect.txt` file

.github/workflows/integration_tests.yml
```diff
@@ -433,7 +436,9 @@ jobs:
         timeout-minutes: 45
         continue-on-error: true
         env:
+          VERSION: ${{ env.VERSION }}
```
christian-kreuzberger-dtx, line 439: This should not be needed. `env.VERSION` indicates that this is already available as a global environment variable, which contains the `matrix.VERSION` (which contains the version of the Kubernetes platform provider)

.github/workflows/integration_tests.yml
```diff
@@ -433,7 +436,9 @@ jobs:
...
           KEPTN_ENDPOINT: ${{ steps.authenticate_keptn_cli.outputs.KEPTN_ENDPOINT }}
+          REMOTE_EXECUTION_PLANE: ${{ matrix.REMOTE_EXECUTION_PLANE }}
```
christian-kreuzberger-dtx, line 441: This should not be needed, as it is already defined as an environment variable above.

test/test_onboard_service.sh
```diff
@@ -4,9 +4,24 @@
...
 echo "Testing onboarding for project $PROJECT"
 
+echo "Using remote execution plane services: ${REMOTE_EXECUTION_PLANE}"
+
+# check if REMOTE_EXECUTION_PLANE is set to true. If yes, scale down the helm-service and jmeter in the keptn namespace and install the services via their helm charts
+if [[ "${REMOTE_EXECUTION_PLANE}" == "true" ]]; then
+  kubectl scale deployment/helm-service -n "${KEPTN_NAMESPACE}" --replicas=0
+  kubectl scale deployment/jmeter-service -n "${KEPTN_NAMESPACE}" --replicas=0
+
+  KEPTN_API_HOSTNAME=$(kubectl -n "${KEPTN_NAMESPACE}" get service api-gateway-nginx -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
```
christian-kreuzberger-dtx, line 19: This should already be available in `KEPTN_ENDPOINT`, no need to fetch it here.

test/test_onboard_service.sh
```diff
@@ -4,9 +4,24 @@
...
+  KEPTN_VERSION_TAG=$(kubectl get deployment -n "${KEPTN_NAMESPACE}" helm-service -o go-template='{{index .metadata.labels "app.kubernetes.io/version"}}')
+
+  helm install helm-service ./helm-service/chart -f ./helm-service/chart/values.yaml -n keptn-helm-service --set helmservice.image.tag="${KEPTN_VERSION_TAG}" --set distributor.image.tag="${KEPTN_VERSION_TAG}" --set remoteControlPlane.enabled=true --set remoteControlPlane.api.protocol=http --set remoteControlPlane.api.hostname="${KEPTN_API_HOSTNAME}" --set remoteControlPlane.api.token="${KEPTN_API_TOKEN}" --create-namespace
+  helm install jmeter-service ./jmeter-service/chart -f ./jmeter-service/chart/values.yaml -n keptn-jmeter-service --set jmeterservice.image.tag="${KEPTN_VERSION_TAG}" --set distributor.image.tag="${KEPTN_VERSION_TAG}" --set remoteControlPlane.enabled=true --set remoteControlPlane.api.protocol=http --set remoteControlPlane.api.hostname="${KEPTN_API_HOSTNAME}" --set remoteControlPlane.api.token="${KEPTN_API_TOKEN}" --create-namespace
```
christian-kreuzberger-dtx, line 23: Please use the charts provided within the CI build artifact `keptn-installer`. They *should* be called
* `dist/keptn-installer/helm-service-${KEPTN_VERSION_TAG}.tgz`
* `dist/keptn-installer/jmeter-service-${KEPTN_VERSION_TAG}.tgz`

see https://github.com/keptn/keptn/suites/2202706220/artifacts/45499321 as a reference

In addition, you should not have to specify those tags (they are specified directly in the chart that's available via the CI build artifact) please remove
* `--set helmservice.image.tag="${KEPTN_VERSION_TAG}" --set distributor.image.tag="${KEPTN_VERSION_TAG}"`
* `--set jmeterservice.image.tag="${KEPTN_VERSION_TAG}" --set distributor.image.tag="${KEPTN_VERSION_TAG}"`

test/test_onboard_service.sh
bacherfl: yes, the `KEPTN_ENDPOINT` variable is set at this point, but with a preceding `http://` and `/api` appended, which is not  accepted by the helm chart of the jmeter/helm service

test/test_onboard_service.sh
```diff
@@ -4,9 +4,24 @@
...
+fi
```
christian-kreuzberger-dtx, line 24: Once installed, please verify the installed services via
```
helm test jmeter-service -n keptn-jmeter-service
helm test helm-service -n keptn-helm-service
```

.github/workflows/integration_tests.yml
bacherfl: good catch - has been removed

.github/workflows/integration_tests.yml
bacherfl: good catch - has been removed

.github/workflows/integration_tests.yml
```diff
@@ -482,6 +487,13 @@ jobs:
         timeout-minutes: 5
         run: echo "y" | keptn uninstall -n ${KEPTN_NAMESPACE}
 
+      - name: Uninstall remote execution plane
+        timeout-minutes: 5
+        if: env.REMOTE_EXECUTION_PLANE == 'true'
+        run: |
+          kubectl delete namespace keptn-helm-service
+          kubectl delete namespace keptn-jmeter-service
+
```
christian-kreuzberger-dtx, line 494: I like that we have this here. Does it make sense to also have the install and verification step outside, right before the test is executed?

test/test_onboard_service.sh
```diff
@@ -4,9 +4,24 @@
...
 
+echo "Using remote execution plane services: ${REMOTE_EXECUTION_PLANE}"
+
+# check if REMOTE_EXECUTION_PLANE is set to true. If yes, scale down the helm-service and jmeter in the keptn namespace and install the services via their helm charts
+if [[ "${REMOTE_EXECUTION_PLANE}" == "true" ]]; then
+  kubectl scale deployment/helm-service -n "${KEPTN_NAMESPACE}" --replicas=0
+  kubectl scale deployment/jmeter-service -n "${KEPTN_NAMESPACE}" --replicas=0
+
+  KEPTN_API_HOSTNAME=$(kubectl -n "${KEPTN_NAMESPACE}" get service api-gateway-nginx -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
+  KEPTN_VERSION_TAG=$(kubectl get deployment -n "${KEPTN_NAMESPACE}" helm-service -o go-template='{{index .metadata.labels "app.kubernetes.io/version"}}')
```
christian-kreuzberger-dtx, line 20: I was hopeing that we could have this variable somewhere else, but unfortunately it's not available yet in our integration tests.

Nevertheless, you don't need it, see comment below regarding `helm install`.

christian-kreuzberger-dtx:
Part of Issue #3214 was also to see this test reported at the end.

Looking at your approach to re-use the existing onboard service tests (which I like, btw!), I recommend adding another step in the yaml file:
```
      - name: Test Continuous Delivery on Remote Exec Plane (with sockshop)
        id: test_continuous_delivery_remote_exec
        if: (env.RUN_CONTINUOUS_DELIVERY_TEST == 'true') && env.REMOTE_EXECUTION_PLANE == 'true'
        ...
        run: |
          test/test_onboard_service.sh
          test/test_new_artifact.sh
          test/test_delete_project.sh
```

test/test_onboard_service.sh
christian-kreuzberger-dtx: Okay, I understand. 
I want to avoid having dependencies within the test files, in this case we depend on api-gateway-nginx being exposed via LoadBalancer. I know we have plenty of other cases where we do that, just trying to make sure we don't add new ones.

To help you with this concrete example, how about extracting `KEPTN_API_HOSTNAME` from `KEPTN_ENDPOINT`?
```
echo http://keptn.1.2.3.4.nip.io:1234/api | awk -F[/:] '{print $4}' # returns keptn.1.2.3.4.nip.io
```
Actually, we most likely would like to also get the port for the hostname, so maybe
```
echo http://keptn.1.2.3.4.nip.io:1234/api | awk -F[/] '{print $3}' # returns keptn.1.2.3.4.nip:1234
```

bacherfl: sounds good, I will try out that approach

.github/workflows/integration_tests.yml
```diff
@@ -429,11 +438,27 @@ jobs:
 
       - name: Test Continuous Delivery (with sockshop)
         id: test_continuous_delivery
-        if: env.RUN_CONTINUOUS_DELIVERY_TEST == 'true' # run only if variable is set
+        if: (env.RUN_CONTINUOUS_DELIVERY_TEST == 'true') && env.REMOTE_EXECUTION_PLANE != 'true'
```
christian-kreuzberger-dtx, line 441: Not sure if you also need to have parentheses around the expressions, e.g.:
```
        if: (env.RUN_CONTINUOUS_DELIVERY_TEST == 'true') && (env.REMOTE_EXECUTION_PLANE != 'true')
```
I've seen it in many examples, though I don't think parentheses is necessary here - you can leave it as it is

.github/workflows/integration_tests.yml
```diff
@@ -374,6 +383,22 @@ jobs:
...
 
+      - name: Install Remote Execution Plane
+        id: install_remote_execution_plane
+        if: env.REMOTE_EXECUTION_PLANE == 'true'
+        timeout-minutes: 5
+        run: |
+          kubectl scale deployment/helm-service -n "${KEPTN_NAMESPACE}" --replicas=0
+          kubectl scale deployment/jmeter-service -n "${KEPTN_NAMESPACE}" --replicas=0
+
+          KEPTN_API_HOSTNAME=$(echo "${KEPTN_ENDPOINT}" | awk -F[/] '{print $3}')
```
christian-kreuzberger-dtx, line 394: I think you need to fetch `KEPTN_API_TOKEN` here too

christian-kreuzberger-dtx (dismissed):
Approved. Please merge as soon as you are comfortable with it :)

christian-kreuzberger-dtx (approved):
LGTM

PR #3466, opened 2021-03-08
Make helm operations atomic

thschue:
The behaviour introduced in this PR ensures that a deployment gets deleted when the initial installation fails and upgrades get reverted when they fail. Further reading: https://medium.com/polarsquad/check-your-helm-deployments-ffe26014804

### DIFF
--- a/helm-service/pkg/helm/helmv3_executor.go
+++ b/helm-service/pkg/helm/helmv3_executor.go
@@ -144,6 +144,7 @@ func (h *HelmV3Executor) UpgradeChart(ch *chart.Chart, releaseName, namespace st
 			iCli.ReleaseName = releaseName
 			iCli.Wait = true
 			iCli.Timeout = time.Minute * 3
+			iCli.Atomic = true
 			release, err = iCli.Run(ch, vals)
 		} else {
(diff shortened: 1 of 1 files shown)


christian-kreuzberger-dtx:
The documentation linked points to Helm v2. We're using v3, is there any difference?

thschue:
> The documentation linked points to Helm v2. We're using v3, is there any difference?

No, i just added some documentation :-)

christian-kreuzberger-dtx:
I'm not sure if this will work together with our example shipyard for blue-green deployments:
https://github.com/keptn/examples/blob/519680b56e270e9a59b09d864415df3d45643a09/onboarding-carts/shipyard.yaml#L40-L45

Here we have
```
        - name: "rollback"
          triggeredOn:
            - event: "staging.delivery.finished"
              selector:
                match:
                  result: "fail"
```

In case any step in `delivery` in stage `staging` fails.

What happens if the `deployment` step fails, but performs a rollback already?

christian-kreuzberger-dtx:
@warber @johannes-b was this investigated during my absence?

agrimmer (approved):
lgtm

Issue #3350, opened 2021-02-23
Delete helm-service and jmeter-service from continuous-delivery Helm Chart and adapt CLI accordingly


johannes-b:
With #3253 the helm-service and jmeter-service provide their individual Helm Chart. Consequently, there is no need to maintain them in the continuous-delivery Helm Chart here: https://github.com/keptn/keptn/blob/master/installer/manifests/keptn/charts/continuous-delivery/templates/continuous-deployment.yaml

**Task**
- Remove the helm-service and jmeter-service from the continuous-delivery Helm Chart
- Adapt the CLI to install the Helm Charts for helm-service and jmeter-service when executing  `keptn install --use-case=continuous-delivery` -> add a separate `helm install` for both charts. 
- Verify building Helm Charts in CI works

**Definition of Done:**
- [ ] When executing `keptn upgrade` from 0.8.0 to 0.8.x the upgrade has to work
- [ ] When executing `keptn install --use-case=continuous-delivery` the control-plane including the helm-service and jmeter-service are installed on my cluster. (No changed behaviour for the end-user) 
- [ ] `keptn install --chart-repo=...` still works so we can install builds directly from GitHub

christian-kreuzberger-dtx:
FYI, this should be as simple as doing this in the CLI:
```
helm install jmeter-service https://github.com/keptn/keptn/releases/download/0.8.0/jmeter-service-0.8.0.tgz -n keptn
helm install helm-service https://github.com/keptn/keptn/releases/download/0.8.0/helm-service-0.8.0.tgz -n keptn
```

At the moment we are also supplying jmeter and helm-service charts at google cloud, so it might even work using the same chart repo.

PR #3471, opened 2021-03-08
#3421 Fix debugging for jmeter and helm-service

### LINKED ISSUE #3421: Debugging helm-service is no longer possible
Keptn 0.8.0-rc1:

After merging of https://github.com/keptn/keptn/pull/3317 it is no longer possible to debug helm-service (and probably also jmeter-service) in GoLand using skaffold / Google Cloud Code.

**To Reproduce**

Install `helm-service` and/or `jmeter-service` using
```console
helm install helm-service https://github.com/keptn/keptn/releases/download/0.8.0-rc1/helm-service-0.8.0-rc1.tgz -n keptn-exec --create-namespace --values=values.yaml
```
where values.yaml looks like this:
```remoteControlPlane:
  enabled: true                              # Enables remote execution plane mode
  api:
    protocol: "http"                        # Used Protocol (http, https)
    hostname: "YOUR-HOST"                             # Hostname of the control plane cluster (and Port)
    apiValidateTls: true                     # Defines if the control plane certificate should be validated
    token: "YOUR-TOKEN"                                # Keptn API Token
```

christian-kreuzberger-dtx:

This PR fixes #3421 by performing some changes to skaffold.yaml, which will allow us to debug jmeter and helm service again.

In addition, I have included similar changes as in PR https://github.com/keptn/keptn/pull/3467, e.g.: using `SKAFFOLD_GO_GCFLAGS` instead of `debugBuild`, memory settings for distributor, fixing distributor tag to 0.8.0

### DIFF
--- a/helm-service/Dockerfile
+++ b/helm-service/Dockerfile
@@ -20,17 +20,15 @@ COPY go.mod go.sum ./
 # Download dependencies
 RUN go mod download
 
-ARG debugBuild
-
-# set buildflags for debug build
-RUN if [ ! -z "$debugBuild" ]; then export BUILDFLAGS='-gcflags "all=-N -l"'; fi
-
 # Copy local code to the container image. 
 COPY . .
 
+# `skaffold debug` sets SKAFFOLD_GO_GCFLAGS to disable compiler optimizations
+ARG SKAFFOLD_GO_GCFLAGS
+
 # Build the command inside the container.
 # (You may fetch or manage dependencies here, either manually or with a tool like "godep".)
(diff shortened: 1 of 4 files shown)


christian-kreuzberger-dtx:
> When only the control plane is installed, debugging the services using skaffold works without a problem. 

Perfect.

> However, when keptn has previously been installed using the --use-case=continuous-delivery option, deploying the service via helm chart will not work due to resource conflicts between the keptn helm chart and the chart of the service. 

That was something I feared, thanks for checking :( 

Anyawy, this issue should be gone once https://github.com/keptn/keptn/issues/3350 is implemented, so it should not be a blocker for long. Though that means #3350 needs to be upped in priority.

PR #3491, opened 2021-03-10
#2289 Show services and recent sequence in stage tile as table

### LINKED ISSUE #2289: Stage tile supports many services
**Status quo:** In an environment screen of a Keptn user, the stage tile (containing 5 services) looked as follows:
image

**Improvement:**
- Each service (including its currently deployed artifact version) is shown in a row
- The services are ordered according to the most recent sequences (meaning that the service with the most recent sequence is on top) 
- Display the last sequence that got executed on that service [Name]
- Display the icon that corresponds to that sequence [Icon]
- Display the last evaluation result executed in the sequence of that stage.
- The `show more Services` button allows loading the next 5 services
- If there is no sequence executed for the service, a `-` is shown for the sequence name and the evaluation result

**Definition of Done**:
image

ermin-muratovic:
Services are shown in the stage tile in the environments screen as table, also showing the recent sequence and the status of that sequence and if available also the evaluation result for that sequence on that stage.

image

closes #2289

### DIFF
--- a/bridge/client/app/_components/ktb-evaluation-info/ktb-evaluation-info.component.html
+++ b/bridge/client/app/_components/ktb-evaluation-info/ktb-evaluation-info.component.html
@@ -1,0 +1,9 @@
+<dt-tag-list aria-label="evaluation-info" *ngIf="evaluation">
+  <dt-tag class="justify-content-center" [dtOverlay]="overlay" [dtOverlayConfig]="overlayConfig" [class.error]="evaluation.isFaulty()" [class.warning]="evaluation.isWarning()" [class.success]="evaluation.isSuccessful()" [textContent]="evaluation.getFinishedEvent().data.evaluation.score | number:'1.0-0'"></dt-tag>
+  <ng-template #overlay>
+    <ktb-evaluation-details [evaluationData]="evaluation.getFinishedEvent()" [showChart]="false"></ktb-evaluation-details>
+  </ng-template>
+</dt-tag-list>
+<dt-tag-list *ngIf="!evaluation">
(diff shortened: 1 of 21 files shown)


Kirdock (changes requested):
- "-"-character for missing evaluation is not centered
image
- if I remember right services should be ordered by latest sequence and there should also be an indicator "show more services" if there are more than 5. This may also be a good start to split data sources instead of using one for each view.
image
- Stages in environment screen are broken, if there is a running evaluation . `Cannot read property 'data' of undefined`.
https://github.com/keptn/keptn/blob/d597f6ae0034ddf7fca0ce18cd9f9ed3561b944a/bridge/client/app/_components/ktb-evaluation-info/ktb-evaluation-info.component.html#L2

ermin-muratovic:
thanks for the comment.

> * if I remember right services should be ordered by latest sequence and there should also be an indicator "show more services" if there are more than 5. This may also be a good start to split data sources instead of using one for each view.

Regarding this one, actually I thought about a separate data source for this view, but it won't make sense, since I would have to do the same polling as in the service screen anyway, because not the last 5 sequences are interesting, but the 5 services with the most recent sequences. So I need to poll the sequences for all services and then apply the sorting. Only for the initial call it would make a difference, since currently we load 20 sequences per service, but it would be enough to load 1 sequence per service for this view. let's see how this works and if we experience performance issues, we can still adapt our strategy.

ermin-muratovic:
I fixed the bugs and in case that there are more than 3 services, it will show a "Show more" button.

image

Kirdock (approved):
LGTM

PR #3494, opened 2021-03-10
#3489: fixed swagger doc

### LINKED ISSUE #3489: Keptn API: Endpoint is missing path parameter and mismatch between parameter name
The Keptn API for the shipyard-controller has two bugs as shown below. In course of fixing this bug, please do a sanity check of the other endpoints.

**Missing path parameter**

image

* Fix: Double-check with implementation, but I assume that `stage` is missing in the URL

**Mismatch between naming**

image

* Fix: Streamlined with the other endpoints

warber:
fixes #3489 

### DIFF
--- a/shipyard-controller/docs/docs.go
+++ b/shipyard-controller/docs/docs.go
@@ -303,62 +303,6 @@ var doc = `{
                 }
             }
         },
-        "/project/{projectName}/stage/{stageName}": {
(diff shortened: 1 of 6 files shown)


bacherfl (approved):
lgtm, to be safe I suggest to execute the integration tests on that branch

PR #3513, opened 2021-03-11
Print warning in case of non-admin rights in helm-service

didiladi:
This commit prevents that the helm-service exists after discovering that
it does not have admin access when checking the available RBAC. Instead,
it logs a warning and continues to operate.

This is a temporary workaround for supporting fine-granular RBAC in the
helm-service to deploy kubernetes workloads (#3511).

### DIFF
--- a/helm-service/main.go
+++ b/helm-service/main.go
@@ -168,7 +168,7 @@ func _main(args []string, env envConfig) int {
 		log.Fatal(fmt.Sprintf("failed to check whether helm-service has admin right: %v", err))
 	}
 	if !adminRights {
-		log.Fatal("helm-service has insufficient RBAC rights.")
+		log.Println("Warning: helm-service is running without admin RBAC rights. See #3511 for details.")
 	}
 
 	ctx := context.Background()


christian-kreuzberger-dtx:
Thanks Didi for the PR!
We have a DCO in place (see checks on GitHub), please sign your commit ;)

helm-service/main.go
```diff
@@ -168,7 +168,7 @@ func _main(args []string, env envConfig) int {
 		log.Fatal(fmt.Sprintf("failed to check whether helm-service has admin right: %v", err))
 	}
 	if !adminRights {
-		log.Fatal("helm-service has insufficient RBAC rights.")
+		log.Println("Warning: helm-service has non-admin RBAC rights. Will continue to operate")
```
christian-kreuzberger-dtx, line 171: Let's change this message to "Warning: helm-service is running without admin RBAC rights. See https://github.com/keptn/keptn/issues/3511 for details."

christian-kreuzberger-dtx:
@agrimmer @thschue with this change we are allowing helm-service to operate without admin rights here. I believe this is not a problem, but wanted to check back with you.

christian-kreuzberger-dtx (approved):
LGTM

PR #3519, opened 2021-03-11
Mongodb-datastore: Remove magiconair dependency in favor of stretchr/testify

christian-kreuzberger-dtx:
During auditing of our go mod files I noticed that mongodb-datastore still uses the old magiconair properties dependency, which should be replaced by stretchr/testify.


christian-kreuzberger-dtx:
superseeded by https://github.com/keptn/keptn/pull/3520

PR #3566, opened 2021-03-18
#3537 Show all labels in service screen correctly

### LINKED ISSUE #3537: Evaluation view in Service View doesnt show all labels as compared to full screen view
When SLI services are providing labels, e.g: the Dynatrace-sli-service those are not shown in the Service View. They are however shown in the Full Screen View of the evaluation.
After talking with @ermin-muratovic its because the service view uses the evaluation.triggered event whereas the full screen view uses the evaluation.finished. evaluation.finished will include additional labels from e.g: sli-providers. therefore the service view should also display the evaluation finished event.

Here is a more detailed overview of the problem
image

ermin-muratovic:
Closes #3537

### DIFF
--- a/bridge/client/app/_components/ktb-event-item/ktb-event-item.component.html
+++ b/bridge/client/app/_components/ktb-event-item/ktb-event-item.component.html
@@ -35,11 +35,11 @@
       </dt-tag-list>
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


Kirdock (approved):
LGTM

Issue #3577, opened 2021-03-19
Allow comma-separated list on event filters for distributors


johannes-b:
**Use-Case:**

![image](https://user-images.githubusercontent.com/729071/111774878-58c72580-88b0-11eb-9592-cc41d8169167.png)

For implementing the above use-case, I need two helm-services: (1) works on control-plane and deploys on `development`, `hardening`, `production`; (2) works on execution-plane and deploys on `production-remote`

**Limitation**: Filters on the distributor just allow a single value. 

**Improvement**: As a user, I would like to configure my distributor to listen to events, e.g., for multiple stages. In the example below, I would like to get events for the `dev` and `hardening` stage: 

```
distributor:
  projectFilter: ""                 
  stageFilter: "development,hardening,production"             
  serviceFilter: ""                     
```

**Implementation details:**
- This should work on project/stage/service-level
- The corresponding function is here: https://github.com/keptn/keptn/blob/master/distributor/cmd/main.go#L681
- Instead of doing a equals comparison, a contains could address this limitation

**Definition of Done:**
- [ ] When setting a filter, a distributor can process events for multiple stags/projects/services
- [ ] Update of Readme.md and keptn/docs (https://keptn.sh/docs/0.8.x/operate/multi_cluster/#how-to-deploy-an-execution-plane-services)

PR #3579, opened 2021-03-19
#3577 Allow a comma-separated list of values for distributor filters

johannes-b:
Signed-off-by: Johannes <johannes.braeuer@dynatrace.com>

### DIFF
--- a/distributor/cmd/main.go
+++ b/distributor/cmd/main.go
@@ -677,19 +677,29 @@ func decodeCloudEvent(data []byte) (*cloudevents.Event, error) {
 	return &event, nil
 }
 
-// Primitive filtering based on project, stage and service properties
+// Primitive filtering based on project, stage, and service properties
 func matchesFilter(e cloudevents.Event) bool {
 
 	keptnBase := &v0_2_0.EventData{}
 	if err := e.DataAs(keptnBase); err != nil {
 		return true
 	}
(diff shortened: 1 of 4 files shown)


distributor/cmd/main.go
```diff
@@ -684,9 +684,9 @@ func matchesFilter(e cloudevents.Event) bool {
 	if err := e.DataAs(keptnBase); err != nil {
 		return true
 	}
-	if env.ProjectFilter != "" && keptnBase.Project != env.ProjectFilter ||
-		env.StageFilter != "" && keptnBase.Stage != env.StageFilter ||
-		env.ServiceFilter != "" && keptnBase.Service != env.ServiceFilter {
+	if env.ProjectFilter != "" && !strings.Contains(env.ProjectFilter, keptnBase.Project) ||
```
bacherfl, line 687: I believe we have to convert the project/stage/service filter to an array and try to find an exact match for `keptnBase.Project`. Otherwise, if we have e.g. a filter with the value:

```
my-project,my-other-project
```

events for projects whose name is a substring of any of those project names would also be forwarded. E.g. if the distributor receives an event with `keptnBase.Project=other` would also fulfill the condition.

johannes-b: Absolutely true! Let me change that.

bacherfl (approved):
lgtm

PR #3581, opened 2021-03-19
#1280 mockup for keptn services and subscriptions

### LINKED ISSUE #1280: Mockup to show installed keptn services and latest version available
**Idea:** Within the Keptn Bridge, the installed Keptn services (e.g., lighthouse, gatekeeper, ...) should be listed. This listing shows the latest version of each service. 

This could mean that we need to list all Kubernetes Deployments within the Keptn Namespaces, parse the images and the tags used, and verify if a newer release is available (e.g., via DockerHub, GitHub, via a JSON file that we put on our S3 Bucket, ...)

Kirdock:
Closes #1280 
image


ermin-muratovic (approved):
lgtm

PR #3584, opened 2021-03-21
Keptn service template generation from cli

### LINKED ISSUE #3034: Keptn CLI Command to generate service/integration based on keptn-sandbox/keptn-service-template-go
**Tasks**

Depending on the keptn cli version, download and use the template from https://github.com/keptn-sandbox/keptn-service-template-go (by matching the Keptn CLI version with a GitHub release of the template; as a fallback: use master branch) and requests the following inputs from the user:

* Service name
* Docker Image Name (organization/image-name)
* comma separated list of cloud-events to listen for (e.g., sh.keptn.event.deployment.triggered)
* TBD

Keptn CLI then creates a directory service-name with the appropriate contents of the template, and replaces the occurrences of several strings according to the README file of the template (using go-templates).

**Definition of Done**

- [ ] Task from above is implemented in the CLI
- [ ] Keptn service template is an actual template with placeholders
- [ ] Keptn service template contains a docker-compose file to run the service locally (distributor + service) 
- [ ] Keptn service template is generated using the CLI command keptn generate keptn-service <service-name> [--keptn-service-template-version=master]
- [ ] Integration Test is added to keptn/keptn: Use the implemented CLI command with the master branch of the keptn-service-template and it should at least do the following:
   - go test ...
   - docker build …
   - docker run … 
   - Run the service and send a cloud-event (e.g., one from the test-events directory of the template) to the service; service should respond with status code 200
   - Update documentation at https://keptn.sh/docs/0.8.x/integrations/custom_integration/ (see keptn/keptn.github.io repo) to include the new command for generating a keptn-service

DarthBenro008:
# Keptn Service Template Generation from CLI

- This project is part of [LFX Mentorship Program](https://mentorship.lfx.linuxfoundation.org/project/630a656a-74ee-4494-9467-b10201ae9b20)
- Aims to create and fixes #3034  

### How to initialise the template?

There were two options to be explored:

- Clone the template
  - Would need to have `git` installed in the developer machine
  - No need to unzip or download, which may cause issues.
  - We can either use a system interop call or make use of *git implementation written in go*, [go-git](https://github.com/go-git/go-git)
- Download and unzip the contents
  - Might be ineffective on some machines
  - maintainance issues
  - does not need `git` as a pre-requiste

**Conclusion:** Cloning the template and inclusion of the `go-git` as a dependency would be the best way to proceed as we can mock and write *unit-testing* and would also not require the developer to have `git` installed as default.

### How to replace strings in the template?

The original issue mentions the use of `go-template` but this would pose some problems such as:

- Refactoring of the service-template that is found [here](https://github.com/keptn-sandbox/keptn-service-template-go) and include `{}` for `go-template` to work.
- User cannot clone the service-template repo directly and work with.

Conclusion: `walkdir` into the cloned folder and treat the files as string, then you can easily replace certain placeholders as mentioned [here](https://github.com/keptn-sandbox/keptn-service-template-go)

A small POC can be found [here](https://gist.github.com/DarthBenro008/dd597623600d3f85a5fa3ca885d36f9c)

### Use of io/ioutil vs io/fs

The new go1.16 comes with the `io/fs` package that makes it much more easier to interact with files and has deprecated `io/ioutil` package.Hence it will be a requirement for the developers to have go1.16 installed in order to compile the CLI, however when shipped as a binary this will not post an issue to the developer.


DarthBenro008:
Milestone 1:
- Research phase completed and a plan of actions have been created.
- CLI Command initialised with the required flags
- `go-git` clone functionality explored.

cli/pkg/git/git_utils.go
```diff
@@ -0,0 +1,55 @@
...
+	NodeID string `json:"node_id"`
+	URL    string `json:"url"`
+	Object struct {
+		Sha  string `json:"sha"`
+		Type string `json:"type"`
+		URL  string `json:"url"`
+	} `json:"object"`
+}
+
+func CloneGitHubUrl(filePath string, url string, tag string) error {
```
bacherfl, line 23: would it be possible to directly use the Github API to download the repository content (see https://docs.github.com/en/rest/reference/repos#download-a-repository-archive-tar)? Adding an additional dependency to `github.com/go-git/go-git/v5` seems to be a bit of an overkill if only the `PlainClone` function from this library will be used

DarthBenro008: Using a tar/zip download from GitHub API, would mean that we will have to unzip/untar the downloaded files ourselves. This would involve OS I/O Operations, in the long run, the unzipping/untaring operational logic might fail maybe due to OS constraints or due to how golang handles the filesystem (something which they did in `go1.16` release)

However I also felt that one additional library would be an overkill for the plain clone, but during the research phase, I had found the following pros and cons and had discussed the same.


> ## Research Phase
> ### How to initialise the template?
> There were two options to be explored:
> 
> * Clone the template
>   
>   * Would need to have `git` installed in the developer machine
>   * No need to unzip or download, which may cause issues.
>   * We can either use a system interop call or make use of _git implementation written in go_, [go-git](https://github.com/go-git/go-git)
> * Download and unzip the contents
>   
>   * Might be ineffective on some machines
>   * maintainance issues
>   * does not need `git` as a pre-requiste
> 
> **Conclusion:** Cloning the template and inclusion of the `go-git` as a dependency would be the best way to proceed as we can mock and write _unit-testing_ and would also not require the developer to have `git` installed as default.


**An alternative solution** could be to assume by default that the user has `git` installed. Then we could simply run an `exec` to clone the required repository.

bacherfl: thank you for the explanation - those are valid points. In my opinion I'd say it's OK to do this using the go-git library, but I would also like to ask @christian-kreuzberger-dtx and @warber for their opinions on this.

warber: @DarthBenro008 I don't quite understand your arguments about why the manual extraction approach involves OS I/O while the other does not. Anyway, I think there is a bigger issue with cloning the repo as you will also receive the entire `.git` folder which is not what we want in this context. 
IMO you can quite easily use the facilities of the `archive/zip` (or `compress/gzip` if you'd rather want to process the tarball ) package to extract the downloaded content. Note, that this can even be done _in-memory_ by creating a `Reader` directly from the response body without temporarily storing the zip/tarball`

``` 
[...]
 zipReader, err := zip.NewReader(bytes.NewReader(body), int64(len(body)))
[...]
``` 
Given that the template repository is quite small this should be enough.
From then on you can just loop over each `File` entry of the `zipReader` and create the corresponding files and directories on disk. (you'll find plenty of examples for the logic of extracting a zip or tarball correctly online). Note that you might also think about writing the files in a temporary directory in the first place and when everything worked move it over to the final destination.

hope this helps...  :slightly_smiling_face:

cli/pkg/file/file_utils.go
```diff
@@ -0,0 +1,113 @@
+package file
```
warber, line 1: please move the most common functionalities of this file to go-utils in the appropriate package:
https://github.com/keptn/go-utils/tree/master/pkg/common 
Note, that some of them already exists. (e.g. DownloadFile)

cli/pkg/git/git_utils.go
DarthBenro008: @warber thank you for the detailed insights, the .git folder is actually removed after the PlainClone as done [here](https://github.com/DarthBenro008/keptn/blob/feat/cli-service-generation/cli/cmd/generate_keptn_service.go#L71)

The `archive/zip` package then should be helpful as we already making use of GitHub API. The only concern was performance. As the repo is small, it should be insignificant. I will try out this approach! 😃

dyladan:
[CLA assistant check](https://cla.oss.dynalabs.io/keptn/keptn?pullRequest=3584) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.oss.dynalabs.io/keptn/keptn?pullRequest=3584) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.oss.dynalabs.io/check/keptn/keptn?pullRequest=3584) it.</sub>

PR #3600, opened 2021-03-24
fix apiGroup

warber:
apigroup field was missing preventing the secret service from creating roles


johannes-b (approved):
lgtm

PR #3650, opened 2021-03-30
#3641 Improve bridge Dockerfile & package.json usage 

### LINKED ISSUE #3641: keptn bridge Dockerfile package.json usage needs improvement
**Environment**

Issue can be reproduced locally when building the Docker image for keptn.

**Affected Component**

keptn bridge -  0.6.1

**Describe the bug**

keptn bridge's Dockerfile contains a multi stage build [1]. This helps to reduce the image size and keep npm dependencies minimal inside the image. To achieve this, node server dependencies are installed through Dockerfile & only Angular distribution get copied to the final image.

However, this makes it difficult to track correct dependencies & versions that end up in the docker image as a typical developer rely on package.json for library changes . Further, developers can be mislead on changing current bridge's root level package.json, thinking that Docker image will have all dependencies added there. 

**To Reproduce**

Add a new library dependency in package.json and use it in node server. Then try to build the Docker image and run the bridge through this image. Though the local setup works (as we perform a npm install against root level package.json), docker image based bridge will fail as dependencies defined in package.json are not used for final image.

**Expected behavior**

We must keep the multi stage build and reduce final image size. However, node server dependencies must be installed through a dedicated package.json, which is easier to maintain by a node/angular developer.

**Solution**

A root level package.json and a node server specific package.json

```
/bridge
|
| -- /server
|     |
|     `-- package.json <- node server specific, minimalistic dependencies 
|
| `-- package.json <- Angular specific & dev setup specific dependencies
```

Considerations 

- We should communicate specific package.json usage through ReadMe
- Update current Dockerfile and use server package.json for building the image 

[1] https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds

Kavindu-Dodan:
This change introduce a new package.json for Express server component of the bridge. Having a dedicated package.json make it easy to add, remove and update dependencies. 

Further, Dockerfile still keeps the multi-stage build and utilise the new server package.json to install Express server dependencies. 

Fixes #3641


bridge/.dockerignore
```diff
@@ -1,5 +1,6 @@
 .vscode/
 .idea/
 deploy/
+dist/
```
Kavindu-Dodan, line 4: This helps local development as it ignore already generated dist folder.

Kavindu-Dodan:
Closing the PR as I have conflicting GIT users. This impacts participants and DCO. I will change the user locally and push a new branch so that my user details are correct.

PR #3675, opened 2021-04-01
#3609 improve trace message formatting

### LINKED ISSUE #3609: Improve formatting of Message field in Bridge to improve readability of output
The bridge does a great job in displaying the result message field of services. But - if messages grow in size it is really hard to read. 
To give you an example. This is the output of the argo-service which provides an overivew of the rollout that just happend. it is passing this as part of the message field:

image

The problem however is that this is displayed without any formatting in the bridge and is therefore not readable
image

In discussion with @ermin-muratovic one quick fix could be to use `<pre>` instead of `<p>` for that block

Kirdock:
Closes #3609 


ermin-muratovic (approved):
lgtm

PR #3678, opened 2021-04-01
Improve x-forward-proto header handling for bridge

### LINKED ISSUE #3672: keptn api-gateway - Improve handling of X-Forwarded-Proto header for bridge

keptn deployments are fronted by a ngnix gateway [1]. And it already handle the _X-Forwarded-Proto_ header [2] which helps to identify the schema used between client and proxy.

However, a typical keptn deployment can be fronted by a load balancer (ex:- CLB for EKS). And when SSL get terminated at that load balancer, we no longer get correct _X-Forwarded-Proto_ that was used by client. 

When working on a solution for #3448 we require cookies, Further, the best practice is to make these cookies secure for production environments (of course conditionally if required). Hence I would like to see improved X-Forwarded-Proto handling for Keptn api-gateway.

I would suggest to use a conditional header generation similar to below (credits to serverfault answer https://serverfault.com/a/516382 ),

```
...
map $http_x_forwarded_proto $forwardproto{
     default $scheme;
     https https; 
}  
...

proxy_set_header X-Forwarded-Proto $forwardproto
....
```

This combining with bridge's Express _trust proxy_ [3] usage, will enable bridge to server secure cookies even when there are load balancers  in front of the keptn deployment. 

[1] https://keptn.sh/docs/concepts/architecture/assets/architecture.png
[2] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
[3] https://expressjs.com/en/guide/behind-proxies.html

Kavindu-Dodan:
Fixes  #3672 by introducing conditional selection of scheme for X-Forwarded-Proto on bridge.

### DIFF
--- a/installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
+++ b/installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
@@ -43,7 +43,7 @@ data:
         '' close;
       }
 
-      map $http_x_forwarded_proto $forwardproto{
+      map $http_x_forwarded_proto $forward_proto {
         default $scheme;
         https https;
       }
@@ -80,7 +80,7 @@ data:
       proxy_http_version 1.1;
       proxy_set_header X-Real-IP $remote_addr;
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-      proxy_set_header X-Forwarded-Proto $forwardproto;
+      proxy_set_header X-Forwarded-Proto $forward_proto;
       proxy_set_header Upgrade $http_upgrade;
       proxy_set_header Connection "Upgrade";
       proxy_set_header Host $host;
--- a/installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
+++ b/installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
@@ -43,6 +43,11 @@ data:
         '' close;
       }
 
+      map $http_x_forwarded_proto $forwardproto{
+        default $scheme;
+        https https;
+      }
+
       server {
           listen       8080;
           server_name  _;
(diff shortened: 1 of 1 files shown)


agrimmer (commented):
lgtm

agrimmer (approved):

bacherfl (dismissed):

installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
```diff
@@ -43,6 +43,11 @@ data:
         '' close;
       }
 
+      map $http_x_forwarded_proto $forwardproto{
```
jsoref, line 46: Can you do this?
```suggestion
      map $http_x_forwarded_proto $forward_proto{
```

If not, you need to add `forwardproto` to `expect.txt`

installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
```diff
@@ -75,7 +80,7 @@ data:
       proxy_http_version 1.1;
       proxy_set_header X-Real-IP $remote_addr;
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-      proxy_set_header X-Forwarded-Proto $scheme;
+      proxy_set_header X-Forwarded-Proto $forwardproto;
```
jsoref, line 83:
```suggestion
      proxy_set_header X-Forwarded-Proto $forward_proto;
```

jsoref (changes requested):

installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
Kavindu-Dodan: Thx for the suggestion, I will update the PR .!

agrimmer (approved):
lgtm

Issue #3671, opened 2021-03-31
Helm Service to wait and validate if user_managed deployment is available on custom deploymentURI


grabnerandi:
In #3659 we extend the capability of helm-service to extract the deploymentURIs from the values.yaml. This is especially useful for user_managed charts.

In order to make sure the helm-service is not just applying the helm chart but at least does a sanity check on the availability of that URL after the helm apply, I suggest that the helm service should validate whether that URL is reachable. 
As it could take some time I suggest retrying the check several times with a wait in between.

If - after several retries (e.g: 10) with a wait time of 5s in between - the URL doesn't return an HTTP 200, the helm-service can respond with a warning that the deployment itself was successful but that the URL couldn't be reached

Issue #3659, opened 2021-03-30
Helm Service shall extract deploymentURI from values.yaml to support custom helm charts


grabnerandi:
Right now the Helm Service only extracts deploymentURI either from Istio's VirtualServices when using the Helm Services for Blue/Green Deployments or when the deploymentURI is passed as part of the incoming deployment.triggered event

As we will see more users using user_managed helm charts, e.g: for Argo Rollouts we need to provide an option for the Helm Service to extract deploymentURIPublic and deploymentURILocal even for those helm charts. The proposal is that these values can be defined in the values.yaml file. Here is an example from my Argo Rollout tutorial:

```yaml
affinity: {}
deploymentURILocal: 
deploymentURIPublic: http://simplenode.demo-rollout-staging.keptn08-agrabner.demo.keptn.sh/
image: docker.io/grabnerandi/simplenodeservice
ingress:
  clusterIssuer: issuer.letsencrypt
  ingressHost: keptn08-agrabner.demo.keptn.sh
  secretName: keptn-tls
pullPolicy: Always
replicaCount: 1
service:
  name: simplenode
  port: 8080
  stage: staging
  type: ClusterIP
tag: 3.0.0
```

The Helm Service should parse the values.yaml and - if those two properties exist - extract them and pass them as part of the deployment.finished event

johannes-b:
@warber When discussing this issue, we should consider: https://github.com/keptn/keptn/issues/3637

johannes-b:
In discussing this issue further, we concluded that the `values.yaml` file is misused with potential problems when deploying a Helm Chart not in a `user_managed` mode. The misuse we see is in using the `values.yaml` to store properties that are not referenced by the Helm Chart itself. This is not a best practice by Helm.  

**User Story:** As a user, I would like to customize my deployment with a deployment URL for public/local access. 

* Instead of adding the `deploymentURILocal` and `deploymentURIPublic` to the `values.yaml`, the proposed way is a new config file: `deploymentURIs.yaml` which needs to be added like any other resource: 
```
keptn add-resource --project=sockshop --stage=production --service=carts --resource=deploymentURI.yaml  --resourceUri=helm/deplyomentURI.yaml
```

* The helm-service reads the properties from that file, as part of:  https://github.com/keptn/keptn/pull/3700/files#diff-7005ba2b40f43bccbf1e2752f13a721bef405433bbfbe9d0a95e2bf5c13ce37eR64

[...]

Issue #3655, opened 2021-03-30
Helm Service causes error during rollback when using user_managed deploymentStrategy


grabnerandi:
With Keptn 0.8 the Helm-Service introduced a deploymentstrategy called "user_managed". This strategy allows the user to define a custom helm chart where the helm-service simply just does a helm apply during deployment.

Right now the Helm-Service also reacts on the rollback.triggered event even for "user_managed" strategy which results in an error because the helm-service assumes it has a generated chart. There is however now generated chart because charts are only created if you "onboard" a service for the direct or blue/green deployment strategy

Here is the error message we see in the bridge:
![image](https://user-images.githubusercontent.com/675886/112980536-496e9480-915a-11eb-8f84-8fe30b4710bc.png)

**How to fix this?**
I think the Helm Service should not react to the rollback in case there is a user_managed deploymentStrategy. Or - when it does - it should not return an error but a pass with a message that the Helm-Service didnt execute any rollback actions because the deploymentstrategy is user_managed

warber:
Working on this item also includes #3610

Issue #3624, opened 2021-03-25
inconsistent usage of user-managed and user_managed causing issues


grabnerandi:
The Helm-Service provides an option of a deploymentstrategy called "user_managed" - but - it seems it is used inconsistently which causes issues. On the one side it used as "user_managed" but then there are parts in the code that uses "user-managed". This leads to problems for instance with tagging and with querying data.

We have to use one consistent string as otherwise end2end workflows that rely on the deploymentstrategy will not work

johannes-b:
Use `user_managed` instead of `user-managed`.

PR #3700, opened 2021-04-07
Helm-Service Improvements to better support user_managed deployments such as Argo Rollouts

grabnerandi:
This PR includes several improvements and fixes that I needed to implement the Keptn -> Argo Rollout integration
When working with the current Helm-Service I ran into issues especially around user_managed charts

Here is the list of issues that this PR addresses
#3655 #3624 #3610 #3659 #3671

### DIFF
--- a/helm-service/controller/deployment_handler.go
+++ b/helm-service/controller/deployment_handler.go
@@ -6,6 +6,8 @@ import (
 	"math"
 	"time"
 
+	"net/http"
+
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	keptnevents "github.com/keptn/go-utils/pkg/lib"
 	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
@@ -96,15 +98,25 @@ func (h *DeploymentHandler) HandleEvent(ce cloudevents.Event) {
 		return
 	}
 
-	// TODO - when we just applied a helm chart we should wait until the deployed service is ready
-	// right now - lets just add a 5 seconds wait but we need to find a better way for this, e.g: query if the deploymentURIs are accessible
-	waitDuration, err := time.ParseDuration("5s")
-	<-time.After(waitDuration)
+	warningMessage := ""
+
+	// If there is a customDeploymentURI specified in values.yaml lets validate that this deployment is accessible!
+	maxRetries := 10
+	checkDeploymentURIPublic := customDeploymentURIPublic
+	if checkDeploymentURIPublic == "" {
+		checkDeploymentURIPublic = customDeploymentURILocal
+	}
+	validateStatusCode, validateErr := h.validateDeploymentURI(checkDeploymentURIPublic, maxRetries)
+	if validateErr != nil {
+		// if it fails we should provide this information as a warning
+		warningMessage = fmt.Sprintf("Validate Custom Deployment URL %s failed. StatusCode=%d, Error: %s", checkDeploymentURIPublic, validateStatusCode, validateErr.Error())
+		h.getKeptnHandler().Logger.Info(warningMessage)
+	}
 
 	// Send finished event - and pass in deploymentURI in case we have a custom defined deploymentURIPublic & Private
 	data, err := h.getFinishedEventDataForSuccess(e, gitVersion,
 		getDeploymentName(deploymentStrategy, false), deploymentStrategy,
-		customDeploymentURIPublic, customDeploymentURILocal)
+		customDeploymentURIPublic, customDeploymentURILocal, warningMessage)
 	if err != nil {
 		h.handleError(ce.ID(), err, keptnv2.DeploymentTaskName, h.getFinishedEventDataForError(e.EventData, err))
 		return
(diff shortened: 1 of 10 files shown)


helm-service/controller/handler_base.go
```diff
@@ -58,6 +61,30 @@ func (h *HandlerBase) existsGeneratedChart(e keptnv2.EventData) (bool, error) {
...
+	// sanity check on values
+	if customDeploymentURIPublic == "null" || customDeploymentURIPublic == "nil" {
+		customDeploymentURIPublic = ""
+	}
+
+	if customDeploymentURIlocal == "null" || customDeploymentURIlocal == "nil" {
+		customDeploymentURIlocal = ""
+	}
+
+	return customDeploymentURIPublic, customDeploymentURIlocal
```
jsoref, line 85: This should make the bot happy. Note that this appears to be the only function where you wrote `URIlocal` instead of `URILocal` and it doesn't match the sibling variable `customDeploymentURIPublic`.
```suggestion
	customDeploymentURILocal := ""

	if chart.Values[CustomDeploymentURIPublicValuesProperty] != nil {
		customDeploymentURIPublic = chart.Values[CustomDeploymentURIPublicValuesProperty].(string)
	}

	if chart.Values[CustomDeploymentURILocalValuesProperty] != nil {
		customDeploymentURILocal = chart.Values[CustomDeploymentURILocalValuesProperty].(string)
	}

	// sanity check on values
	if customDeploymentURIPublic == "null" || customDeploymentURIPublic == "nil" {
		customDeploymentURIPublic = ""
	}

	if customDeploymentURILocal == "null" || customDeploymentURILocal == "nil" {
		customDeploymentURILocal = ""
	}

	return customDeploymentURIPublic, customDeploymentURILocal
```

jsoref (changes requested):

helm-service/controller/deployment_handler.go
```diff
@@ -257,3 +288,38 @@ func (h *DeploymentHandler) getFinishedEventDataForNoDeployment(eventData keptnv
 		EventData: eventData,
 	}
 }
+
+func (h *DeploymentHandler) validateDeploymentURI(deploymentURI string, maxRetries int) (int, error) {
```
bacherfl, line 292: Can you please add a Unit test for this function? Also, I think this func does not have to be a pointer receiver for `DeploymentHandler`, as it does not use any of its properties

helm-service/controller/handler_base.go
```diff
@@ -58,6 +61,30 @@ func (h *HandlerBase) existsGeneratedChart(e keptnv2.EventData) (bool, error) {
 	return helm.DoesChartExist(e, genChartName, h.getConfigServiceURL())
 }
 
+func (h *HandlerBase) getCustomDeploymentURLs(chart *chart.Chart) (string, string) {
```
bacherfl, line 64: please add a unit test for this function. Also in this case, it doesn't have to be a pointer receiver

helm-service/deploy/service.yaml
```diff
@@ -27,7 +27,7 @@ spec:
       serviceAccountName: keptn-helm-service
       containers:
       - name: helm-service
-        image: keptn/helm-service:latest
+        image: grabnerandi/helm-service:0.8.1
```
bacherfl, line 30: please revert this change

releasenotes/releasenotes_V0.8.1.md
```diff
@@ -0,0 +1,146 @@
+# Release Notes 0.8.1
```
bacherfl, line 1: I guess this should not be part of the PR

helm-service/controller/rollback_handler.go
```diff
@@ -26,32 +42,55 @@ func NewRollbackHandler(keptnHandler Handler,
...
+		}
+
+		data := r.getFinishedEventData(e.EventData, keptnv2.StatusSucceeded, e.Result, "Finished rollback", gitVersion)
+		if err := r.sendEvent(ce.ID(), keptnv2.GetFinishedEventType(keptnv2.RollbackTaskName), data); err != nil {
+			r.handleError(ce.ID(), err, keptnv2.RollbackTaskName, r.getFinishedEventDataForError(e.EventData, err))
+			return
+		}
+		r.getKeptnHandler().Logger.Info(fmt.Sprintf("Finished release for service %s in stage %s and project %s", e.Service, e.Stage, e.Project))
+	} else {
+		finishedMessage := fmt.Sprintf("No rollback executed because deployment strategy is %s!", e.Deployment.DeploymentStrategy)
```
bacherfl, line 83: can we add a test case for this scenario in `rollback_handler_test.go`?

helm-service/controller/handler_base.go
warber: Two remarks: 
1. The current implementation will `panic` if the value is of type other than `string` 
2. We should avoid testing **unexported** functions, these are implementation details and could change anytime. As an alternative, you could provide a more generic form of this function (e.g. `func GetValueAsStr(chart chart.Chart , val string) (string,error)` in `helm_helper.go`, test it via unit tests, and make use of it here in `handler_base.go`.

warber (changes requested):

warber:
Incorporation of this PR will be done in the scope of #3766

dyladan:
[CLA assistant check](https://cla.oss.dynalabs.io/keptn/keptn?pullRequest=3700) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla.oss.dynalabs.io/keptn/keptn?pullRequest=3700) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: grabnerandi<br/>:x: web-flow<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.oss.dynalabs.io/check/keptn/keptn?pullRequest=3700) it.</sub>

PR #3752, opened 2021-04-09
CI workflow tweaks

jsoref:
1. the `types` field is not spelled `type` -- this could mean that the CI pull_request case has been unintentionally doing something totally out of character. -- per https://github.com/keptn/keptn/pull/3752#discussion_r614576353 we've decided that we want `labeled`, `unlabeled`, and the normal `types` defaults (`opened`,`synchronize`, `reopened`).
2. the `runs-on` field wants either a string for a GH hosted runner, or an array for a self-hosted. But shellcheck was using an array for a GH hosted.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -10,7 +10,7 @@ on:
     branches:
       - 'master'
       - 'release-*'
-    type: [labeled]
+    types: [opened, synchronize, reopened, labeled, unlabeled]
   workflow_dispatch: # run CI when triggered manually
 
 defaults:
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   lint_shellscripts:
-    runs-on: [ubuntu-latest]
+    runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
     - name: Run ShellCheck


jsoref:
Sigh, the paths-ignore won't work properly for `expected` things. So, scratch all of them.

johannes-b (dismissed):
I needed to double-check the change from `type` to `types`, but looks good (see: https://github.com/marketplace/actions/label-actions) 

LGTM

.github/workflows/CI.yml
```diff
@@ -10,7 +10,7 @@ on:
     branches:
       - 'master'
       - 'release-*'
-    type: [labeled]
+    types: [labeled]
```
johannes-b, line 13: @jsoref I think you found a bug: `type: [labeled]` never worked. Consequently, your change to `types` fixes the behavior but, it just triggers the action when a PR has a label attached. By default, we don't put labels on a PR. 

Can you please change it to: `types: [labeled, unlabeled]`

jsoref: I'm pretty sure you don't want that either. Speaking from experience, the GitHub documentation is immensely frustrating...

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request

I suspect you want this to be omitted entirely.

`type: [labeled]` would result in only processing pull requests when they gain a label.
`type: [unlabeled]` would result in only processing pull requests when they lose a label.
`type: [labeled, unlabeled]` would result in only processing pull requests when they gain or lose a label.

The default is `type: [opened, synchronize, reopened]`.

You could conceivably want to add `labeled` to that list. But, it's more likely that you just want the line removed so you get the default (which is what you have now).

lmk, and I'll adjust.

jsoref:
```suggestion
    types: [opened, synchronize, reopened, labeled, unlabeled]
```

johannes-b: To make it explicit when the action runs, I like your suggestion with: `types: [opened, synchronize, reopened, labeled, unlabeled]`

Can you please adapt your PR accordingly? (I can`t commit your suggestion, because  I don`t get it signed for DCO)

jsoref: @johannes-b: done

johannes-b: Thanks!

johannes-b (approved):

PR #3780, opened 2021-04-14
Feature/3448/bridge oidc login

### LINKED ISSUE #3448: OAuth/OpenID Connect based login for Keptn bridge
**Background**

Keptn bridge relies on basic authentication (or no authentication if credentials are not set). 

It would be beneficial to Keptn users if we can extend authentication & authorization capability with OAuth [1]/OpenID Connect [2]. This will help organisations to allow their users to login to Keptn Bridge, without the need of sharing single set of credentials. Further, we can build some permissions on top of this solution (ex:- View vs Admin users) as future goals.

**Considerations**

Solution should be flexible and provide the ability to integrate with different OAuth/OpenID Connect providers (ex:- Login with GitHub or login with corporate identity server). This means, Keptn Bridge cannot have vendor specific logic or validations. Hence, OAuth/OpenID Connect related functionality should be outsourced to a dedicated (micro)service. 

Furthermore, rather than relying on tokens to establish authenticated & authorised state in Keptn Bridge, it would be beneficial to utilise sessions. This makes Keptn Bridge independent of the token contents. 

**Solution**

Keptn Bridge contains a switch to enable/disable OAuth/OpenID Connect based authentication. This switch also enable session cookies. Following diagram shows the flow of requests & responses

KeptnOAuth

Keptn Bridge should check the existence of OAuth service and also validate responses obtained from it.

For example, following are requests and responses Keptn Bridge can expect from OAuth service,

To obtain authorization request,
```
Request :

   Get <OAuth_Service>/authorization

Response :

  {
   "location": <authorization_request>
  }
```

To validate authorization code & permissions
```
Request :

   POST <OAuth_Service>/token

   {
     "state" : <state>,
     "code" : <authorization_code>
   }

Response :

  {
   "status": "SUCCESS",
   "user" : <USER_IDENTIFIER>
  }
```

OAuth service must implement logic to generate authorization request, obtain tokens for valid authorization codes and validate permission checks based on token contents or user information. Furthermore, authorization request and token request should be linked ideally through `state`.  OAuth service is expected to follow recommended best practices such as use of `PKCE`[3] and JWK validations where applicable. 

**Note on OAuth 2.0 & OpenID Connect**

OAuth 2.0 define a protocol for authorization. OpenID Connect extends OAuth 2.0 and provide authentication capability on top of JWT ID Token. 

[1] - https://tools.ietf.org/html/rfc6749
[2] - https://openid.net/specs/openid-connect-core-1_0.html
[3] - https://tools.ietf.org/html/rfc7636

Kavindu-Dodan:
Fixes #3448

This PR introduce the OAuth base login to Keptn bridge. This needs to be enabled by settings environment variable _OAUTH_ENABLED_ to value _true_

While user identification and authentication is done using OAuth 2.0/OpenID connect, bridge uses cookies to maintain user logged in state.

As explained in issue #3448 , bridge rely on an external service to generate and handle the actual OAuth 2.0/OpenID Connect details. The exact details of the service contract is explained through this comment (including swagger documentation) - https://github.com/keptn/keptn/issues/3448#issuecomment-819535178 

To enable asynchronous discovery data loading at start-up, I had to refactor our server code base. 

Following are new environmental variables and their default values 

- OAUTH_ENABLED : Default to false. 
- OAUTH_DISCOVERY : undefined
- SECURE_COOKIE : Default to false
- TRUST_PROXY : undefined

OAUTH_ENABLED  & OAUTH_DISCOVERY  are mandatory to enable OAuth based login. 

SECURE_COOKIE can be ignored but advisable to use in production. Also, if enabling secure cookies, make sure to add correct number to TRUST_PROXY (ex:- 2) as described at - http://expressjs.com/en/guide/behind-proxies.html


johannes-b:
Hi @Kavindu-Dodan, 
thanks for fixing the misspelling of `Initialising`. 
I assume that the other words are correct. Can you please add them to the file `.github/actions/spelling/expect.txt` - then the spell checker does not complain about it.

bridge/client/app/_components/ktb-user/ktb-user.component.html
```diff
@@ -0,0 +1,6 @@
+<div class="user">
+  <p>User : {{user}}</p>
```
ermin-muratovic, line 2: image

Can we make it `User: {{user}}` as it is with `Keptn version: {{version}}` and Bridge `version: {{bridgeVersion}}` ?

ermin-muratovic (changes requested):

bridge/client/app/_components/ktb-user/ktb-user.component.html
Kavindu-Dodan: Thanks for the review. Corrected with the latest commit

ermin-muratovic (dismissed):
LGTM, although as discussed with @Kavindu-Dodan we need an environment to do an integration test and also consider an error page in the bridge in case the user is unauthorized. That could be checked in a follow-up issue.

Kavindu-Dodan:
@ermin-muratovic  that's absolutely correct. I see three key follow-up tasks here,

1. We need error pages in bridge 

Not dedicated to OAuth login flow, but common, well formatted ideally context aware (ex:- error message can be altered) error pages

2. A test service which implement the contract of the OAuth service. 

For example, an OAuth service which can provide OIDC login flow against GitHub. This can be used for testing and as an example for adopters. 

3. OIDC logout 

We can further extend the solution to provide RP initiated logout [1].

I will try to contribute where possible. 

[1] https://openid.net/specs/openid-connect-rpinitiated-1_0.html

bridge/Dockerfile
```diff
@@ -26,6 +26,13 @@ ENV VERSION="${version}"
 ENV API_URL "http://api-gateway-nginx.keptn.svc.cluster.local"
 ENV API_TOKEN ""
 
+# OAuth and session cookie related configurations
```
agrimmer, line 29: We should move this ENV-variables into deploy/bridge.yaml and the Keptn installer Helm-chart.

Kavindu-Dodan: Thank you @agrimmer . I fixed this with the latest commit.

agrimmer (approved):
LGTM

jsoref:
It looks like some of your previous pushes passed. Also, oddly, some of the commits that failed here passed for me in my fork which is quite puzzling. That said, I think the latest changes are better anyway.

agrimmer (approved):
lgtm

ermin-muratovic (dismissed):
lgtm

bridge/client/app/_components/ktb-user/ktb-user.component.ts
```diff
@@ -0,0 +1,15 @@
...
+  selector: 'ktb-user',
+  templateUrl: './ktb-user.component.html',
+  styleUrls: ['./ktb-user.component.scss']
+})
+export class KtbUserComponent {
+
+  @Input() user;
+
+  logout(): void {
+    window.location.href = '/logout';
```
christian-kreuzberger-dtx, line 13: That's not what I would expect in an Angular App. Could you explain why we are using `window.location.href` here instead of using routes/states to perform a logout?

bridge/client/app/_interceptors/http-error-interceptor.ts
```diff
@@ -1,23 +1,37 @@
...
     return next.handle(request)
       .pipe(
         retryWhen(genericRetryStrategy()),
         catchError((error: HttpErrorResponse) => {
+
+          if (error.status === 401) {
+            if (!this.isReloading) {
+              this.isReloading = true;
+              this.toast.create('Login required. Redirecting to login.');
+              setTimeout(() => window.location.href = '/login', 1000);
```
christian-kreuzberger-dtx, line 29: Also here.

Neither `setTimeout` nor `window.location.href` is something that I would expect in an Angular app.

bridge/client/app/app.module.ts
```diff
@@ -94,6 +94,7 @@ import { KtbSubscriptionComponent } from './_components/ktb-subscription/ktb-sub
 import { DtFilterFieldModule } from '@dynatrace/barista-components/filter-field';
 import { KtbSubscriptionItemComponent } from './_components/ktb-subscription-item/ktb-subscription-item.component';
 import { KtbDeploymentListComponent } from './_components/ktb-deployment-list/ktb-deployment-list.component';
+import {KtbUserComponent} from './_components/ktb-user/ktb-user.component';
```
christian-kreuzberger-dtx, line 97: Please align the formatting with the other imports in this file (I know, we're missing linting in our CI...)

christian-kreuzberger-dtx (changes requested):
I left some comments, please address them

bridge/client/app/_components/ktb-user/ktb-user.component.ts
ermin-muratovic: Because the `/logout` route is not handled by the Angular app, but only by the node app and removes the session in the backend. We could add a route in the Angular app and make an API call there, but I don't know if we really need that?

bridge/client/app/_interceptors/http-error-interceptor.ts
ermin-muratovic: Here we have the same as with the `/logout` route, also the `/login` route is not handled by the Angular app, but for login we only redirect to the SSO handler, which provides the login mask. The timeout is there, to allow the user to read the Toast with the message `'Login required. Redirecting to login.'` and I think setTimeout is still the best solution therefore.

bridge/client/app/_components/ktb-user/ktb-user.component.ts
christian-kreuzberger-dtx: Okay, then let's add this as a comment:
```
// redirect user to the logout page, which is handled by the express app
```
One further question:
What if `bridge` is hosted behind a proxy, e.g., https://myserver.com/my-namespace/keptn/bridge/? In this case navigating to `/logout` would mean that we navigate to https://myserver.com/logout, instead of https://myserver.com/my-namespace/keptn/bridge/logout.

bridge/client/app/_interceptors/http-error-interceptor.ts
christian-kreuzberger-dtx: Thanks for the explanation. I would have expected rxjs timeout here, instead of `setTimeout`. Anyway, let's add a comment here for setTimeout, e.g.:
```
// wait for 1 second such that the user can see the toast notification
```

Kavindu-Dodan: @ermin-muratovic 's explanation is correct.

When approaching the solution, I had two options to implement the service interactions and routing

1. Implement routing in Angular
2. Implement routing in Express

I went ahead with the second option. Justification is that this approach reduced the complexity of the solution. 

If I went ahead with Angular, then I would have to introduce another API layer, which works independent of session cookie. And the login would need to be implemented in three layers - Angular : Express : OAuth Service  (For example, authorization request would need to be obtained by Angular application from Express layer. To serve this, express layer interacts with OAuth service. Same goes for redirect handling).

Regardless,  window.location.href is the desired way to navigate to a non angular route. And since navigation destroy the context, I had to delay the navigation so we can show a toast and make user aware what happens.

bridge/client/app/app.module.ts
Kavindu-Dodan: @christian-kreuzberger-dtx  solved through the latest commit

bridge/client/app/_components/ktb-user/ktb-user.component.ts
Kavindu-Dodan: @christian-kreuzberger-dtx Thanks for pointing this out.

Latest commit address this issue in both angular and express layer

For angular, since we have defined the base-href , I now use Location service [1]

For express, I have a conditional navigation. PREFIX_PATH was introduced to bridge and if it is present, root navigation is altered to PREFIX_PATH/bridge. 

please have a look on the changes. 

[1] - https://angular.io/api/common/Location

christian-kreuzberger-dtx: Nice, thanks

christian-kreuzberger-dtx (approved):
thanks for the changes, approval from my side. @ermin-muratovic you have the last word ;)

ermin-muratovic (approved):
lgtm

PR #3807, opened 2021-04-18
Add new 'env' Helm chart value to add additional environment variables

### LINKED ISSUE #3806: Helm chart should support ability to set HTTP proxy for configuration-service
In a corporate network environment where public internet access is limited and all external traffic must go through a HTTP proxy, some services, like the `configuration-service`, are not able to communicate to GitHub (where the Git upstream repo is hosted) or other external resources.

It's possible to enable access via a HTTP proxy for services, like the `configuration-service`, by adding additional environment variables to the [`Deployment`](https://github.com/keptn/keptn/blob/1e7b9b8276721c58a50ecc440ccd1decb9b352b8/installer/manifests/keptn/charts/control-plane/templates/core.yaml#L552) as below:

```yaml
apiVersion: apps/v1
kind: Deployment
...
metadata:
  name: configuration-service
spec:
...
    spec:
      containers:
      - env:
        # ---
        - name: HTTPS_PROXY
          value: http://proxy:3128
        - name: NO_PROXY
          value: 172.20.0.1,localhost,127.0.0.1
        # ---
        - name: PREFIX_PATH
        - name: POD_NAMESPACE
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
        - name: DATASTORE_URI
          value: mongodb-datastore:8080
        image: docker.io/keptn/configuration-service:0.8.1
...
```

Therefore, it would be beneficial to enable the ability to add additional environment variables to the various services `Deployment` manifests via [chart values](https://github.com/keptn/keptn/blob/1e7b9b8276721c58a50ecc440ccd1decb9b352b8/installer/manifests/keptn/charts/control-plane/values.yaml#L66), to satisfy this use case and potentially others in future.

donovanmuller:
Adds new `env` value for `configurationService` and `bridge` which allows additional environment variables to be added to the `Deployment` manifests. This allows `HTTPS_PROXY` and other environment variables to be passed to the services.

E.g.

```console
$ pwd
xxx/keptn/installer/manifests/keptn

$ helm template . -f values.yaml \
    --set control-plane.configurationService.env.HTTPS_PROXY=https://proxy:3128 \
    --set control-plane.configurationService.env.NO_PROXY=localhost \
    --set control-plane.bridge.env.HTTPS_PROXY=https://proxy:3128 \
    --set control-plane.bridge.env.NO_PROXY=localhost \
    | grep -a3 "HTTPS_PROXY"
              value: "true"
            - name: KEPTN_INSTALLATION_TYPE
              value: "QUALITY_GATES,CONTINUOUS_OPERATIONS"
            - name: HTTPS_PROXY
              value: "https://proxy:3128"
            - name: NO_PROXY
              value: "localhost"
--
--
                  fieldPath: metadata.namespace
            - name: DATASTORE_URI
              value: mongodb-datastore:8080
            - name: HTTPS_PROXY
              value: "https://proxy:3128"
            - name: NO_PROXY
              value: "localhost"

$ # Negative test
$ helm template . -f values.yaml | grep "HTTPS_PROXY"

```

Adding the new value to:

* `configurationService` to solve for use case described in #3806 
* `bridge` for the [version check](https://github.com/keptn/keptn/blob/069dd0f5c7b6f37a3737f4c0c9c7cf07a801b039/bridge/server/api/index.js#L61)

Not sure if any other services would require this but can be added in future.

Resolves #3806

### DIFF
--- a/installer/manifests/keptn/charts/control-plane/templates/core.yaml
+++ b/installer/manifests/keptn/charts/control-plane/templates/core.yaml
@@ -204,6 +204,10 @@ spec:
               value: "{{ .Values.bridge.showApiToken.enabled }}"
             - name: KEPTN_INSTALLATION_TYPE
               value: "{{ .Values.bridge.installationType |default (print "QUALITY_GATES,CONTINUOUS_OPERATIONS") }}"
+            {{- range $key, $value := .Values.bridge.env }}
+            - name: {{ $key }}
+              value: {{ $value | quote }}
+            {{- end }}
           envFrom:
             - secretRef:
                 name: bridge-credentials
@@ -558,6 +562,10 @@ spec:
                   fieldPath: metadata.namespace
             - name: DATASTORE_URI
               value: mongodb-datastore:8080
+            {{- range $key, $value := .Values.configurationService.env }}
+            - name: {{ $key }}
+              value: {{ $value | quote }}
+            {{- end }}
           ports:
             - containerPort: 8080
           resources:


christian-kreuzberger-dtx:
Thanks for the PR. 

1. Is this considered best practice in Kubernetes/helm? A quick search reveals to me that there is multiple ways to handle additional env variables, e.g., [Wordpress helm chart](https://github.com/helm/charts/blob/7e45e678e39b88590fe877f159516f85f3fd3f38/stable/wordpress/values.yaml#L128-L134), [MySQL helm chart](https://github.com/helm/charts/blob/7e45e678e39b88590fe877f159516f85f3fd3f38/stable/mysql/values.yaml#L65-L68) 
2. Could you please provide an example in a comment on this PR, so we can add it to our documentation at https://keptn.sh/docs/0.8.x/operate/advanced_install_options/ (see https://github.com/keptn/keptn.github.io for source code).

christian-kreuzberger-dtx:
Yet another "alternative" approach to this provided by harbor: https://github.com/goharbor/harbor-helm/blob/2a1742269ea4e1e7e68587c3c753896264fa582f/values.yaml#L349-L358

donovanmuller:
@christian-kreuzberger-dtx I actually logged a similar issue on the VPA chart (https://github.com/FairwindsOps/charts/issues/425) and I used the same implementation that team chose. If there are better ways, I'm not against it at all. As you are seeing, there seems to be varying ways to achieve this.

I chose to allow generic environment variables over specific proxy values in the values file as it might come in handy in future to define other non Keptn related environment variables but I also see the value in applying YAGNI and constraining this to explicit proxy settings. I have no strong feelings with either approach.

You asked for an example, is the one in the PR description ok?

```
$ helm template . -f values.yaml \
    --set control-plane.configurationService.env.HTTPS_PROXY=https://proxy:3128 \
    --set control-plane.configurationService.env.NO_PROXY=localhost \
    --set control-plane.bridge.env.HTTPS_PROXY=https://proxy:3128 \
    --set control-plane.bridge.env.NO_PROXY=localhost \
...
```

I'll wait for a decision on how to proceed given the above two considerations before adding any docs 👍

christian-kreuzberger-dtx:
Thanks for the example, I must have overlooked it in the PR :D 

I agree, the generic approach of allowing env variables is (in the current situation) the one that provides the most benefits (less maintenance for us, most flexibility for end-users).

Regarding naming (`env` vs. `extraEnv` vs. `extraEnvVars`) I'm pinging @thschue and @TannerGabriel , maybe they have an opinion :) 
I do like the touch of `... --set control-plane.configurationService.env.HTTPS_PROXY=https://proxy:3128  ` though.

donovanmuller:
@thschue @TannerGabriel do you have any strong opinions on the naming of the the new value?

Would love to get this into the `0.8.2` release if possible?

christian-kreuzberger-dtx:
> Would love to get this into the 0.8.2 release if possible?
Sorry, this won't make it into 0.8.2.

Nevertheless, if we don't hear back anyone, I assume they are okay with the solution, so I'll approve the PR.

christian-kreuzberger-dtx (approved):
LGTM

christian-kreuzberger-dtx:
Hi @donovanmuller , after talking to @johannes-b, @thschue and @DTMad we came to the conclusion that introducing `env` might be a viable option for certain things (e.g., temporarily overwriting a variable for debug/troubleshooting purposes, demo-ing a feature, etc...), but it might not be a good match for proxy settings.

In other words, typical use-cases (like setting a proxy setting) should be directly supported in our values.yaml, similar to the approach that [Harbor](https://github.com/goharbor/harbor-helm/blob/2a1742269ea4e1e7e68587c3c753896264fa582f/values.yaml#L349-L358) uses.
 
Setting `env` should be more of a "I already know what I'm doing, don't prevent me from doing it".

**In other words**, lets get this change merged, but lets not use it for setting proxy configuration :D

PR #3813, opened 2021-04-19
Set error to null when re-routing to project to hide previously shown…

### LINKED ISSUE #3804: "Project does not exist" pane does not vanish when navigating to a existing project
**Environment**
* 0.8.1 Bridge

**Affected Component**
* [ ] Bridge

**Describe the bug**
Project doesn't exist pane does not vanish when navigating to a existing project

**To Reproduce**

1. Navigate to not existing project e.g. /bridge/project/sockshoppe
    The screen shows, that the project does not exist
2. Click on 'Choose project' and select an existing project
3. The screen shows now both, the 'Project does not exist' and the project overview

**Expected behavior**
The 'Project does not exists' pane should vanish and only the project overview is visible

**Screenshots**
image

laneli:
Fixes #3804 

### DIFF
--- a/bridge/client/app/project-board/project-board.component.ts
+++ b/bridge/client/app/project-board/project-board.component.ts
@@ -29,7 +29,6 @@ export class ProjectBoardComponent implements OnInit, OnDestroy {
     this.route.params
       .pipe(takeUntil(this.unsubscribe$))
       .subscribe(params => {
-        this.error = null;
         if (this.route.snapshot.url[0].path === 'trace') {
           this.dataService.loadTracesByContext(params.shkeptncontext);
           this.dataService.traces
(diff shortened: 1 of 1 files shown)


bridge/client/app/project-board/project-board.component.ts
```diff
@@ -29,6 +29,7 @@ export class ProjectBoardComponent implements OnInit, OnDestroy {
     this.route.params
       .pipe(takeUntil(this.unsubscribe$))
       .subscribe(params => {
+        this.error = null;
```
ermin-muratovic, line 32: Maybe it would be better to put it as `else` statement at https://github.com/keptn/keptn/blob/a2dd0b1b971ccfaa8cac00e8fc9c3139c7ea7d32/bridge/client/app/project-board/project-board.component.ts#L66-L67, so in case that the "Retry"-Button for some reason once returns `null` and then later a `project`, the error is also set to `null`

ermin-muratovic (changes requested):

ermin-muratovic (approved):
lgtm

PR #3829, opened 2021-04-21
#3826 Run pull_request CI workflow regardless of type

### LINKED ISSUE #3826: CI workflow tweaks cause multiple CI runs when opening a Pull Request
This is a follow-up of https://github.com/keptn/keptn/pull/3752 - pinging @jsoref  on this as well.

I noticed that CI runs are sometimes duplicated, e.g., when opening a PR:
image

image

I'm pretty sure it has to do with the following line:
https://github.com/keptn/keptn/blob/dde0b7171afc3c34edb165761055627a097d84f6/.github/workflows/CI.yml#L13

Any opinions in how to fix this?
Maybe we don't need `synchronize` or `opened`?

christian-kreuzberger-dtx:
Fixes #3826

As discussed in #3826, I have completely removed `types` in the CI workflow to not have duplicate runs when opening a PR with a label.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -10,7 +10,6 @@ on:
     branches:
       - 'master'
       - 'release-*'
-    types: [opened, synchronize, reopened, labeled, unlabeled]
   workflow_dispatch: # run CI when triggered manually
 
 defaults:


johannes-b (approved):
lgtm

PR #3832, opened 2021-04-21
Cleanup repo, adapt dependabot

christian-kreuzberger-dtx:
We don't use `upgrader` at all anymore, this is all handled via the CLI and the respective helm-charts, hence I have removed this directory.
Also, writeManifest.sh is no longer in use (this is now all handled via a GH action).

Last but not leased, I added secret-service to dependabot configuration


warber (approved):
:+1:

PR #3839, opened 2021-04-22
Bug/3815/evaluation nan shown as number

### LINKED ISSUE #3815: Bridge showing a NaN as a number
When I trigger an evaluation and the data behind this evaluation is actually responding with a "NaN" the bridge shows a number (0) which took me quite a while to find the error.

image

image

The bridge should show that the data has not been retrieved correctly instead of showing the number 0.

### LINKED ISSUE #3838: Cannot read property 'keptnContext' of undefined
**Environment**
* Keptn Version 0.8.1

**Affected Component**
* [x ] Bridge

**Describe the bug**
image
image

**To Reproduce**
In Bridge
* Navigate to /project/<projectname>
* Click on stage
* Error shows up in console
* Stage info stays empty

**Expected behavior**
When keptn context is undefined, handle this accordingly

laneli:
Fixes #3815
Fixes #3838

### DIFF
--- a/bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.ts
+++ b/bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.ts
@@ -350,7 +350,7 @@ export class KtbEvaluationDetailsComponent implements OnInit, OnDestroy {
               x: x,
               y: index,
               z: s.indicatorResult.score,
-              color: this._evaluationColor[s.indicatorResult.status]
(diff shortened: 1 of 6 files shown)


Kirdock:
We should show an error-icon if `success` is `false`. Additionally on hover there should appear an overlay, which shows the error-message

laneli:
image

I re-worked the view. Now there is an icon shown, the error message is in an overlay like with the violated criteria. Additionally the criteria is shown, but only informative. There is no error styling applied there.

According the heatmap, I would propose to keep the "grey" state as semantically the evaluation neither failed or passed.

Kirdock (approved):
LGTM

PR #3840, opened 2021-04-22
Fixed issue with existing target url reachability check

warber:
Added more reliable check for http endpoint reachability.

### DIFF
--- a/jmeter-service/main.go
+++ b/jmeter-service/main.go
@@ -232,6 +232,10 @@ func runWorkload(serviceURL *url.URL, testInfo *TestInfo, workload *Workload, lo
 }
 
 func checkEndpointAvailable(timeout time.Duration, serviceURL *url.URL) error {
+	if serviceURL == nil {
+		return fmt.Errorf("url to check for reachability is nil")
+	}
+
 	client := http.Client{
 		Timeout: timeout,
 	}
--- a/jmeter-service/main.go
+++ b/jmeter-service/main.go
@@ -231,11 +231,11 @@ func runWorkload(serviceURL *url.URL, testInfo *TestInfo, workload *Workload, lo
 	return executeJMeter(testInfo, workload, resultDirectory, serviceURL, resultDirectory, breakOnFunctionalIssues, logger)
 }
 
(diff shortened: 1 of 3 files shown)


bacherfl (approved):

bacherfl (approved):

christian-kreuzberger-dtx (changes requested):
This is not the same. `LookupIP` checks if the DNS lookup works, `client.Get(serviceURL.String())` is trying to access the host via HTTP

jmeter-service/main.go
```diff
@@ -232,6 +231,18 @@ func runWorkload(serviceURL *url.URL, testInfo *TestInfo, workload *Workload, lo
...
 
+func checkEndpointAvailable(timeout time.Duration, serviceURL *url.URL) error {
+	if serviceURL == nil {
+		return fmt.Errorf("url to check for reachability is nil")
+	}
+
+	client := http.Client{
+		Timeout: timeout,
+	}
+	_, err := client.Get(serviceURL.String())
```
christian-kreuzberger-dtx, line 242: It is not guaranteed that `GET serviceURL` (e.g, `GET HTTP://carts.sockshop-dev:80/`) should work. There are plenty of services that don't return a positive HTTP result when accessing `/`.

jmeter-service/main.go
```diff
@@ -100,8 +100,7 @@ func runTests(event cloudevents.Event, shkeptncontext string, data keptnv2.TestT
 	var res bool
 	healthCheckWorkload, err = getWorkload(jmeterconf, TestStrategy_HealthCheck)
 	if healthCheckWorkload != nil {
-
-		_ , err := net.LookupIP(serviceUrl.String())
+		err := checkEndpointAvailable(5*time.Second, serviceUrl)
```
christian-kreuzberger-dtx, line 103: This introduces an unexpected change of behaviour. Instead of looking up the IP address of the service, we are checking whether the service is reachable via HTTP.

christian-kreuzberger-dtx (approved):
LGTM

Issue #3781, opened 2021-04-14
 Service account `keptn-configuration-service` does not need full permissions on secret management


johannes-b:
The service account `keptn-configuration-service` does not need all permissions (create, get, delete, update); it just needs GET permissions because no other usage is found in: https://github.com/keptn/keptn/blob/0.8.1/configuration-service/common/git.go#L45

**Task**
* Create a role with just `get` permissions
* Bind the role to the service account: `keptn-configuration-service`

PR #3842, opened 2021-04-22
#3781 Rename rolebinding to get-secrets

christian-kreuzberger-dtx:
An error was introduced in https://github.com/keptn/keptn/pull/3782 whereas the roleRef of a rolebinding was changed.

This leads to a problem when upgrading from an older version, e.g.:
```
Start upgrading Helm Chart keptn in namespace keptn
Error: Could not complete Keptn upgrade: Error when installing/upgrading Helm Chart keptn in namespace keptn: cannot patch "keptn-configuration-service-manage-secrets" with kind RoleBinding: RoleBinding.rbac.authorization.k8s.io "keptn-configuration-service-manage-secrets" is invalid: roleRef: Invalid value: rbac.RoleRef{APIGroup:"rbac.authorization.k8s.io", Kind:"Role", Name:"keptn-get-secrets"}: cannot change roleRef 
```

Together with @bacherfl we decided that renaming the rolebinding should fix this problem (and this also ensures that the rolebinding is properly named, in this case).

### DIFF
--- a/installer/manifests/keptn/charts/control-plane/templates/rbac.yaml
+++ b/installer/manifests/keptn/charts/control-plane/templates/rbac.yaml
@@ -315,9 +315,9 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  name: keptn-configuration-service-manage-secrets
+  name: keptn-configuration-service-get-secrets
   labels:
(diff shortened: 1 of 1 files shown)


johannes-b (commented):
lgtm, but I would like to ask @bacherfl for an additional approval

bacherfl (approved):

PR #3867, opened 2021-04-25
Set HTTP_SSL_VERIFY based on remoteControlPlane.api.apiValidateTls

### LINKED ISSUE #3865: Helm/JMeter charts do not honour 'remoteControlPlane.api.apiValidateTls: false' in template
**Environment**

* Client OS (e.g., Linux, macOS, Windows): MacOS 11.2.3
* Keptn Version (`keptn version`): `Keptn CLI version: 0.8.2, Keptn cluster version: 0.8.2`
* Kubernetes Cloud Provider (e.g., GKE, AKS): EKS
* Kubernetes version (`kubectl version`): 1.19

**Affected Component**

* [ ] Docs
* [ ] CLI
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [x] Helm
* [ ] Istio
* [ ] Bridge
* [ ] Approval
* [ ] Datastore
* [ ] REST API
* [x] jmeter

**Describe the bug**

The `remoteControlPlane.api.apiValidateTls` value in both the Helm and JMeter Helm charts is not actually honoured. Only the value `"true"` will ever be rendered despite the value being  set to `false` or `true`. I.e. the `distributor` container's `HTTP_SSL_VERIFY` environment variable can never set to `"false"` via the Helm chart values.

**To Reproduce**

If you set the `helm-service` chart's value as below:

```yaml
remoteControlPlane:
  enabled: true
  api:
    protocol: "https"
    hostname: "keptn.replaceme.com"
    apiValidateTls: false
    token: "..."
```

the rendered manifests will always have the `distributor` container's `HTTP_SSL_VERIFY` environment variable set to `"true"`.

**Expected behavior**

Setting the `remoteControlPlane.api.apiValidateTls` value to `false` should render the `HTTP_SSL_VERIFY` environment variable's value as `"false"`.

**Additional context**

The reason this happens is because the conditional on `remoteControlPlane.api.apiValidateTls` in the template evaluates the  _value_ and if it's `true` then renders the `true` value but if it's `false` it renders the `default "true"`:

https://github.com/keptn/keptn/blob/069dd0f5c7b6f37a3737f4c0c9c7cf07a801b039/helm-service/chart/templates/deployment.yaml#L118-L119

Without changing the template, a solution would be to use a string value of `remoteControlPlane.api.apiValidateTls: "false"`, however this is not currently possible due to the JSON schema only allowing a boolean type:

https://github.com/keptn/keptn/blob/069dd0f5c7b6f37a3737f4c0c9c7cf07a801b039/helm-service/chart/values.schema.json#L40-L41

donovanmuller:
Sets the value of `HTTP_SSL_VERIFY` based on the string representation of the boolean value set with `.Values.remoteControlPlane.api.apiValidateTls`.

E.g. usage:

```console
$ pwd
.../keptn/jmeter-service/chart

$ # Still works with apiValidateTls=true
$ helm template . -f values.yaml \
  --set remoteControlPlane.enabled=true \
  --set remoteControlPlane.api.hostname=keptn.replaceme.com \
  --set remoteControlPlane.api.token=xxx \
  --set remoteControlPlane.api.apiValidateTls=true \
    | grep -a1 "HTTP_SSL_VERIFY"
              value: "xxx"
            - name: HTTP_SSL_VERIFY
              value: "true"

$ # Now works with apiValidateTls=false
$ helm template . -f values.yaml \
  --set remoteControlPlane.enabled=true \
  --set remoteControlPlane.api.hostname=keptn.replaceme.com \
  --set remoteControlPlane.api.token=xxx \
  --set remoteControlPlane.api.apiValidateTls=false \
    | grep -a1 "HTTP_SSL_VERIFY"
              value: "xxx"
            - name: HTTP_SSL_VERIFY
              value: "false"
```

Resolves #3865

### DIFF
--- a/helm-service/chart/templates/deployment.yaml
+++ b/helm-service/chart/templates/deployment.yaml
@@ -116,7 +116,8 @@ spec:
             - name: KEPTN_API_TOKEN
               value: "{{ .Values.remoteControlPlane.api.token }}"
             - name: HTTP_SSL_VERIFY
-              value: "{{ .Values.remoteControlPlane.api.apiValidateTls | default "true" }}"
+              {{- $apiValidateTls := .Values.remoteControlPlane.api.apiValidateTls | ternary "true" "false" }}
+              value: "{{ $apiValidateTls }}"
             {{- end }}
 
       {{- with .Values.nodeSelector }}
--- a/jmeter-service/chart/templates/deployment.yaml
+++ b/jmeter-service/chart/templates/deployment.yaml
@@ -84,7 +84,8 @@ spec:
             - name: KEPTN_API_TOKEN
               value: "{{ .Values.remoteControlPlane.api.token }}"
             - name: HTTP_SSL_VERIFY
-              value: "{{ .Values.remoteControlPlane.api.apiValidateTls | default "true" }}"
+              {{- $apiValidateTls := .Values.remoteControlPlane.api.apiValidateTls | ternary "true" "false" }}
+              value: "{{ $apiValidateTls }}"
             {{- end }}
 
       {{- with .Values.nodeSelector }}


christian-kreuzberger-dtx (approved):
Thanks for the PR, looks good to me!

thschue:
lgtm

PR #3873, opened 2021-04-26
#3843 Fixing CD integration test

### LINKED ISSUE #3843: Integration Tests for continuous-delivery should fail when delivery fails
Right now integration tests for continuous-delivery didn't fail, which lead to an undetected error which we only discovered during manual testing of the 0.8.2 release.

## Example

image (13)
image (14)

## Details

It seems that checking if carts-db has been deployed does not not fail for sockshop-staging:
https://github.com/keptn/keptn/blob/069dd0f5c7b6f37a3737f4c0c9c7cf07a801b039/test/test_new_artifact.sh#L15-L17

In addition, carts was also not found in sockshop-staging, which should be checked here:
https://github.com/keptn/keptn/blob/069dd0f5c7b6f37a3737f4c0c9c7cf07a801b039/test/utils/trigger_delivery_sockshop.sh#L41-L44
and
https://github.com/keptn/keptn/blob/069dd0f5c7b6f37a3737f4c0c9c7cf07a801b039/test/utils.sh#L502-L538

## Actions

- [ ] Ensure integration tests detect when an error like this happens, and make sure to exit

christian-kreuzberger-dtx:
Fixes #3843

* made sure that errors are logged using `print_error` rather than just `echo` (`print_error`does an `echo "::error::...."` in the background)
* added verify steps that were missing in several places
* added echo statements that help troubleshooting in the future
* added filename and line number for integration test errors (print_error function)

**Example (from current master branch):**
image

image

**Example with release-0.8.2 branch**
https://github.com/keptn/keptn/actions/runs/785731421

### DIFF
--- a/test/utils.sh
+++ b/test/utils.sh
@@ -285,7 +285,7 @@ function verify_image_of_deployment() {
   if [[ "$CURRENT_IMAGE_NAME" == "$IMAGE_NAME" ]]; then
     echo "Found image ${CURRENT_IMAGE_NAME} in deployment ${DEPLOYMENT} in namespace ${NAMESPACE}"
   else
-    echo "ERROR: Found image ${CURRENT_IMAGE_NAME} but expected ${IMAGE_NAME}  in deployment ${DEPLOYMENT} in namespace ${NAMESPACE}"
+    print_error "ERROR: Found image ${CURRENT_IMAGE_NAME} but expected ${IMAGE_NAME}  in deployment ${DEPLOYMENT} in namespace ${NAMESPACE}"
     exit 1
   fi
 }
@@ -316,7 +316,7 @@ function wait_for_deployment_in_namespace() {
   done
 
   if [[ $RETRY == "$RETRY_MAX" ]]; then
-    echo "Error: Could not find deployment ${DEPLOYMENT} in namespace ${NAMESPACE}"
+    print_error "Could not find deployment ${DEPLOYMENT} in namespace ${NAMESPACE}"
     exit 1
   fi
 }
@@ -346,7 +346,7 @@ function wait_for_deployment_with_image_in_namespace() {
   done
 
   if [[ $RETRY == "$RETRY_MAX" ]]; then
-    echo "Error: Could not find deployment ${DEPLOYMENT} in namespace ${NAMESPACE}"
+    print_error "Could not find deployment ${DEPLOYMENT} in namespace ${NAMESPACE}"
     exit 1
   fi
 }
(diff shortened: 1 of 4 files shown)


test/utils.sh
```diff
@@ -507,32 +531,61 @@ function verify_sockshop_deployment() {
...
+    # verify that a carts deployment is up and running
+    wait_for_deployment_with_image_in_namespace "carts" "${PROJECT}-${STAGE}" "${ARTIFACT_IMAGE}:${ARTIFACT_IMAGE_TAG}"
+    verify_test_step $? "Deployment ${ARTIFACT_IMAGE}:${ARTIFACT_IMAGE_TAG} carts not up in ${PROJECT}-${STAGE}, exiting ..."
+
+    # verify that a carts pod is up and running
+    verify_pod_in_namespace "carts" "${PROJECT}-${STAGE}"
+    verify_test_step $? "Pod carts not found, exiting ..."
+
+    # verify image name of carts deployment
+    verify_image_of_deployment "carts" "${PROJECT}-${STAGE}" "${ARTIFACT_IMAGE}:$ARTIFACT_IMAGE_TAG"
```
bacherfl, line 579: do we still need this check if we have verified the deployment/image with `wait_for_deployment_with_image_in_namespace` above?

test/utils.sh
```diff
@@ -507,32 +531,61 @@ function verify_sockshop_deployment() {
...
+    # verify that a blue-green carts deployment is up and running
     wait_for_deployment_with_image_in_namespace "carts-primary" "${PROJECT}-${STAGE}" "${ARTIFACT_IMAGE}:${ARTIFACT_IMAGE_TAG}"
+    verify_test_step $? "Deployment carts ${ARTIFACT_IMAGE}:${ARTIFACT_IMAGE_TAG} not up in ${PROJECT}-${STAGE}, exiting ..."
+
+    # verify that a blue-green carts pod is up and running
     verify_pod_in_namespace "carts-primary" "${PROJECT}-${STAGE}"
     verify_test_step $? "Pod carts-primary not found, exiting ..."
+
+    # verify image name of carts-primary deployment
+    verify_image_of_deployment "carts-primary" "${PROJECT}-${STAGE}" "${ARTIFACT_IMAGE}:$ARTIFACT_IMAGE_TAG"
```
bacherfl, line 567: do we still need this check if we have verified the deployment/image with `wait_for_deployment_with_image_in_namespace` above?

christian-kreuzberger-dtx: I double checked, you are correct, we don't need this here.

test/utils.sh
christian-kreuzberger-dtx: I double checked, you are correct, we don't need this here.

bacherfl (approved):

PR #3881, opened 2021-04-27
Feature/3681/implement task delay

### LINKED ISSUE #3681: Implementation of a waiting functionality to trigger a task after x minutes
**Context:** A remediation sequence triggers an action that should resolve a problem. However, the effect of the action may not be visible immediately. Consequently, it is required to wait for a few minutes before triggering the next task of a rem. sequence. 

*Option 1:*

```
- name: "production" 
  sequences: 
  - name: remediation
      tasks:
      - name: get-action 
      - name: action
      - name: wait
        properties:
        - time: "10m"
      - name: evaluation
```

*Option 2:*

```
- name: "production" 
  sequences: 
  - name: remediation
      tasks:
      - name: get-action 
      - name: action
      - name: evaluation
        triggeredAfter: "10m"
        - properties:
```

Based on the above examples, the sequence waits 10 minutes before triggering the `evaluation`. 

❔ Implementing a separate service for just calling the wait function is maybe over-engineered. Can we integrate the functionality into an existing service (shipyard-controller) OR should we model it as an extra property that is controlled by the shipyard controller? 

**Decision:**  We go with option 2 for implementing the waiting functionality.

**Task:**
- Implement the chosen approach in the current shipyard-controller
- Provide unit tests for this new functionality
- Implement an integration test that verifies this use-case

**Definition of Done:**
- [ ] Spec is up-dated
- [ ] Unit test and integration test cover the new functionality
- [ ] The shipyard controller waits for x minutes before triggering the tasks

:shield: This story will also be used to do major refactorings of the *ShipyardController*

warber:
closes #3681 

also included:
* refactorings, and introduction of eventdispatcher
* replacement of logger with global logger


test/test_delayed_tasks.sh
```diff
@@ -0,0 +1,74 @@
...
+
+echo "Creating a new service"
+keptn create service ${SERVICE} --project ${PROJECT}
+verify_test_step $? "keptn create service ${SERVICE} failed"
+sleep 2
+
+echo "Sending trigger echosequence event"
+keptn_context_id=$(send_event_json ./test/assets/trigger_echosequence_event_2.json)
+
+echo "Wating for sequence triggered event"
```
jsoref, line 59:
```suggestion
echo "Waiting for sequence triggered event"
```

test/test_delayed_tasks.sh
```diff
@@ -0,0 +1,74 @@
...
+taskTriggeredEvent=$(get_event_with_retry sh.keptn.event.echo.triggered "${keptn_context_id}" "${PROJECT}")
+taskTriggeredEventTime=$(echo "$taskTriggeredEvent" | jq -r '.time')
+taskTriggeredEventTimeUNIX=$(date -d "$taskTriggeredEventTime" +%s)
+
+diff="$(($taskTriggeredEventTimeUNIX-$sequenceTriggeredEventTimeUNIX))"
+
+if (( $diff < 20 )); then
+  print_error "Test failed. Task was triggered too early"
+  exit 2
+fi
```
jsoref, line 74: Could you add the missing end of line here (and as applicable throughout)?

bacherfl (approved):
lgtm, @christian-kreuzberger-dtx please also have a look

shipyard-controller/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Use the official Golang image to create a build artifact.
 # This is based on Debian and sets the GOPATH to /go.
-FROM golang:1.13.7-alpine as builder
+FROM golang:1.16.3-alpine as builder
```
christian-kreuzberger-dtx, line 3: Why this udpate?

shipyard-controller/go.mod
```diff
@@ -1,33 +1,20 @@
 module github.com/keptn/keptn/shipyard-controller
 
-go 1.13
+go 1.15
```
christian-kreuzberger-dtx, line 3: Why go 1.15?

shipyard-controller/Dockerfile
bacherfl: we needed to upgrade because one of the dependencies of https://github.com/benbjohnson/clock requires at least 1.15

christian-kreuzberger-dtx: Thanks Florian! We should not have multiple Go versions across our services, so we should specify which one we want to have (for the next couple of months).
We have a separate issue for updating all our services to Go 1.15 (1.16 respectively): https://github.com/keptn/keptn/issues/2936 - @warber we need to up the priority on that, and switch to 1.15 or 1.16 for all services.

christian-kreuzberger-dtx (approved):
As discussed, we need to up the priority for https://github.com/keptn/keptn/issues/2936 to change Go Versions to 1.15/1.16 on all services, not just this one.
Other then that, LGTM

PR #3885, opened 2021-04-27
#3884 Quote env var values

### LINKED ISSUE #3884: Keptn install fails on master
The current state of the master branch cannot be installed because there seems to be an error in the helm chart. Currently the installation fails with the following error message:

```
Error: Could not complete Keptn installation: Error when installing/upgrading Helm Chart keptn in namespace keptn-test: Deployment in version "v1" cannot be handled as a Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Env: []v1.EnvVar: v1.EnvVar.Value: ReadString: expects " or n, but found f, error found in #10 byte of ...|,"value":false},{"na|..., bigger context ...|H","value":null},{"name":"OAUTH_ENABLED","value":false},{"name":"OAUTH_DISCOVERY","value":null},{"na|... 
```

bacherfl:
Closes #3884

### DIFF
--- a/installer/manifests/keptn/charts/control-plane/templates/core.yaml
+++ b/installer/manifests/keptn/charts/control-plane/templates/core.yaml
@@ -205,13 +205,13 @@ spec:
             - name: KEPTN_INSTALLATION_TYPE
               value: "{{ .Values.bridge.installationType |default (print "QUALITY_GATES,CONTINUOUS_OPERATIONS") }}"
             - name: PREFIX_PATH
-              value: {{ .Values.prefixPath }}
+              value: "{{ .Values.prefixPath }}"
             - name: OAUTH_ENABLED
-              value: {{ .Values.bridge.oauth.enabled }}
+              value: "{{ .Values.bridge.oauth.enabled }}"
             - name: OAUTH_DISCOVERY
-              value: {{ .Values.bridge.oauth.discovery }}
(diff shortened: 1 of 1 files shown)


installer/manifests/keptn/charts/control-plane/templates/core.yaml
```diff
@@ -205,13 +205,13 @@ spec:
...
+              value: "{{ .Values.bridge.oauth.enabled }}"
             - name: OAUTH_DISCOVERY
-              value: {{ .Values.bridge.oauth.discovery }}
+              value: "{{ .Values.bridge.oauth.discovery }}"
             - name: SECURE_COOKIE
-              value: {{ .Values.bridge.oauth.secureCookie }}
+              value: "{{ .Values.bridge.oauth.secureCookie }}"
             # Correlates to trust proxy number of hops as defined at http://expressjs.com/en/guide/behind-proxies.html
             - name: TRUST_PROXY
               value: {{ .Values.bridge.oauth.trustProxy }}
```
christian-kreuzberger-dtx, line 217: do we also need to quote trust_proxy?

christian-kreuzberger-dtx (approved):
LGTM

Issue #3760, opened 2021-04-13
Click on deployment opens Service screen


johannes-b:
**User story:** As a user, I would like to get to the details of a service when clicking the deployment in a stage. Example: When clicking `carts:0.9.11`, the service screen opens with the `carts:0.9.11` pre-selected. 

![image](https://user-images.githubusercontent.com/729071/114375058-96079600-9b84-11eb-92b9-51d8839fa8d8.png)

--> click opens: (Please note, the two mocks are not in sync - the `production` stage should be selected) 

![image](https://user-images.githubusercontent.com/729071/113976146-ec9d6900-9840-11eb-9a1e-5cb0216207e7.png)

**Definition of Done:**
- [ ] From the environment screen, I can navigate to a specific deployment by clicking on it.

PR #3913, opened 2021-04-29
Link deployed image to stage in services

laneli:
Implements #3760 

image


Kirdock:
This should also work for quality-gates only use cases. Note that not every service in the overview has an image.
`service.deploymentContext` is the `contextId` of the latest `deployment.finished` or `evaluation.finished` sequence. It could also be that the service is shown in the stage but still does not have an evaluation or a deployment.

Kirdock (approved):
LGTM

PR #3920, opened 2021-04-29
#3019 added golang ci lint

### LINKED ISSUE #3019: Introduce golangci-lint into build chain
## Action Item
As we want to switch from `golint` to `golangci-lint` we need to adapt our build chain.
Besides the linters which are enabled by default, the following linters shall be enabled as well:

`bodyclose, exportloopref, funlen, gocyclo, gosec, nestif, unparam`

## Definition of Done
- [ ] golangci-lint is used on each PR to analyze the code changes contained in the PR
- [ ] current golang-lint execution is removed

## Resources
[golangci-lint-action](https://github.com/golangci/golangci-lint-action)
[reviewdog/golangci](https://github.com/reviewdog/action-golangci-lint)

christian-kreuzberger-dtx:
Fixes #3019

I also disabled spellchecks for .reviewdog.yml, as I believe this is a file that does not need to be spell checked but will regularly introduce false positive spelling errors.

Example:
image

### DIFF
--- a/.github/actions/spelling/excludes.txt
+++ b/.github/actions/spelling/excludes.txt
@@ -18,3 +18,4 @@ ignore$
 \.sum$
 \.xlf$
 _test\.go$
+.reviewdog.yml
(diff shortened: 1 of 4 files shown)


warber (approved):
:+1:

PR #3924, opened 2021-04-30
OIDC RP Initiated logout 

### LINKED ISSUE #3880: Extend OAuth/OpenID Connect solution to support logout
**Background**

#3448 focused on supporting user login using OAuth/OpenID Connect. #3780 provided Keptn bridge interface and API contract to implement this.

Logout implemented through #3780 merely focus on ending bridge session. From OpenID Connect perspective, it is desired to see a logout for Authorization Server (aka OpenID Provider - OP)

**Considerations**

Similar to #3448 , the solution must be vendor neutral. From Keptn bridge point of view, we must only fulfil an API contract, facilitated by a supportive service. 

**Solution**

OIDC RP initiated logout [1] is the most straightforward implementation bridge can easily support. This can be built on top of the OAuth service proposed trough #3448. 

Logout interaction can be simplified as below

1. If OAuth service support logout, it can publish through discovery document (i.e - logout_endpoint ) 
2. If logout at OAuth service is available, when bridge perform the logout, bridge can invoke the logout_endpoint. 
3. OAuth service can then generate the logout request for the specific configurations (Note - id_token_hint can be an optional input)
4. Bridge simply redirect the user agent to the specific logout request 

Specification recommends to utilise id_token_hint [1] and for the first iteration we can support this parameter to be stored against user session in bridge. This makes interactions with OAuth service stateless (except for state usage in login flow).

[1] https://openid.net/specs/openid-connect-rpinitiated-1_0.html

Kavindu-Dodan:
The PR fixes #3880

PR contains the implementation of the contract as explained by the solution proposal - [Proposal](https://github.com/keptn/keptn/issues/3880#issuecomment-829871496)

As a summary,

- Bridge checks for rp_logout support status when reading OAuth service discovery
- Bridge stores logout_hint if present in the token decision response to user session
- When user logout, bridge consumes the rp_logout endpoint to obtain the required redirection location

### DIFF
--- a/bridge/server/user/oauth.js
+++ b/bridge/server/user/oauth.js
@@ -1,12 +1,18 @@
 const express = require('express');
 const router = express.Router();
 const axios = require('axios');
-const sessionAuthentication = require('./session').setAuthenticatedPrincipal;
+const sessionAuthentication = require('./session').setAuthenticatedSession;
 const removeSession = require('./session').removeSession;
+const isAuthenticated = require('./session').isAuthenticated;
+const getLogoutHint = require('./session').getLogoutHint;
 
 const AUTHORIZATION = 'authorization';
 const AUTH_URL = 'authorization_url';
 const TOKEN_DECISION = 'token_decision';
+const USER = 'user';
+const LOGOUT_HINT = 'logout_hint';
+const RP_LOGOUT = 'rp_logout';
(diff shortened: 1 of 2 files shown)


ermin-muratovic (approved):
lgtm

Issue #3417, opened 2021-03-02
Set Git upstream URL via Settings page


johannes-b:
**Problem:** When a project has no upstream configured, the user has to use the Keptn CLI to update the upstream config for the project. 

**Improvement:**
A user should be able to set the upstream without a context switch (CLI vs Keptn Bridge). Therefore, a settings page is provided that allows configuring the upstream config for a project. 

**Task:**
* On the project tile, show the info: `Set Git upstream repository`
* When clicking on the `+` icon (the icon can be another one) the user gets forwarded to the settings page.
* The settings page provides a form to enter the information
  * `Git upstream URL`
  * `Git user`
  * `Token` (password field) 
* The "Set Git upstream" button allows setting the upstream.

--> When the user has set the Git upstream, a user can use the settings page to change the configuration
* Rename the button from `Set Git upstream` to `Update Git upstream`

## Definition of Done

**Project tile:**
![image](https://user-images.githubusercontent.com/729071/110595480-57924c00-817e-11eb-9ef5-dc66e2fecd8f.png)

**Settings page: (ignore the Check buttons)**
![image](https://user-images.githubusercontent.com/729071/109645771-94938880-7b57-11eb-98eb-35aa5231df67.png)

![image](https://user-images.githubusercontent.com/729071/109646229-37e49d80-7b58-11eb-84b9-00793c3cc4a9.png)

PR #3931, opened 2021-05-03
Set git upstream url via project settings page

laneli:
Implements #3417 
* The text and link for git upstream configuration changed:
image

* There is a new project settings page containing the git upstream form
image

* Validation and disabling of button is build in accordingly
image

* If a user already has an upstream url set, the data is loaded and displayed in the form
* The token field is filled with placeholder data and emptied on focus
image


Kirdock:
We should re-think our `http-error-interceptor`. Currently if an error happens, we retry it up to 3 times, which is not needed in this case because there is some server-side validation (invalid URL, username not found, etc.) that may cause it.

We can at least check if the provided URL is valid client-side

ermin-muratovic (approved):
lgtm

laneli:
@Kirdock can you file a new issue for the http-interceptor?

Kirdock:
> @Kirdock can you file a new issue for the http-interceptor?

done. Issue: #3950

PR #3932, opened 2021-05-03
#3927 Broken UI in quality gates only

### LINKED ISSUE #3927: Bridge in Quality gates only use-case breaks on same sequence and task name "evaluation"
If you have a quality gates only configuration and the sequence and the task name are both "evaluation", the bridge gets an error while trying to retrieve the evaluation result. It should load the `evaluation.finished` event from the task, but gets the `evaluation.finished` event of the sequence.

This leads to a broken UI, see screenshots below.

image

image

image

ermin-muratovic:
Retrieve correct evaluation event in case that the sequence and the evaluation event have the same name.

Fixes #3927

### DIFF
--- a/bridge/client/app/_models/root.ts
+++ b/bridge/client/app/_models/root.ts
@@ -98,7 +98,7 @@ export class Root extends Trace {
   }
 
   getEvaluation(stageName: String): Trace {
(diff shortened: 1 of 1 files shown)


Kirdock (approved):
LGTM

PR #3947, opened 2021-05-04
Adapt spec to show correct configure monitoring event type and payload

laneli:


warber (approved):
:+1:

Issue #3925, opened 2021-04-30
Keptn bridge - Error pages


Kavindu-Dodan:
**Background**

Keptn bridge is a component which consists of  an Angular UI and an Express server. Right now, Express server is minimalistic and acts like a proxy between UI and other Keptn components (ex:- Include X Token for requests and facilitate responses).

However, with feature implementations such as #3448 & #3880 there are considerable logic introduced in server layer. And these features require error handling.

**Solution**

I can think about two approaches to implement error handling,

1. Redirect to Angular layer : Angular expose error routing
2. Minimalistic error templates in Express layer : Pug [1] is a good candidate

If we can agree on the approach, I am willing to contribute the first PRs for this feature request. 

[1] - https://www.npmjs.com/package/pug

Kavindu-Dodan:
Discussed the topic with @ermin-muratovic & following was decided for the first iteration,

- Use a minimalistic Express template to generate basic error pages 
- Use following design, inspired by Bridge Angular UI error page

![image](https://user-images.githubusercontent.com/8186721/116693844-baee6c80-a9be-11eb-8904-27d50e1adb04.png)

 I will investigate about pug [1] and work on a flexible solution. 

[1] - https://www.npmjs.com/package/pug

johannes-b:
Implemented by: https://github.com/keptn/keptn/pull/3952

PR #3952, opened 2021-05-04
Introduce server views to support better server errors

Kavindu-Dodan:
Introduce views based on Pug [1] for server errors. Solves #3925

**Previews** 

image

image

**Features**

- Customisable error title & error message
- Optional location and locationMessage

[1] https://pugjs.org/api/getting-started.html


christian-kreuzberger-dtx (approved):
Looks good to me.
Pinging @johannes-b and @ermin-muratovic on this so that they are aware of the change.

bridge/server/views/error.pug
```diff
@@ -0,0 +1,18 @@
+html
+  head
+    link(rel="stylesheet", href="/static/error.v1.css")
```
Kavindu-Dodan, line 3: Need to update relative paths as we could have different deployment base paths

christian-kreuzberger-dtx (approved):
Approving again :)
LGTM

ermin-muratovic (approved):
lgtm

PR #3954, opened 2021-05-04
#3893 Added ceCache.remove functionality

### LINKED ISSUE #3893: Distributor for remote execution plane needs to handle the case of slowly responding services
As a side-effect of https://github.com/keptn/keptn/issues/3888, we should make sure that distributor in remote-execution plane adds the CloudEvent to the `ceCache` **before** it is sent to the respective service.

If sending fails, distributor should remove the CloudEvent from  `ceCache`.

https://github.com/keptn/keptn/blob/01ca31a892e783c537ee4efe709267e022f9b943/distributor/cmd/main.go#L471-L476

# Actions

- [ ] Implement and Unit Test `ceCache.Remove` 
- [ ] Change implementation of distributor to add CloudEvents early to `ceCache`, and remove them on error
- [ ] Unit Test the new behaviour

christian-kreuzberger-dtx:
Fixes #3893 by 
* adding a `ceCache.remove` function
* Unit testing the `remove` function
* Extending the unit test for testing the polling/forwarding behaviour

In addition, I've applied boyscout rule and slightly improved the code in main.go

### DIFF
--- a/distributor/cmd/main.go
+++ b/distributor/cmd/main.go
@@ -478,7 +478,7 @@ func pollEventsForTopic(endpoint string, token string, topic string) {
 					// Sending failed, remove from cloudevents cache
 					ceCache.Remove(*event.Type, event.ID)
 				}
-				logger.Infof("Number of sent events for topic %s: %d", topic, ceCache.Length(topic))
+				logger.Infof("CloudEvent sent! Number of sent events for topic %s: %d", topic, ceCache.Length(topic))
 			}()
 		}
 	}
--- a/distributor/cmd/main.go
+++ b/distributor/cmd/main.go
@@ -470,9 +470,13 @@ func pollEventsForTopic(endpoint string, token string, topic string) {
 
 		if e != nil {
 			logger.Infof("Sending CloudEvent with ID %s to %s", event.ID, env.PubSubRecipient)
+			// add to cloudevents cache
+			ceCache.Add(*event.Type, event.ID)
 			go func() {
-				if err := sendEvent(*e); err == nil {
-					ceCache.Add(*event.Type, event.ID)
+				if err := sendEvent(*e); err != nil {
+					logger.Errorf("Sending CloudEvent with ID %s to %s failed: %s", event.ID, env.PubSubRecipient, err.Error())
+					// Sending failed, remove from cloudevents cache
+					ceCache.Remove(*event.Type, event.ID)
 				}
 				logger.Infof("Number of sent events for topic %s: %d", topic, ceCache.Length(topic))
 			}()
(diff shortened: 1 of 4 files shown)


distributor/cmd/main.go
```diff
@@ -442,39 +442,60 @@ func pollHTTPEventSource(endpoint string, token string, topics []string) {
...
 	logger.Infof("Received %d new .triggered events", len(events))
+
+	// ensure CloudEvents Cache exists
+	if ceCache == nil {
+		logger.Debug("Cache containing sent CloudEvents is nil. Creating a new one")
+		ceCache = lib.NewCloudEventsCache()
+	}
+
+	// iterate over all events, discard the event if it has already been sent
 	for index, _ := range events {
```
warber, line 461: I know, it's not part of the PR, but could you please replace line 461 and 462 with
```
	for index := range events {
		event := *events[index]
```

distributor/cmd/main.go
```diff
@@ -442,39 +442,60 @@ func pollHTTPEventSource(endpoint string, token string, topics []string) {
...
 func pollEventsForTopic(endpoint string, token string, topic string) {
 	logger.Infof("Retrieving events of type %s", topic)
 	events, err := getEventsFromEndpoint(endpoint, token, topic)
 	if err != nil {
 		logger.Errorf("Could not retrieve events of type %s from endpoint %s: %v", topic, endpoint, err)
 	}
 	logger.Infof("Received %d new .triggered events", len(events))
+
+	// ensure CloudEvents Cache exists
+	if ceCache == nil {
```
warber, line 455: I think you could even move the variable declaration to the top in line 69 and remove this check. It's a global variable.

christian-kreuzberger-dtx: `ceCache` is already initialized in `createHTTPConnection`, so I guess we don't need the check here.
```
func createHTTPConnection() {
	if env.PubSubRecipient == "" {
		logger.Error("No pubsub recipient defined")
		return
	}

	ceCache = lib.NewCloudEventsCache()

	eventEndpoint := getHTTPPollingEndpoint()
	topics := strings.Split(env.PubSubTopic, ",")

	pollingInterval, err := strconv.ParseInt(env.HTTPPollingInterval, 10, 64)
	if err != nil {
		pollingInterval = defaultPollingInterval
	}

	pollingTicker := time.NewTicker(time.Duration(pollingInterval) * time.Second)

	for {
		<-pollingTicker.C
		pollHTTPEventSource(eventEndpoint, env.KeptnAPIToken, topics)
	}
}
```

distributor/cmd/main.go
christian-kreuzberger-dtx: Will do.

distributor/cmd/main.go
christian-kreuzberger-dtx: I think we need it for running unit tests on the `pollEventsForTopic` function.

christian-kreuzberger-dtx: I've moved it.

warber (approved):

Issue #3755, opened 2021-04-12
RemediationService: Use rootCause field instead problemTitle


warber:
**User Story**:
As a user, I want the *RemediationService* to use the `rootCause` property
so that the correct remediation action gets determined.

**Details**:
The *RemediationService* shall be adapted to use the `rootCause` field instead of the `problemTitle` when receiving a `sh.keptn.event.get-action.triggered` event in order to determine the proper remediation action. A fallback to using `problemTitle` shall be in place.

**Definition of Done**
- *RemediationService* is processing the `rootCause` field in order to provide the remediation action.
- *RemediationService* is uses `problemTitle` if there is not `rootCause` as a fallback
- Integration tests for problem remediations are adjusted and working

warber:
reading the root cause and falling back to problem title in the Remediation service is part of: https://github.com/keptn/keptn/pull/3995

Issue #3698, opened 2021-04-06
Clean-up of Bridge due to changed remediation use-case


johannes-b:
The Bridge should support the adapted remediation use-case, since it can deal with any custom sequence. However, the Bridge has some custom adoptions for the remediation use-case like icons and how to deal with problem events. 

**Tasks:**
- Remove the custom implementations for the remediation use-case in the bridge
- Adapt the bridge to work with the new approach
- Environment screen: **Problem icon** + list rem. actions that have been executed + result of action
- Service screen: **Status of the artifact** 
- Sequence screen: Icon of the remediation use-case 

**Definition of Done:**
- [x] The bridge supports the remediation use-case - this can contain an "opinionated" flavor

Issue #3683, opened 2021-04-01
Implement looping-mechanism via shipyard-controller


johannes-b:
**Use-case:** A remediation config can contains multiple actions that wait to get executed as response to an open problem, e.g.:
`remediation.yaml`: 

```
apiVersion: spec.keptn.sh/0.1.4
kind: Remediation
metadata:
  name: carts-remediation
spec:
  remediations:
    - problemType: Response time degradation
      actionsOnOpen:
        - action: scaling
          name: scaling
          description: Scale up
          value: "1"
        - action: toggle-feature
          name: Toogle feature flag
          description: Toogle feature flag EnablePromotion to OFF
          value: 
            EnablePromotion: "off"
```

A remediation sequence will execute just one action: 

```
- name: "production" 
  sequences: 
  - name: remediation
      tasks:
      - name: get-action                           < get action - "scaling"
      - name: action                                 < execute action - "scaling"
      - name: evaluation
```

To achieve a looping mechanism, the `remediation` sequence should get re-triggered when the `evaluation` task **failed**.

**Proposal:**
- Advance the trigger filters to consider the result of an individual task: 

```
- name: "production" 
  sequences: 
  - name: remediation
      triggeredOn: 
      - event: production.remediation.finished
          selector:
            match:
              evaluation.result: fail                             < see here
      tasks:
      - name: get-action 
      - name: action
      - name: evaluation 
```

**Tasks:**
- Update of shipyard in keptn/spec: https://github.com/keptn/spec/blob/master/shipyard.md
- Extend the shipyard-controller to filter events on task level

**Definition of Done:**
- [ ] When I have multiple actions in a `remediation.yaml` configured, they get executed by a single remediation sequence (get-action, action, evaluation) run.

PR #3965, opened 2021-05-05
#3682: processing get-action.triggered events

### LINKED ISSUE #3682: Clean-up remediation-service to not control the remediation sequence
**Current status:** In its current implementation, the remediation-service controls the execution of the remediation sequence by sending out the `action.triggered` event and waiting for `action.finished` and `evaluation.finished`. Besides, waiting for 10 minutes is hard-coded into the services. 

Should we call the task: `get-action` ? Since we have the same discussion here: https://github.com/keptn/spec/issues/57 --> `get-action`

**Clean-up remediation-service:** The remediation-service does not control the sequence, but:
* it subscribes to a `get-action.triggered` event,
* it analysis the event payload and determines the next action to execute by looking into the `remediation.yaml`.
* If the `remediation.yaml` defines multiple actions for one problem and the one was already be executed, the remediation-service must be smart enough to take the next one in the list of actions. 

**Clean-up of spec:**
- By implementing this approach, the reserved Keptn Task `remediation` is not needed anymore: https://github.com/keptn/spec/blob/master/shipyard.md#remediation

As a result of this clean-up task, the shipyard-controller takes over the responsibility for sending `action.triggered`, and `evaluation.triggered` - since this is modeled in the shipyard. 

**Design decisions:**
* The [sh.keptn.event.problem.open](https://github.com/keptn/spec/blob/master/cloudevents.md#problem) event does not work as trigger for the `remediation` sequence --> it will be the `sh.keptn.event.remediation.triggered` that kicks off the sequence. 
* Can the issue https://github.com/keptn/spec/issues/59 be resolved then?

**Definition of Done**: 
- [ ] Adaptation of integration tests
- [ ] The remediation sequence execution is explicitly modeled by the shipyard, e.g.:
```
- name: "production" 
  sequences: 
  - name: remediation
      tasks:
      - name: get-action
      - name: action
      - name: evaluation
        triggeredAfter: "10m"
```

warber:
closes #3682 

According to the issue, with this PR the behavior of the remediation service was changed to only process `sh.keptn.event.get-action` events


warber:
@check-spelling-bot apply changes

warber:
@check-spelling-bot apply [changes](https://api.github.com/repos/keptn/keptn/issues/comments/832510526)

warber:
> @check-spelling-bot apply [changes](https://api.github.com/repos/keptn/keptn/issues/comments/832510526).

warber:
> @check-spelling-bot apply [changes](https://api.github.com/repos/keptn/keptn/issues/comments/832510526).

warber:
> @check-spelling-bot apply [changes](https://api.github.com/repos/keptn/keptn/issues/comments/832728474).

warber:
We should merge this PR as soon all related stories are ready:

https://github.com/keptn-contrib/dynatrace-service/issues/287
https://github.com/keptn/keptn/issues/3683
https://github.com/keptn/keptn/issues/3698
https://github.com/keptn/keptn/issues/3755

remediation-service/handler/remediation.go
```diff
@@ -0,0 +1,53 @@
...
+	return remediationData, nil
+}
+
+func GetNextAction(remediation *v0_1_4.Remediation, problemDetails v0_2_0.ProblemDetails, actionIndex int) (*v0_2_0.ActionInfo, error) {
+	rootCause := problemDetails.RootCause
+	problemTitle := problemDetails.ProblemTitle
+
+	var actions []v0_1_4.RemediationActionsOnOpen
+	for _, r := range remediation.Spec.Remediations {
+		if r.ProblemType == rootCause || r.ProblemType == problemTitle {
```
bacherfl, line 30: would it make sense to first go through all remediations and check if a remediation for the root cause is available?
This way, having a remediation that matches the problem title would override a more specific one, if it is defined earlier in the shipyard. e.g.:

```
problemTitle = "response time degradation"
rootCause = "failure rate increase"

remediations:
  - problemType: "response time degradation"
    actionsOnOpen:
      - name: "some action"
      
  - problemType: "failure rate increase"
    actionsOnOpen:
      - name: "other action that would be better suited"
```

In this example, the service would determine that the correct action to take would be the first one (based on the problem title), while the other one that addresses the root cause will not be returned.

bacherfl (changes requested):

remediation-service/handler/remediation.go
warber: good catch, will do it

PR #3966, opened 2021-05-05
Fix unit test in Bridge

ermin-muratovic:
Unit tests are failing randomly, because of "keptnInfo" being null. This PR should fix this.


christian-kreuzberger-dtx (approved):
LGTM

PR #3967, opened 2021-05-05
Report failing unit tests

christian-kreuzberger-dtx:
This PR enhances our karma configuration for Bridge such that it reports failing unit tests by name (generated out of the spec file).

For this I had to add the karma-spec-reporter.
In addition, I enabled `sourceMap` such that we get stacktraces with typescript files, not the generated js files.


ermin-muratovic (approved):
lgtm

PR #3979, opened 2021-05-05
Turn off talk-to-bot feature

jsoref:
It's experiential. And this repository identified a logic issue:

There's code to check for:
* Is the commenter allowed to make a request
* Is the comment addressed to the bot
* Does the comment have an action

Unfortunately, right now, it provides feedback for the first thing, but it should only do that if it knows it's being addressed.

I'll fix it (hopefully tonight after work), but it'll need a new release.


christian-kreuzberger-dtx (approved):
LGTM

Issue #3647, opened 2021-03-29
Support of more advanced staging environments (in environment screen)


johannes-b:
As a user, I have a Shipyard that looks as follows: 

```
apiVersion: "spec.keptn.sh/0.2.0"
kind: "Shipyard"
metadata:
  name: "shipyard-sockshop"
spec:
  stages:

    - name: "quality-assurance"
      sequences:
        - name: "delivery"
          tasks:
            - name: "deployment"
            - name: "test"
            - name: "evaluation"
            - name: "release"

    - name: "dev"
      sequences:
        - name: "delivery"
          triggeredOn:
            - event: "quality-assurance.delivery.finished"
          tasks:
            - name: "deployment"
            - name: "test"
            - name: "evaluation"
            - name: "release"

    - name: "staging"
      sequences:
        - name: "delivery"
          triggeredOn:
            - event: "dev.delivery.finished"
          tasks:
            - name: "deployment"
            - name: "test"
            - name: "evaluation"
            - name: "release"

    - name: "production"
      sequences:
        - name: "delivery"
          triggeredOn:
            - event: "staging.delivery.finished"
          tasks:
            - name: "deployment"
            - name: "release"

    - name: "production-A"
      sequences:
        - name: "delivery"
          triggeredOn:
            - event: "project.delivery.finished"
          tasks:
            - name: "deployment"
            - name: "release"

    - name: "production-B"
      sequences:
        - name: "delivery"
          triggeredOn:
            - event: "project.delivery.finished"
          tasks:
            - name: "deployment"
            - name: "release"

    - name: "production-C"
      sequences:
        - name: "delivery"
          triggeredOn:
            - event: "project.delivery.finished"
          tasks:
            - name: "deployment"
            - name: "release"

```

--> This environment has 5 stages in horizontally aligned and 3 vertically aligned: 

![image](https://user-images.githubusercontent.com/729071/112843648-15846800-90a3-11eb-8a55-2c6e98687f55.png)

**Tasks:**
- Provide the parent stage in the response of the `/stage` (`/project`) endpoint  
- Display "parallel" stages (e.g., production-customer-A, production-customer-B, production-customer-C) in vertical order
- Set a fixed width of the "main part" of this screen (the stage overview) and add a horizontal scrolling bar
- Reduce the width of the right pane (info pane) 

![image](https://user-images.githubusercontent.com/729071/112832376-83c22e00-9095-11eb-9df3-9b3e3c5894b9.png)

**Definition of Done:**
[...]

PR #3989, opened 2021-05-06
#3647 Support more advanced staging environments

ermin-muratovic:
The environments view now will group all stages, that have the same parent stages.

image


laneli (approved):
lgtm

PR #4003, opened 2021-05-07
#3969 address potential memleak issues

### LINKED ISSUE #3969: distributor of shipyard-controller OOM crash after flooding it with events
With Keptn 0.8.2 and dynatrace-sli-service, I executed the following command (spamming evaluation events) for about 2 hours and distributor of shipyard-controller crashed multiple times due to Out Of Memory (OOM) errors:
```
while true; do keptn trigger evaluation --project=sockshop --stage=staging --service=carts  --timeframe=5m; sleep 5; done;
```

image
image

# Analysis

I believe what happened is that distributor is sending every event to the /event endpoint of shipyard-controller. As I got a response time degradation in shipyard-controller (see https://github.com/keptn/keptn/issues/3962 ), those events were queing up in distributor, which eventually lead to an OOM in distributor of shipyard-controller.

https://github.com/keptn/keptn/blob/c007df69a606c8a3017e4914d8541c77a13e3a5d/distributor/cmd/main.go#L680-L696

# Proposed Actions

- Reproduce this scenario (locally) with a test (e.g., by spamming events to distributor, and making sure that the receiving service is slow in responding)
- Improve event handling in distributor, such that it has a steady memory usage (e.g., by following "fire and forget" strategy and introducing a 5-10 second timeout on the sendEvent function)

bacherfl:
Closes #3969

elow is the time series graph for the memory consumption with these changes in place (in the background, an evaluation was triggered every 5 seconds)
Screenshot 2021-05-07 at 12 25 38

Integration test run: https://github.com/keptn/keptn/runs/2545876419?check_suite_focus=true


shipyard-controller/handler/eventhandler.go
```diff
@@ -114,15 +115,17 @@ func (eh *EventHandler) HandleEvent(c *gin.Context) {
...
-			SetInternalServerErrorResponse(err, c)
+	go func() {
+		err := eh.ShipyardController.HandleIncomingEvent(*event)
+		if err != nil {
+			if err == errNoMatchingEvent {
+				log.Infof("no matching event found for event %s", event.ID)
+			} else {
+				log.Errorf("error processing request: %s", err.Error())
+			}
+			return
```
christian-kreuzberger-dtx, line 126: I wonder if this is the right place for creating a new goroutine. 

I'd much rather do it as late as possible in the request-response cycle, e.g. here (`shipyardcontroller.go`)
```
	switch statusType {
	case string(common.TriggeredEvent):
		return sc.handleTriggeredEvent(event)
	case string(common.StartedEvent):
		return sc.handleStartedEvent(event)
	case string(common.FinishedEvent):
		return sc.handleFinishedEvent(event)
	default:
		return nil
	}
```

shipyard-controller/handler/shipyardcontroller.go
```diff
@@ -384,12 +382,17 @@ func (sc *shipyardController) handleFinishedEvent(event models.Event) error {
 		}
 		log.Infof("Task sequence related to eventID %s: %s.%s", event.Triggeredid, eventToSequence.Stage, eventToSequence.TaskSequenceName)
 		log.Info("Trying to fetch shipyard and get next task")
-		shipyard, err := common.GetShipyard(eventScope.Project)
+		project, err := sc.projectRepo.GetProject(eventScope.Project)
```
christian-kreuzberger-dtx, line 385: Please add some comments here, e.g., `// fetching cached shipyard file from project repo (materialized view)`

shipyard-controller/handler/eventhandler.go
bacherfl: agreed, will change that

christian-kreuzberger-dtx (approved):
LGTM

PR #4004, opened 2021-05-07
#4000 Fix golangci-lint problem with Go 1.16

### LINKED ISSUE #4000: Reviewdog fails to find .go files (after Go 1.16 upgrade)
In the most recent PRs, reviewdog always fails with
image

christian-kreuzberger-dtx:
Fixes #4000

The fix is as simple as setting `GO111MODULE=off ` (see https://github.com/golangci/golangci-lint/issues/1833)


warber (approved):
:+1:

PR #4008, opened 2021-05-07
#3585 Allow service names up to 43 characters

### LINKED ISSUE #3585: Length of service names is too restrictive
A Helm Release's name is limited to 53 characters, which is correctly checked by the ShipyardController when a new Service is created. The Helm release name is generated following the scheme:
`<projectname>-<stagename>-<servicename>[-generated]`, leaving not much room for the length of the service name in practice. Thus, it is way too easy to hit that limit.

Further, many Keptn projects don't use Helm at all (quality gates, performance testing, delivery with other tools than helm,...) where enforcing such strict naming conventions is too restrictive.

Finally, the user is confronted with the error message `[...] service name needs to be less than x characters [...]` without further context. Hence, users are wondering why the same service name is too short in project X, while it is fine in project Y.

> The helm Chart name can be reduced to: `<service_name>[-gen]` ... this results in a service length of 53-4 = 49.

Note: The solution needs to be backward compatible (especially keeping in mind `helm-service` and Keptn installations before 0.8.2)

Tasks:
- [ ] Come up and implement a less restrictive template for helm release names 
- [ ] Adapt the name checks in the CLI/shipyard-controller

Definition of done:
- [ ] Implementation of new helm release naming scheme leaving significantly more room for service name lengths.
- [ ] Solution is well documented and all tests are adapted

christian-kreuzberger-dtx:
Fixes #3585

I chose an approach that would ensure that existing helm releases stay as they are. Only if the existing release name template does not stay below 53 chars, some parts of the template are dropped.

I also added an integration test (written in Go) to verify that creating a service works.


helm-service/pkg/helm/helm_helper.go
```diff
@@ -66,13 +70,40 @@ func GetChartName(service string, generated bool) string {
...
-// GetReleaseName returns the name of the Helm release
+// GetReleaseName returns the name of the Helm release, e.g., ${PROJECT}-${STAGE}-${SERVICE}[-generated]
 func GetReleaseName(project string, stage string, service string, generated bool) string {
 	suffix := ""
 	if generated {
 		suffix = "-generated"
 	}
-	return project + "-" + stage + "-" + service + suffix
+
+	proposedReleaseName := project + "-" + stage + "-" + service + suffix
```
bacherfl, line 80: Imho this would be easier to read than nesting those `if`s: 

```
if len(proposedReleaseName) <= serviceNameMaxLen {
  return proposedReleaseName
}

proposedReleaseName = service + suffix
if len(proposedReleaseName) <= serviceNameMaxLen {
  return proposedReleaseName
}
// ... you get the idea ;)
```

christian-kreuzberger-dtx: Agreed, will adapt

test/go-tests/createservice_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	output, err := ExecuteCommand(fmt.Sprintf("keptn create project %s --shipyard=./%s", projectName, file))
+
+	require.Nil(t, err)
+	require.Contains(t, output, "created successfully")
+
+	output, err = ExecuteCommand(fmt.Sprintf("keptn create service %s --project=%s", serviceName, projectName))
+
+	// ToDo: Should fail because of shipyard-controller
```
christian-kreuzberger-dtx, line 65: I'll need to check how I can test this here
maybe using helm cli?

test/go-tests/createservice_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	output, err := ExecuteCommand(fmt.Sprintf("keptn create project %s --shipyard=./%s", projectName, file))
+
+	require.Nil(t, err)
+	require.Contains(t, output, "created successfully")
+
+	output, err = ExecuteCommand(fmt.Sprintf("keptn create service %s --project=%s", serviceName, projectName))
+
+	require.Nil(t, err)
+	require.Contains(t, output, "created successfully")
```
christian-kreuzberger-dtx, line 96: I'll need to check how I can test this here
maybe using helm cli?

bacherfl (approved):
lgtm

PR #4013, opened 2021-05-10
#3876 do not cache index.html

### LINKED ISSUE #3876: Bridge index.html is delivered with a 7 day cache header, which causes caching issues when deploying a new version
I just manually upgraded from Keptn 0.8.2 to our current master build. 
Unfortunately I was unable to review changes made in Keptn Bridge due to a caching issue.

Temporary Resolution: Clear your browser cache.

## Analysis

The way our Angular app is designed, we deliver it using index.html and several .js and .css files.

Unfortunately it seems that index.html responds with the following headers:
image

Especially the cache-contorl header suggests that index.html (as well as other files) can be cached for up to 7 days by any browser.

Angular uses so called cache-busters for JavaScript files. They are in the form of a hash-sum of the file, e.g.:
```
runtime-es2015.0dae8cbc97194c7caed4.js
main-es2015.a3fb3542b4bd3307ed8f.js
```
These are included in index.html. Therefore loading those resources from cache is fine.

However, if we ever upload an updated JS file, it's hash-sum changes and so do the imports in index.html.
If index.html is cached, we will never get an information of those updated javascript files.

## Action Items

As index.html is only 1.87 Kilobyte big, there is no harm in not caching it, therefore

- [ ] Ensure index.html is always delivered with an appropriate cache header (e.g., `cache-control: no-cache, no-store, must-revalidate`) to not cache it.

*Note*: not sure if this is something we need to control in bridge container, or in api-gateway-nginx.

## Acceptance Criteria
- [ ]  index.html is leaded with the headers mentioned above and no caching is happening

christian-kreuzberger-dtx:
Fixes #3876 by introducing a function that prevents .HTML files from being cached.

Installed it on my Kubernetes cluster and verified it works.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -217,6 +217,7 @@ errstr
 eslint
 esnext
 esource
+etag
 Eter
 euo
 evaluationdetails
--- a/bridge/server/app.js
+++ b/bridge/server/app.js
@@ -34,7 +34,18 @@ module.exports = (async function (){
   app.use('/static', express.static(path.join(__dirname, 'views/static'), {maxAge: oneWeek}));
 
   // UI static files - Angular application
-  app.use(express.static(path.join(__dirname, '../dist'), {maxAge: oneWeek}));
+  app.use(express.static(path.join(__dirname, '../dist'), {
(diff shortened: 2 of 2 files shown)


ermin-muratovic (approved):
lgtm

PR #4015, opened 2021-05-10
#4011 move setup of remote execution plane

bacherfl:
Installing the remote execution plane at the beginning of the tests messed up parts of the integration tests where those were not needed. Therefore, the setup of the remote execution plane has been moved to be executed right before this feature is tested

Integration test run: https://github.com/keptn/keptn/runs/2543064370?check_suite_focus=true

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -388,27 +388,6 @@ jobs:
         timeout-minutes: 1
         run: keptn status
 
-      - name: Install Remote Execution Plane
-        id: install_remote_execution_plane
-        if: env.REMOTE_EXECUTION_PLANE == 'true'
-        timeout-minutes: 5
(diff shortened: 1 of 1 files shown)


bacherfl:
> > Installing the remote execution plane at the beginning of the tests messed up parts of the integration tests where those were not needed.
> 
> This sounds like a critical bug that needs to be fixed in Keptn core, not in the integration tests. What if a customer runs into this problem?

In that case it is a setup problem that was caused by our integration test scripts - at one point we scale down certain services to mock their responses, and afterwards we scale them up again (this is done in the sequence state integration test). Therefore we end up having two deployments of the helm service (one in the keptn-test namespace, the other one in the keptn-helm-service namespace).

christian-kreuzberger-dtx (approved):
LGTM

Issue #4018, opened 2021-05-10
Integration tests: Report outcome of tasks test_sequence_states, test_delayed_tasks, test_manage_secrets


christian-kreuzberger-dtx:
Right now the following integration test tasks are not reported in the integration test summary:

* test_sequence_states
* test_delayed_tasks

https://github.com/keptn/keptn/blob/d90eb91ffe6d967090c130d11bb7e2101e49c0cd/.github/workflows/integration_tests.yml#L603-L610

In addition, the following task is reported but there is no header in the table:

* test_manage_secrets

https://github.com/keptn/keptn/blob/d90eb91ffe6d967090c130d11bb7e2101e49c0cd/.github/workflows/integration_tests.yml#L657-L662

Example: https://github.com/keptn/keptn/issues/4014

There might even be more tasks that are not reported properly, let's check on that too.

# Definition of Done

- [ ] All integration tests currently executed are properly reported with their status and a proper header

PR #4024, opened 2021-05-10
#3683 Implement looping mechanism for shipyard controller, reference specific task in sequence trigger

### LINKED ISSUE #3683: Implement looping-mechanism via shipyard-controller

bacherfl:
Closes #3683

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -404,6 +404,14 @@ jobs:
           KEPTN_ENDPOINT: ${{ steps.authenticate_keptn_cli.outputs.KEPTN_ENDPOINT }}
         run: cd test/go-tests && go test -run SequenceStateIntegrationTest
 
+      - name: Test Sequence Loop
+        id: test_sequence_states
+        timeout-minutes: 5
+        continue-on-error: true
+        env:
+          KEPTN_ENDPOINT: ${{ steps.authenticate_keptn_cli.outputs.KEPTN_ENDPOINT }}
+        run: cd test/go-tests && go test -run SequenceLoopIntegrationTest
+
       - name: Test Delayed Tasks
         id: test_delayed_tasks
         timeout-minutes: 5
--- a/shipyard-controller/handler/shipyardcontroller.go
+++ b/shipyard-controller/handler/shipyardcontroller.go
@@ -486,7 +486,7 @@ func (sc *shipyardController) proceedTaskSequence(eventScope *models.EventScope,
 			return err
 		}
 
-		return sc.triggerNextTaskSequences(eventScope, taskSequence, shipyard, eventHistory, inputEvent)
+		return sc.triggerNextTaskSequences(eventScope, taskSequence, shipyard, eventHistory, inputEvent, previousTask)
 	}
 	return sc.sendTaskTriggeredEvent(eventScope, taskSequence.Name, *task, eventHistory)
 }
@@ -520,8 +520,8 @@ func (sc *shipyardController) appendTriggerEventProperties(eventScope *models.Ev
 	return inputEvent, eventHistory, nil
 }
 
-func (sc *shipyardController) triggerNextTaskSequences(eventScope *models.EventScope, completedSequence *keptnv2.Sequence, shipyard *keptnv2.Shipyard, eventHistory []interface{}, inputEvent *models.Event) error {
-	nextSequences := getTaskSequencesByTrigger(eventScope, completedSequence.Name, shipyard)
+func (sc *shipyardController) triggerNextTaskSequences(eventScope *models.EventScope, completedSequence *keptnv2.Sequence, shipyard *keptnv2.Shipyard, eventHistory []interface{}, inputEvent *models.Event, previousTask string) error {
+	nextSequences := getTaskSequencesByTrigger(eventScope, completedSequence.Name, shipyard, previousTask)
 
 	if len(nextSequences) == 0 {
 		sc.onSequenceFinished(*inputEvent)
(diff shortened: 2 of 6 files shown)


test/go-tests/sequencetrigger_test.go
```diff
@@ -0,0 +1,289 @@
...
+	keptncommon "github.com/keptn/go-utils/pkg/lib/keptn"
+	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
+	"github.com/stretchr/testify/require"
+	"net/http"
+	"os"
+	"testing"
+	"time"
+)
+
+const sequenceTriggerShipyard = `apiVersion: "spec.keptn.sh/0.2.0"
```
christian-kreuzberger-dtx, line 14: I think this is the wrong spec version, as spec 0.2.0 does not support the selector to match a task

test/go-tests/test_utils.go
```diff
@@ -25,6 +29,56 @@ func (sender *APIEventSender) SendEvent(event v2.Event) error {
 	return err
 }
 
+func EnsureProjectExists(projectName, shipyardFilePath string) error {
```
christian-kreuzberger-dtx, line 32: I think this function is not doing what it is called.

1. it checks whether the project exists or not
2. if the project exists, the project is deleted
3. it creates a new project

I think it should be handled differently, e.g., via a param:

```golang
func CreateProject(projectName, shipyardFilePath, recreateIfAlreadyExists boolean) error {
...

	if resp.Response().StatusCode != http.StatusNotFound {
                if (recreateIfAlreadyExists) {
		// delete project if it exists
		_, err = ExecuteCommand(fmt.Sprintf("keptn delete project %s", projectName))
		if err != nil {
			return err
		}
                } else {
                    return error.Error("Project already exists")
                }
	}
...
```

test/go-tests/sequencetrigger_test.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+		state := states.States[0]
+
+		if !IsEqual(t, "state.State", "finished", state.State) {
+			return false
+		}
+
+		if !IsEqual(t, "len(state.Stages)", 1, len(state.Stages)) {
+			return false
+		}
```
christian-kreuzberger-dtx, line 275: It's super unintuitive to read this code here. It's in the reverse order of what stretchr/testify assert would do, e.g.
```golang
  // assert equality
  assert.Equal(t, 123, 123, "they should be equal")
```

Why not use `` here? According to the docs it should work:
```golang
		if !assert.Equal(t, "finished", state.State, "state.State") {
			return false
		}
```

.github/workflows/integration_tests.yml
```diff
@@ -404,6 +404,14 @@ jobs:
...
         run: cd test/go-tests && go test -run SequenceStateIntegrationTest
 
+      - name: Test Sequence Loop
+        id: test_sequence_loop
+        timeout-minutes: 5
+        continue-on-error: true
+        env:
+          KEPTN_ENDPOINT: ${{ steps.authenticate_keptn_cli.outputs.KEPTN_ENDPOINT }}
+        run: cd test/go-tests && go test -run SequenceLoopIntegrationTest
+
```
christian-kreuzberger-dtx, line 414: we need to add these tests to the test report - see issue https://github.com/keptn/keptn/issues/4018

shipyard-controller/handler/shipyardcontroller.go
```diff
@@ -520,8 +520,8 @@ func (sc *shipyardController) appendTriggerEventProperties(eventScope *models.Ev
 	return inputEvent, eventHistory, nil
 }
 
-func (sc *shipyardController) triggerNextTaskSequences(eventScope *models.EventScope, completedSequence *keptnv2.Sequence, shipyard *keptnv2.Shipyard, eventHistory []interface{}, inputEvent *models.Event) error {
-	nextSequences := getTaskSequencesByTrigger(eventScope, completedSequence.Name, shipyard)
+func (sc *shipyardController) triggerNextTaskSequences(eventScope *models.EventScope, completedSequence *keptnv2.Sequence, shipyard *keptnv2.Shipyard, eventHistory []interface{}, inputEvent *models.Event, previousTask string) error {
+	nextSequences := getTaskSequencesByTrigger(eventScope, completedSequence.Name, shipyard, previousTask)
```
christian-kreuzberger-dtx, line 524: You added `previousTask` as a parameter, but it is not used anywhere?

christian-kreuzberger-dtx (approved):
LGTM

PR #4098, opened 2021-05-18
Added header for get-action event

bacherfl:
The header for the get-action event was missing in the cloudevent spec generation


warber (approved):

christian-kreuzberger-dtx:
Thanks! Didn't know that there needs to be a header :D

PR #4111, opened 2021-05-19
Helm Schema: Repository might contain a digit on the start

thschue:
As AWS repositories often start with digits and the current JSON schema does not allow this at the moment, this has been added.


johannes-b (approved):
lgtm

PR #4133, opened 2021-05-19
#3698 Remediation clean up in Bridge

ermin-muratovic:
Showing remediation actions in environments screen correctly again:

image

Sequence screens also shows remediation sequences with a error icon and shows the last evaluation results in overview:

image

### DIFF
--- a/bridge/client/app/_models/trace.ts
+++ b/bridge/client/app/_models/trace.ts
@@ -164,6 +164,14 @@ class Trace {
     return this.type.endsWith(EventTypes.REMEDIATION_TRIGGERED_SUFFIX);
   }
 
+  public isRemediationAction(): boolean {
+    return this.type === EventTypes.ACTION_TRIGGERED;
+  }
+
+  public getRemediationActionDetails(): string {
+    return this.data.action.description || this.data.action.name;
+  }
+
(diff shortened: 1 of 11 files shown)


bridge/client/app/_models/trace.ts
```diff
@@ -158,6 +160,18 @@ class Trace {
     return this.type === EventTypes.PROBLEM_DETECTED || this.type === EventTypes.PROBLEM_OPEN;
   }
 
+  public isRemediation(): boolean {
+    return this.type.endsWith(EventTypes.REMEDIATION_TRIGGERED_SUFFIX);
+  }
+
+  public isRemediationAction(): boolean {
+    return this.type === EventTypes.ACTION_TRIGGERED;
```
laneli, line 168: Is the type always sh.keptn.event.action.triggered only for remediation actions? It seems very generic for me - could it also be triggered with "any" action?

bacherfl: for remediation, the type is always `sh.keptn.event.action.triggered`, according to KEP 37: https://github.com/keptn/enhancement-proposals/pull/37

laneli: Ok thanks :)

laneli (approved):

PR #4153, opened 2021-05-21
#4065 Added goroutines to lighthouse-service

### LINKED ISSUE #4065: "Reponse time degradation in lighthouse-service" when spamming get-sil-events
Using Keptn 0.8.2.

After a quick and dirty load-test (`while true; do keptn trigger evaluation --project=sockshop --stage=staging --service=carts  --timeframe=5m; sleep 5; done`) with Dynatrace monitoring active, I noticed that lighthouse-service performs poorly under sustained load:

image

We are talking of response times of over a minute before lighthouse sends a `get-sli.triggered` event.

I believe this is caused by fetching the SLO file here:
https://github.com/keptn/keptn/blob/e72dd50418ce70efc1b565f1ecf738b05f0ed750/lighthouse-service/event_handler/start_evaluation_handler.go#L51-L82

christian-kreuzberger-dtx:
Fixes #4065

Please note: I tried to ensure that my changes touch as little code as possible, therefore I just split the existing func into two funcs, and called it via `go ...`.

With this change I was able to improve handling of lighthouse-service such that HTTP calls are non-blocking.

Verified using git upstream set to github.com.

image

First test (12:10): No git upstream set
Second test: 12:25, 12:55: Git upstream set
Third test: 13:10: Git upstream set, changes in code applied

Proof it works:
image


bacherfl (approved):
lgtm

warber (approved):

PR #4181, opened 2021-05-27
#4041 Uniform registration endpoint

### LINKED ISSUE #4041: API: Register/Unregister endpoint for registering a Keptn Service that connects to Keptn control-plane
# User Story:

As a user, I need to know which integrations/Keptn-services are currently subscribed to a certain event type, and whether they are active.

## Details

When a Keptn-service/integration is started up, the distributor registers the service via the API endpoint `register`, and sends certain metadata:

* which event types is the Keptn-service subscribed to (e.g., `sh.keptn.event.>` or `sh.keptn.event.deployment.triggered`)
* which project, service, and stage filters are active/inactive
* distributor version
* kubernetes metadata like namespace, pod name, deployment name (might need to be passed as environment variables to the distributor)
* additional metadata (e.g., hostname, deployment name, service name, service version) 

Each service that is registered should get an identifier. Re-registering the same service should lead to the same identifier.

The identifier could be a STRING that consists of the following metadata:

`${SERVICENAME}-${NAMESPACE}[-${PROJECT}][-${STAGE}][-${SERVICE}]`

### Example

Think of a setup, where we have three execution-plane clusters, one for each stage: dev, hardening, prod

On each of those clusters, we deploy `helm-service` to namespace `keptn-exec`.
We configure the distributor to filter for `deployment.triggered` and `stage=dev` on cluster 1, `stage=hardening` on cluster 2, and `stage=prod` on cluster 3.
This means we have three instances of helm-service running.

## Assumptions

Data model is already available and implemented in Go: https://github.com/keptn/keptn/issues/4033

## Actions

- Implement `register` and `unregister` endpoints in the Keptn API (Proposal: shipyard-controller)
- Write unit tests for those endpoints

## Definition of Done

- [ ] Endpoints are available and unit tested

## Dependencies

- https://github.com/keptn/keptn/issues/4033

bacherfl:
Closes #4041

The Endpoint for the uniform integration registrations will be available at:

```
/controlPlane/v1/uniform/registration
```

Screenshot 2021-05-31 at 09 06 31
Screenshot 2021-05-31 at 09 07 14
Screenshot 2021-05-31 at 09 07 42
Screenshot 2021-05-31 at 09 07 56

https://github.com/keptn/keptn/actions/runs/892724225


test/go-tests/uniform_test.go
```diff
@@ -0,0 +1,50 @@
+package go_tests
```
warber, line 1: can we fire a `DELETE` request of the already registered integration at the end of the test as well?

bacherfl: sure, I'll add that

shipyard-controller/db/mongodb_connection.go
```diff
@@ -12,11 +12,20 @@ import (
...
 var mutex = &sync.Mutex{}
 
+var mongoDBConnectionInstance *MongoDBConnection
+
 // MongoDBConnection takes care of establishing a connection to the mongodb
 type MongoDBConnection struct {
 	Client *mongo.Client
 }
 
+func GetMongoDBConnectionInstance() *MongoDBConnection {
```
warber, line 22: This is not thread-safe.
I would propose something like this:
```
var mongoConnectionOnce sync.Once
...
func GetMongoDBConnectionInstance() *MongoDBConnection {
	mongoConnectionOnce.Do(func() {
		mongoDBConnectionInstance = &MongoDBConnection{}
		mongoDBConnectionInstance.EnsureDBConnection()
	})
	return mongoDBConnectionInstance
}
```
We could also skip the calls to `EnsureDBConnection()` in all the repos.

warber (approved):

PR #4205, opened 2021-05-31
Trigger change detection for heatmap

### LINKED ISSUE #4007: Selection change in heatmap does not always update SLO table - needs second click
## Enhancement request

*Selection change in heatmap doesn't always update SLO table - needs a second click*

I noticed this when clicking through the heatmap to view historical results. It seems that while the click in the heatmap will change the selection in the heatmap the table underneath doesn't always get updated. 

**Situation:** Every second time I have to click twice for the table to be refreshed

* **Click 1**:
image

* **Click 2**: *Still shows the result from previously selected entry*
image

* Either mouse move or Click 3 will now get me to the right table information:
image

## Action item
- [ ] Investigation why the SLI breakdown does not get updated

## Acceptance criteria
- [ ] As a user, I can click through the heatmap and get the current SLI breakdown displayed.

laneli:
Fixes #4007 


ermin-muratovic (approved):
lgtm

Issue #4095, opened 2021-05-18
Bridge downloads and uses customized look and feel on startup


warber:
## User Story

As a user, I want to be able to customize the look and feel of the Keptn Bridge.

### Details
The User interface of the Keptn Bridge shall be customizable *without forking the bridge*.
This means that on startup/bootstrap of the Keptn bridge, it needs to download the assets to be used
for customizing the bridge (CSS files, etc...). Note that the outcome of #3886 defines what files can be used for customizing the look and feel of the bridge.

**To be decided:** Where/What component is the provider/host of the assets which are downloaded by the Keptn Bridge?

The Keptn Bridge shall be configurable (via an environment variable) that points to the location of the assets to be downloaded.
Note that, most probably the assets to be downloaded are not a single file, but multiple files or an archive.

### Acceptance Criteria
Note: As part of this story it is enough to be able to customize the Logo and the "keptn" Label in the header of the Bridge UI.

Given the Keptn bridge is not deployed
When the User is deploying the Keptn Bridge with the environment variable `KEPTN_LOOK_AND_FEEL_URL` set to `X`
Then the Bridge downloads all assets from location `X` on startup and uses them for the look and feel of the UI.

Given the Keptn Bridge is not deployed
When the User is deploying the Keptn Bridge without setting the environment variable `KEPTN_LOOK_AND_FEEL_URL`
Then the Bridge uses its default set of styles used for the look and feel of the UI.

### Definition of Done
- [ ] Feature implemented and (manually) tested
- [ ] Documentation updated

### Dependencies
#3886

PR #4221, opened 2021-06-01
#4095 Download and use customized look and feel for Bridge

ermin-muratovic:
Bridge downloads and uses custom look and feel on startup.

For details how to apply custom look and feel see [bridge/README.md](https://github.com/keptn/keptn/tree/feature/4095%2Fbridge-downloads-custom-look-and-feel/bridge#custom-look-and-feel)

Example with custom logo, app header background in dark and custom title.
image


bridge/server/app.js
```diff
@@ -27,6 +31,52 @@ if(!integrationsPageLink) {
...
+
+try {
+  console.log("Installing default Look-and-Feel");
+
+  let destDir = path.join(__dirname, '../dist/assets/branding');
+  let srcDir = path.join(__dirname, '../client/assets/default-branding');
+
+  let brandingFiles = ["app-config.json", "logo.png", "logo_inverted.png"];
+  brandingFiles.forEach((file) => {
+    if(!fs.existsSync(destDir)) {
```
Kirdock, line 43: Move this check outside the loop or change the path to `path.join(srcDir, file)`.
I would also move `destDir` and the check if it exists one layer up and remove the declaration and check during the custom look and feel installation

Because the variables are not overwritten, I would change it to `const`

bridge/client/environments/environment.prod.ts
```diff
@@ -1,3 +1,16 @@
-export const environment = {
-  production: true
-};
+import {DynamicEnvironment} from "./environment.dynamic";
+
+class Environment extends DynamicEnvironment {
+
+  public appTitle: string;
+  public logoUrl: string;
+  public logoInvertedUrl: string;
```
Kirdock, line 7: Remove unused variables `appTitle`, `logoUrl` and `logoInvertedUrl`. They are accessed via `this.config.`

bridge/client/environments/environment.ts
```diff
@@ -1,3 +1,16 @@
-export const environment = {
-  production: false
-};
+import {DynamicEnvironment} from "./environment.dynamic";
+
+class Environment extends DynamicEnvironment {
+
+  public appTitle: string;
+  public logoUrl: string;
+  public logoInvertedUrl: string;
```
Kirdock, line 7: Remove unused variables `appTitle`, `logoUrl` and `logoInvertedUrl`. They are accessed via `this.config.`

Kirdock:
At the moment `https` is not supported
`Error while downloading custom Look-and-Feel file. Cause : TypeError [ERR_INVALID_PROTOCOL]: Protocol "https:" not supported. Expected "http:"`

bridge/client/assets/default-branding/app-config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "appTitle": "keptn",
+  "logoUrl": "assets/branding/logo.png",
+  "logoInvertedUrl": "assets/branding/logo_inverted.png"
```
johannes-b, line 4: Is the property: `"stylesheetUrl": "assets/branding/style.css"` missing here?

bridge/client/environments/environment.prod.ts
ermin-muratovic: They are necessary for Typescripts type check.

bridge/client/assets/default-branding/app-config.json
ermin-muratovic: The `stylesheetUrl` is optional and for the default theme not necessary.

bridge/server/app.js
ermin-muratovic: You are right, the check if the directory exists can be done once, instead of inside the loop.

bridge/server/app.js
```diff
@@ -27,6 +33,59 @@ if(!integrationsPageLink) {
...
+          try {
+            let zip = new admZip(destFile);
+            zip.extractAllTo(destDir, true);
+          } catch (err) {
+            console.error(`[ERROR] Error while extracting custom Look-and-Feel file. ${err}`);
+          }
+        });
+      });
+    }).on('error', function(err) {
+      fs.unlink(destFile);
```
Kirdock, line 82: The server crashes if an invalid url is provided
```
TypeError [ERR_INVALID_CALLBACK]: Callback must be a function. Received undefined
```

Kirdock:
At the moment the default-configuration is not loaded when the dev environment is started with `start:server:dev` for example.

Cause: The server first copies the files into the dist folder. After that `ng build` is executed and recreates the dist folder. Therefore the copied files from the server are deleted.

bridge/server/app.js
ermin-muratovic: This should work now as well. If the file is not downloadable, extractable or the provided URL is invalid at all, there will be an error log message, but the Bridge should start and work with the default look and feel.

Kirdock (approved):

PR #4222, opened 2021-06-01
Bug/4149/scrollbar behaviour environment screen

### LINKED ISSUE #4149: Unexpected behavior of scrollbars in environment screen
## Current Behavior

1. Even if you only have one stage horizontally, there is a horizontal scrollbar on a Full HD display.
2. The vertical scrollbar is not draggable/clickable.

image

## Definition of Done
 - [x] The horizontal scrollbar only appears if the stages exceed the horizontal width.
 - [x] The vertical scrollbar can be clicked and moved with the cursor.

laneli:
Fixes #4149


laneli:
image

image

Horizontal scrollbar is now hidden except when the content is bigger than before.
Vertical scrollbar is now interactable by using the pointer-events: none CSS property.

Kirdock (approved):

PR #4252, opened 2021-06-07
#4165 Do not load versions.json if ENABLE_VERSION_CHECK is set to false

### LINKED ISSUE #4165: Bridge runs version check although ENABLE_VERSION_CHECK env is set to "false"
**Describe the bug**
Bridge runs version check although the environment variable ENABLE_VERSION_CHECK is set to "false". 

Before 0.8.0, the version check was only executed after the user opted in for that and the ENABLE_VERSION_CHECK flag was used to determine, whether the Bridge should ask the user or not. Since 0.8.0 the version check is executed, unless the user opted out, but if ENABLE_VERSION_CHECK is set to "false" there is no way to opt-out, since the version check settings are hidden.

**To Reproduce**
1. Set environment variable `ENABLE_VERSION_CHECK` to "false
2. Start Keptn Bridge
3. The `version.json` is loaded and the info about new Keptn version is displayed

**Expected behavior**
If ENABLE_VERSION_CHECK is set to "false", the `version.json` shouldn't be loaded and the info about new Keptn version shouldn't be displayed.

laneli:
Fixes #4165

### DIFF
--- a/bridge/client/app/_services/data.service.ts
+++ b/bridge/client/app/_services/data.service.ts
@@ -112,25 +112,29 @@ export class DataService {
   }
 
   public loadKeptnInfo() {
-    forkJoin({
-      availableVersions: this.apiService.getAvailableVersions(),
-      bridgeInfo: this.apiService.getKeptnInfo(),
-      keptnVersion: this.apiService.getKeptnVersion(),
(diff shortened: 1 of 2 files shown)


Kirdock (approved):

ermin-muratovic (approved):
lgtm

PR #4259, opened 2021-06-08
#4002 show sequence execution in service screen

### LINKED ISSUE #4002: Improve status information in Bridge Service View for failed deployments
## Enhancement request

In my scenario, I ran build number 2 which failed to deploy in production. However, it is very hard to see from here why it failed and also no link to the sequence screen to give me more details

* Service screen: here is what I see in the bridge:

image

* Sequence screen: here is the information that shows that the release step actually failed.
image

Now, I think we should bring this information to the service screen as this is important for the service owner. Here is a suggestion on how to improve the service screen
1. provide information that a certain stage failed including the quality gate result
2. provide the reason for the failure -> show the failed task with a link to the sequence
image

## Action item

- [ ] The issue https://github.com/keptn/keptn/issues/3941 adds the **[ View sequence execution ]** button to this view. Next, add the start time of the sequence, and the label: Sequence *delivery*  in front of the button

image

- [ ] When the sequence is finished, change the sequence icon according to the status (red/green) and show the message of the finished event:

image

- [ ] If no task is running (because the sequence finished), show the result of the last task in this sequence. If everything went well, the result is `passed` and green. If a problem occurred, the result is `failed` and red. 

image

## Acceptance criteria
- [ ] As a user, I can see the status/message of the sequence and subsequent sequences that were triggered.

Kirdock:
Closes #4002 
image

Remediations are now on top of the sequences instead in a separate list
image

The "View sequence execution"-button is removed because the sequence name now redirects to the sequence.


laneli (approved):
Looks good to me :)

Issue #4155, opened 2021-05-21
Bridge breaks on "sh.keptn.event.evaluation.triggered" root event


ermin-muratovic:
**Describe the bug**
The Bridge breaks on the environment screen, in case that there is a root event "sh.keptn.event.evaluation.triggered", which is not executed as part of a Sequence, but directly by sending the event to keptn via API or command line e.g.

![image](https://user-images.githubusercontent.com/6098219/119139770-b3613700-ba43-11eb-8505-da4b40efea0b.png)

**To Reproduce**
1. Create event
```
{
  "data": {
    "image": "docker.io/keptnexamples/carts:0.12.3",
    "project": "sockshop",
    "service": "carts",
    "stage": "quality-gate",
    "test": {
      "end": "2021-05-20T15:55:21.224Z",
      "start": "2021-05-20T15:53:32.708Z"
    },
    "teststrategy": "manual"
  },
  "id": "10740038-d80b-4ed4-b43a-2a3981bf926d",
  "source": "User",
  "specversion": "1.0",
  "type": "sh.keptn.event.evaluation.triggered"
}
```
2. Trigger keptn send event -f=./event.json
2. Open Bridge and navigate to environments screen

**Expected behavior**
The evaluation result gets displayed correctly and doesn't break the Bridge

christian-kreuzberger-dtx:
Please note that this is **not** a proper way to trigger an evaluation, as it does **not** contain the stage name in the event type. Evaluation result should **not** get displayed.

```
{
  "data": {
    "image": "docker.io/keptnexamples/carts:0.12.3",
    "project": "sockshop",
    "service": "carts",
    "stage": "quality-gate",
    "test": {
      "end": "2021-05-20T15:55:21.224Z",
      "start": "2021-05-20T15:53:32.708Z"
    },
    "teststrategy": "manual"
  },
  "id": "10740038-d80b-4ed4-b43a-2a3981bf926d",
  "source": "User",
  "specversion": "1.0",
  "type": "sh.keptn.event.evaluation.triggered" <--- MUST BE  "sh.keptn.event.quality-gate.evaluation.triggered" 
}
```


The way to properly trigger an evaluation is documented here: https://github.com/keptn/keptn/pull/4153

PR #4265, opened 2021-06-08
#4155 Bridge breaks on "sh.keptn.event.evaluation.triggered" root event

ermin-muratovic:
Call hasFailedEvaluation() on root event to prevent endless loop of calling "isFailedEvaluation()"


laneli (commented):
Isn't it just the same with a different name? :D

Kirdock:
> Isn't it just the same with a different name? :D

It is, but now it isn't an endless loop anymore because it calls `isFailedEvaluation` from `Trace` instead of `Root`

Kirdock (approved):

PR #4272, opened 2021-06-09
Feature/4034/uniform screen api implementation

### LINKED ISSUE #4034: Show uniform screen with data fetched from Uniform Backend
## User Story
As a User, I want to be able to see all Keptn services (integrations) that are currently connected/subscribed to the control plane

### Details
The current implementation uses a mocked data model to present the Uniform screen.
When the Backend API is implemented, the data for this screen shall be fetched using the real backend API for the uniform state.

Note, that the real data model returned by the API endpoint might be different from the currently mocked data model.
The current UI for _editing_ subscriptions of services must be disabled as this is currently not supported. However, showing the
subscriptions shall still be possible.

### Definition of Done
- [x] Feature implemented and (manually) tested
- [x] Feature flag for uniform screen removed

## Dependencies
#4035

laneli:
Closes #4034

The API endpoint differs a bit from the mock implementation:
* Sequences are not an object, there is only a list of subscribed topics. So the detail was removed as it holds no additional information.
* There is currently no endpoint for registering / removing cloud event subscriptions for an event - this is currently also removed and will added again when the feature will be implemented
* The list of services is expanded on the whole view height for more convenience.
* Sorting has to be adapted to be manually sorted

Screenshot with adapted mock data, as the endpoint does not deliver useful data.
image

Signed-off-by: Elisabeth Lang elisabeth.lang@dynatrace.com


bridge/client/app/_components/ktb-keptn-services-list/ktb-keptn-services-list.component.ts
```diff
@@ -40,20 +39,38 @@ export class KtbKeptnServicesListComponent implements OnInit {
...
-  public formatSubscription(subscriptions: Subscription[]): string {
-    return subscriptions.reduce((events, subscription) => {
-      if (!events.includes(subscription.event)) {
-        events.push(subscription.event);
-      }
-      return events;
-    }, []).join(', ');
+  public formatSubscriptions(subscriptions: string[]): string {
+    const formatted = []
+    subscriptions.forEach(subscription => {formatted.push(subscription.replace('sh.keptn.', ''))});
```
Kirdock, line 51: Change `sh.keptn.`  to `sh.keptn.events.`

Kirdock:
I would also move the column `Deployment` between `Name` and `Version`.

laneli:
> Even if there is just one subscription, we still need the information on which project, service and stage this subscription is.

As decided today in All eyes meeting this will be skipped for now

> I would also move the column `Deployment` between `Name` and `Version`.

Deployment was removed - as it will be the same as name

> The custom sort should be changed to be case-insensitive. Change the comparison to `localeCompare`.

Done

> We also should consider changing the way, the items are ordered. If we sort by `namespace` and items have the same 
namespace, the items should then be ordered by `name`.

Done

Kirdock (approved):

PR #4285, opened 2021-06-10
#4006 improve service image overflow

### LINKED ISSUE #4006: Environment layout improvement for service versions
## Enhancement request

This is more an enhancement request than a feature request. I noticed the following bridge layout when looking at our demo environment where we deploy hipster store and all its different services. 

I was hoping we can come up with a better way to visualize the names of the services and the last deployed version. Right now the `service:version` name is cut off if it is too long. Also, when hovering over the name with the mouse there is no tooltip.

I am not a UI layout expert but was hoping we can address this to better visualize a large list of services per environment without just cutting off the version information:

image

## Action Item

- [ ] For the service name, introduce a linebreak after the colon, for example:
```
frontend
v2.2.2
```
- [ ] If the service name is still too long, then cut it off using ... 
- [ ] + add a tooltip showing the entire service name

## Acceptance criteria

- [ ] As a user, I can see the service name (cut-off in case it is too long) + the version of the artifact in the line below

Kirdock:
Closes #4006 

Deployed image name: 
image

Deployed image version:
image

### DIFF
--- a/bridge/client/app/_components/ktb-services-list/ktb-services-list.component.html
+++ b/bridge/client/app/_components/ktb-services-list/ktb-services-list.component.html
@@ -4,9 +4,18 @@
 
       <ng-container dtColumnDef="serviceName" [dtColumnProportion]="2" uitestid="keptn-services-list-problems">
         <dt-cell *dtCellDef="let row">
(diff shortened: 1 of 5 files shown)


laneli (approved):

ermin-muratovic (approved):
lgtm

PR #4286, opened 2021-06-10
Use image version tag with timestamp for PR builds

### LINKED ISSUE #4191: For non release-builds, use timestamps for containers in helm-charts
Right now in our CI build we build Docker Images with the following tags:

* `0.8.3-dev` (`x.y.z-dev`)
* `0.8.3-dev.202105280620` (`x.y.z-dev.timestamp`)

In our helm-charts, we always use the first version string (`x.y.z-dev`), which can lead to problem when trying to install and/or upgrade from a dev version to another dev version, as the Docker Images are not re-pulled.

This has partially already been done for master builds (the else branch is for release builds):
https://github.com/keptn/keptn/blob/0f24233122a216b6bf597731f8673f345b841e0c/.github/workflows/CI.yml#L603-L612

# Task

- Add a check for PR builds and also use the `$VERSION.$TIMESTAMP`

# Definition of Done

- [ ] Master/PR builds use the version string **with timestamp**
- [ ] Release builds still use the version string **without** timestamp

mowies:
**This PR**
* adds the timestamp to built images that come from a PR build

Fixes #4191

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -603,7 +603,7 @@ jobs:
       - name: Build Helm Charts
         id: build_helm_charts
         run: |
-          if [[ "$BRANCH" == "master" ]]; then
+          if [[ "$BRANCH" == "master" ]] || [[ "$GITHUB_EVENT_NAME" == "pull_request" ]]; then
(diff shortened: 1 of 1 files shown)


christian-kreuzberger-dtx (approved):
LGTM

PR #4300, opened 2021-06-11
#3943 show every nth label starting from the end

### LINKED ISSUE #3943: Show every nth label in heatmap starting from end
Related to #3811 
In consultation with @Gevorgisk we should show every nth label on the heatmap on small display sizes but starting from the end.
This means, that the last label is always visible - the evaluation the user is probably most interested in. Also see https://github.com/keptn/keptn/issues/3811#issuecomment-831838832

Kirdock:
Closes #3943 

- Tick interval is now determined by component width.
- Last and first label are always displayed

image

### DIFF
--- a/bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.ts
+++ b/bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.ts
@@ -138,10 +138,11 @@ export class KtbEvaluationDetailsComponent implements OnInit, OnDestroy {
       },
       tickPositioner: function() {
         const positions = [],
+          labelWidth = 70,
           ext = this.getExtremes(),
           xMax = Math.round(ext.max),
           xMin = Math.round(ext.min),
-          maxElements = window.innerWidth / 70,
+          maxElements = (document.querySelector('dt-chart')?.clientWidth || labelWidth) / labelWidth,
           tick = Math.floor(xMax / maxElements) || 1;
 
         for (let i = xMax; i > xMin; i -= tick) {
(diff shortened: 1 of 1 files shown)


bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.ts
```diff
@@ -136,6 +136,26 @@ export class KtbEvaluationDetailsComponent implements OnInit, OnDestroy {
...
+          xMin = Math.round(ext.min),
+          maxElements = (document.querySelector('dt-chart')?.clientWidth || labelWidth) / labelWidth,
+          tick = Math.floor(xMax / maxElements) || 1;
+
+        for (let i = xMax; i > xMin; i -= tick) {
+          positions.push(i);
+        }
+        if(positions[positions.length-1] - tick < xMin){ //if the second item is too near to the first one, remove the second one
+          positions[positions.length-1] = xMin;
+        }
```
christian-kreuzberger-dtx, line 153: I guess this is responsible for the visual gap on the left side?
image

I'm not sure if that's a good thing. Why is this?

Kirdock: There is not enough space to display the first item, that's why the second one is removed. It could look like this
image

In order to remove this gap, we need to remove the label of the first evaluation. Should we do that?

christian-kreuzberger-dtx: Visually this looks like there is an error with the calculation or spacing. We should obviously show the first and last label, and evenly spread out the labels in between. I'll take a look at the calculation.

christian-kreuzberger-dtx: Okay, I see what's going on here.
This is because we have 50 columns. Therefore we could have 25 labels shown.
If we show the first label though, we will need to hide the second one.

From a visual standpoint, I'd propose to make sure the distance is the same between labels, and therefore, as suggested by @Kirdock, remove the label of the first evaluation. @johannes-b is that okay with you?

ermin-muratovic (approved):
lgtm

PR #4302, opened 2021-06-11
#4095 Custom look-and-feel support in Bridge

ermin-muratovic:


warber (approved):
tested also on my local installation, works :ok:

PR #4310, opened 2021-06-14
#3890 keptn install: check for pending operations

### LINKED ISSUE #3890: Keptn installation fails when upgrading to newer version
When I try to upgrade from `0.8.1-dev-PR-3529` to `0.8.2` via the `install` command the following error appears 

```
Error: Could not complete Keptn installation: Error when installing/upgrading Helm Chart keptn in namespace keptn: another operation (install/upgrade/rollback) is in progress
```

OS: Windows 10
CLI Version: 0.8.2

christian-kreuzberger-dtx:
Fixes #3890

I added a check and a proper error message in case an upgrade is still pending.

### DIFF
--- a/cli/pkg/helm/helm_helper.go
+++ b/cli/pkg/helm/helm_helper.go
@@ -176,7 +176,7 @@ func (c Helper) UpgradeChart(ch *chart.Chart, releaseName, namespace string, val
 		}
 
 		histClient := action.NewHistory(cfg)
-		var release *release.Release
+		var newRelease *release.Release
 
 		timeoutInMinutes := 10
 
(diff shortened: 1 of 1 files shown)


cli/pkg/helm/helm_helper.go
```diff
@@ -178,21 +178,35 @@ func (c Helper) UpgradeChart(ch *chart.Chart, releaseName, namespace string, val
...
 			iCli.Wait = true
-			iCli.Timeout = 10 * time.Minute
+			iCli.Timeout = time.Duration(timeoutInMinutes) * time.Minute
 			release, err = iCli.Run(ch, vals)
 		} else {
+			logging.PrintLog("Found existing installation, overwriting...", logging.InfoLevel)
+
+			// check if the previous installation is still pending (e.g., waiting to complete)
+			for _, release := range releases {
+				if release.Info.Status == "pending-upgrade" {
```
bacherfl, line 196: should we also check for `release.StatusPendingInstall, release.StatusPendingUpgrade and release.StatusPendingRollback` here?

christian-kreuzberger-dtx: Good idea. As we were already using `release` as a variable, I renamed it to `newRelease` and changed the if clause.

bacherfl (approved):

PR #4316, opened 2021-06-14
#2479 Switch to nats 0.8.0

### LINKED ISSUE #2479: Run docker container as non-root users
Over the time we have had several issues with restrictive OpenShift environments, whereas the root-cause was the fact that almost all of our containers run processes as root, but the persistent volumes on OpenShift require access as a non-root user.

https://github.com/keptn/keptn/issues/2455
https://github.com/keptn/keptn/issues/2453
https://github.com/keptn/keptn/issues/2157
https://github.com/keptn/keptn/issues/1951

To fix this, multiple approaches need to be considered:

1. Fix the Docker images by adding the following:
```
FROM alpine:3.x
RUN adduser -D myuser
USER myuser
```

2. Ensure all third-party images (e.g., nginx, mongodb, nats, ...) run as non-root users (we believe they already do, but let's double check)

3. Ensure all PVs and PVCs (this will mainly affect configuration-service) still work with those changes

4. Ensure Upgrading from an old Keptn version to the new Keptn version works

**Additional info:**

So far, we found a StackOverflow post and a script that might fix the problem (using an initcontainer):

* https://stackoverflow.com/questions/55419927/permission-denied-when-changing-permissions-on-pv-with-init-container
* https://github.com/sclorg/s2i-base-container/blob/master/core/root/usr/bin/fix-permissions#L21

## DoD 
- [ ] List the images that are running as a root user

christian-kreuzberger-dtx:

Finally fix #2479 by switching to nats 0.8.0.
Apparently something has changed with the service-naming, hence I needed to add `fullnameOverride`.

Integration Tests: https://github.com/keptn/keptn/actions/runs/938419801

### DIFF
--- a/installer/manifests/keptn/charts/control-plane/Chart.yaml
+++ b/installer/manifests/keptn/charts/control-plane/Chart.yaml
@@ -27,5 +27,5 @@ dependencies:
     version: 0.1.0
     condition: mongodb.enabled
   - name: nats
-    version: 0.7.4
+    version: 0.8.0
     repository: https://nats-io.github.io/k8s/helm/charts/
--- a/installer/manifests/keptn/charts/control-plane/values.yaml
+++ b/installer/manifests/keptn/charts/control-plane/values.yaml
@@ -11,6 +11,7 @@ keptnSpecVersion: latest
 
 nats:
   nameOverride: keptn-nats-cluster
+  fullnameOverride: keptn-nats-cluster
   nats.cluster.replicas: 3
 
   natsbox:


warber (approved):

christian-kreuzberger-dtx:
We ran into an issue when upgrading from an older version to this:
```
Error: Could not complete Keptn installation: Error when installing/upgrading Helm Chart keptn in namespace keptn: cannot patch "keptn-nats-cluster" with kind StatefulSet: StatefulSet.apps "keptn-nats-cluster" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden 

```

PR #4317, opened 2021-06-14
#3801 allow text break for info-group content

### LINKED ISSUE #3801: Bridge Service Screen - On small screens the "Service not deployed" text is partly vanishing
Problem:
When the screen on browser is resized to be smaller, part of the text on the service screen for "Service not deployed yet" is hidden by the pane borders.

Solution:
The text should break, so that the whole text is shown. The pane resizes with the longer text.

Note: Investigate whether there are other screens with UI components that have the same problem.

image

Kirdock:
Closes #3801
Service screen:
image

Environment screen:
image

### DIFF
--- a/bridge/client/styles.scss
+++ b/bridge/client/styles.scss
@@ -38,6 +38,10 @@ app-root {
   white-space: nowrap;
 }
 
+.dt-info-group-content > *{
+  white-space: normal;
+}
+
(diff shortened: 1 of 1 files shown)


christian-kreuzberger-dtx (approved):
LGTM

PR #4320, opened 2021-06-14
#4030 forward logs to log api

### LINKED ISSUE #4030: Forward log messages of execution plane Keptn-services to Keptn core
# User Story: Forward log messages of execution plane Keptn-services to Keptn core

As a user, I want to see error (log) message of my execution plane Keptn-services without having to use `kubectl`. Therefore, those error messages need to be forwarded to Keptn core.

## Detail

Execution plane Keptn-services are usually started on Kubernetes and have a distributor as a sidecar. The service itself does not know about the sidecar, though it communicates with it via some REST API Endpoints (that are basically forwarded to Keptn control-plane).
Currently, log messages of the Keptn-service are just printed to stdout/stderr (depending on the logging framework). Those log messages can only be seen by calling `kubectl -n some-namespace logs deployment/my-service my-service`.

### Proposal
Capturing errors can be done in the Distributor of a Keptn Service, which would capture error logs (sent to std::err) of the side-car container (the service) and forward them to the keptn control plane
:exclamation: Note that calling the errors API shall be limited to one call per minute. (Implement `throttler` in go-utils)
:exclamation: Beware of possible refactoring tasks in the distributor
:exclamation: This operation must be allowed to fail (e.g. old Keptn version vs. new Distributor)

## Definition of Done

- [ ] Error Log Messages of Keptn-services are forwarded to Keptn core in a way that does not require implementation changes in the Keptn-service.

## Dependencies
#4032 

## Resources

- How to capture logs from as a side-car container https://superuser.com/a/1530503

bacherfl:
Closes #4030


distributor/deploy/distributor.yaml
```diff
@@ -42,6 +79,18 @@ spec:
         - name: PUBSUB_URL
           value: 'nats://keptn-nats-cluster'
         - name: PUBSUB_TOPIC
-          value: 'sh.keptn.event.configuration.change,sh.keptn.events.deployment-finished'
+          value: 'sh.keptn.event.get-action.triggered'
         - name: PUBSUB_RECIPIENT
           value: 'helm-service'
+        - name: MY_POD_NAME
```
warber, line 85: what about these changes?

distributor/pkg/lib/uniformlog.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}()
+}
+
+func (l *EventUniformLog) OnEvent(event cloudevents.Event) error {
+	keptnEvent, err := keptnv2.ToKeptnEvent(event)
+	if err != nil {
+		return fmt.Errorf("could not decode CloudEvent to Keptn event: %v", err.Error())
+	}
+	if keptnEvent.Type == nil {
```
warber, line 55: should this check be made inside ToKeptnEvent ?

distributor/pkg/lib/uniformlog.go
```diff
@@ -0,0 +1,103 @@
...
+		return errors.New("received Keptn Event has no type")
+	}
+	if strings.HasSuffix(*keptnEvent.Type, ".finished") {
```
warber, line 58: should we introduce a func in go-utils for checking if an event is a finished event?

distributor/pkg/lib/uniformlog.go
bacherfl: I just had a look at the implementation of `ToKeptnEvent` - this function already makes sure that the Type is not nil, so this check was unneccesary anyway

warber (approved):

PR #4326, opened 2021-06-14
#4022 introduced nested routes for project board

### LINKED ISSUE #4022: Refactor Angular Router Usage in Bridge
During implementation of https://github.com/keptn/keptn/issues/3944 I noticed that Angular Routing is heavily misused in the project-board component.

For instance, handling the final view/screen is handled via the following switch case

https://github.com/keptn/keptn/blob/da5859fb8291894c6b019d60781834db4192eb09/bridge/client/app/project-board/project-board.component.html#L64-L71

in combination with a subscription
```
    this.route.params
      .subscribe(params => {
          // ...  

          if(this.route.snapshot.url.length > 2) {
            this.view = this.route.snapshot.url[2].path;
          } else {
            this.view = 'environment';
          }
        }
      });
```

as well as the following "links" in the menu:
https://github.com/keptn/keptn/blob/da5859fb8291894c6b019d60781834db4192eb09/bridge/client/app/project-board/project-board.component.html#L38-L48
which use `selectView` or `redirectView` depending on what is clicked...

Weirdly enough we do have all sub-pages listed as valid routes in app.routing.ts:
https://github.com/keptn/keptn/blob/da5859fb8291894c6b019d60781834db4192eb09/bridge/client/app/app.routing.ts#L12-L23

This leads to all kind of strange behaviours, as we are effectively circumventing Angular Router in a way it's not meant to be.

# Proposed Solution

Use nested routes. Period.

This will require some refactoring, but it will allow us

* to grow without having to consider every special case for routing
* to use links in the menu instead of buttons/click actions (allowing us to right-click and "open in new tab")

In addition, the code should get better maintainable.

Kirdock:
Closes #4022 

### DIFF
--- a/bridge/client/app/app.routing.ts
+++ b/bridge/client/app/app.routing.ts
@@ -1,26 +1,37 @@
 import {NgModule} from '@angular/core';
-import {Routes, RouterModule} from '@angular/router';
+import {Routes, RouterModule, ExtraOptions} from '@angular/router';
 import {DashboardComponent} from './dashboard/dashboard.component';
 import {ProjectBoardComponent} from './project-board/project-board.component';
 import {EvaluationBoardComponent} from './evaluation-board/evaluation-board.component';
 import {ForwarderGuard} from './_forwarder/forwarder_guard';
+import {KtbUniformViewComponent} from './_views/ktb-uniform-view/ktb-uniform-view.component';
+import {KtbIntegrationViewComponent} from './_views/ktb-integration-view/ktb-integration-view.component';
+import {KtbSettingsViewComponent} from './_views/ktb-settings-view/ktb-settings-view.component';
+import {KtbServiceViewComponent} from './_views/ktb-service-view/ktb-service-view.component';
+import {KtbSequenceViewComponent} from './_views/ktb-sequence-view/ktb-sequence-view.component';
+import {KtbEnvironmentViewComponent} from './_views/ktb-environment-view/ktb-environment-view.component';
 
+const routingConfiguration: ExtraOptions = {
+  paramsInheritanceStrategy: 'always'
+};
 
 const routes: Routes = [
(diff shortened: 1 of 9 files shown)


bridge/client/app/app.routing.ts
```diff
@@ -1,26 +1,37 @@
...
-  {path: 'project/:projectName/service', component: ProjectBoardComponent},
-  {path: 'project/:projectName/service/:serviceName', component: ProjectBoardComponent},
-  {path: 'project/:projectName/service/:serviceName/context/:shkeptncontext', component: ProjectBoardComponent},
-  {path: 'project/:projectName/service/:serviceName/context/:shkeptncontext/stage/:stage', component: ProjectBoardComponent},
-  {path: 'project/:projectName/sequence', component: ProjectBoardComponent},
-  {path: 'project/:projectName/sequence/:shkeptncontext', component: ProjectBoardComponent},
-  {path: 'project/:projectName/sequence/:shkeptncontext/event/:eventId', component: ProjectBoardComponent},
-  {path: 'project/:projectName/sequence/:shkeptncontext/stage/:stage', component: ProjectBoardComponent},
+  {path: 'project/:projectName', component: ProjectBoardComponent, children: [
+      {path: '', pathMatch: 'full', component: KtbEnvironmentViewComponent},
```
ermin-muratovic, line 22: I would suggest to provide the environment view at `path: 'environment'` and use a redirect rule for `path: ''`

ermin-muratovic (changes requested):

ermin-muratovic (approved):
lgtm

bridge/client/app/app.routing.ts
ermin-muratovic: After a discussion with @Kirdock, we leave it as it is for now, unless we introduce deep links for the environment view.

Kirdock: I would change it when we really need it

PR #4358, opened 2021-06-16
#3423 Fix apikey name in swagger docs

### LINKED ISSUE #3423: Swagger-UI: Share Authorization/API Token between services
With 0.7.x it was enough to authenticate once in our Swagger-UI, and the token was shared between all other microservices (mongodb, configuration-service):

image

With 0.8.0-rc1 we have introduced new microservices (e.g., control-plane, statistics-service) that use a different Token Name, therefore we lose authentication in those two microservices within swagger-ui:

image

An easy fix would be to ensure that all microservices use the same token name in swagger-ui.

# Definition of Done

- [ ] When authenticating in swagger-ui, the auth token is shared across all microservices

christian-kreuzberger-dtx:
Fixes #3423 

With this PR, I've added some more fixes related to #3423 
* remove `@security` definitions from all endpoints
* remove `swagger.yaml/json` files from `docs/` subfolder
* Auto-generate `swagger.yaml/json/docs.go` files in `docs/` subfolder during docker build
* Update README of services accordingly

In addition, I noticed that the `version` field was populated with 1.0 all the time. I've aligned it with our other swagger docs to be replaced by the actual Keptn version.


christian-kreuzberger-dtx:
Closing this as the issue itself is marked as wontfix/not fixable.

PR #4378, opened 2021-06-16
Align gin gonic based services to use Keptn version, fix Dockerfiles and swagger-ui

christian-kreuzberger-dtx:

While trying to fix issue https://github.com/keptn/keptn/issues/3423 I discovered that there are several things that need to be aligned in gin gonic based services.

* version is properly set (instead of 1.0)
* Dockerfile synced
* Generation of docs / swagger.yaml moved to Dockerfile
* Duplicated docs.go / swagger.yaml / swagger.json in secret-service removed

# Proof it works
image

image

image


bacherfl (approved):
lgtm

PR #4395, opened 2021-06-17
#4188 refactoring of observables

### LINKED ISSUE #4188: Refactor observables inside of router parameter subscription
Currently there are some observables and subscriptions inside of `router.params.subscribe(...)`.
Because of that there may be several registered observables instead of one, because the parameters may change without calling `onDestroy()` in the component.
One example would be the `loadRoots`-interval in the project-board. This timer is registered every time but never removed when the project is changed via the dropdown in the header.

*Research*: Is it possible to count open subscriptions? 

## Definition of Done
- [x] Code is reformatted in a way, that observables are only registered once.
- [ ] Unit test that checks the `unsubscribe mechanism.

Kirdock:
Closes #4188  

### DIFF
--- a/bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.ts
+++ b/bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.ts
@@ -95,8 +95,8 @@ export class KtbSequenceViewComponent implements OnInit, OnDestroy {
       .pipe(
         map(params => params.projectName)
       );
-    this.roots$ = this.dataService.roots;
-    const roots$ = this.dataService.roots
+
+    this.roots$ = this.dataService.roots
       .pipe(
         takeUntil(this.unsubscribe$),
         filter(roots => roots?.length > 0)
@@ -130,7 +130,7 @@ export class KtbSequenceViewComponent implements OnInit, OnDestroy {
     });
 
     // init; set parameters
-    combineLatest([this.route.params, roots$])
+    combineLatest([this.route.params, this.roots$])
       .pipe(
         takeUntil(this.unsubscribe$),
         take(1)
@@ -146,7 +146,7 @@ export class KtbSequenceViewComponent implements OnInit, OnDestroy {
         }
     });
 
-    roots$.subscribe(roots => {
+    this.roots$.subscribe(roots => {
       this.updateFilterSequence(roots);
       this.refreshFilterDataSource();
       // Set unfinished roots so that the traces for updates can be loaded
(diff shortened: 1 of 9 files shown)


bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.ts
```diff
@@ -91,64 +91,68 @@ export class KtbSequenceViewComponent implements OnInit, OnDestroy {
   constructor(private _changeDetectorRef: ChangeDetectorRef, private dataService: DataService, private route: ActivatedRoute, public dateUtil: DateUtil, private router: Router, private location: Location) { }
 
   ngOnInit() {
-    this.currentSequence = null;
-    this.route.params
-      .pipe(takeUntil(this.unsubscribe$))
-      .subscribe(params => {
+    const projectName$ = this.route.params
```
ermin-muratovic, line 94: `projectName$` is used only once, so is it really necessary? it's difficult to read, having first the declaration of `projectName$`, then some lines of other code in between and then a one-time usage of `projectName$`

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.ts
```diff
@@ -91,64 +91,68 @@ export class KtbSequenceViewComponent implements OnInit, OnDestroy {
...
+      .pipe(
+        map(params => params.projectName)
+      );
+    this.roots$ = this.dataService.roots;
+    const roots$ = this.dataService.roots
```
ermin-muratovic, line 99: What's the difference between `this.roots$` and `const roots$` and can we combine those to prevent confusion?

bridge/client/app/project-board/project-board.component.ts
```diff
@@ -25,67 +25,79 @@ export class ProjectBoardComponent implements OnInit, OnDestroy {
...
-              }
-            });
+    const projectName$ = this.route.params.pipe(
+      map(params => params.projectName),
+      filter(projectName => projectName)
+    );
+
+    this.project$ = projectName$.pipe(
+      switchMap(projectName => this.dataService.getProject(projectName))
+    );
```
ermin-muratovic, line 38: Same as the comment above, do we need the `const projectName$` if it is used only once or could we combine those lines in one `pipe()`?

ermin-muratovic (changes requested):

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.ts
Kirdock: The difference is that the second one skips empty roots. This is necessary because later I set the parameters for this page and only take 1. Because we use `BehaviorSubject` the initial value is also emitted

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.ts
Kirdock: I don't think it's harder to read. You clearly see on what `this.project$` depends on, when you assign it.

ermin-muratovic (approved):
lgtm

PR #4411, opened 2021-06-18
Improve sequence selection UX

### LINKED ISSUE #3976: Better UX to show which sequence is currently selected
For a selected sequence the background of the tile header gets a bit of a darker grey shade. This is very unreadable - especially on screens with bad contrast. Also if you read the content on the right and want to know in which sequence you are - it's nearly impossible to recognize the difference between the greys.

My suggestion here is to use a darker grey and to add some sort of bar on the end. This would improve readability of the selected state a lot.

Frame 1

laneli:
Fixes #3976 

image

* Improved hovering
* Improved visibility of selected item

### DIFF
--- a/bridge/client/app/_components/ktb-root-events-list/ktb-root-events-list.component.scss
+++ b/bridge/client/app/_components/ktb-root-events-list/ktb-root-events-list.component.scss
@@ -3,6 +3,20 @@
 
 .ktb-root-events-list {
 
+  .ktb-selectable-tile.ktb-tile-error,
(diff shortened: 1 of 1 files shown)


ermin-muratovic (changes requested):
I would suggest to make the left border wider on selection instead of adding a right border, in the color it is right now (green or red depending on the state) and just make the background color one tick darker on hover.

ermin-muratovic (approved):
lgtm

PR #4434, opened 2021-06-22
#3999 Add github actor to integration test report

### LINKED ISSUE #3999: Integration Test Report: Note down the Actor or whether it was scheduled
Right now when triggering integration tests manually (e.g., for a release branch) and an issue is created because of failing tests, it is hard to find out who triggered it, e.g.:

image

It should be possible to add the so called `GITHUB_ACTOR`, see https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables.

Even better would be if the actor is also mentioned, e.g., @christian-kreuzberger-dtx 

# Definition of Done

- [ ] Integration test error issues contain the actor

mowies:
**This PR**
* adds a line for who triggered the integration tests in the report
* fixes issue creation step for failing integration tests

Closes #3999


christian-kreuzberger-dtx (approved):
LGTM

Issue #4331, opened 2021-06-15
Unify Go versions in CI workflow


mowies:
The CI flow has the currently supported Go version defined in multiple places. This could be unified to make Go version upgrades easier and less prone to errors.

Occurences:
https://github.com/keptn/keptn/blob/master/.github/workflows/CI.yml#L308
https://github.com/keptn/keptn/blob/master/.github/workflows/CI.yml#L461
https://github.com/keptn/keptn/blob/master/.github/workflows/CI.yml#L525

PR #4444, opened 2021-06-22
#4331 Unify go versions in CI workflow

mowies:
**This PR**
* replaces all hardcoded occurrences of a go version with a global env variable that is set at the beginning of the CI workflow


christian-kreuzberger-dtx (approved):
LGTM

PR #4484, opened 2021-06-23
Fix e2e test timeout

laneli:
Protractor waits for async tasks to complete - in case of rxjs timer they do not finish so the tests time out
To tackle this it is necessary to run this kind of tasks in ngZone.runOutsideAngular in order to not block Angular.
https://github.com/angular/protractor/blob/master/docs/timeouts.md#waiting-for-angular

### DIFF
--- a/bridge/client/app/dashboard/dashboard.component.ts
+++ b/bridge/client/app/dashboard/dashboard.component.ts
@@ -15,7 +15,7 @@ export class DashboardComponent implements OnInit, OnDestroy{
   public logoInvertedUrl = environment?.config?.logoInvertedUrl;
   public isQualityGatesOnly: boolean;
 
-  private readonly _projectTimerInterval = 5 * 1000;
+  private readonly _projectTimerInterval = 30 * 1000;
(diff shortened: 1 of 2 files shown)


bridge/client/app/dashboard/dashboard.component.ts
```diff
@@ -15,10 +15,10 @@ export class DashboardComponent implements OnInit, OnDestroy{
   public logoInvertedUrl = environment?.config?.logoInvertedUrl;
   public isQualityGatesOnly: boolean;
 
-  private readonly _projectTimerInterval = 30 * 1000;
+  private readonly _projectTimerInterval = 5 * 1000;
```
christian-kreuzberger-dtx, line 18: Is this on purpose? I think 30 seconds was already okay, wasn't it?

laneli: Yes you are right, this was a leftover from testing

christian-kreuzberger-dtx (approved):
LGTM

ermin-muratovic (approved):
lgtm

Issue #4385, opened 2021-06-17
Fix random error messages in integration test workflow


christian-kreuzberger-dtx:
Currently, our integration test spits out random errors that are hard to trace:
![image](https://user-images.githubusercontent.com/56065213/122366657-0a333100-cf5c-11eb-8aef-f64d86ba668d.png)
Example: https://github.com/keptn/keptn/actions/runs/945026997

After some investigation, I found that those errors are coming from the following step:
![image](https://user-images.githubusercontent.com/56065213/122372611-18d01700-cf61-11eb-826e-5d04b9be814a.png)

# Definition of Done

- [ ] Code for this step is fixed and errors are no longer presented, unless there really is an error. 
- [ ] Errors presented from this step show a clear message

PR #4495, opened 2021-06-24
#4385 Remove debug build steps

mowies:
**This PR**
* removes some debug steps from the integration test pipeline

**Notes**
Removing the debug steps should be ok since they were there for minishift and some were expected to fail. This resulted in error annotations on the integration test builds that didn't contain any useful information


christian-kreuzberger-dtx (approved):
LGTM

PR #4563, opened 2021-07-01
#4384 fallback if configuration change image is not a string

### LINKED ISSUE #4384: Shipyard-controller and Bridge run into errors when using an `image` object for a configuration change
## User story

As a user, I have a Helm Chart with a `values.yaml` file that looks as follows:
```
image:
  repository: docker.io/keptnexamples/carts
  tag: 0.11.1
```

Consequently, I would like to trigger a new delivery for this Helm Chart by sending an event: 

```
{
  "data": {
    "configurationChange": {
      "values": {
        "image": {
          "repository": "docker.io/keptnexamples/carts",
          "tag": "0.11.3"
        }
      }
    },
    "deployment": {
      "deploymentURIsLocal": null,
      "deploymentstrategy": ""
    },
    "project": "sockshop",
    "service": "carts",
    "stage": "dev"
  },
  "source": "https://github.com/keptn/keptn/cli#configuration-change",
  "specversion": "1.0",
  "time": "2021-06-17T07:44:29.173Z",
  "type": "sh.keptn.event.dev.delivery.triggered"
}
```

### Bug Shipyard-controller

* The shipyard-controller runs into an error since it cannot deal with an `image` object in `values`. 
* If I rename `image` to `my_image`:

``` 
      "values": {
          "my_image": {
            "repository": "docker.io/keptnexamples/carts",
            "tag": "0.11.3"
        }
      }
```

everything works fine. Nevertheless, this does not reflect the `values.yaml` file of the Helm Chart. 

### Bug Bridge

When sending the above-mentioned event to Keptn, the Bridge runs into an Error: 

image

* Implement a fallback to not show the image:tag if it is not available. > The bridge should still continue to work

### Acceptance criteria: 

- [ ] I can send the above-mentioned event to Keptn and the image gets deployed by the helm-service.

Kirdock:
Fixes #4384 

### DIFF
--- a/bridge/client/app/_models/trace.ts
+++ b/bridge/client/app/_models/trace.ts
@@ -64,7 +64,7 @@ class Trace {
     };
     configurationChange: {
       values: {
-        image: string
+        image: any
       }
     };
 
@@ -274,10 +274,14 @@ class Trace {
       else if(this.data.image)
         this.image = this.data.image.split("/").pop();
       else if(this.data.configurationChange?.values)
-        this.image = this.data.configurationChange.values.image?.split("/").pop();
+        this.image = this.getConfigurationChangeImage();
     }
 
     return this.image;
+  }
+
+  public getConfigurationChangeImage(): string {
+    return typeof this.data.configurationChange.values.image === 'string' ? this.data.configurationChange.values.image.split('/').pop() : ''
   }
 
   getProject(): string {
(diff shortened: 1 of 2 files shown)


bridge/client/app/_models/trace.ts
```diff
@@ -64,7 +64,7 @@ class Trace {
     };
     configurationChange: {
       values: {
-        image: string
+        image: any
```
laneli, line 67: Use also here `string | undefined`

laneli: 👌

laneli (approved):

PR #4567, opened 2021-07-01
Improve tile width for environment screen

### LINKED ISSUE #4146: Environment screen always has scrollbars when having more than 2 stages
This is with Keptn 0.8.3.

# Current Behaviour

I am looking at the environment screen for sockshop, and it always uses up more horizontal space than it has, leading to scrollbars showing.
This is really horrible on a widescreen, e.g.:
image

At the same time, it shrinks down but still shows scrollbars when I am on a Full HD display

image

# Expected Behaviour

Scrollbars on the Full HD screen are okay when there are 3 stages. But when I move to a bigger screen, I expect the stage-tiles to not grow in size, but stay at their current size, such that the scrollbar disappears.

## Two possible options:
- Define a fixed size for the stage tile
- Define a fixed size until a specified screen size and then have a variable tile width.

Please consider: Three stages must be shown completely. 

# Defnition of Done

- [ ] Stage tiles don't grow depending on the screen size, but stay at a fixed size (or at least the sizing is reasonable depending on the screensize -> responsive layout).

laneli:
Closes #4146 

When only 3 tiles are available the size of the tiles is adapted so that all of them fit on the screen:
image

When more than 3 tiles are available in the row, the panel size is reduced to show that it can be scrolled:
image

### DIFF
--- a/bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.html
+++ b/bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.html
@@ -12,7 +12,7 @@
     ></dt-filter-field>
   </div>
   <div class="container p-0" fxFlex fxLayout="column">
-    <div class="stage-list" fxLayout="row" fxLayout.lt-sm="column" fxLayoutGap="15px" fxLayoutAlign="start">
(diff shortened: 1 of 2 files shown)


Kirdock (changes requested):
Consider a `min-width` (`350px` or more) for stage tiles.
For example if the screen width is 1200px, stages look like this:
image

Kirdock (approved):

PR #4627, opened 2021-07-12
#3803 hover text for icon-buttons

### LINKED ISSUE #3803: Show alt text / tooltip for icon buttons
## User Story
As a user, I want to see what action will be performed by the Bridge when hovering over the buttons so that I can easily learn how to use Keptn Bridge and avoid triggering unwanted actions.

### Details
* Add either alt texts or tooltips that show on hovering over the buttons. This addresses buttons like:
  * Copy to clipboard
  * Reveal values
  * Open in new window
  * Show raw / payload
  * Show chart
  * etc.

* Also, think about using better icons at some places if necessary.

* Show tooltip that explains the meaning of the score: `The score represents the portion an SLI contributes to the overall evaluation result between 0 and 100.` 

image

Kirdock:
Closes #3803 
Also fixes traces being in the wrong sequence.


ermin-muratovic (approved):
lgtm, please just make sure that the _bugfix_ is also mentioned in the merge message

PR #4651, opened 2021-07-14
#4443 rewrote bridge server to typescript

### LINKED ISSUE #4443: Update Bridge server to TS and ESM
A dependabot PR introduced a new version of a dependency that relies solely on ESM.
As our Bridge server does not support ESM yet, we should consider to switch to be future proof. In the same step it would make sense to migrate to TS in order to be type safe.

The related PR:
https://github.com/keptn/keptn/pull/4208

Also see https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c of the 4.0.0 crypto-random-string release notes.

Kirdock:
Closes #4443


ermin-muratovic (approved):
lgtm

PR #4659, opened 2021-07-15
#4658 wrong http-status-code

### LINKED ISSUE #4658: Bridge server returns wrong http-status-code
If for example a project is not found, the bridge server returns `500` instead of `404`.

Change `error.status` to `error.response.status`
https://github.com/keptn/keptn/blob/fdcd06bb3f62e7b8f487763145b651d39db64eb0/bridge/server/app.ts#L235

Kirdock:
Fixes #4658 


agrimmer (approved):
LGTM

PR #4670, opened 2021-07-16
fix(cli): Fixed `KubeConfigCheck` configuration flag

### LINKED ISSUE #4553: CLI to remove warning when KUBECONFIG file is missing
I thought I may have seen this issue/request reported before - but - couldnt find it

As a User I find it very misleading when I use the Keptn CLI on a terminal where I dont have a kubeconfig. Many Keptn users will use the Keptn CLI on ANY terminal withouth any access to the k8s cluster where Keptn runs on. Therefore the warning that KUBECONFIG file cannot be open is very misleading.
Here is an example:

```
$ keptn trigger delivery --project=myproject --service=simplenode --stage=staging --image=myimage
Warning: could not open KUBECONFIG file: Cannot find file /home/ec2-user/.kube/config
```

Especially when using the keptn CLI in automation script where it gets called a lot you end up with a lot of warning text output that is very misleading to the end user

When the error is printed, it should also notify the user about the possibility to run 
```
keptn set config kubeContextCheck false
```

When set `keptn set --help` get the list of config that I can disable

### What to fix

- [x] When `kubeContextCheck` is set to `false`, the Keptn CLI should not look for a `kubeconfig`.
- [ ] When a `kubeconfig` is expected but it cannot be found, the Keptn CLI should print the possibility to run `keptn set config kubeContextCheck false`
- [x] `keptn set --help` should print the list of config entries that a user can disable

warber:
fixes #4553 

With this PR without having a`~/.kube/config` file, the CLI behaves as follows.

**`KubeContextCheck` enabled (default)**
```
$ ~ ./keptn set config KubeContextCheck true
$ ~ ./keptn get project
Warning: could not parse KUBECONFIG file: Cannot find file /home/bernd/.kube/config
Hint: If you don't have 'kubectl' installed you can disable this check via 'keptn-local set config KubeContextCheck false'
NAME			CREATION DATE			SHIPYARD VERSION
...
```

**`KubeCOntextCheck` disabled**
```
$ ~ ./keptn set config KubeContextCheck false
$ ~ ./keptn get project
NAME			CREATION DATE			SHIPYARD VERSION
...
```

**`--help` text**
```
$ ~ keptn set config --help
Sets flags of the CLI configuration, which is stored in $HOME/.keptn/config.

This command takes a key and a new value as arguments and applies them to the local keptn CLI configuration
Supported keys:
 * AutomaticVersionCheck: true|false
 * KubeContextCheck: true|false

Usage:
  keptn set config [flags]

Examples:
keptn set config AutomaticVersionCheck false

Global Flags:
  -h, --help               help
      --mock               Disables communication to a Keptn endpoint
  -n, --namespace string   Specify the namespace where Keptn should be installed, used and uninstalled in (default "keptn")
  -q, --quiet              Suppresses debug and info messages
  -v, --verbose            Enables verbose logging to print debug messages
  -y, --yes                Assume yes for all user prompts

```

### DIFF
--- a/cli/pkg/credentialmanager/handler.go
+++ b/cli/pkg/credentialmanager/handler.go
@@ -178,9 +178,9 @@ func getCurrentContextFromKubeConfig() {
 	fileContent, err := fileutils.ReadFile(kubeconfig)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Warning: could not parse KUBECONFIG file: "+err.Error()+"\n")
-		fmt.Println("Hint: you can disable this check via 'keptn set config KubeContextCheck true'")
-	} else if err = yaml.Unmarshal(fileContent, &kubeConfigFile); err != nil {
 		fmt.Println("Hint: If you don't have 'kubectl' installed you can disable this check via 'keptn-local set config KubeContextCheck false'")
+	} else if err = yaml.Unmarshal(fileContent, &kubeConfigFile); err != nil {
+		fmt.Fprintf(os.Stderr, "Warning: could not parse KUBECONFIG file: "+err.Error()+"\n")
 	}
 }
 
--- a/cli/pkg/credentialmanager/handler.go
+++ b/cli/pkg/credentialmanager/handler.go
@@ -147,22 +147,23 @@ func handleCustomCreds(configLocation string, namespace string) (url.URL, string
 
 // initChecks needs to be run when credentialManager is called or initialized
 func initChecks(autoApplyNewContext bool) {
-	if !GlobalCheckForContextChange {
-		cliConfigManager := config.NewCLIConfigManager()
-		err := getCurrentContextFromKubeConfig()
-		if err != nil {
-			log.Fatal(err)
-		}
+	cliConfigManager := config.NewCLIConfigManager()
+	cliConfig, err := cliConfigManager.LoadCLIConfig()
+	if err != nil {
+		log.Fatal(err)
+	}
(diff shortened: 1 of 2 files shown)


cli/pkg/credentialmanager/handler.go
```diff
@@ -175,17 +176,12 @@ func getCurrentContextFromKubeConfig() error {
...
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "Warning: could not open KUBECONFIG file: "+err.Error()+"\n")
-		return nil
-	}
-
-	err = yaml.Unmarshal(fileContent, &kubeConfigFile)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Warning: could not parse KUBECONFIG file: "+err.Error()+"\n")
-		return nil
+		fmt.Println("Hint: If you don't have 'kubectl' installed you can disable this check via 'keptn-local set config KubeContextCheck false'")
```
mowies, line 181: shouldn't this message refer to the kubeconfig file or the kube context instead of `kubectl`?

mowies (changes requested):

cli/pkg/credentialmanager/handler.go
thisthat:
```suggestion
		fmt.Println("Hint: If you don't have 'kubectl' installed you can disable this check via 'keptn set config KubeContextCheck false'")
```

mowies (approved):

PR #4674, opened 2021-07-19
Fix sonar sanitization security hotspot

laneli:

### DIFF
--- a/bridge/tslint.json
+++ b/bridge/tslint.json
@@ -89,7 +89,7 @@
     ],
     "trailing-comma": false,
     "no-conflicting-lifecycle": true,
-    "no-host-metadata-property": true,
+    "no-host-metadata-property": false,
     "no-input-rename": true,
     "no-inputs-metadata-property": true,
     "no-output-native": true,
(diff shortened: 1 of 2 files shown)


bridge/tslint.json
```diff
@@ -89,7 +89,7 @@
     ],
     "trailing-comma": false,
     "no-conflicting-lifecycle": true,
-    "no-host-metadata-property": true,
+    "no-host-metadata-property": false,
```
mowies, line 92: instead of disabling this rule globally, you could just add the tslint disable comment for the single line where it's needed

mowies (changes requested):

bridge/tslint.json
laneli: I taked with Ermin about this ... we have this host metadata for our components on several files. I researched a bit and it totally valid to use this instead of @HostListener decorator in angular. We want to continue to use this as it is and therefore it makes sense to disable the rule in tslint.

mowies (approved):

PR #4675, opened 2021-07-19
feat(shipyard-controller): Automatic expiration of Uniform integrations (#4615)

### LINKED ISSUE #4615: De-Registration of uniform services fails on upgrade scenario
## Problem 

When **upgrading** a Keptn installation, the deregistration process of keptn services fails.
The reason for this is, that also the `ShipyardController` which is providing the registration interface is restarted at the same time frame in which the keptn services have the chance to deregister themselves.

This results in multiple entries of the same service(s) in the uniform screen.

## Workaround

The duplicated entries could be identified and deleted "manually" using the `DELETE` endpoint of the uniform API.

## Solution

The current concept is too fragile and needs to be changed.

### Proposal
Services should have the possibility to periodically register themselves (as is right now) but don't have the need to explicitly deregister themselves on service shutdown. Instead, there should be a TTL in place which expires when the given service stops refreshing the registration.

bacherfl:
Closes #4615 

The TTL value for uniform integrations can be defined using the following property of the `values.yaml` of the control plane helm-chart:

```
shipyardController.config.uniformIntegrationTTL // default "2m"
```


warber (dismissed):

warber (approved):

shipyard-controller/main.go
```diff
@@ -166,6 +168,11 @@ func main() {
 	watcher.Run(context.Background())
 
 	uniformRepo := createUniformRepo()
+	err = uniformRepo.SetupTTLIndex(getDurationFromEnvVar(envVarUniformIntegrationTTL, envVarUniformTTLDefault))
+	if err != nil {
+		log.WithError(err).Error("could not setup TTL index for log repo entries")
```
warber, line 173: wrong error message

warber (changes requested):

warber (changes requested):
looks good, however, changing the `TTL` value results in an `IndexOptionsConflict` error, hence the index will not be changed.

warber (approved):

thisthat (approved):

PR #4679, opened 2021-07-19
feat(api): rate limiter for auth endpoint (#4323)

### LINKED ISSUE #4323: Authenticating (1) a user on Bridge or (2) Keptn CLI on `/auth` endpoint are not throttled
## Action Item

*Details:* A user can make limited use of the APIs: `bridge login` (to login to the Bridge using basic auth) and `/auth` (to authenticate Keptn CLI). Hence, a brute-force attack could be used to guess the password. 

**Task:**
- [x] Throttle basic auth requests in bridge-server 
- [x] Throttle requests to /auth in api-service

Throttling shall be done per IP address.

### Definition of Done
- [x] Implement Throttling of requests to avoid Brute Force attacks

bacherfl:
Closes #4323

The rate limiting can be configured using the following values in the `values.yaml` file of the control plane helm-chart:

```
apiService.config.maxAuthRequestsPerSecond // the maximum number of requests per second (default=1.0)
apiService.config.maxAuthRequestBurst // maximum burst size of auth requests (default=2)
```

Note: The rate limiter is only applied for failed authentication requests, otherwise the bridge would run into rate limit errors when loading sequence traces


api/restapi/configure_keptn_test.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func Test_getEnvConfig(t *testing.T) {
+	_ = os.Setenv("MAX_AUTH_REQUESTS_PER_SECOND", "0.5")
+	_ = os.Setenv("MAX_AUTH_REQUESTS_BURST", "1")
+	config, err := getEnvConfig()
+	require.Nil(t, err)
+	require.Equal(t, 0.5, config.AuthRequestsPerSecond)
+	require.Equal(t, 1, config.AuthRequestMaxBurst)
+}
```
warber, line 16: this test has a race condition with the other one in this file.
Let's just unset these env vars after the test.

```suggestion
func Test_getEnvConfig(t *testing.T) {
	defer os.Unsetenv("MAX_AUTH_REQUESTS_PER_SECOND")
	defer os.Unsetenv("MAX_AUTH_REQUESTS_BURST")
	_ = os.Setenv("MAX_AUTH_REQUESTS_PER_SECOND", "0.5")
	_ = os.Setenv("MAX_AUTH_REQUESTS_BURST", "1")
	
	config, err := getEnvConfig()
	require.Nil(t, err)
	require.Equal(t, 0.5, config.AuthRequestsPerSecond)
	require.Equal(t, 1, config.AuthRequestMaxBurst)
}
```

warber (changes requested):

warber (approved):

PR #4686, opened 2021-07-20
Fix integration test

Kirdock:
Fixes error on `npm ci` used in E2E tests


mowies:
Is there already an integration test run that worked because of the changes here?

Kirdock:
> Is there already an integration test run that worked because of the changes here?

Will be done after the images are built.

mowies (approved):

PR #4726, opened 2021-07-21
Backport of #4658

thisthat:
Thanks to @Kirdock for the original commit

### DIFF
--- a/bridge/server/app.js
+++ b/bridge/server/app.js
@@ -198,7 +198,7 @@ module.exports = (async function (){
     // render the error page
     if(err.response && err.response.data && err.response.data.message)
       err.message = err.response.data.message;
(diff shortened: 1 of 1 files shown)


thisthat:
SonarCloud complains of code smell issues due to the complexity of some functions. They can be ignored since in `master` we have already refactored that component.

bridge/server/app.js
```diff
@@ -198,7 +198,7 @@ module.exports = (async function (){
     // render the error page
     if(err.response && err.response.data && err.response.data.message)
       err.message = err.response.data.message;
-    res.status(err.status || 500).send(err.message);
+    res.status(err.response === null || err.response == undefined ? 500 : err.response.status).send(err.message);
```
Kirdock, line 201: I would change it to `err.response && err.response.status ? err.response.status : 500`

Kirdock (approved):

PR #4737, opened 2021-07-22
fix: Improve set Git upstream error / success handling (#4374)

### LINKED ISSUE #4374: Setting upstream failed, but the error is not shown in Bridge
Using Keptn 0.8.3,

I added a git upstream in Bridge, but it seems that the API responded with a 500, which is not shown prominently in the UI:

image
image

**Note**: A short toast notification in the bottom-center is shown, but it disappears so quickly, that it is barely noticeable.

Probably related to https://github.com/keptn/keptn/issues/4375

laneli:
Fixes #4374 

image
image


ermin-muratovic:
Could we align the buttons on the right side? And do we have the same issue with global notifications (new version notification)?

Inked126630051-15999d52-bd34-4ddf-bb5b-34ac659ac60b_LI

laneli:
> Could we align the buttons on the right side? And do we have the same issue with global notifications (new version notification)?

I will check about the globals

laneli:
It's now aligned globally and component-wise. I also updated the hover effect of the icon button, so that it does look a bit more pretty.

image

image

ermin-muratovic (approved):
lgtm

PR #4747, opened 2021-07-22
Bug fixes in secrets view

### LINKED ISSUE #4622: Create secret view: The bridge breaks if the first key-value pair is deleted
If a second key-value pair is created and the first one is deleted, the bridge runs into an error
image

### LINKED ISSUE #4633: Secrets list is not updated after deleting a secret
### Bug

After deleting a secret, the list still contains the secret. I have to refresh the page.

image

### Expected behaviour

The secret is gone, after deleting it.

### LINKED ISSUE #4660: Secrets view is empty after deleting one secret
When having multiple secrets created and deleting one, all the secrets are gone, but they re-appear after refreshing the UI.

Step1 (delete):
image

Step2:
image

Step3 (after refresh):
image

ermin-muratovic:
Fixes #4622 
Fixes #4660 
Fixes #4633

### DIFF
--- a/bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.html
+++ b/bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.html
@@ -31,7 +31,7 @@
                   <input formControlName="value" type="password" required dtInput placeholder="Value" autocomplete="false" />
                   <dt-error>Must not be empty</dt-error>
                 </dt-form-field>
-                <button type="button" dt-icon-button variant="nested" aria-label="Remove key-value pair" title="Remove key-value pair" (click)="removePair(i)">
+                <button *ngIf="data?.controls.length > 1" type="button" dt-icon-button variant="nested" aria-label="Remove key-value pair" title="Remove key-value pair" (click)="removePair(i)">
                   <dt-icon name="removerow"></dt-icon>
                 </button>
               </div>
--- a/bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.html
+++ b/bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.html
@@ -9,28 +9,34 @@
       <form [formGroup]="createSecretForm">
         <div class="mb-3" fxLayout="row" fxLayoutAlign="none end" fxLayoutGap="15px">
           <dt-form-field>
-            <dt-label>Name</dt-label>
-            <input [(ngModel)]="secret.name" formControlName="name" type="text" required dtInput placeholder="Secret name" />
-            <dt-error *ngIf="getFormControl('name')!.errors?.required">Must not be empty</dt-error>
-            <dt-error *ngIf="getFormControl('name')!.errors?.pattern">Name must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character</dt-error>
+            <dt-label>Name *</dt-label>
+            <input formControlName="name" type="text" required dtInput placeholder="Secret name" />
+            <dt-error *ngIf="getFormControl('name').errors?.required">Must not be empty</dt-error>
+            <dt-error *ngIf="getFormControl('name').errors?.pattern">Name must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character</dt-error>
           </dt-form-field>
         </div>
 
-        <div class="mb-1" fxLayout="row" fxLayoutAlign="none end" fxLayoutGap="15px" *ngFor="let data of secret.data; let index = index;">
-          <dt-form-field>
-            <dt-label>Key</dt-label>
-            <input [(ngModel)]="data.key" [formControlName]="'key'+(index+1)" type="text" required dtInput placeholder="Key" />
-            <dt-error *ngIf="getFormControl('key' + (index+1))?.errors?.required">Must not be empty</dt-error>
-            <dt-error *ngIf="getFormControl('key' + (index+1))?.errors?.pattern">Key must consist of alphanumeric characters, '-', '_' or '.'</dt-error>
-          </dt-form-field>
-          <dt-form-field>
-            <dt-label>Value</dt-label>
-            <input [(ngModel)]="data.value" [formControlName]="'value'+(index+1)" type="password" dtInput placeholder="Value" />
-            <dt-error>Must not be empty</dt-error>
-          </dt-form-field>
-          <button dt-icon-button variant="nested" aria-label="An example nested button containing an icon" title="Remove key-value pair" (click)="removePair(index)">
-            <dt-icon name="removerow"></dt-icon>
-          </button>
+        <div formArrayName="data">
+          <ng-container *ngFor="let dataGroup of data.controls; let i = index;">
+            <form [formGroup]="dataGroup">
+              <div class="mb-1" fxLayout="row" fxLayoutAlign="none end" fxLayoutGap="15px">
+                <dt-form-field>
+                  <dt-label>Key *</dt-label>
+                  <input formControlName="key" type="text" required dtInput placeholder="Key" autocomplete="false" />
+                  <dt-error *ngIf="dataGroup.get('key').errors?.required">Must not be empty</dt-error>
+                  <dt-error *ngIf="dataGroup.get('key').errors?.pattern">Key must consist of alphanumeric characters, '-', '_' or '.'</dt-error>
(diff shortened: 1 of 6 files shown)


bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.html
```diff
@@ -9,43 +9,51 @@ <h2>Create Secret</h2>
...
+                  <input formControlName="key" type="text" required dtInput placeholder="Key" autocomplete="false" />
+                  <dt-error *ngIf="dataGroup.get('key').errors?.required">Must not be empty</dt-error>
+                  <dt-error *ngIf="dataGroup.get('key').errors?.pattern">Key must consist of alphanumeric characters, '-', '_' or '.'</dt-error>
+                </dt-form-field>
+                <dt-form-field>
+                  <dt-label>Value *</dt-label>
+                  <input formControlName="value" type="password" required dtInput placeholder="Value" autocomplete="false" />
+                  <dt-error>Must not be empty</dt-error>
+                </dt-form-field>
+                <button dt-icon-button variant="nested" aria-label="An example nested button containing an icon" title="Remove key-value pair" (click)="removePair(i)">
```
laneli, line 34: The aria-label is still from the example ;)

bridge/client/app/_models/secret.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { Secret } from './secret';
+import { waitForAsync } from '@angular/core/testing';
+
+describe('Secret', () => {
+
+  it('should create with default scope "keptn-default"', waitForAsync(() => {
+    const secret: Secret = new Secret();
+
+    expect(secret).toBeTruthy();
+    expect(secret.scope).toBe('keptn-default');
```
laneli, line 10: `expect(secret.data.length).toEqual(0)` ?

bridge/client/app/_models/secret.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+    secret.addData('key1', 'value1');
+    secret.addData('key2', 'value2');
+
+    expect(secret.data.length).toBe(2, 'Secret should have 2 key-value pairs');
+    expect(secret.getData(0).key).toBe('key1', 'Key at index 0 should be "key1"');
+    expect(secret.getData(0).value).toBe('value1', 'Value at index 0 should be "value1"');
+    expect(secret.getData(1).key).toBe('key2', 'Key at index 1 should be "key2"');
+    expect(secret.getData(1).value).toBe('value2', 'Value at index 1 should be "value2"');
+  }));
+});
```
laneli, line 25: Please also add tests for removeData, getData, setName

laneli (changes requested):

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.ts
```diff
@@ -1,57 +1,67 @@
...
 
-  constructor(private dataService: DataService, private router: Router, private route: ActivatedRoute) { }
+  constructor(private dataService: DataService, private router: Router, private route: ActivatedRoute, private fb: FormBuilder) {
+  }
 
-  ngOnInit(): void {
-    this.secret = new Secret();
-    this.addPair();
+  get data(): FormArray | null {
+    return this.createSecretForm.get('data') as FormArray;
```
laneli, line 33: Please remove the code smell here :) I think `as FormArray` is not necessary.

Kirdock: Because of the initialization of `createSecretForm` we can say it is not null. So the solution here would be to remove `| null` from the return type because it is not needed if we cast it to FormArray anyways.

bridge/client/app/_components/ktb-secrets-list/ktb-secrets-list.component.ts
```diff
@@ -15,7 +15,7 @@ export class KtbSecretsListComponent implements OnInit, OnDestroy {
   private readonly unsubscribe$ = new Subject<void>();
   private closeConfirmationDialogTimeout?: ReturnType<typeof setTimeout>;
 
-  public tableEntries: DtTableDataSource<object> = new DtTableDataSource();
+  public tableEntries: DtTableDataSource<Secret> = new DtTableDataSource();
   public currentSecret?: Secret;
 
   public deleteSecretDialogState: string | null = null;
```
Kirdock, line 21: Maybe change it to `public deleteSecretDialogStage?: string` to have a default value of `undefined`, else we would have two different styles on how to deal with default/null values in the project. Do we stick to `undefined` or `null` as default value?

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.ts
ermin-muratovic: Without `as FormArray` TypeScript complains that `Type 'AbstractControl | null' is not assignable to type 'FormArray | null'.`
With it Sonar complains that `assertion is unnecessary since it does not change the type of the expression`.

Any ideas? Maybe @Kirdock has a suggestion?

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.html
```diff
@@ -9,43 +9,51 @@ <h2>Create Secret</h2>
...
+                  <input formControlName="key" type="text" required dtInput placeholder="Key" autocomplete="false" />
+                  <dt-error *ngIf="dataGroup.get('key')?.errors?.required">Must not be empty</dt-error>
+                  <dt-error *ngIf="dataGroup.get('key')?.errors?.pattern">Key must consist of alphanumeric characters, '-', '_' or '.'</dt-error>
+                </dt-form-field>
+                <dt-form-field>
+                  <dt-label>Value *</dt-label>
+                  <input formControlName="value" type="password" required dtInput placeholder="Value" autocomplete="false" />
+                  <dt-error>Must not be empty</dt-error>
+                </dt-form-field>
+                <button dt-icon-button variant="nested" aria-label="Remove key-value pair" title="Remove key-value pair" (click)="removePair(i)">
```
Kirdock, line 34: Add `type="button"` here. Else on delete it tries to submit the form but is not possible, which results in this message in the console
image

Kirdock:
At the moment it is possible to add a secret without a key-value pair (if you remove the first one). Is this behavior intended?

ermin-muratovic:
> At the moment it is possible to add a secret without a key-value pair (if you remove the first one). Is this behavior intended?

You are right. Would it be enough to hide the "remove"-Button if there is only one key-value pair, so that you can't remove all?

Kirdock:
> > At the moment it is possible to add a secret without a key-value pair (if you remove the first one). Is this behavior intended?
> 
> You are right. Would it be enough to hide the "remove"-Button if there is only one key-value pair, so that you can't remove all?

That would have been my suggestion, yes.

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.html
```diff
@@ -9,43 +9,51 @@ <h2>Create Secret</h2>
...
           <dt-icon name="addrowonbottom"></dt-icon>
         </button>
 
         <div fxLayout="row" fxLayoutAlign="none end" fxLayoutGap="15px" class="mt-3">
-          <button (click)="createSecret()" [disabled]="isLoading" dt-button>
+          <button (click)="createSecret()" [disabled]="createSecretForm.invalid || isLoading" dt-button>
             <dt-loading-spinner *ngIf="isLoading" aria-label="Create secret"></dt-loading-spinner>
             Add secret
           </button>
           <button dt-button variant="secondary" routerLink="../../secrets">
```
Kirdock, line 51: Same warning message here when you click the button. Add `type="reset"` here

Kirdock:
Sorry if I suddenly change my point of view, but after looking at the changes for disabling the possibility to remove the first key-value pair, I would prefer disabling the button instead of removing it completely. The view would be more static. What do you think?

ermin-muratovic:
> Sorry if I suddenly change my point of view, but after looking at the changes for disabling the possibility to remove the first key-value pair, I would prefer disabling the button instead of removing it completely. The view would be more static. What do you think?

Right, that's definitely better :)

Kirdock (approved):

laneli (approved):

PR #4775, opened 2021-07-28
fix(bridge): fix error in sequence screen if a sequence does not have any events (#4774)

### LINKED ISSUE #4774: Sequence screen breaks if a sequence does not have any events
At the moment if the sequence does not have a recent event, the sequence screen breaks
image

Kirdock:
Fixes #4774 


ermin-muratovic (approved):
lgtm

PR #4783, opened 2021-07-28
feat(bridge): Fetch all data in environment screen (#4538)

### LINKED ISSUE #4538: The environment does not always show the right information
The environment screen does not always show the right information for services, because all kind of information is taken from root events. At the moment we fetch about 25 root events, but there could still be some services without any and this wouldn't reflect the current status.

We have two options now.
# Option 1
A new endpoint because we need additional information that is missing in the sequence endpoint and the result/response should be one sequence for each service in a stage, which is completely different than the sequence endpoint.

# Option 2
We use the sequence endpoint. Then we need additional information:

- Approval traces. It is possible to accept decline approvals. Should they just be fetched if they are needed? Additional information needed would be `data.labels` and `data.message` of the approval trace because this information is passed on to the next tasks in the sequence.
image.
- Evaluation traces. Currently we have an overlay for all evaluations which shows the whole evaluation result. Either we disable this overlay or we fetch this information.
image

- Remediations. The screen shows which remediation is currently running and also which actions have been executed. So additional information would be `problemTitle` or `ImpactedEntity` and `actions[]`
- DeploymentUri. At the moment `deploymentUri` is taken from the latest deployment trace if there is any. My suggestion would be to add it to the service in the `GET /project` endpoint.

Contra:
- The result has to be adjusted or we need an additional parameter like `?isService=true` because we need a sequence for each service in each stage

# Option 3
The project endpoint is extended. The environment screen needs the following information

## V1
```json
"stages": [
	{
		"stageName": "dev",
		"parentStages": [],
		"services": [
			{
				"image": "carts:0.12.3",
				"deploymentUrl": "http://...",
				"serviceName": "carts",
				"shkeptncontext": "",
				"latestSequence": {
					"name": "evaluation",
					"state": "triggered" | "finished",
					"result": "pass" | "fail" | "warning",
					"evaluationResult?": {
						"score": 100,
						"result": "pass" | "fail"
					}
				},
				"openApprovals": [
					{
						"shkeptncontext": "",
						"id": "",
						"labels": {
							"key": "value"
						},
						"message": ""
					}
				],
				"openRemediations": [
					{
						"problemTitle": "Failure rate increase",
						"shkeptncontext": "",
						"actions": [
							{
								"name": "Toogle feature flag EnableItemCache to ON",
								"state": "triggered" | "finished",
								"result": "pass" | "fail"
							}
						]
					}
				]
			}
		]
	}
]
```

## V2:
Remediation sequences and approval traces can be fetched on demand
```json
"stages": [
	{
		"stageName": "dev",
		"parentStages": [],
		"services": [
			{
				"image": "carts:0.12.3",
				"deploymentUrl": "http://...",
				"serviceName": "carts",
				"shkeptncontext": "",
				"latestSequence": {
					"name": "evaluation",
					"state": "triggered" | "finished",
					"result": "pass" | "fail" | "warning",
					"evaluationResult?": {
						"score": 100,
						"result": "pass" | "fail"
					}
				},
				"openApprovals": [
					"context1",
					"context2"
				],
				"openRemediations": [
					"context1",
					"context2"
				]
			}
		]
	}
]
```
We also need all evaluation results because of the overlay, but this can be fetched on demand.
image

# Option 4
Each additional information as suggested in Option 3 is fetched and provided through the backend of the bridge.

After discussion we decided to go for option 4.

Kirdock:
Closes #4538


bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
```diff
@@ -1,93 +1,47 @@
...
-  }
-
-  ngOnDestroy(): void {
-    this.unsubscribe$.next();
-    this._rootsTimer.unsubscribe();
+        map(params => params.projectName),
+        filter(projectName => !!projectName),
+        switchMap(projectName => this.dataService.getProject(projectName))
+      );
+    project$.subscribe(project => {
```
ermin-muratovic, line 37: It looks like an unsubscribe is missing here. Either add it to the pipe above or here.

Kirdock: done

ermin-muratovic (approved):
lgtm

PR #4843, opened 2021-08-06
fix(bridge): Show error message when project can not be loaded (#4765)

### LINKED ISSUE #4765: Redirect to dashboard if project is deleted
When you are in the project-screen and the project is deleted (through the CLI or another tab), the bridge still shows the project-screen and still tries to poll the project even if it receives a 404 status code.

# Definition of Done
- [ ] The bridge redirects to the dashboard if the project endpoint returns 404

laneli:

Fixes #4765


bridge/client/app/_services/data.service.ts
```diff
@@ -197,13 +207,24 @@ export class DataService {
...
+    }, err => {
+      if (err.status === 404) {
         const projects = this._projects.getValue();
-        const existingProject = projects?.find(p => p.projectName === project.projectName);
-        if (existingProject){
-          const {roots, sequences, ...copyProject} = project;
-          Object.assign(existingProject, copyProject);
+        const projectIdx = projects?.findIndex(p => p.projectName === projectName);
+        if (projectIdx && projectIdx > 0) {
+          projects?.splice(projectIdx);
```
Kirdock, line 222: change it to `splice(projectIdx, 1)`.
Also index 0 is excluded, is there a reason why?

bridge/client/app/_services/data.service.mock.ts
```diff
@@ -46,6 +47,16 @@ export class DataServiceMock extends DataService {
     return of({});
   }
 
+  public projectExists(projectName: string): Observable<boolean | undefined> {
```
Kirdock, line 50: There is no difference between this function and the one in the DataService, therefore we don't need to overwrite it here.

bridge/client/app/project-board/project-board.component.ts
```diff
@@ -16,51 +17,57 @@ export class ProjectBoardComponent implements OnInit, OnDestroy {
...
+  public hasProject$: Observable<boolean | undefined>;
   public contextId?: string;
-  private readonly _projectTimerInterval = 30_000;
   private _errorSubject: BehaviorSubject<string | undefined> = new BehaviorSubject<string | undefined>(undefined);
   public error$: Observable<string | undefined> = this._errorSubject.asObservable();
   public isCreateMode$: Observable<boolean>;
 
-  constructor(private router: Router, private route: ActivatedRoute, private dataService: DataService) {
-    const projectName$ = this.route.paramMap.pipe(
+  constructor(private router: Router, private route: ActivatedRoute, public dataService: DataService, @Inject(INITIAL_DELAY_MILLIS) private initialDelayMillis: number) {
```
Kirdock, line 27: What's the reason behind `public` here?

bridge/client/app/project-board/project-board.component.ts
```diff
@@ -79,33 +86,49 @@ export class ProjectBoardComponent implements OnInit, OnDestroy {
...
-              }
-            }
-          } else {
-            this._errorSubject.next('trace');
+        this.navigateToTrace(traces, eventselector);
+      });
+    }
+  }
+
+  public navigateToTrace<S, U>(traces: Trace[] | undefined, eventselector: string | null) {
```
Kirdock, line 94: Are the type aliases needed here (`<S, U>`)?

bridge/client/app/project-board/project-board.component.ts
```diff
@@ -16,51 +17,57 @@ export class ProjectBoardComponent implements OnInit, OnDestroy {
   private readonly unsubscribe$ = new Subject<void>();
 
   public logoInvertedUrl = environment?.config?.logoInvertedUrl;
-  public project$: Observable<Project | undefined>;
+  public projectName$: Observable<string>;
```
Kirdock, line 20: Do we need this as a global variable?

Kirdock (changes requested):

bridge/client/app/_services/data.service.ts
laneli: Changed it. No reason, just a mistake. Glad you saw it.

bridge/client/app/project-board/project-board.component.ts
laneli: A leftover from tryouts for testing. I will set it to private again.

bridge/client/app/project-board/project-board.component.ts
laneli: No, I don't think so. The IDE created it when automatically refactoring. I will remove it.

bridge/client/app/project-board/project-board.component.ts
laneli: No, you are right. Also a leftover from testing tryouts.  I will remove it.

Kirdock (approved):

PR #4875, opened 2021-08-10
ci: Retag unchanged images in build-everything builds (#4684)

### LINKED ISSUE #4684: Speed up CI workflow for PR integration tests
Currently, when integration tests for a PR should be executed, the following steps have to be taken:

For the PR, the label `ci:trigger-build-everything` has to be added. This will ensure that not only the image for any service that contains actual code changes is built - instead, all services will be built and tagged. Additionally, a helm-chart that references the built images is uploaded as an artifact. When that is done, the integration tests can be started.

However, rebuilding all images does take up a lot of time. Therefore, I suggest to only rebuild the services that contain changes, and simply re-tag the ones that are the same as on the target branch of the PR.

mowies:
## This PR

- changes the "build-everything" build's behaviour so that unchanged docker images or not rebuilt. Instead, the last successfully built image will be pulled and retagged. Only artifacts that actually have changes are built fresh.
- moves the build preparation with changed files detection into a separate shell script to clean up the pipeline

### Related Issues

Fixes #4684

### Follow-up Tasks

- [ ] Instead of pulling, retagging and pushing unchanged docker images, the docker registry API could be used to just pull the manifest and push it with a new tag. This would be much faster and would use far less bandwidth (#4905)


bacherfl (approved):
lgtm

PR #4908, opened 2021-08-11
fix(api): Apply rate limiter before token is validated (#4906)

### LINKED ISSUE #4906: Rate limiting in auth endpoint: 429 responses contain information on whether token was valid or not
The rate limiter in the `/auth` endpoint (see #4323) currently returns HTTP 429 (Too Many Requests) responses if too many failed authentication requests are sent to the API. Requests with a valid token are not throttled because this would quickly lead to the problem of the bridge running in a rate limit when loading information via the API, since every external request to the API has to pass the `/auth` endpoint. 
Currently, the rate limiter middleware first checks the incoming auth token for its validity in order to make the decision if the rate limiter should be applied or not. If the token is invalid, the limiter is applied. This however has the implication that when a 429 Response is returned, it indirectly tells the caller of the API that the provided token of the request is invalid.

To fix that, the better approach would be the following:

The limiter should applied before doing any token validation. 
If the limiter does not allow the current request to be handled, a 429 response is returned - in this case no token validation has been done yet, so the client does not know whether the provided token is valid or not.
If the limiter does allow the request to be handled, the token should be checked - if it's valid, the limiter will be reset, and subsequent valid requests to the API will not be affected by rate limiting. If the token was invalid, the limiter will not be reset, and subsequent requests will be throttled if they exceed the maximum allowed rate.

bacherfl:
## This PR

- Changes the logic of the `/auth` endpoint's rate limiter to be applied before any token validation is done.
- If the incoming API token is valid, this will reset the limiter in order to prevent subsequent valid requests to be throttled
- With this logic, the caller of the API cannot automatically assume that the provided token was invalid when receiving a HTTP 429 (Too many requests) response since now also a request with a valid token can be throttled if there were too many invalid requests before.

### Related Issues

Closes #4906

### DIFF
--- a/api/middleware/ratelimiter.go
+++ b/api/middleware/ratelimiter.go
@@ -52,25 +52,22 @@ func (r *RateLimiter) Handle(handler http.Handler) http.Handler {
 }
 
 func (r *RateLimiter) Apply(w http.ResponseWriter, req *http.Request, handler http.Handler) {
-	// check if authentication is valid
-	_, err := r.tokenValidator.ValidateToken(req.Header.Get("x-token"))
-	if err != nil {
-		// if not, apply rate limiting
-		ipAddress := getRemoteIP(req)
-		limiter := r.getIPBucket(ipAddress)
-		if limiter.Allow() == false {
-			http.Error(w, http.StatusText(429), http.StatusTooManyRequests)
-			return
-		}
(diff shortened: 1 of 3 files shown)


mowies (approved):
LGTM

PR #4928, opened 2021-08-12
feat(bridge): Prevent expanding empty tile (#4057)

### LINKED ISSUE #4057: Prevent from expanding tile when there is no content
All the expanded tiles are expanding even if there are no contents in it. This is very counter-intuitive and also the user can think that there is something not working properly.
This is the case in nearly every screen:

image

image

image

## Acceptance criteria:

- [ ] When no contents are in an expanded tile, I would propose to:
  * To hide the chevron icon
  * To not expand it
  * To have the cursor set to default (`cursor: default`)

## Definition of Done:
- [ ] It is not possible to expand a tile without content
- [ ] Test: Provide a component test or an end-2-end test (depends on implementation)

### LINKED ISSUE #4075: Closed Loop Remediation fails due to missing start and end time in evaluation.triggered event
During https://github.com/keptn/examples/issues/165 we've encountered a problem with the execution of a remediation, more spcifically the evaluation task ,which has no Evaluation field set.

The Lighthouse service fails with the error: `evaluation.triggered event does not contain evaluation timeframe`

image

Remediation sequence in `shipyard.yaml`:

```
        - name: "remediation"
          triggeredOn:
            - event: "production.remediation.finished"
              selector:
                match:
                  evaluation.result: "fail"
          tasks:
            - name: "get-action"
            - name: "action"
            - name: "evaluation"
              triggeredAfter: "2m"
```

ermin-muratovic:
## This PR
- allows to set expandable tile as disabled, which will prevent the user to expand an empty tile
- sets all expandable tiles (environments screen, service screen and sequence screen) to disabled if empty

image

### Related Issues
Closes #4057

### DIFF
--- a/bridge/client/app/_components/ktb-expandable-tile/ktb-expandable-tile.component.ts
+++ b/bridge/client/app/_components/ktb-expandable-tile/ktb-expandable-tile.component.ts
@@ -97,8 +97,11 @@ export class KtbExpandableTileComponent {
     return this._disabled;
   }
   set disabled(value: boolean) {
-    if (this._disabled && this._selected) {
-      this._selected = false;
+    if (this._disabled !== value) {
+      this._disabled = value;
+      if (this._disabled) {
+        this.selected = false;
+      }
       this._changeDetectorRef.markForCheck();
     }
   }
--- a/bridge/client/app/_components/ktb-expandable-tile/ktb-expandable-tile.component.ts
+++ b/bridge/client/app/_components/ktb-expandable-tile/ktb-expandable-tile.component.ts
@@ -23,8 +23,6 @@ export class KtbExpandableTileHeader {}
     '[class.ktb-tile-error]': 'error',
     '[attr.aria-success]': 'success',
     '[class.ktb-tile-success]': 'success',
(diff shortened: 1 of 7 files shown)


bridge/client/app/_components/ktb-expandable-tile/ktb-expandable-tile.component.ts
```diff
@@ -97,8 +97,11 @@ export class KtbExpandableTileComponent {
     return this._disabled;
   }
   set disabled(value: boolean) {
-    if (this._disabled && this._selected) {
-      this._selected = false;
+    if (this._disabled !== value) {
+      this._disabled = value;
+      if (this._disabled) {
+        this.selected = false;
```
Kirdock, line 103: Do we even need this property? Looks like it is never set to `true`

ermin-muratovic: Right. at some time we split this component into `ktb-selectable-tile` and this `ktb-expandable-tile`. For the KtbSelectableTile it does make sense, but here we can drop it.

Kirdock (approved):

PR #4930, opened 2021-08-13
feat: Remove override of evaluation result using previous test result

agrimmer:
Signed-off-by: Andreas Grimmer <andreas.grimmer@dynatrace.com>

## This PR
- removes deprecated logic which sets the evaluation result depending on a previous test result


bacherfl (approved):
lgtm

PR #4938, opened 2021-08-16
feat: Introduced WebHook Service (#4736)

### LINKED ISSUE #4736: Add a WebHook Service
## User Story

As a user, I would like to configure a webhook that gets called for specific events.

### Details

The webhook service is a default Keptn-service that will live in the @keptn-contrib organization.

When the webhook service gets triggered by an event, it first sends a `sh.keptn.event.[task].started` event. After that, it looks for the `webhook.yaml` file. The lookup strategy searches the `webhook.yaml` in well-defined directories, based on the following priority:
- `/project/{projectName}/stage/{stageName}/service/{serviceName}/resource`
- `/project/{projectName}/service/{serviceName}/resource`
- `/project/{projectName}/stage/{stageName}/resource`
- `​/project​/{projectName}​/resource`

If no file is found, the webhook service sets the result of the `.finished` event to `fail`.

The configuration (`webhook.yaml`) of the webhook service looks as follows:
```yaml
kind: WebhookConfig
metadata:
  name: webhook-configuration
spec:
  webhooks:
  - type: "sh.keptn.event.test.triggered"
    envFrom:
    - secretRef:
      name: mysecret
    requests:
    - "curl --header 'Content-Type: application/json' --header 'x-token: ${SECRET_KEY}' --request POST –data '{\"event\":\"{{.event.type}}\",\"service\":\"{{.event.data.service}}\",\"environment\":\"{{.event.data.stage}}\" https://api.mycompany.com"
```

When the file is found, the webhook service executes the `curl` command. Mind that, replacing `{{.value}}` with proper data will be done in (#4742).

The webhook service will set the result of the `.finished` event to `pass` if the HTTP webhook returned 2xx (200-299); if not, it will return `fail`.
The response body of the webhook call is added to the `data.[task]` section of the `.finished` event.

## Acceptance Criteria

- [ ] Webhooks are called for an event
- [ ] Secrets via env vars are visible by curl
- [ ] Make sure that only `curl` commands can be set in `webhook.yaml`

## DoD

- [ ] The webhook service is implemented
- [ ] Unit tests
- [ ] Integration tests

---

### Dependencies

- #4437

### LINKED ISSUE #4742: Define custom payload for webhook integration
## User Story

As a user, I would like to define a custom payload that is forwarded to the webhook URL

### Details

The payload field allows adding any kind of JSON payload. To map data from the incoming Keptn event to the payload, a templating mechanism should be made available.

In the template, a property from the received event is referenced by using `{{.event.[property]}}`. 
The templating system will then replace this part with the actual value.
Example, of custom payload:

```json
{ 
   "eventType":"{{.event.type}}",
   "service":"{{.event.data.service}}",
   "environment":"{{.event.data.stage}}",
   "deploymentUrl":"{{.event.data.deployment.deploymentUrl}}"
}
```

The template system should also verify that:

- [ ] All properties exist in the event

If an error occurs, an `Error Log` event should be generated to notify the problem. 
Furthermore, the webhook service should report the error in the `.finished` event and set the result field to `fail`.

## Acceptance Criteria

- [ ] Templating system is in place
- [ ] Integrated into the webhook service

## DoD

- [ ] Unit Tests
- [ ] Propert error handling

### LINKED ISSUE #4743: Filter available URLs for the webhook integration
## Action Item

The webhook service must ensure that an HTTP webhook cannot access the K8s API. 

## Acceptance Criteria

- [ ] Network Policy must decline cluster-internal calls

bacherfl:
Closes #4736 
Closes #4742
Closes #4743


johannes-b:
I'm playing around with the webhook-service, and it already looks really good.

I just encountered some changes requests: 

### Webhook.conf in ./webhook/

imageThe 
- Right now, the `webhook.conf` is stored directly in the folder of the service. 
- [ ] However, it should be stored in the `webhook` folder. 

### Adoption in `finished` event 

* Right now, the webhook-service returns a `finished` event as follows:
```
{
  "data": {
    "labels": null,
    "project": "sockshop",
    "result": "pass",
    "service": "carts",
    "sh.keptn.event.test.triggered": {
      "responses": [
        "ok"
      ]
    },
    "stage": "dev",
    "status": "succeeded"
  },
  "id": "a3248326-892b-43d0-bda7-0e338c108de0",
  "source": "webhook-service",
  "specversion": "1.0",
  "time": "2021-09-02T12:07:20.593Z",
  "type": "sh.keptn.event.test.finished",
  "shkeptncontext": "49206abf-a201-40d5-945e-4d21d5a76500",
  "triggeredid": "8c8f4598-5cb0-4f11-bf94-21acdc5c6ece"
}
```

- [ ] Change the `data` payload to contain just `test` and not the full event type `sh.keptn.event.test.triggered`

johannes-b:
## Error log when no webhook.conf available

- [ ] When the webhook-service has a subscription on, e.g., `sh.keptn.event.test.triggered` for all stages (and all services), but not `webhook.conf` file is available, I would like to get an error log telling me about the missing `webhook.conf`.

image

webhook-service/deploy/service.yaml
```diff
@@ -0,0 +1,94 @@
...
+            requests:
+              memory: "16Mi"
+              cpu: "25m"
+            limits:
+              memory: "32Mi"
+              cpu: "100m"
+          env:
+            - name: PUBSUB_URL
+              value: 'nats://keptn-nats-cluster'
+            - name: PUBSUB_TOPIC
```
Kirdock, line 62: Initially, the webhook service should not be subscribed to any event.

Kirdock (changes requested):

warber (dismissed):

johannes-b:
I encountered a problem with the webhook-service as explained in this image: 

image

warber (approved):

PR #4948, opened 2021-08-17
fix(bridge): fixes session fixation by regenerating the session cookie after login (#4947)

### LINKED ISSUE #4947: OAuth login - Session fixation
**Affected Component**

- [ ] Bridge

**Describe the bug**

OAuth login does not provide a new session cookie. This opens up session fixation vulnerability for OAuth enabled Keptn instances. 

However, bridge already utilize signed session cookies with a secret. Further, session id candidate is a crypto random. Hence, exploiting this vulnerability is not straightforward.

**Expected behaviour**

A new session cookie must be generated and authenticate for successful logins.

Kavindu-Dodan:
### Related Issues

Fixes #4947

### How to test

- Enabled OAuth based login 
- Check session cookie value before the login 
- Check session cookie value after the login
- Validate that cookie value has changed

### DIFF
--- a/bridge/server/user/session.ts
+++ b/bridge/server/user/session.ts
@@ -1,4 +1,4 @@
-import { Router, Request, Express } from 'express';
+import { Express, Request, Router } from 'express';
 import expressSession from 'express-session';
 import mS from 'memorystore';
 import random from 'crypto-random-string';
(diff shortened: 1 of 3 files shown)


bridge/server/user/session.ts
```diff
@@ -127,7 +132,7 @@ function sessionRouter(app: Express)  {
 
 export { sessionRouter };
 export { isAuthenticated };
-export { setAuthenticatedSession as sessionAuthentication };
+export { authenticateSession as authenticateSession };
```
Kirdock, line 135: Remove `as authenticateSession`. Renaming is not needed if they are the same.

Kirdock (changes requested):

bridge/server/user/session.ts
Kavindu-Dodan: Perfect thanks :)

Kirdock (approved):

PR #4976, opened 2021-08-18
chore(bridge): Migrate test framework to Jest (#4841)

### LINKED ISSUE #4841: Bridge: Switch from Jasmine / Karma to Jest as test framework
We consider switching from Jasmine to Jest for Bridge testing.
Points that would speak for it:

- There is more knowledge about it
- Mocking and stubbing is more comfortable
- Runs faster
- Is more stable than Karma and doesn't need a headless browser

There is the documentation for migration to Jest:
https://jestjs.io/docs/migration-guide

There is also a third-party package that makes the necessary code transformations: jest-codemods.
See the documentation linked above for more information.

Additional sources:
https://jestjs.io/docs/getting-started
https://medium.com/angular-in-depth/migrate-your-angular-library-to-jest-faba9579591a

## DoD
- [ ] Tests are running locally and in CI

laneli:

## This PR
Migrates the testing framework from Jasmine / Karma to Jest

### Related Issues
Closes #4841


bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.spec.ts
```diff
@@ -1,27 +1,34 @@
...
 
   beforeEach(waitForAsync(() => {
     TestBed.configureTestingModule({
-      declarations: [ KtbCreateSecretFormComponent ],
       imports: [
         AppModule,
-        HttpClientTestingModule,
-      ]
+        HttpClientTestingModule
+      ],
```
ermin-muratovic, line 16: Let's add the trailing comma here? Because on other places we also have it.
```suggestion
        HttpClientTestingModule,
      ],
```

bridge/tslint.json
```diff
@@ -1,5 +1,10 @@
 {
   "extends": "tslint:recommended",
+  "linterOptions": {
+    "exclude": [
+      "client/app/_models/trace.mock.ts"
```
ermin-muratovic, line 5: Is here an wildcard like `*.mock.ts` possible?

ermin-muratovic (dismissed):

Kirdock (changes requested):
At the moment there are about 2 different ways, how the `beforeEach` in each test is defined. It has the same functionality, it's just a very minor thing that does not need to be fixed.
v1:
```ts
beforeEach(async () => )
```
Here we also don't need to call `then()` if we put `await` before and the code inside of `then()` could be moved to the row below

v2:
```ts
beforeEach(waitForAsync() => )
```

bridge/tslint.json
laneli: I added it, but only for models. The mocks like DataServiceMock I would not exclude from tslint.

bridge/client/app/_utils/app.utils.ts
```diff
@@ -1,11 +1,15 @@
...
     if (delay === 0) {
       return of(undefined);
     }
 
-    return timer(0, delay);
+    return timer(dueTime, delay);
   }
 }
+
+export const INITIAL_DELAY_MILLIS = new InjectionToken<number>('Initial delay in millis');
```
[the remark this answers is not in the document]
laneli: Is now called `POLLING_INTERVAL_MILLIS`

bridge/client/app/_components/ktb-root-events-list/ktb-root-events-list.component.spec.ts
```diff
@@ -37,20 +37,18 @@ describe('KtbEventsListComponent', () => {
...
-      });
-  }));
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(KtbRootEventsListComponent);
+    component = fixture.componentInstance;
+    dataService = fixture.debugElement.injector.get(DataService);
+    dataService.loadProjects(); // reset project.sequences
+    // @ts-ignore
+    dataService.getProject(projectName).subscribe((pr: Project) => {
```
Kirdock, line 47: This one could eventually be problematic, because `beforeEach` does not wait for its completion, but it should be fine in the test. Could be changed to `project = await dataService.getProject(projectName).toPromise()` or we use the `done` parameter of `beforeEach` and add it inside the `subscribe`

laneli: Added `done`

Kirdock (approved):

PR #5092, opened 2021-08-31
fix(bridge): project delete dialog was not closed (#5091)

### LINKED ISSUE #5091: Project delete dialog is not closed
**Describe the bug**
If you delete a project, you are instantly redirected to the dashboard and the delete-dialog does not disappear. 

**Expected behavior**
The redirect occurs after the project is deleted and the delete-dialog should be closed.

**Screenshots**
image

**Additional context**
root cause is probably 
https://github.com/keptn/keptn/blob/78e55bb74191a8bd1673ae01653a751487c38775/bridge/client/app/_components/ktb-project-settings/ktb-project-settings.component.ts#L190-L196

This event is triggered instantly because we have a `BehaviorSubject` here, which emits it's current value.

Kirdock:
Fixes #5091
Also fixes an issue that causes the delete action to be canceled


ermin-muratovic (approved):
lgtm

PR #5119, opened 2021-09-01
fix(bridge): faded-out integrations were not excluded from unread-error-event check (#5118)

### LINKED ISSUE #5118: Unread error event indicator for integrations is shown even if there aren't any
At the moment the unread error event indicator is shown if there are faded-out integrations with unread error events. Faded-out integrations should be excluded here.

Kirdock:
Fixes #5118 


ermin-muratovic (approved):
lgtm

PR #5120, opened 2021-09-01
fix(bridge): faded-out integrations were not excluded from unread-error-event check (#5118)

### LINKED ISSUE #5118: Unread error event indicator for integrations is shown even if there aren't any

Kirdock:
# Related PR
https://github.com/keptn/keptn/pull/5119


ermin-muratovic (approved):
lgtm

PR #5122, opened 2021-09-01
fix(shipyard-controller): Remove log noise in sequence migrator (#5096)

### LINKED ISSUE #5096: Exclude events from sequence migration that usually don't have a stage
At the moment the shipyard-controller throws some errors at startup during the sequence migration if an event does not contain a stage.
`msg="could not migrate sequence with shkeptncontext 6173b673-c701-4c21-bfa9-ee1a9f3702c0" error="could not determine scope of task sequence: event does not contain a stage"`
There are some events like `sh.keptn.event.service.delete.finished` or `sh.keptn.event.service.delete.started` that don't have a stage and therefore they should be excluded and they should not raise any errors.

## Definition of Done
- [ ] Identify all events that will never have a stage
- [ ] The shipyard-controller does not raise any errors if events like `service.deleted` don't have a stage

bacherfl:
Closes #5096

### DIFF
--- a/shipyard-controller/handler/sequence_migrator.go
+++ b/shipyard-controller/handler/sequence_migrator.go
@@ -47,7 +47,8 @@ func (sm *SequenceMigrator) MigrateSequences() {
 	for _, project := range projects {
 		// migrate sequences of projects in parallel
 		log.Infof("migrating sequences of project %s", project.ProjectName)
-		go sm.migrateSequencesOfProject(project.ProjectName, wg)
+		sm.migrateSequencesOfProject(project.ProjectName, wg)
+		log.Infof("finished migration of sequences of project %s", project.ProjectName)
 	}
 	wg.Wait()
 }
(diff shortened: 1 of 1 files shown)


shipyard-controller/handler/sequence_migrator.go
```diff
@@ -56,7 +57,6 @@ func (sm *SequenceMigrator) migrateSequencesOfProject(projectName string, wg *sy
 	pageSize := int64(50)
 
 	for {
```
agrimmer, line 59: Can you please explain why this does not produce an endless loop? 
I cannot see how the `NextPageKey` is used?

bacherfl: right, I have fixed that now

agrimmer: Thank you!

thisthat (approved):

PR #5139, opened 2021-09-02
feat(secret-service): Add list of keys within secrets created by the secret-service (#4749)

### LINKED ISSUE #4749: Extend GET /secret API to return secret keys
## Action Item

The `/secrets/v1/secret` API endpoint should also report the list of keys for each secret. 

## DoD

- [ ] For each secret, the list of available keys is reported
- [ ] Unit Test
- [ ] Test backward compatibility of the API

bacherfl:
Closes #4749


secret-service/pkg/backend/secretbackend_k8s.go
```diff
@@ -220,6 +235,16 @@ func remove(s []string, r string) []string {
 	return s
 }
 
+func insert(s []string, i string) []string {
```
mowies, line 238: More out of curiosity: Would this not be a nice fit to use pointers or do the operation in place instead of using a return?

mowies (approved):

secret-service/pkg/backend/secretbackend_k8s.go
bacherfl: I wanted to stay in line with similar functions of the golang core library like `append()`, which also returns the value of the resulting array, instead of modifying the pointer

PR #5152, opened 2021-09-03
feat(bridge): Configure webhook service in Bridge (#4750)

### LINKED ISSUE #4750: Controls to configure a webhook
## User Story

As a user, I can enter the configuration for the webhook call and the configuration is stored in the Git repo. 

### Details

webhook-basic

Besides the task (e.g., `evaluation.triggered`) and filter (e.g. stage/service) of a normal integration, 
the Bridge should provide for the webhook service the possibility to configure the following properties:

- request method: [`POST`, `PUT`]
- request URL: the webhook endpoint that is called
- list of custom headers that are passed over to the endpoint (optional)
- payload
- proxy (optional)

The Bridge should notify the user about misconfiguration of the webhook service when:
- not a valid URL is provided in the URL and proxy fields
- ~localhost/internal cluster IP is used :question:~

After a user entered the information and clicked the [Save] button, a `webhook.yaml` is stored in the Git repository of this project. The content of the `webhook.yaml` file is the following:

```yaml
apiVersion: webhook.keptn.sh/v1alpha1
kind: Webhook
metadata:
  name: webhook-configuration
spec:
  requests:
  - "curl --header "Content-Type: application/json" --request POST –data '{"event":"{{.event.type}}","service":"{{.event.data.service}}","environment":"{{.event.data.stage}}' https://api.mycompany.com"
```

The stage and service filters define the exact location where the `webhook.yaml` is stored.

- If no service filter is selected, the config is stored on project level: `​/project​/{projectName}​/resource`
- If a stage filter is selected, it is stored on stage-level: `/project/{projectName}/stage/{stageName}/resource`
- If a service filter is selected, it is stored on service-level: `/project/{projectName}/service/{serviceName}/resource`
- If a stage & service filters are selected, it is stored on service-level: `/project/{projectName}/stage/{stageName}/service/{serviceName}/resource`

Furthermore, the Bridge should provide an example of the events the webhook service is subscribed to. This will facilitate the creation of a custom payload. For more details, see #4742.

## Acceptance Criteria

- [x] The webhook service can be configured
- [ ] A link with an event example is provided

## DoD

- [x] `webhook.yaml` is created via the Bridge
- [x] Unit tests
- [x] UI tests

## Dependencies

- #4736

ermin-muratovic:
## This PR
- Adds the capability to configure the webhook-service in the bridge, when editing a subscription for the webhook-service

image

### Related Issues
Closes #4750


ermin-muratovic:
Please note following known issues:

- Currently it is allowed to save a subscription with `Use for all projects` active and applying `stage` and `service` filters. The Bridge will create a `webhook.yaml` in every project, in case `Use for all projects` is activated. But when applying a `stage` and `service` filter, it might throw an error, because it could not create the `webhook.yaml` for a `stage` or `service` in another project if that `stage` or `service` is not existing.
- If a subscription is created with the option `Use for all projects` active the Bridge will create the `webhook.yaml` for all existing projects. If a new project is created afterwards, because of the configured subscription the `webhook-service` might receive the `triggered` event it subscribed to, but the `webhook.yaml` might not exist. The result would be a failing sequence in that project, saying that the webhook configuration could not be found. The fix would be to edit the subscription and save it with the same configuration.
- The `webhook-service` can be configured to trigger multiple requests, but the Bridge allows the configuration only of one request. If there is an `webhook.yaml` with multiple requests configured, changing that via the Bridge will read only the first request and save only that one, overriding the old configuration and loosing the other requests.
- Currently it is possible to configure multiple subscriptions with the same or overlapping filter configurations. If the `webhook.yaml` was created by another subscription, but a new subscription has such an overlapping filter configuration, the Bridge will override the old configuration file without warning.

johannes-b:
Hello, 

I will collect here some findings, I propose for a change: 

---

- [ ] Change order of fields: 
image

---

- [ ] Don't change the size of the field (stay with a fixed size) 

image


---

- [ ] Common style across the page (and across all other pages) 

image

--- 

- [ ] Broader URL field with

image

---

- [ ] Use font-family: monospace for `Custom payload`

image

johannes-b:
### Webhook.yaml stored in in ./webhook/

- [ ] Please consider the changes in the webhook-service, which stores the `webhook.yaml` file in the `webhook` folder. https://github.com/keptn/keptn/pull/4938#issuecomment-911617505

johannes-b:
I would like to express my opinion on the two concerns mentioned by Ermin: 

---
- Currently it is allowed to save a subscription with `Use for all projects` active and applying `stage` and `service` filters. The Bridge will create a `webhook.yaml` in every project, in case `Use for all projects` is activated. But when applying a `stage` and `service` filter, it might throw an error, because it could not create the `webhook.yaml` for a `stage` or `service` in another project if that `stage` or `service` is not existing.

- If a subscription is created with the option `Use for all projects` active the Bridge will create the `webhook.yaml` for all existing projects. If a new project is created afterwards - because of the configured subscription - the `webhook-service` might receive the `triggered` event it subscribed to, but the `webhook.yaml` might not exist. The result would be a failing sequence in that project, saying that the webhook configuration could not be found. The fix would be to edit the subscription and save it with the same configuration.
---

Due to the fact that a webhook has the characteristic of being project-specific, I propose to disable the `Use for all projects` check box when creating a webhook configuration.
* Implications (workaround): 
  * Subscription/config is not shared, but a user has to create the same subscription in another project again. This is acceptable.   
* Benefit: 
  * It is obvious and more intuitive for a user whether a webhook-config is available or not. (As stated by Ermin, if a stage filter is in place but the stage is missing in a project, the webhook-config will not be created.) Besides, for a new project a shared *webhook subscription*  does not work anyways and will fail the first time since we don`t generate the `webhook.conf` automatically. 
  
 **Adaption** 
 - When creating a subscription for a webhook-service, the `Use for all projects` is hidden:
 
 image

johannes-b:
Regarding this concern: 

---
- The `webhook-service` can be configured to trigger multiple requests, but the Bridge allows the configuration only of one request. If there is an `webhook.yaml` with multiple requests configured, changing that via the Bridge will read only the first request and save only that one, overriding the old configuration and loosing the other requests.
--- 

The scope of this issue is focused on allowing to configure **one** webhook request. If there is a need for a sequence of webhooks, a follow-up story should tackle that. 

*My expectation:* If there is the following `webhook.yaml`, a user gets the first request presented in the Bridge. A change of this request will override the first entry, but the second entry stays un-touched. 

```
kind: WebhookConfig
metadata:
  name: webhook-configuration
spec:
  webhooks:
    - type: "sh.keptn.event.test.triggered"
      envFrom:
        - name: "secretKey"
          secretRef:
            name: "slack-secret"
            key: "endpoint"
      requests:
        - "curl -X POST -H 'Content-type: application/json' --data '{\"text\":\"First call"}' https://hooks.slack.com/services/{{.env.secretKey}}"
        - "curl -X POST -H 'Content-type: application/json' --data '{\"text\":\"Second call\"}' https://hooks.slack.com/services/{{.env.secretKey}}"
```

**Scenario:**

* The Bridge displays the first request ("First call")
* A user changes the custom payload from "First call" to "Updated" and presses the Update button
* The first request gets updated: 

```
kind: WebhookConfig
metadata:
  name: webhook-configuration
spec:
  webhooks:
    - type: "sh.keptn.event.test.triggered"
      envFrom:
        - name: "secretKey"
          secretRef:
            name: "slack-secret"
            key: "endpoint"
      requests:
        - "curl -X POST -H 'Content-type: application/json' --data '{\"text\":\"Updated"}' https://hooks.slack.com/services/{{.env.secretKey}}"
        - "curl -X POST -H 'Content-type: application/json' --data '{\"text\":\"Second call\"}' https://hooks.slack.com/services/{{.env.secretKey}}"
```

johannes-b:
Regarding this issue: 

---
- Currently it is possible to configure multiple subscriptions with the same or overlapping filter configurations. If the `webhook.yaml` was created by another subscription, but a new subscription has such an overlapping filter configuration, the Bridge will override the old configuration file without warning.
--- 

Given the situation that the following subscription is already available: 
image

and wants to create another subscription: 
image

Then we have an overlap for the `dev` stage. 

**Expected behavior:**

When clicking the `Create subscription` button, a warning pops up that informs that user that the already available webhook-config for the `dev` stage will be overridden when continuing. This story should be addressed in a dedicated issue.

johannes-b:
### Editing a webhook config

I would like to change a webhook config, but the `Update subscription` button does not get enabled: 


* From: 
image

* To: 
image

johannes-b:
### Editing a webhook config

1. I create this webhook config: 

image

When editing the config again: 
image

The escaping is wrong: 
image

johannes-b:
### Payload optional

- [ ] Please change the field to be optional and not mandatory: 

image

johannes-b:
### Deleting a subscription

I have a webhook.conf like this: 

image

* When deleting the subscription: 

image

* The webhook.conf still looks the same: 

image

johannes-b:
### Editing a webhook conf

- [ ] When opening the Webhook conf, `data` instead of the `payload` is shown:

image

bridge/client/app/_components/ktb-webhook-settings/ktb-webhook-settings.component.ts
```diff
@@ -0,0 +1,119 @@
...
+  public _subscription?: UniformSubscription;
+  public webhookConfigForm = new FormGroup({
+    method: new FormControl('', [Validators.required]),
+    url: new FormControl('', [Validators.required, Validators.pattern(FormUtils.URL_PATTERN)]),
+    payload: new FormControl('', []),
+    header: new FormArray([]),
+    proxy: new FormControl('', [Validators.pattern(FormUtils.URL_PATTERN)]),
+  });
+
+  public webhookMethods: WebhookConfigMethod[] = ['POST', 'PUT'];
```
johannes-b, line 26: Can we please add GET too?

bridge/client/app/_utils/form.utils.ts
```diff
@@ -1,6 +1,8 @@
 import { AbstractControl, ValidatorFn } from '@angular/forms';
 
 export class FormUtils {
+  public static URL_PATTERN = /^(?:http(s)?:\/\/)?[\w.-]+(?:\.[\w\.-]+)+[\w\-\._~:/?#[\]@!\$&'\(\)\*\+,;=.]+$/;
```
thisthat, line 4: I would keep the escape consistent:
```suggestion
  public static URL_PATTERN = /^(?:http(s)?:\/\/)?[\w\.\-]+(?:\.[\w\.\-]+)+[\w\-\._~:\/?#[\]@!\$&'\(\)\*\+,;=\.]+$/;
```

I think this would allow to match `https://mywebsite..com-`
```suggestion
  public static URL_PATTERN = /^(?:http(s)?:\/\/)?[\w\.-]+(?:[\w\.-]+)+[\w\._~:/?#[\]@!\$&'\(\)\*\+,;=]+$/;
```

thisthat: I would suggest to follow the RegEx: `^http(s)?:\/\/([\w-]+.)+[\w-]+(\/[\w\- .\/?%&=])?$`
source: https://stackoverflow.com/a/5717338

Kirdock: This one is still wrong. Something like "mywebsite.com", "http://mywebsite.com/" or "http://mywebsite.com/?q=asdf" is not possible

johannes-b:
- [ ] Please provide a more user-guided error message for: 

image

Suggestion: `URL contains a special character causing it to be invalid. Please remove the special character.`

bridge/client/app/_components/ktb-webhook-settings/ktb-webhook-settings.component.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  get headerControls(): FormGroup[] {
+    return this.header.controls as FormGroup[];
+  }
+
+  constructor(private dataService: DataService) {
+  }
+
+  private getWebhook(): void {
+    if (this.subscriptionExists && this.subscription && this.projectName) {
```
laneli, line 77: Instead of doing this, wouldn't it make more sense to load the webhook in the parent component, where we should be aware of all the inputs and then provide the webhook config also as an input?

laneli (changes requested):

bridge/client/app/_components/ktb-modify-uniform-subscription/ktb-modify-uniform-subscription.component.ts
```diff
@@ -143,18 +199,61 @@ export class KtbModifyUniformSubscriptionComponent {
...
-      update = this.dataService.createUniformSubscription(integrationId, subscription);
+      updates.push(this.dataService.createUniformSubscription(integrationId, subscription));
+    }
+
+    if (this.webhookSettings) {
+      const webhookConfig: WebhookConfig = new WebhookConfig();
+      webhookConfig.type = subscription.event;
+      webhookConfig.filter = subscription.filter;
+      webhookConfig.prevConfiguration = this._previousFilter;
+      webhookConfig.method = this.webhookSettings.getFormControl('method').value;
```
laneli, line 217: Also here if we provide an `@Output` for e.g. webhookChanged in the webhook settings, we then have a defined API and there is no need to include it with ViewChild.

laneli (approved):

Issue #5138, opened 2021-09-02
Shipyard-controller: Every 10 seconds an error is produced with "could not load queued sequences"


agrimmer:
**Environment**

* Client OS (e.g., Linux, macOS, Windows): any
* Keptn Version (`keptn version`): Migrated from Keptn 0.8.6 to 0.9.0
* Kubernetes Cloud Provider (e.g., GKE, AKS): any
* Kubernetes version (`kubectl version`): any

**Affected Component**
* [ ] Docs
* [ ] CLI
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [ ] Helm
* [ ] Istio
* [ ] Bridge
* [ ] Approval
* [ ] Datastore
* [ ] REST API
* [ ] eventbroker / distributor
* [ ] jmeter
* [x] shipyard-controller

**Describe the bug**
Every 10 seconds the shipyard-controller produces the following error log
`time="2021-09-02T15:46:12Z" level=error msg="could not load queued sequences" error="no matching event found"`

**To Reproduce**
Look at the logs of the shipyard-controller in version 0.9.0 after an upgrade of Keptn 0.8.6.

What is the root-cause of this error message?

PR #5166, opened 2021-09-07
fix(shipyard-controller): Adapted log output when no queued sequence is found (#5138)

bacherfl:
Backport of #5138

This PR adapts the log output of the shipyard controller by distinguishing between an `ErrNoEventFound` and any other type of error that might happen when retrieving queued sequences from the mongodb collection. In the case of an `ErrNoEventFound`, which of course can always be the case when there are currently no events in the queue, it doesn't make sense to generate an error log - in this case, no log output is generated to reduce the noise in the logs.


RealAnna (approved):
LGTM

thisthat (approved):

PR #5171, opened 2021-09-07
fix: Fall back to previous git credentials when updating upstream fails (#5064)

### LINKED ISSUE #5064: Providing an invalid git token disables the access to the repository
**Describe the bug**
If you provide an invalid token via POST `controlPlane/v1/project`, accessing the git repository is not possible anymore. If you try to fetch some resources like the shipyard file, the API returns an error `could not check out branch`.

Another thing that may be interesting, but does not need to be changed, is the response message if an invalid token is provided. At the moment it looks like that it is scanned for the provided token and replaces it with some asterisks. If you provide a token like "a", the response looks like this
image

**Expected behavior**
An invalid provided token does not affect or overwrite the git upstream

bacherfl:
Closes #5064

### DIFF
--- a/configuration-service/handlers/project.go
+++ b/configuration-service/handlers/project.go
@@ -36,6 +36,13 @@ func PostProjectHandlerFunc(params project.PostProjectParams) middleware.Respond
 	common.LockProject(params.Project.ProjectName)
 	defer common.UnlockProject(params.Project.ProjectName)
 
+	rollbackFunc := func() {
+		logger.Infof("Rollback: try to delete created directory for project %s", params.Project.ProjectName)
+		if err := os.RemoveAll(config.ConfigDir + "/" + params.Project.ProjectName); err != nil {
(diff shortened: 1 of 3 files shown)


configuration-service/handlers/project.go
```diff
@@ -36,6 +36,13 @@ func PostProjectHandlerFunc(params project.PostProjectParams) middleware.Respond
 	common.LockProject(params.Project.ProjectName)
 	defer common.UnlockProject(params.Project.ProjectName)
 
+	rollbackFunc := func() {
```
thisthat, line 39: I would also add a test that verifies `rollbackFunc` is triggered in case of errors

thisthat (approved):

configuration-service/handlers/project.go
thisthat: As discussed with @bacherfl, tests will be added in a follow up

PR #5193, opened 2021-09-08
feat(bridge): Show list of files and link to git repo per stage for a service (#4506)

### LINKED ISSUE #4506: Manage files for services
## User Story

As a user, I would like to add, delete, update files (e.g., SLO, remediation file, Helm Chart) 

#### Description

Building a UI for managing files (e.g., adding, deleting, updating) would be a wrapper around solutions (GitHub, Bitbucket, etc.) that are already available. To address the above user story: 

* List the resources (files) that are stored for the service. 
* To manage those, the user should get redirected to the branch and folder of the upstream repository (see stage `dev` and `staging` in mockup). 
* If no upstream repo is configured, a user has to set that first to use the add/delete/update capabilities. (see stage `production` in mockup). (For clarification, if a project has a git upstream repo all stages share the same repo. The situation as shown in the mockup (two stages have an upstream, one stage does not) can never be the case.) 

image

❓ The technical difficulty is how to derive the links to the branches for GitHub, Bitbucket, Azure DevOps?

### Acceptance criteria
- [x] Icon button "Settings" on settings page
- [x] Services settings page
- [x] Services -> List of services in the settings navigation
- [x] List of resources separated by stage
- [x] Link to branch and folder on the upstream repo
- [x] Link to configure Git repo when no upstream repo is set

### Definition of done
- [x]  Everything in acceptance criteria implemented and manually tested
- [x]  Added E2E Test
- [ ]  Documentation updated

laneli:
Closes #4506
Signed-off-by: Elisabeth Lang elisabeth.lang@dynatrace.com


bridge/client/app/_components/ktb-edit-service/ktb-edit-service.component.ts
```diff
@@ -40,6 +46,73 @@ export class KtbEditServiceComponent implements OnDestroy {
       this.eventService.deletionProgressEvent.next({isInProgress: true});
       this.deleteService();
     });
+
+    this.project$?.pipe(
+      takeUntil(this.unsubscribe$),
+    ).subscribe(project => {
+      if (!this.fileTree && project && (project.gitRemoteURI || project.gitRemoteURI !== '') && this.serviceName) {
```
Kirdock, line 53: Is the remoteURI check wrong? It enters the `if`, if it is `undefined`. I guess you meant `&&` instead of `||` but even then `project.gitRemoteURI !== ''` is not needed.
`if(project.gitRemoteURI)` is false if it is empty string

bridge/client/app/_components/ktb-edit-service/ktb-edit-service.component.ts
```diff
@@ -40,6 +46,73 @@ export class KtbEditServiceComponent implements OnDestroy {
...
+        this.getResourcesAndTransform(project, this.serviceName);
+      }
+    });
+  }
+
+  public getResourcesAndTransform(project: Project, serviceName: string): void {
+    this.dataService.getServiceResourceForAllStages(project.projectName, serviceName).subscribe((resources) => {
+      this.fileTree = [];
+      project.stages.forEach(stage => {
+        this.fileTree?.push({stage: stage.stageName, files: this.getFileTreeForStage(resources, stage.stageName)});
```
Kirdock, line 63: You could change it to a simple `for-loop` (`for const stage of project.stages`). Then it is not necessary to add optional chaining here.
Another option would be `this.fileTree = project.stages.map(stage => ...)`

bridge/server/services/api-service.ts
```diff
@@ -104,4 +104,12 @@ export class ApiService {
   public getShipyard(projectName: string): Promise<AxiosResponse<Resource>> {
     return this.axios.get<Resource>(`${this.baseUrl}/configuration-service/v1/project/${projectName}/resource/shipyard.yaml`);
   }
+
+  public getServiceResource(projectName: string, stageName: string, serviceName: string, nextPageKey?: number): Promise<AxiosResponse<ResourceResponse>> {
+    let url = `${this.baseUrl}/configuration-service/v1/project/${projectName}/stage/${stageName}/service/${serviceName}/resource`;
+    if (nextPageKey) {
+      url += `?nextPageKey=${nextPageKey}`;
+    }
+    return this.axios.get<ResourceResponse>(url);
```
Kirdock, line 113: I don't really like the approach to add query parameters this way because it could easily raise some problems (caused by invalid provided characters like space or percent). The way how to add query parameters is:
```ts
const url = `${this.baseUrl}/configuration-service/v1/project/${projectName}/stage/${stageName}/service/${serviceName}/resource`;
const params: {[key: string]: string} = {};
if (nextPageKey) {
  prams.nextPageKey = String(nextPageKey);
}
return this.axios.get<ResourceResponse>(url, params);
```

bridge/server/services/data-service.ts
```diff
@@ -236,6 +238,35 @@ export class DataService {
...
+          const serviceRes: ServiceResource = {...res, stageName: stage.stageName};
+          resources.push(serviceRes);
+        }
+
+        currentPage++;
+        // tslint:disable-next-line:radix
+      } while ((currentPage - 1) !== parseInt(nextPage));
+    }
+
+    return resources;
```
Kirdock, line 267: Resources are wrong concatenated if I see correctly. Either the `for loop` inside of the `do while` should be moved outside or the `resourceResponses` should not be concatenated

Kirdock (changes requested):

bridge/client/app/_components/ktb-edit-service/ktb-edit-service.component.ts
laneli: Ok I was not sure about that and added it just in case. I will remove it.

bridge/client/app/_components/ktb-edit-service/ktb-edit-service.component.ts
```diff
@@ -63,8 +103,97 @@ export class KtbEditServiceComponent implements OnDestroy {
...
+      entry.children = this._getSubTree(uriParts);
+      tree.push(entry);
+    }
+
+    return tree;
+  }
+
+  private _mergeTrees(t1: TreeEntry[], t2: TreeEntry, parent: TreeEntry): void {
+    if (t1.length === 0) {
+      parent?.children?.push(t2);
```
Kirdock, line 151: Could you add a test to cover this `if` statement?

Kirdock (changes requested):
I would move the data aggregation to `ktb-edit-service-file-list`. You can first split the resources by stage and then provide them to the component. You still have a check here if there are any resources.

In my point of view, it would be better to move the merging of the tree to the bridge-server

bridge/shared/interfaces/serviceResource.ts
```diff
@@ -0,0 +1,5 @@
+import { Resource } from './resource';
```
Kirdock, line 1: This file is not used anymore and can be removed

bridge/client/app/ktb-edit-service-file-list/ktb-edit-service-file-list.component.ts
```diff
@@ -0,0 +1,101 @@
...
+  private getEntryLevel(entry: TreeFlatEntry): number {
+    return entry.level;
+  }
+
+  private isEntryExpendable(entry: TreeFlatEntry): boolean {
+    return entry.expandable;
+  }
+
+  private getEntryChildren(entry: TreeEntry): TreeEntry[] {
+    if (!entry.children) {
```
Kirdock, line 75: this if-statement could be removed if `children` is always set from the bridge-server

bridge/client/app/ktb-edit-service-file-list/ktb-edit-service-file-list.component.ts
```diff
@@ -0,0 +1,101 @@
...
+      return [];
+    }
+    return entry.children;
+  }
+
+  private treeTransformer(node: TreeEntry, level: number): TreeFlatEntry {
+    const flatNode = new TreeFlatEntry();
+    flatNode.fileName = node.fileName;
+    flatNode.level = level;
+    flatNode.expandable = !!node.children;
```
Kirdock, line 85: if we decide to always return `children: []`, this should be changed to `node.children.length !== 0`

bridge/server/services/data-service.ts
```diff
@@ -245,4 +276,50 @@ export class DataService {
...
+    const tree: TreeEntry = {fileName, children: [] as TreeEntry[]};
+    const dict: { [key: string]: boolean } = {_: true};
+    const folders: TreeEntry[] = [];
+    const files: TreeEntry[] = [];
+    for (const key in currentDirectory) {
+      if (dict[key]) {
+        files.push(...currentDirectory._.map(item => {
+          return {
+            fileName: item,
+          } as TreeEntry;
```
Kirdock, line 313: If we add `children: []` here, we don't need to consider this case client-side.

bridge/shared/interfaces/resourceFileTree.ts
```diff
@@ -0,0 +1,9 @@
+export interface FileTree {
+  stageName: string;
+  tree: TreeEntry[];
+}
+
+export interface TreeEntry {
+  fileName: string;
+  children?: TreeEntry[];
```
Kirdock, line 8: If we decide to always return `children` the `?` can be removed

Kirdock (changes requested):

bridge/client/app/ktb-edit-service-file-list/ktb-edit-service-file-list.component.ts
laneli: We leave it as is :)

bridge/client/app/ktb-edit-service-file-list/ktb-edit-service-file-list.component.ts
laneli: We leave it as is :)

bridge/server/services/data-service.ts
laneli: We leave it as is :)

bridge/shared/interfaces/resourceFileTree.ts
laneli: We leave it as is :)

Kirdock (approved):

PR #5194, opened 2021-09-08
fix(shipyard-controller): Adapted sequence state representation for case where sequence can not be started (#5137)

### LINKED ISSUE #5137: Bridge: A failed sequence is shown as it would be running and cannot be aborted
**Environment**

* Client OS (e.g., Linux, macOS, Windows): any
* Keptn Version (`keptn version`): 0.9.0
* Kubernetes Cloud Provider (e.g., GKE, AKS): any
* Kubernetes version (`kubectl version`): any

**Affected Component**

* [ ] Docs
* [ ] CLI
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [ ] Helm
* [ ] Istio
* [x] Bridge
* [ ] Approval
* [ ] Datastore
* [ ] REST API
* [ ] jmeter

**Describe the bug**
If a sequence fails, the Bridge shows the sequence as it would be still running. That means, the Bridge shows the spinning wheel for the sequence and also the control buttons for pausing and stopping. See Screenshot:
Screen Shot 2021-09-02 at 16 50 20

**To Reproduce**
* Preparation: Delete the shipyard.yaml in Git repo
* Trigger an evaluation
* Sequence is finished with result `fail` but Bridge still shows that this sequence is running

**Expected behavior**
Sequence should be shown as failed in Keptn Bridge.

bacherfl:
Closes #5137

This PR makes sure that event when a sequence cannot be started by a `<sequence>.triggered` event, the state of the failed attempt is properly reflected and can be displayed by the bridge.
The following scenarios have been handled and are covered by integration tests:

- Shipyard file could not be retrieved
- Shipyard could be retrieved, but does not have the correct version
- Valid Shipyard is available, but no matching sequence definition found

In these cases, a sequence state will be created and immediately be set to `finished`, and the `lastFailedEvent` property will be set, so the bridge can display it as a failed sequence. The resulting `<sequence>.finished` event then contains information about why it has failed in the `data.message` property.

### DIFF
--- a/shipyard-controller/handler/shipyardcontroller.go
+++ b/shipyard-controller/handler/shipyardcontroller.go
@@ -531,7 +531,24 @@ func (sc *shipyardController) startTaskSequence(event models.Event) error {
 
 	taskSequence, err := sc.getTaskSequenceInStage(eventScope.Stage, taskSequenceName, shipyard)
 	if err != nil {
-		return err
+		msg := fmt.Sprintf("could not get definition of task sequence %s: %s", taskSequenceName, err.Error())
+		finishedEvent := event
+		finishedEventData := keptnv2.EventData{
+			Project: eventScope.Project,
+			Stage:   eventScope.Stage,
+			Service: eventScope.Service,
+			Labels:  eventScope.Labels,
+			Status:  keptnv2.StatusErrored,
+			Result:  keptnv2.ResultFailed,
+			Message: msg,
+		}
+		finishedEvent.Data = finishedEventData
+
+		sc.onSequenceFinished(finishedEvent)
+		return sc.sendTaskSequenceFinishedEvent(&models.EventScope{
+			EventData:    finishedEventData,
+			KeptnContext: event.Shkeptncontext,
+		}, taskSequenceName, event.ID)
 	}
 	sc.onSequenceStarted(event)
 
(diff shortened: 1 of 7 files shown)


shipyard-controller/handler/shipyardcontroller.go
```diff
@@ -537,7 +542,24 @@ func (sc *shipyardController) startTaskSequence(event models.Event) error {
...
+			Result:  keptnv2.ResultFailed,
+			Message: msg,
+		}
+		finishedEvent.Data = finishedEventData
+
+		sc.onSequenceFinished(finishedEvent)
+		return sc.sendTaskSequenceFinishedEvent(&models.EventScope{
+			EventData:    finishedEventData,
+			KeptnContext: event.Shkeptncontext,
+		}, taskSequenceName, event.ID)
```
thisthat, line 562: should also this code be pushed to its own function? This way it can be shared with the one in `onTriggerSequenceFailed()`

bacherfl: you're right - that can be reused here as well

thisthat (approved):

shipyard-controller/handler/shipyardcontroller.go
thisthat: Thank you :shipit:

PR #5230, opened 2021-09-14
fix(distributor): fix issue of having no initial pubsub topic defined

### LINKED ISSUE #5189: Distributor: Updating subscriptions does not work when no initial topic has been defined
Currently, when the `PUBSUB_TOPIC` env var is empty during the startup of the distributor, it will not create a NATS connection. The problem here is that, when a subscription is defined afterwards via the uniform API, the new subscription will not be applied to the distributor.

**Steps to reproduce:**

1. Deploy a keptn service with a distributor with `PUBSUB_TOPIC` set to an empty value
2. Add a subscription for the service
3. Send an event for that topic - Nothing is received  

**Workaround:**
At the moment, this problem can be avoided using the following workaround for services that should not be subscribed to any topic initially, but may be configured to receive events afterwards:
In this case, the `PUBSUB_TOPIC` env var can be set to any value that does not reflect a Keptn event type (e.g. `empty-topic`)

warber:
closes #5189

### DIFF
--- a/distributor/pkg/config/envconfig.go
+++ b/distributor/pkg/config/envconfig.go
@@ -142,3 +142,10 @@ func (env *EnvConfig) GetPubSubRecipientURL() string {
 	}
 	return recipientService + ":" + env.PubSubRecipientPort + path
 }
+
+func (env *EnvConfig) GetPubSubTopics() []string {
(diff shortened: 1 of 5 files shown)


bacherfl (approved):
lgtm

PR #5249, opened 2021-09-15
feature(distributor): exclusive message processing for multiple distributors

### LINKED ISSUE #4689: Research how to avoid multiple event processing when multiple pods of the same service are running
## Action Item

In the case of multiple Keptn core pods running, messages could be processed multiple times.
Only a single pod of the same replica should be able to process a given event. Note that, it shall still be possible to process messages multiple times because *distinct* keptn integrations are subscribed to it. (e.g. two keptn integrations reacting to `*.test.triggered`)

NATS provides the feature of distributing / load balancing the messages between multiple subscribers in a queue group. This feature is promising for addressing this issue. 

### Resources
https://docs.nats.io/nats-concepts/queue

## DoD 

- [ ] Document how we could limit the event received by replicas of the same service.
- [ ] Prototype implementation that shows the chosen solution works

warber:
closes #4689


mowies (approved):

bacherfl (approved):
lgtm

PR #5258, opened 2021-09-16
fix(jmeter-service): prevent failure if deploymentURIs does not end with a '/'(#3612)

### LINKED ISSUE #3612: JMeter service fails if deploymentURI doesnt end with a / or a proper path!
The JMeter-service assumes that the passed url in deploymentURI ends with a path, e.g: / - if that path is not there the JMeter service will call the script with an incorrect CHECK_PATH variable which can cause issues. e.g the default jmeter script basicCheck.jmx relies on the CHECK_PATH variable

To give you an example. When the deploymentURI looks like this everything is fine:
```
"deploymentURIsPublic": [
  "http://yourappdomain.abc/"
]
```

however - when the URI doesnt include the path at the end it will fail
```
"deploymentURIsPublic": [
  "http://yourappdomain.abc"
]
```

## DoD

- [x] Check for a correct URI and append `/` if missing
- [x] Unit Tests
- [x] Update the documentation addressing #4513

RealAnna:
## This PR
If a valid parsed deployment url has an empty path add '/' to it, added corresponding one line in the unit test to check this 

Fixes #3612

### DIFF
--- a/jmeter-service/main.go
+++ b/jmeter-service/main.go
@@ -196,10 +196,18 @@ func getTestInfo(data keptnv2.TestTriggeredEventData, shkeptncontext string) *Te
 //
 func getServiceURL(data keptnv2.TestTriggeredEventData) (*url.URL, error) {
 	if len(data.Deployment.DeploymentURIsLocal) > 0 && data.Deployment.DeploymentURIsLocal[0] != "" {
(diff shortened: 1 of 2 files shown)


thisthat (approved):
LGTM

PR #5263, opened 2021-09-16
fix(shipyard-controller): Adapted sequence state representation for case where sequence can not be started (#5137) 

bacherfl:
Backport of #5137


thisthat (approved):
Thank you 🎉

Issue #5259, opened 2021-09-16
Tasks might interfere if another sequence was paused but task still running


ermin-muratovic:
**Environment**

* Keptn Version (`keptn version`): 0.9.0

**Describe the bug**
If a sequence is paused, the currently running task might still take some time. If another sequence is triggered meanwhile, the sequence would start and it might happen that two tasks are executed at the same time.

As an example, in my case the helm-service failed installing/upgrading chart, because the deployment task in the paused sequence before was still running.

![image](https://user-images.githubusercontent.com/6098219/133598396-551d93b0-7a01-4a43-bd2e-18be6fa758ce.png)

**To Reproduce**

1. Trigger a delivery sequence
2. Pause the sequence, while the deployment task is running
3. Trigger a new delivery sequence
4. See the new sequence starting and if both deployment task are running at the same time you can see the error

**Expected behavior**
I would expect the new sequence to wait for the task in the paused sequence to be finished.

**Screenshots**
![image](https://user-images.githubusercontent.com/6098219/133598681-8e4523ca-4179-40c1-b21f-879e27e6d55f.png)

PR #5275, opened 2021-09-17
fix(bridge): grouping sequence after pause #5154

### LINKED ISSUE #5154: Bridge not grouping paused sequences correctly
When pausing and resuming in Keptn 0.9 the bridge does not show the grouping of the tasks, in fact, there are shown as individual task.

image

ermin-muratovic:
## This PR
- makes sure that the abort/resume/pause buttons are disabled correctly in case sequence is timedOut
- notifies the user about waiting status in case other sequences are blocking the execution

I could not reproduce the issue with grouping sequence tasks, which feels actually like a bug in the shipyard-controller.
But I reported another bug to the shipyard-controller, see #5259.

### Related Issues
Fixes #5154 

image

### DIFF
--- a/bridge/client/app/_components/ktb-root-events-list/ktb-root-events-list.component.html
+++ b/bridge/client/app/_components/ktb-root-events-list/ktb-root-events-list.component.html
@@ -10,14 +10,15 @@
         <div class="container">
           <div fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="5px">
             <div fxFlex fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="5px">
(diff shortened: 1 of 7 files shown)


laneli:
A general improvement that comes to my mind:
Can we have a status that it is waiting also on the list. Like idk 3 dots like https://github.com/tnhu/loading-dots
just to indicate that it is not in progress atm.

ermin-muratovic:
> A general improvement that comes to my mind:
> Can we have a status that it is waiting also on the list. Like idk 3 dots like https://github.com/tnhu/loading-dots
> just to indicate that it is not in progress atm.

Now it is showing the `idle` icon and the text `waiting`.

laneli (approved):

PR #5287, opened 2021-09-17
fix(bridge): Show proper error messages if not OAUTH is configured and prevent login loop (#5086)

### LINKED ISSUE #5086: Bridge: If 401 for API token request is returned, endless redirect loop occurs
How to reproduce:
* Configure API_URL
* Configure a wrong API_TOKEN
* Start bridge locally (as it is easiest to reproduce it here)
* Navigate to the Bridge
* There is now an error message that login is required and that you are redirected to login
* Redirect occurs, and the same procedure is done again -> redirect loop
* At bridge server we can see this error: `data: { code: 401, message: 'incorrect api key auth' }`

What to expect:
* Show an error message that the configured API token is invalid and prevent from redirect loop when no OAuth is configured.

laneli:

Closes #5086


Kirdock (changes requested):
There are two error messages if you try to access a project. This should be reduced to one.
image

Kirdock (approved):

PR #5290, opened 2021-09-17
fix(cli): WIP switch automatic version check based on config

vadasambar:
## This PR

- adds this new feature

### Related Issues

Fixes #1234523


cli/cmd/root.go
```diff
@@ -81,27 +82,20 @@ func initConfig() {
 		logging.LogLevel = logging.QuietLevel
 	}
 
+	cfgMgr := config.NewCLIConfigManager()
```
vadasambar, line 85: I have used `CLIConfigManager` here instead of `viper` because we already have it in place and it works well. 
TODO for the future can be to re-write `NewCLIConfigManager` with viper.

cli/cmd/root_test.go
```diff
@@ -242,6 +244,34 @@ func Test_runVersionCheck(t *testing.T) {
...
 		},
+		{
+			name:           "show version check if `AutomaticVersionCheck` is set to true",
+			cliVersion:     "0.8.0",
+			metadataStatus: http.StatusOK,
+			metadataResponse: keptnapimodels.Metadata{
+				Keptnversion: "0.8.1-dev",
+			},
+			wantOutput: "* Warning: Your Keptn CLI version (0.8.0) and Keptn cluster version (0.8.1-dev) don't match.",
+			cliConfig: func() config.CLIConfig {
```
vadasambar, line 255: Doing this because changing `rootCLIConfig` directly can have impact on other tests.

Issue #5257, opened 2021-09-16
Core: make use of secrets' scopes


thisthat:
## User Story

As a user, I would like to bind a secret to be accessible only to a specific service

## Technical details

The secret-service has the concept of [scope](https://github.com/keptn/keptn/blob/069dd0f5c7b6f37a3737f4c0c9c7cf07a801b039/secret-service/pkg/model/scope.go#L4). The list of hardcoded [scopes](https://github.com/keptn/keptn/blob/master/secret-service/scopes.yaml) should be extended to support also `webhook-service` and `dynatrace-service`, both with `get` permissions. When a secret is created, the secret-service should add the scope in the secret metadata and create the role binding. When a secret gets deleted, the respective role binding is also deleted. 

### DoD
- [x] Webhook service creates a Service Account during its deployment matching the scope name
- [x] Secrets are annotated with scopes 
- [x] Role bindings are created together with secrets

PR #5300, opened 2021-09-20
feat(secret-service): added creation of rolebinding based on scope name

warber:
closes: #5257


secret-service/pkg/backend/secretbackend_k8s.go
```diff
@@ -219,6 +282,7 @@ func (k K8sSecretBackend) createK8sSecretObj(secret model.Secret, namespace stri
 			Namespace: namespace,
 			Labels: map[string]string{
 				"app.kubernetes.io/managed-by": "keptn-secret-service", // add a 'managed-by' label so we can identify secrets managed by the secret-service
```
bacherfl, line 284: let's introduce a constant for `"keptn-secret-service"`

secret-service/pkg/backend/secretbackend_k8s_test.go
```diff
@@ -223,9 +222,25 @@ func TestDeleteK8sSecret(t *testing.T) {
 		ScopesRepository:       scopesRepository,
 	}
 
+	kubernetes.Fake.PrependReactor("list", "secrets", func(action k8stesting.Action) (handled bool, ret runtime.Object, err error) {
```
bacherfl, line 225: If no special logic should be applied when a request for retrieving an object from the fake K8s service should be executed, the object can instead be passed to the `NewSimpleClientSet()` function - afterwards, it can be retrieved, updated and deleted without implementing a specific reactor for each request. Example:

```
kubernetes := k8sfake.NewSimpleClientset(&corev1.Secret{
		ObjectMeta: metav1.ObjectMeta{
			Name:      "my-secret",
			Namespace: "keptn_namespace",
			Labels:    map[string]string{"app.kubernetes.io/scope": "my-scope"},
		},
	},
		&rbacv1.Role{
			Rules: []rbacv1.PolicyRule{
				{
					Resources:     []string{"secrets"},
					ResourceNames: []string{"my-other-secret", "my-secret"},
				},
			},
		},
	)
```

bacherfl (approved):

PR #5366, opened 2021-09-22
chore(bridge): adapt retry-mechanism (#4867)

### LINKED ISSUE #4867: Remove error-retry-mechanism of api calls
At the moment there are 3 retries if an api call fails. For most of the cases this is not needed, like a 404 response for a fetching a project. Either remove the retry-mechanism completely or limit it to error responses with a status code of 500.

Kirdock:
## This PR
- adapts the retry mechanism for HTTP calls to only retry if they are one of the following status codes: 500, 502, 503, 504

### Related Issues
Closes #4867 

### DIFF
--- a/bridge/client/app/_interceptors/http-error-interceptor.ts
+++ b/bridge/client/app/_interceptors/http-error-interceptor.ts
@@ -2,7 +2,7 @@ import { Inject, Injectable } from '@angular/core';
 import { HttpErrorResponse, HttpEvent, HttpHandler, HttpInterceptor, HttpRequest } from '@angular/common/http';
 import { Observable, throwError } from 'rxjs';
 import { catchError, retryWhen } from 'rxjs/operators';
-import { genericRetryStrategy } from './http-generic-retry-strategy';
+import { genericRetryStrategy, RetryParams } from './http-generic-retry-strategy';
 import { DtToast } from '@dynatrace/barista-components/toast';
 import { Location } from '@angular/common';
 import { RETRY_ON_HTTP_ERROR } from '../_utils/app.utils';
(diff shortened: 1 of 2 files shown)


bridge/client/app/_interceptors/http-error-interceptor.ts
```diff
@@ -22,24 +22,20 @@ export class HttpErrorInterceptor implements HttpInterceptor {
...
       .pipe(
         retryWhen(genericRetryStrategy(params)),
         catchError((error: HttpErrorResponse) => {
-
           if (error.status === 401) {
             this._handleUnauthorizedError(error);
           } else if (error.error instanceof ErrorEvent) {
             // A client-side or network error occurred. Handle it accordingly.
-            console.error('An error occurred:', error.error.message);
             this.toast.create(`${error.error.message}`);
```
ermin-muratovic, line 35: The toast disappears after a few seconds and the `console.error` allows to find the error in the console logs again. Not sure if we should remove this? I would actually keep it.

ermin-muratovic (changes requested):

bridge/client/app/_interceptors/http-error-interceptor.ts
laneli: I don't think that we log something internal here just an API response. For debugging and traceablity I would also suggest to keep it.

ermin-muratovic: After some investigation, we found out that the browser logs failed HTTP calls anyway and since we don't provide any additional details to the user, we can skip this `console.error`.

For our refactoring we should definitly consider to provide better error messages to the user, like "`Project xyz could not be found`", instead of "`HTTP GET /project/xyz failed with 404`".

@laneli @thisthat can you add this to our refactoring task?

ermin-muratovic (approved):

PR #5384, opened 2021-09-23
fix(bridge): handle incorrect remediation sequences (#5383)

### LINKED ISSUE #5383: Bridge server breaks if invalid remediation sequence is retreived
There could be the case that a sequence is triggered that is not present in the shipyard file. This leads to a sequence state that does not have any stages and this runs into an error on the bridge server because it is assumed that remediation sequences always have a stage.

Kirdock:
Closes #5383 


ermin-muratovic (approved):
lgtm

PR #5386, opened 2021-09-23
feat(bridge): Show loading bar only on initial data fetch #4910

### LINKED ISSUE #4910: Bridge: Loading bar should be shown only on initial data fetch
## Description
The loading bar on top of the page appears every n seconds when some polling occurs.
This bar suggests somehow, that the Bridge is loading something - which should not be the case because the polling is a background task that should be done silently.
Also, the loading bar that is appearing repeatedly is distracting. If the user examines some data and the bar appears, the attention is drawn away.

My suggestion is that the bar appears only at the initial fetching of data, e.g. after reload or when additional data has to be fetched to visualize the data on the screen. For example, after reload, there are a few seconds where data is not yet shown on the environment screen. Showing a loading bar here would be an indicator that something is still in progress, which totally makes sense. We should therefore limit the use of the loading bar to this.

## Acceptance criteria
- [ ] The bar only appears only on an initial fetch or when additional data is fetched
- [ ] The bar is not shown on background polling tasks

## DoD
- [ ] Acceptance criteria implemented and manually tested

ermin-muratovic:
## This PR
- makes sure that the loading indicator is shown only on initial data fetch

### Related Issues
Fixes #4910


bridge/client/app/_components/ktb-http-loading-bar/ktb-http-loading-bar.component.ts
```diff
@@ -16,70 +16,80 @@ export class KtbHttpLoadingBarComponent implements OnInit, OnDestroy {
   private hideLoadingTimer?: ReturnType<typeof setTimeout>;
   private animateLoadingBarInterval?: ReturnType<typeof setInterval>;
 
-  public loading = false;
+  public loading = 0;
```
Kirdock, line 19: Could you rename this variable to know what it is used for? Because `loading` sounds more like a `boolean` if it is loading or not.

Kirdock (changes requested):
We don't have any tests for this component. Could you add some?

At the moment displaying the progress bar is a bit weird. Maybe we can improve its behavior.
- Only show the progress bar if API call to "XYZ" has never been executed.
=> It is showing up in the uniform registration screen the first time but then never again but the screen fetches this information every time the page is accessed.
- Currently, query parameters are ignored in the URL.
=> The first time a trace is loaded, the progress bar is shown, then never again

ermin-muratovic:
As discussed with @laneli and @Kirdock, we are removing the loading bar. Loading indicators should be used in place where data is loaded.

Kirdock:
Could you align the other loading icons then?
They are different across the screens.
For example:
uniform:
image

services:
image


And some screens don't have a loading indicator like the environment or secret screen.

PR #5388, opened 2021-09-23
feat(bridge): Allow to create secret with scope selection #5269

### LINKED ISSUE #5269: Bridge: secret screen should have a scope selection
## Action Item

The secret screen should be updated with a dropdown menu to select the scope of a secret. 

## Technical details

As a first solution, the dropdown is statically filled with the following options:
 - `keptn-default` (default)
 - `keptn-webhook-service`
 - `dynatrace-service`

The selected scope should be used as a parameter for the `scope` property of the `POST /secret` API call.

## DoD

- [ ] Secrets are created with a scope 
- [ ] Unit test
- [ ] Update the doc (#4745)

ermin-muratovic:
## This PR
- adds a dropdown to select a scope while creating a secret

### Related Issues
Fixes #5269 

image

### DIFF
--- a/bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.html
+++ b/bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.html
@@ -9,10 +9,16 @@
       <form [formGroup]="createSecretForm">
         <div class="mb-3" fxLayout="row" fxLayoutAlign="none end" fxLayoutGap="15px">
           <dt-form-field>
-            <dt-label>Name *</dt-label>
+            <dt-label class="required">Name</dt-label>
             <input formControlName="name" type="text" required dtInput placeholder="Secret name" />
             <dt-error *ngIf="getFormControl('name')?.errors?.required">Must not be empty</dt-error>
             <dt-error *ngIf="getFormControl('name')?.errors?.pattern">Name must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character</dt-error>
+          </dt-form-field>
+          <dt-form-field>
+            <dt-label class="required">Scope</dt-label>
+            <dt-select formControlName="scope" class="mr-2 item" placeholder="Choose scope" aria-label="Choose your task">
+              <dt-option *ngFor="let scope of scopes" [value]="scope" [textContent]="scope"></dt-option>
+            </dt-select>
           </dt-form-field>
         </div>
 
@@ -21,13 +27,13 @@
             <form [formGroup]="dataGroup">
               <div class="mb-1" fxLayout="row" fxLayoutAlign="none end" fxLayoutGap="15px">
                 <dt-form-field>
-                  <dt-label>Key *</dt-label>
+                  <dt-label class="required">Key</dt-label>
                   <input formControlName="key" type="text" required dtInput placeholder="Key" autocomplete="false" />
                   <dt-error *ngIf="dataGroup.get('key')?.errors?.required">Must not be empty</dt-error>
                   <dt-error *ngIf="dataGroup.get('key')?.errors?.pattern">Key must consist of alphanumeric characters, '-', '_' or '.'</dt-error>
                 </dt-form-field>
                 <dt-form-field>
-                  <dt-label>Value *</dt-label>
+                  <dt-label class="required">Value</dt-label>
                   <input formControlName="value" type="password" required dtInput placeholder="Value" autocomplete="false" />
                   <dt-error>Must not be empty</dt-error>
                 </dt-form-field>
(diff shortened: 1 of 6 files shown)


bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.spec.ts
```diff
@@ -18,10 +21,51 @@ describe('KtbCreateSecretFormComponent', () => {
...
+
+    // when
+    component.getFormControl('name')?.setValue('test');
+    component.getFormControl('scope')?.setValue(component.scopes[1]);
+    component.data?.controls[0].get('key')?.setValue('testKey');
+    component.data?.controls[0].get('value')?.setValue('testValue');
+    createButton.click();
+
+    // then
+    expect(spy).toHaveBeenCalled();
```
Kirdock, line 64: I would change this to
```ts
const secret: Secret = new Secret();
secret.name = 'test';
secret.scope = component.scopes[1];
secret.data.push({key: 'testKey', value: 'testValue'});
expect(spy).toHaveBeenCalledWith(secret);
```
and remove the code below.

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.spec.ts
```diff
@@ -18,10 +21,51 @@ describe('KtbCreateSecretFormComponent', () => {
...
+    const spy = jest.spyOn(dataService, 'addSecret');
+
+    // when
+    component.getFormControl('name')?.setValue('test');
+    component.data?.controls[0].get('key')?.setValue('testKey');
+    component.data?.controls[0].get('value')?.setValue('testValue');
+    createButton.click();
+
+    // then
+    expect(spy).toHaveBeenCalled();
```
Kirdock, line 43: I would change this to
```ts
const secret: Secret = new Secret();
secret.name = 'test';
secret.scope = component.scopes[1];
secret.data.push({key: 'testKey', value: 'testValue'});
expect(spy).toHaveBeenCalledWith(secret);
```
and remove the code below.

Maybe also make this secret global or create a mock-file because the values are re-used.

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.html
```diff
@@ -9,26 +9,32 @@ <h2>Create Secret</h2>
...
-            <dt-label>Name *</dt-label>
-            <input formControlName="name" type="text" required dtInput placeholder="Secret name" />
+            <dt-label class="required">Name</dt-label>
+            <input formControlName="name" type="text" required dtInput placeholder="Secret name" uitestid="keptn-secret-name-input" />
             <dt-error *ngIf="getFormControl('name')?.errors?.required">Must not be empty</dt-error>
             <dt-error *ngIf="getFormControl('name')?.errors?.pattern">Name must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character</dt-error>
           </dt-form-field>
+          <dt-form-field>
+            <dt-label class="required">Scope</dt-label>
+            <dt-select formControlName="scope" class="mr-2 item" placeholder="Choose scope" aria-label="Choose your task" uitestid="keptn-secret-scope-input">
```
Kirdock, line 19: Could you set a fixed width for it? Depending on which item is selected, the width of the control changes.

Kirdock (changes requested):
When I execute the tests, they run through but there are still error messages. Could you look into it?
`Cannot read property '_syncPendingControls' of null`.

Could you add some tests to the `ktb-create-secret-form.component.ts` component so we reach about 100% test coverage?

Kirdock (commented):
I know I'm annoying but could you add some unit tests for successfully and unsuccessfully adding a secret (if `navigate` is called or `loading` is set to `false` again)?
https://github.com/keptn/keptn/pull/5388/files#diff-51538cc4a74f0adb4ecfbfff9feb7285c9c694e519318db871569ba809d5604eR56-R59

laneli:
I took a look at it and it's a bit too unsteady to look at. I would propose a solution like this:

image

bridge/client/app/_models/secret.ts
```diff
@@ -1,3 +1,9 @@
+export enum Scopes {
+  KeptnDefault = 'keptn-default',
```
laneli, line 2: Just a remark ... I added it also in my PR here:  `shared\interfaces\secret.ts`
as: 
```
export enum SecretScope {
  DEFAULT = 'keptn-default',
  WEBHOOK = 'keptn-webhook-service'
}
```
This is needed to retrieve secrets only for a specific scope, so as server and bridge make use of it, can we streamline it?

laneli: Also I would vote for enum name `SecretScope` as there could be other scopes too.

Kirdock (approved):
LGTM

PR #5437, opened 2021-09-29
feat(bridge): Add secrets to webhook configuration (#4751)

### LINKED ISSUE #4751: Add secrets to webhooks
## User Story

As a user, I would like to use a secret value for the webhook call.

### Details

webhook-secret

Detailed mockup:
Add Subscription4

When a user clicks on the :lock: icon, the value field is replaced with the possibility to select a secret and one of its keys. The list of secrets and their keys is implemented by #4749. 
As shown in the mockup for the "Custom payload" a popup overlay is opened that provides a list of secrets. The list is expandible, so for every secret, there will be the related keys in the expanded section.
❗ Only secrets with `scope` equal to `webhook-service` shall be presented. 

When a secret name (e.g., `mysecret`) and a secret key (e.g., `SECRET_KEY`) is selected, they should be reported in the `webhook.yaml` as follows:

```yaml
apiVersion: webhook.keptn.sh/v1alpha1
kind: Webhook
metadata:
  name: webhook-configuration
spec:
  envFrom:
  - secretRef:
    name: mysecret
  requests:
  - "curl --header "Content-Type: application/json" --header "x-token: ${SECRET_KEY}" --request POST –data '{"event":"{{.event.type}}","service":"{{.event.data.service}}","environment":"{{.event.data.stage}}' https://api.mycompany.com"
```

## Acceptance Criteria

- [x] The webhook service can be configured with secrets

## DoD

- [x] `webhook.yaml` with secrets is created via the Bridge
- [x] Only secrets with `scope` equal to `keptn-webhook-service` can be selected
- [x] Unit tests
- [x] UI tests

## Dependencies

- #4749 
- #4750

laneli:
Closes #4751


Kirdock (changes requested):
- Can this be improved? I would expect the expand-icon to be on the bottom right corner.
image
- At the moment a secret is only replaced once (bridge to webhook format). A payload like `{{.mysecret.key}}  {{.mysecret.key}}` would be changed to `{{.env.secret_mysecret_key}}  {{.mysecret.key}}` instead of `{{.env.secret_mysecret_key}}  {{.env.secret_mysecret.key}}`

bridge/shared/models/webhook-config.ts
```diff
@@ -10,8 +24,9 @@ export class WebhookConfig implements wc {
   public method: WebhookConfigMethod;
   public url: string;
   public payload: string;
-  public header: { name: string; value: string }[];
+  public header: WebhookHeader[];
   public proxy?: string;
+  public secrets?: WebhookSecret[];
```
Kirdock, line 29: Currently, the secrets are also sent to the client but the client does not need them

Kirdock (changes requested):
We now have two different kinds of secrets.
`Secret 1` (includes the `data` property) that is being sent from client to secret-service
`Secret 2` (includes the `keys` property) that is being received from the secret service.
In my opinion, we should distinguish between these two and split them in order to have a cleaner interface on how data is sent and retrieved. Then we also don't have to always check if the property `key `or `data` of `Secret 1` or `Secret 2` exists

Kirdock (approved):

PR #5446, opened 2021-09-30
doc(maintainers): add vadasambar as member

vadasambar:
- https://github.com/keptn/community/issues/56

## This PR
- adds vadasambar as member

### Notes
I have set the affiliation to `No Affiliation` for now (I might change this in the future; Will add an organization if that doesn't work)


vadasambar:
@thisthat I think I need someone to merge this PR for me because I don't have the permission. And a comment here https://github.com/keptn/community/issues/56

PR #5452, opened 2021-10-01
feat(bridge): Allow multiple webhooks with same subscription configuration (#5267)

### LINKED ISSUE #5267: (WIP) Bridge: Allow multiple subscription to change a single webhook.yaml file
## Action Item

Update the WebHook screen to leverage the subscription id.

### Technical Details

When two subscriptions have a webhook integration configured on the same service but for different stages, they write on the same `webhook.yaml` configuration file.  With #5264, the webhook service can parse the subscription id.
![][https://user-images.githubusercontent.com/729071/132518770-9349183b-113d-4edd-b717-72cc7f26c928.png]

The Bridge should be updated to leverage the subscription id to determine which configuration is being changed and only adapt that part of the `webhook.yaml` file. 

## DoD

- [ ] `webhook.yaml` is correctly updated based on the subscription id
- [ ] Unit tests

## Dependencies
- [ ] https://github.com/keptn/keptn/issues/5264

Kirdock:
Closes #5267 

### DIFF
--- a/bridge/server/services/data-service.ts
+++ b/bridge/server/services/data-service.ts
@@ -16,7 +16,7 @@ import { UniformRegistrationLocations } from '../../shared/interfaces/uniform-re
 import { WebhookConfig, WebhookConfigFilter } from '../../shared/models/webhook-config';
 import { UniformRegistrationInfo } from '../../shared/interfaces/uniform-registration-info';
 import { WebhookConfigYaml } from '../interfaces/webhook-config-yaml';
-import { UniformSubscriptionFilter } from '../../shared/interfaces/uniform-subscription';
+import { UniformSubscription, UniformSubscriptionFilter } from '../../shared/interfaces/uniform-subscription';
 import axios from 'axios';
 import { Resource } from '../../shared/interfaces/resource';
 import { FileTree, TreeEntry } from '../../shared/interfaces/resourceFileTree';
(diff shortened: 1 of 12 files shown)


bridge/server/services/data-service.ts
```diff
@@ -327,6 +327,17 @@ export class DataService {
     return Yaml.parse(shipyard);
   }
 
+  public async createSubscription(integrationId: string, subscription: UniformSubscription, webhookConfig: WebhookConfig): Promise<void> {
+    const response = await this.apiService.createSubscription(integrationId, subscription);
+    const subscriptionId = response.data.id;
+    await this.saveWebhookConfig(webhookConfig, subscriptionId);
```
laneli, line 333: Following the path from component to this point webhookConfig is always optional, except here. Because when `req.body.webhookConfig` in `api/index.ts` is undefined it will be undefined here.

Did you also consider that subscriptions are created / updated without a webhook config? Now it seems like it tries to always save a webhook config. Also in `saveWebhookConfig` I could see no mechanism to just not save the config as it is also not optional there.

laneli (changes requested):

bridge/server/services/data-service.ts
Kirdock: You are right. Forgot to make it optional. Fixed it

laneli (approved):

PR #5454, opened 2021-10-01
fix(bridge): Evaluation board only updates if there are new evaluations (#5396)

### LINKED ISSUE #5396: Follow-up on: Service evaluation screen gets refreshed every time after polling data
## Bug

In https://github.com/keptn/keptn/issues/4491, the bug of the refreshing service evaluation is explained.  With Keptn 0.9.1, this bug still exists even though it should have been fixed.

* I expand the SLI to see the compared with value: 
image

* The next poll closes the expandable: 
image

* The next poll jumps to the top: 
image

## Details

- It could be clarified that the bug fix has been merged into `master` branch using: https://github.com/keptn/keptn/pull/4582

## Definition of Done
- [ ] Do not refresh the page when no new evaluation is available. 
- [ ] While a user is investigating the evaluation heatmap and details, the components on the top (stage overview and list of sequences) are updated in the background.  
- [ ] A "Refresh ..." appears when a new evaluation is available. Very similar to what GitHub does on open PRs. The user can then do an update (but will not happen automatically)
image

Kirdock:
Fixes #5396


bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.html
```diff
@@ -1,7 +1,11 @@
 <div *ngIf="evaluationData">
   <div class="mt-1 mb-1" *ngIf="showChart && !isInvalidated">
     <div fxLayout="row" class="pr-2">
-      <div fxFlex></div>
+      <div fxFlex>
+        <button dt-button variant="secondary" color="warning" *ngIf="updateResults" (click)="refreshEvaluationBoard(updateResults)">
```
laneli, line 5: We already have a refresh button on several screens (dashboard, sequence view, ...). I would suggest reusing that button to keep consistent. Also, I think the warning color is not needed here.
```
<button dt-button (click)="refreshEvaluationBoard(updateResults)" *ngIf="updateResults">
  <dt-icon name="refresh"></dt-icon>
  Reload
</button>
```

laneli (changes requested):

laneli (approved):

PR #5456, opened 2021-10-02
feat: customize image registry

### LINKED ISSUE #4933: Allow installs/upgrades from private, authenticated container registries
Allow installs/upgrades from private, authenticated container registries.
In the enterprise, this is the only way some of us are allowed to get container images.
No where in the [helm charts](https://github.com/keptn/keptn/blob/cc13a460e6ec79337f95bd41d645f5a5046cee79/installer/manifests/keptn/charts/control-plane/templates/core.yaml) is there a way to set `imagePullSecrets` via values, for example.

Allow us to pass in parameters so install/upgrades pull container images from private, authenticated container registries.

Either forking your branches & adding template fields for this, or letting installs fail and updating deployment manifests via `kubectl` to correct.
Not ideal.

### LINKED ISSUE #5248: Image of containers are not set properly when doing a `keptn upgrade`
Due to the recent changes made in https://github.com/keptn/keptn/pull/4984 , upgrading from Keptn <= 0.9.0 fails because the old value for the `repository` property of each service is not updated. Previously, this property included the docker registry, and looked like this for example: `docker.io/keptn/shipyard-controller`. In the referenced PR, this has been split up into the properties `registry` and `repository`, so the default values in the `values.yaml` now look like this:

```
    image:
      registry: docker.io
      repository: keptn/shipyard-controller
```

This works fine when a new installation of Keptn is created on a cluster. However, when upgrading a previous release, the resulting values then look like this:

```
    image:
      registry: docker.io
      repository: docker.io/keptn/shipyard-controller
```

Therefore, in this case, the resulting image will be set to `docker.io/docker.io/keptn/shipyard-controller` and will therefore fail.

lmmarsano:
# This PR
allow user to globally customize image pull registry & pull secrets

## Related Issues
closes #4933
fixes #5248

## Notes
follows [bitnami example](https://github.com/bitnami/charts/tree/master/bitnami/common/)
refactoring repetitive code into a [library chart](https://helm.sh/docs/topics/library_charts/) may be warranted

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -75,3 +75,4 @@ For a realtime overview of current contributors to the Keptn project, we refer t
 * [Kolla](https://github.com/krkolla)
 * [Alasdair Patton](https://github.com/alipatton10)
 * [Brad McCoy](https://github.com/bradmccoydev)
+* [Luis Marsano](https://github.com/lmmarsano)
--- a/helm-service/chart/README.md
+++ b/helm-service/chart/README.md
@@ -11,14 +11,18 @@ The following table lists the configurable parameters of the Helm-service chart 
 
 | Parameter                | Description             | Default        |
 | ------------------------ | ----------------------- | -------------- |
-| `helmservice.image.repository` | Container image name | `"docker.io/keptn/helm-service"` |
+| `global.imagePullSecrets` | An array of image registry secrets | `[]` |
+| `global.imageRegistry` | Registry where all images are located |  |
+| `helmservice.image.registry` | Registry where the image is located |  |
+| `helmservice.image.repository` | Container image name | `"keptn/helm-service"` |
 | `helmservice.image.pullPolicy` | Kubernetes image pull policy | `"IfNotPresent"` |
 | `helmservice.image.tag` | Container tag | `""` |
 | `helmservice.service.enabled` | Creates a kubernetes service for the helm-service | `true` |
 | `distributor.stageFilter` | Sets the stage this helm service belongs to | `""` |
 | `distributor.serviceFilter` | Sets the service this helm service belongs to | `""` |
 | `distributor.projectFilter` | Sets the project this helm service belongs to | `""` |
-| `distributor.image.repository` | Container image name | `"docker.io/keptn/distributor"` |
+| `distributor.image.registry` | Registry where the image is located |  |
+| `distributor.image.repository` | Container image name | `"keptn/distributor"` |
 | `distributor.image.pullPolicy` | Kubernetes image pull policy | `"IfNotPresent"` |
 | `distributor.image.tag` | Container tag | `""` |
 | `remoteControlPlane.enabled` | Enables remote execution plane mode | `false` |
--- a/helm-service/chart/templates/_helpers.tpl
+++ b/helm-service/chart/templates/_helpers.tpl
@@ -61,3 +61,90 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Return the proper image name
+{{ dict "imageRoot" .Values.path.to.the.image "context" $ | include "helm-service.images.image" }}
+*/}}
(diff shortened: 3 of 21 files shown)


warber:
thx @lmmarsano, can you pls do a rebase/merge of master and resolve the conflict? thx

checkelmann:
Hi,

from user perspective it is not clear for my understanding what this change will affect in our current system. 

I'm rolling out my keptn installation using terraform with helm where I've set my internal images like so

```
set {
 name = "control-plane.mongodb.image.repository"
 value = "${var.container_registry}/centos/mongodb-36-centos7"
 }
```

Do I now need to set additional values or will this work as well?

What about upgrades of existing installations, and how should this be done now?
Is there any kind of documentation available regarding this change?

Sorry @lmmarsano for these questions, but this PR is quite big and could result in upgrade issues if users rolling out keptn using any kind of automation which could break.

Thank you!

lmmarsano:
> Do I now need to set additional values or will this work as well?
> 
> What about upgrades of existing installations, and how should this be done now? Is there any kind of documentation available regarding this change?
> 
> Sorry @lmmarsano for these questions, but this PR is quite big

The [individual commits](https://github.com/keptn/keptn/pull/5456/commits) answer some of these questions.
The large revert commit links back to its original pull request broken down to cleaner commits, too.
You can quickly preview resolved configurations with a `helm template` command like
```shell
helm template --set global.imageRegistry=$REGISTRY keptn ./keptn | grep image:.
```

and whatever value you want to substitute for `$REGISTRY`.

The new parameters behave similar to [those in bitnami charts](https://github.com/bitnami/charts/tree/master/bitnami/wordpress#parameters).

Nothing additional needs to be set: you can continue as you already do and leave new parameters unset.
Alternatively, if you find repeating yourself revolting, then you can set `global.imageRegistry`, instead: when set, that (or the higher precedence `repository` parameter for the image), `/`, and `repository` join to form the full image name for each `keptn` chart.
New defaults for `repository` omit the registry hostname, which [was always implicitly Docker public registry](https://kubernetes.io/docs/concepts/containers/images/#image-names).

# Caveats
- Redundantly setting registry in `repository` & either `registry` parameter of course results in repetition, so don't do that.
- External chart dependencies (ie, `nats`) still need their parameters set as before. Someone should probably tell them to improve their charts.

checkelmann:
lgtm

checkelmann (approved):
lgtm

thisthat (dismissed):
Thank you for your contribution @lmmarsano and sorry for taking so long to approve your PR.
Can I ask you to do a final rebase of your branch with `keptn:master` so we can merge it?

mowies:
Hey @lmmarsano !

Thanks so much for your contribution here!
Sorry for the long wait. We needed some internal discussions on how to proceed with your PR.

The features in the PR are a great addition to keptn but I am afraid that we can't merge it as it is.
The PR has too much going on and actually has 2 different features inside that we would like to split. (usage of bitnami chart patterns/customizing our installer to use the global registry setting)

So, we came up with a plan to still integrate your changes in Keptn in a different way.
The plan is the following:
We will take your PR as inspiration and create 2 separate issues that will be estimated and worked on as usual in our agile process.
This means that it will take some more time until we get around to implementing the tickets, but we feel more confident doing this work ourselves since we want to implement the features a little bit different from your PR.

You will definitely still be added as a contributor to the repo though since you did so much groundwork here!

We hope you're happy with this solution.

lmmarsano:
@mowies If you're going to do that, then as suggested before, a more serious design should probably refactor reusable template logic into a [library chart](https://helm.sh/docs/topics/library_charts/) instead of copy identical code to multiple charts.
That'd cut down repetition and ease maintenance.
It'd require creating an additional chart that a more concerted plan like this could direct.

mowies:
Exactly! So my plan was to either use the bitnami common chart directly (since it is under Apache-2.0 License) or create our own library chart with only the things we need.

mowies:
Here are the 2 tickets that I created from this PR:
* #6042
* #6043

PR #5459, opened 2021-10-04
fix(installer): readiness liveness probes (#5303)

### LINKED ISSUE #5303: Core: Readiness and Liveness probes
## Action Item

Make sure all Keptn core services have a readiness and liveness probe.

## DoD

- [ ] All services have a readiness and liveness probe configured in their deployment

RealAnna:
## This PR
- adds health endpoints to core services 
- verifies all services have readiness and liveness probes working 'properly'
- modifies installer yaml files accordingly

### Related Issues

Fixes #5303 

### Notes
a probe is 'proper' if defined on an existing service endpoint and not on a fictitious one. The health checks are shallow since they do not verify any dependency but the endpoint is setup as the very last thing. In most cases readiness and liveness still correspond.

### DIFF
--- a/helm-service/main.go
+++ b/helm-service/main.go
@@ -7,14 +7,13 @@ import (
 	"github.com/keptn/keptn/helm-service/pkg/configurationchanger"
 	"github.com/keptn/keptn/helm-service/pkg/helm"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"net/http"
 	"net/url"
 
+	api "github.com/keptn/go-utils/pkg/api/utils"
 	"github.com/keptn/keptn/helm-service/pkg/namespacemanager"
+	keptnutils "github.com/keptn/kubernetes-utils/pkg"
 	"log"
 	"os"
-
-	keptnutils "github.com/keptn/kubernetes-utils/pkg"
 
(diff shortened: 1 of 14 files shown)


helm-service/main.go
```diff
@@ -191,6 +191,12 @@ func _main(args []string, env envConfig) int {
 	return 0
 }
 
+func healthEndpointHandler(w http.ResponseWriter, r *http.Request) {
+	if r.URL.Path == "/health" {
+		w.WriteHeader(http.StatusOK)
+	}
+}
+
```
thisthat, line 199: These lines are used in multiple places. Would it make sense to move it in a common package, such as `go-utils`?

thisthat (commented):
Sonar cloud complains of the following string not being a constant: https://sonarcloud.io/project/issues?id=keptn_keptn&open=AXxKkWofNVUo1Qq9zweA&pullRequest=5459&resolved=false&types=CODE_SMELL

Could you please fix this? :)

helm-service/main.go
RealAnna: back then I was not 100% sure if I will end up differentiating between health/ready and health/lives but since now they are the same, sure I can add it to that one

PR #5473, opened 2021-10-04
feat(bridge): Show payload of last event in subscription configuration

### LINKED ISSUE #5265: Bridge: Show payload of the last event for webhook config
## User story
As I user, I want to show the payload for the last event I selected for configuration, to then be able to configure the custom payload with the properties I need.

## Mockup
Add Subscription2

### Acceptance criteria
* Show a button "Show example payload" below the subscription configuration
* The button is enabled when the "Acts on" and "When" is already selected
* The button is disabled when "Acts on" and "When" are not selected yet. Show a hint that the fields have to be selected first.
* Clicking on this button opens an overlay with a backdrop
* The overlay shows an example payload for the selected event
* The overlay shows a mashup of payloads when the wildcard * is selected for an event ❓ Needs clarification
* Clicking on the backdrop or on "Close" closes the backdrop

ermin-muratovic:
## This PR
- allows the user to view the payload of the last event during the subscription configuration

### Related Issues
Fixes #5265 

image


ermin-muratovic:
> * If there is not any filter applied, the stage and service information (`This is the latest sh.keptn.event.... from stage X for service Y`) for the payload viewer should be taken from the received event.

The message should represent the applied filter, not the response, as it is described in the issue.

Issue #5490, opened 2021-10-04
Use Viper to manage `CLIConfig`


vadasambar:
**Environment**

* Client OS (e.g., Linux, macOS, Windows): any
* Keptn Version (`keptn version`): any
* Kubernetes Cloud Provider (e.g., GKE, AKS): any
* Kubernetes version (`kubectl version`): any

**Affected Component**

* [ ] Docs
* [x] CLI
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [ ] Helm
* [ ] Istio
* [ ] Bridge
* [ ] Approval
* [ ] Datastore
* [ ] REST API
* [ ] eventbroker / distributor
* [ ] jmeter

**Describe the problem**

- Right now, we have our own code to read in `CLIConfig` [here](https://github.com/keptn/keptn/blob/4a1be7531523273ff4cbc76d0c7d07b124d57f6a/cli/pkg/config/cli_config.go#L40-L55).
- It would be good if we could replace it with `Viper`. We had `Viper` in `root.go` but I removed it in [this](https://github.com/keptn/keptn/pull/5290) PR. This is because there was code around `Viper` but it wasn't actually used for anything.
i.e., `cfgFile` [here](https://github.com/keptn/keptn/pull/5290/files#diff-eb154bbd601602eeae895e7af6a12b88cbdb7d585a2d32216c398f68cabcdff5L84) is actually empty because we aren't reading it [anywhere](https://github.com/keptn/keptn/pull/5290/files#diff-eb154bbd601602eeae895e7af6a12b88cbdb7d585a2d32216c398f68cabcdff5L60-R66) and the [default `cfgFile` path is incorrect](https://github.com/keptn/keptn/pull/5290/files#diff-eb154bbd601602eeae895e7af6a12b88cbdb7d585a2d32216c398f68cabcdff5L89-L97) (should be `~/home/username/.keptn/config` like in [this](https://github.com/keptn/keptn/blob/4a1be7531523273ff4cbc76d0c7d07b124d57f`6a/cli/pkg/config/cli_config.go#L31-L36) function)

**To Do**

- [ ] Replace the functions to manage `CLIConfig` with Viper
- [ ] Fix `cfgFile` path (expose it as a flag)

PR #5694, opened 2021-10-13
refactor(cli):  Use Viper to manage Keptn config

vadasambar:
Fixes https://github.com/keptn/keptn/issues/5490

This PR uses Viper internally to manage the Keptn Config. 
## Why
- [Viper](https://github.com/spf13/viper) allows reading config fields from the environment variables
- Viper allows aliasing fields
- Changing file type of the config is relatively easier
- Viper decouples the CLI from the internal representation of the Config

### How to test

- You can do `keptn <any-command> --cli-config ~/.foo/my-keptn-config` to use a config stored at path other than `~/.keptn/config`

### Things to note
- Please check `go.sum`. Adding viper 1.9.0 lead to many changes in `go.sum` (additive. I have not removed any dependency)


cli/cmd/root.go
```diff
@@ -62,6 +62,8 @@ func init() {
 	rootCmd.PersistentFlags().BoolVarP(&mocking, "mock", "", false, "Disables communication to a Keptn endpoint")
 	rootCmd.PersistentFlags().StringVarP(&namespace, "namespace", "n", "keptn",
 		"Specify the namespace where Keptn should be installed, used and uninstalled in")
+	rootCmd.PersistentFlags().StringVarP(&cfgFile, "config-file", "", "",
```
vadasambar, line 65: I tried `-c` and `-f` shorthands but both are being used in the subcommands.

cli/cmd/set_config.go
```diff
@@ -32,7 +32,7 @@ Supported keys:
 			cmd.SilenceUsage = false
 			return errors.New("required arguments KEY and VALUE")
 		}
-		configMng = config.NewCLIConfigManager()
+		configMng = config.NewCLIConfigManager("")
```
vadasambar, line 35: Function signature `NewCLIConfigManager` has changed to support custom config file path. I have done a global replace on this function `NewCLIConfigManager()` -> `NewCLIConfigManager("")`

cli/go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/mitchellh/mapstructure v1.4.2
 	github.com/spf13/cast v1.4.1 // indirect
 	github.com/spf13/cobra v1.2.1
+	github.com/spf13/viper v1.8.1
```
vadasambar, line 23: Latest version of Viper at the time of this PR
image
https://pkg.go.dev/github.com/spf13/viper?tab=versions

cli/go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/mitchellh/mapstructure v1.4.2
 	github.com/spf13/cast v1.4.1 // indirect
 	github.com/spf13/cobra v1.2.1
+	github.com/spf13/viper v1.9.0
```
vadasambar, line 23: Latest version of Viper at the time of this PR
image
https://pkg.go.dev/github.com/spf13/viper?tab=versions

cli/go.sum
```diff
@@ -34,6 +34,7 @@ cloud.google.com/go/bigquery v1.8.0/go.mod h1:J5hqkt3O0uAFnINi6JXValWIb1v0goeZM7
 cloud.google.com/go/datastore v1.0.0/go.mod h1:LXYbyblFSglQ5pkeyhO+Qmw7ukd3C+pD7TKLgZqpHYE=
 cloud.google.com/go/datastore v1.1.0/go.mod h1:umbIZjpQpHh4hmRpGhH4tLFup+FVzqBi1b3c64qFpCk=
 cloud.google.com/go/firestore v1.1.0/go.mod h1:ulACoGHTpvq5r8rxGJ4ddJZBZqakUQqClKRT5SZwBmk=
+cloud.google.com/go/firestore v1.6.0/go.mod h1:afJwI0vaXwAG54kI7A//lP/lSPDkQORQuMkv56TxEPU=
```
vadasambar, line 37: All the changes in this file are because of two reasons:
1. I added viper as a dependency (CI was failing) with
```sh
# In the keptn root directory
$ go get cli/pkg/config
```
2. I updated the dependency from viper 1.8.1 -> 1.9.0 in the `go.mod` because 1.9.0 is the latest stable version at the time of this comment.  1.8.1 was added automatically on running the `go get` command above. After updating the `go.mod` file I ran the same command again
```sh
# In the keptn root directory
$ go get cli/pkg/config
```

cli/pkg/config/cli_config.go
```diff
@@ -1,39 +1,57 @@
...
+	CurrentContext        string     `json:"current-context" mapstructure:"current-context"`
 }
 
 // CLIConfigManager manages the path of the CLI config
 type CLIConfigManager struct {
 	CLIConfigPath string
+	viper         *viper.Viper
 }
 
+var mgr *CLIConfigManager
```
vadasambar, line 31: Maintain an internal `CLIConfigManager` struct to avoid re-initializing new `viper` instances.

cli/pkg/config/cli_config.go
```diff
@@ -43,11 +61,28 @@ func (c *CLIConfigManager) LoadCLIConfig() (CLIConfig, error) {
...
-	if err != nil {
+	// Allow setting file path after CLIConfigManager is initialized
+	c.viper.SetConfigFile(c.CLIConfigPath)
+	if err := c.viper.ReadInConfig(); err != nil {
 		return cliConfig, fmt.Errorf("error when reading config file: %w", err)
 	}
-	if err := json.Unmarshal(data, &cliConfig); err != nil {
+
+	if err := c.viper.Unmarshal(&cliConfig, func(dConfig *mapstructure.DecoderConfig) {
+		dConfig.DecodeHook = mapstructure.StringToTimeHookFunc(time.RFC3339)
```
vadasambar, line 71: [`StringToTimeHookFunc`](https://pkg.go.dev/github.com/mitchellh/mapstructure#StringToTimeHookFunc) decode hook was added because `viper.Unmarshal` does not know how to unmarshal json time string into `*time.Time` (it throws an error if we don't add the decode hook)

cli/pkg/config/cli_config.go
```diff
@@ -43,11 +61,28 @@ func (c *CLIConfigManager) LoadCLIConfig() (CLIConfig, error) {
...
+	}); err != nil {
+		return cliConfig, fmt.Errorf("error when unmarshalling config file: %w", err)
+	}
+
+	return cliConfig, nil
+}
+
+// GetCLIConfig gets the already loaded configuration
+func (c *CLIConfigManager) GetCLIConfig() (CLIConfig, error) {
```
vadasambar, line 80: Many places in the code use `LoadCLIConfig` which reads in the config file every time `LoadCLIConfig` function is called. `GetCLIConfig` is lighter version of the function. It only returns config already loaded in the Viper.

cli/pkg/config/cli_config.go
```diff
@@ -56,12 +91,40 @@ func (c *CLIConfigManager) LoadCLIConfig() (CLIConfig, error) {
...
+	// This mismatch causes Viper.MergeConfigMap() to throw an error
+	// Note 2: We are using json.Marshal because we have set viper config type to json
+	// Please change this if we are using a different config type
 	data, err := json.Marshal(config)
 	if err != nil {
 		return fmt.Errorf("error when marshalling config file: %w", err)
 	}
-	if err := ioutil.WriteFile(c.CLIConfigPath, []byte(data), 0644); err != nil {
+
+	if err := c.viper.MergeConfig(bytes.NewReader(data)); err != nil {
```
vadasambar, line 110: `config` passed as parameter overwrites config stored in the Viper

cli/pkg/config/cli_config_test.go
```diff
@@ -60,6 +64,12 @@ func TestStoreCLIConfig(t *testing.T) {
 	}
 
 	data, err := fileutils.ReadFileAsStr(mng.CLIConfigPath)
+
+	// This is to remove the indentation viper adds
+	data = strings.ReplaceAll(data, " ", "")
```
vadasambar, line 69: Viper stores the config like this
```json
{
  "automatic_version_check":true,
  "current-context":"",
  "kube_context_check":true,
  "last_version_check":"2020-02-20T00:00:00Z"
}
```
But we want to flatten it out (remove indentations) so that we can compare it with the `testConfig`

cli/pkg/config/cli_config.go
```diff
@@ -1,39 +1,57 @@
...
+	AutomaticVersionCheck bool       `json:"automatic_version_check" mapstructure:"automatic_version_check"`
+	KubeContextCheck      bool       `json:"kube_context_check" mapstructure:"kube_context_check"`
+	LastVersionCheck      *time.Time `json:"last_version_check" mapstructure:"last_version_check"`
+	CurrentContext        string     `json:"current-context" mapstructure:"current-context"`
 }
 
 // CLIConfigManager manages the path of the CLI config
 type CLIConfigManager struct {
 	CLIConfigPath string
+	viper         *viper.Viper
```
vadasambar, line 28: I have kept the field private for now. We can make this public in the future if required.

cli/pkg/config/cli_config.go
```diff
@@ -1,39 +1,57 @@
...
+		var err error
+		cfgFile, err = GetKeptnDefaultConfigPath()
+		if err != nil {
+			log.Fatal("could not get default config path", err)
+		}
 
-	dir, err := keptnutils.GetKeptnDirectory()
-	if err != nil {
-		log.Fatal(err)
+		fmt.Printf("no value for --config-file was specified\n defaulting to: %s\n", cfgFile)
```
bacherfl, line 45: please use `logging.PrintLog()` with `VerboseLevel` instead of `fmt.Printf`

vadasambar: I will change this. Thanks for the review :pray:

bacherfl (approved):
lgtm

PR #5824, opened 2021-10-28
chore: Cancel integration tests when mismatch between CLI and kube context is detected (#5743)

### LINKED ISSUE #5743: go-tests: CreateProject has two different ways of accessing the Keptn installation, might break clusters without knowing
I just ran into a weird problem when executing integration tests locally:

`CreateProject("podtato-head", true)` should check if the project exists, and if it does, delete it. Then create the new one.

However, this didn't work.

The reason behind this is simple: 
I configured `KEPTN_ENDPOINT` to point to Keptn Installation A, while my Keptn CLI is currently pointed to Installation B.

In the source code you can see that we are accessing the API directly for checking whether the project exists (`ApiGETRequest` which uses `KEPTN_ENDPOINT` internally), while we then delete and create a project using the CLI:

https://github.com/keptn/keptn/blob/6a83829e3c54fb5a52d0b8405f1936337d454801/test/go-tests/test_utils.go#L48-L82

# Current behaviour

This will potentially create/delete projects on the wrong cluster (not linked to `KEPTN_ENDPOINT`)

# Desired behaviour

`KEPTN_ENDPOINT` should be the cluster I am running my tests against, this includes the `keptn` CLI during those integration tests.

# Notes

1. We might need to check if there are other occurances of using the `keptn` CLI within the go-tests.
2. We could try to avoid using the `keptn` CLI within go-tests. On the other hand, I think an integration test should also cover the CLI.

bacherfl:
Closes #5743

### DIFF
--- a/test/go-tests/init_test.go
+++ b/test/go-tests/init_test.go
@@ -1,0 +1,54 @@
+package go_tests
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestMain(m *testing.M) {
+	if err := setup(); err != nil {
+		os.Exit(-1)
+	}
+	code := m.Run()
+	os.Exit(code)
+}
+
+func setup() error {
+	match, err := endpointsMatch()
+	if err != nil {
+		return fmt.Errorf("could not compare endpoints of kubectl context and keptn CLI: %s", err.Error())
+	}
+	if !match {
+		return errors.New("endpoint mismatch between CLI and kubectl detected")
+	}
+	return nil
+}
+
+func endpointsMatch() (bool, error) {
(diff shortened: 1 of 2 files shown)


test/go-tests/init_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+func TestMain(m *testing.M) {
+	if err := setup(); err != nil {
+		os.Exit(-1)
+	}
+	code := m.Run()
+	os.Exit(code)
+}
+
+func setup() error {
```
warber, line 19: please write comments why we are doing this

warber (changes requested):

warber (approved):

PR #5888, opened 2021-11-02
fix(cli): Make sure the release version is set in command descriptions (#5762)

### LINKED ISSUE #5762: Keptn CLI: `keptn generate docs` creates docu link without release version
### Bug

When executing the command `keptn generate docs` the docu for the CLI commands is generated which is used here: https://keptn.sh/docs/0.9.x/reference/cli/commands/

Some of the commands reference content from the keptn.sh page; e.g.: https://keptn.sh/docs/0.9.x/reference/cli/commands/keptn_delete_project/ with a link to: *Clean-up after deleting a project*

This link is broken since the release version is missing in the path: `https://keptn.sh/docs//continuous_delivery/deployment_helm/#clean-up-after-deleting-a-project` 

The variable which is used for generating the proper link is set here: https://github.com/keptn/keptn/blob/master/cli/cmd/version.go#L103

### Expected behavior: 
- When running the command `keptn generate docs` the generated docs contain links that work for that release: https://keptn.sh/docs/0.9.x/reference/cli/commands/

### Definition of done: 
- [ ] Please update keptn.sh for 0.8.x, 0.9.x, and 0.10.x

bacherfl:
Closes #5762

The reason why the keptnReleaseDocsURL was not set properly was the order of the initialization within the cli project:

The commands, including their description including the `keptnReleaseNotesURL` are located in the `cmd` subpackage and are therefore initialized before the `main` package, where the `Version` property was set using the ldflags.

### DIFF
--- a/cli/cmd/auth.go
+++ b/cli/cmd/auth.go
@@ -122,7 +122,7 @@ func verifyAuthParams(authParams *authCmdParams) error {
 
 		err = smartFetchKeptnAuthParameters()
 		if err != nil {
-			return fmt.Errorf(err.Error()+parametersRequiredMessage, keptnReleaseDocsURL, namespace, namespace)
+			return fmt.Errorf(err.Error()+parametersRequiredMessage, getReleaseDocsURL(), namespace, namespace)
 		}
 	}
 	return nil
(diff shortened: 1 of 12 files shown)


RealAnna (approved):
LGTM

PR #5915, opened 2021-11-05
refactor(bridge): Refactoring of service screen (#4918) (#5244)

### LINKED ISSUE #4918: Refactoring of service screen
At the moment there is a lot of data-manipulating and there are a lot of API-Calls for deployments. This should be moved to the bridge server.

- Introduce a new endpoint `GET service`. This endpoint returns all services and all deployments. Deployments should contain at least `name/version`, `stages` and `hasOpenRemediations` for each stage.
- Introduce a new endpoint `GET service/{serviceName}/deployments/{keptnContext}`. This endpoint returns all deployment information: `sequence`, `openRemediations`, `approvalTrace` (open ones) and `evaluationTraces`. Return 404 if `keptnContext` is not found.

# Definition of Done
- [x] The bridge client uses the new endpoint
- [x] Unit/UI tests

### LINKED ISSUE #5244: Bridge: Move data aggregation from service screen to bridge-server
## Action Item

Move data aggregation logic from the service screen into the bridge-server. 

### Technical details

Think to add a caching layer to improve response time and reduce the number of calls to the backend. 
Keep in mind that in the future, we could run a single bridge-server for multiple Keptn installations. Hence we should prevent cross-tenant visibility of events.

## DoD

- [ ] Bridge-server performs the aggregation of data
- [ ] UI only shows the data

Kirdock:
Fixes #4918 
Fixes #5244 

~~Note: `fromTime` can't be implemented for the service states endpoint, because of `hasRemediations` for each stage. The previous and current value should be checked then. Always returning all service states is less effort (fewer API calls).~~

Note2: A `fromTime` parameter for the service states endpoint can actually be implemented without any additional calls. Just need to check if there are any `remediation.*` events in `service.latestEvents` with a timestamp higher than `fromTime`.
Needs discussion if we should implement this (including an update algorithm for service states, which should already exist in the commit history)


bridge/client/app/_components/ktb-sequence-list/ktb-sequence-list.component.html
```diff
@@ -19,21 +19,24 @@ <h2>Sequences</h2>
...
           <dt-icon
-            *ngIf="!trace.isLoading() || trace.isApproval() || trace.getLastTrace().isApproval(); else loading"
+            *ngIf="subSequence.state === 'finished' || subSequence.hasPendingApproval; else loading"
             class="event-icon"
-            [class.error]="trace.isFaulty()"
-            [class.success]="trace.isSuccessful()"
-            [class.highlight]="trace.isApproval() || trace.getLastTrace().isApproval()"
-            [name]="trace.isFinished() ? trace.getIcon() : trace.getLastTrace().getIcon()"
+            [class.error]="subSequence.result === 'fail'"
+            [class.success]="subSequence.result === 'pass'"
```
ermin-muratovic, line 27: I would introduce methods here, `isFaulty()` and `isSuccessful()` or similar, because it reduces necessary changes if the model changes and also use the `ResultTypes` enum.

bridge/client/app/_components/ktb-sequence-list/ktb-sequence-list.component.html
```diff
@@ -19,21 +19,24 @@ <h2>Sequences</h2>
         </div>
       </ng-container>
       <ng-template #trace>
-        <ng-container *ngIf="isTrace(row) as trace">
+        <ng-container *ngIf="isSubsequence(row) as subSequence">
           <dt-icon
-            *ngIf="!trace.isLoading() || trace.isApproval() || trace.getLastTrace().isApproval(); else loading"
+            *ngIf="subSequence.state === 'finished' || subSequence.hasPendingApproval; else loading"
```
ermin-muratovic, line 24: I would introduce a method `isFinished()` here. It is easier to read and also reduces necessary changes if the model or the result changes. And we have a `SequenceState` enum that can be used.

laneli (changes requested):

ermin-muratovic (changes requested):

ermin-muratovic:
Before the refactoring we had a info "Last processed artifact:"
image

This was removed during the refactoring:
image

Was this on purpose?

ermin-muratovic:
When a sequence in a stage was still running the stage was highlighted green when selected, thus reading the stage name was quite okay:
image

Now the background of the stage stays gray, while the label is white, which makes reading the label difficult because of little contrast:
image

Can we fix this?

bridge/client/app/_components/ktb-sequence-list/ktb-sequence-list.component.html
Kirdock: Because we just have one single property here, I would just use equals. Nevertheless I'll use the enum here

ermin-muratovic:
While staying on the service screen, after the polling worked (which seems to not update the screen immediately), the newer service is added at the bottom.

Before:
image

After refactoring:
image

After refreshing the refactored page:
image

I would expect to have always the same sorting, but not sure which is more correct, the latest on top or the latest on bottom.

Kirdock:
Regarding the latest processed artifact: Is the environment variable `CONTINUOUS_DELIVERY` set? Else the artifact will not show up because it's set to quality gates only environment.

Kirdock:
@ermin-muratovic 
- order of services is now by name, also on update
- fixed selection color of stages that are currently running

ermin-muratovic (approved):
lgtm

ermin-muratovic (approved):
lgtm

laneli (approved):

PR #5918, opened 2021-11-05
refactor(bridge): Refactoring of services settings (#5100)

### LINKED ISSUE #5100: Bridge: Update settings screen data polling
## Action Item

The polling result of the project is not used in every view and some views only require one information of the project instead of the whole one.

### Implementation details

The settings screen does not need any polling and only Git-related data.

--- 

### DoD
- [ ] Unit tests
- [ ] Introduce a new endpoint in the bridge-server that returns only the information needed.

laneli:
Closes #5100


bridge/client/app/_components/ktb-service-settings-list/ktb-service-settings-list.component.ts
```diff
@@ -1,36 +1,46 @@
...
+        filter((projectName): projectName is string => !!projectName)
       )
-      .subscribe((project) => {
-        const services: string[] = project?.getServices()?.map((service) => service.serviceName) ?? [];
-        this.dataSource = new DtTableDataSource<string>(services);
+      .subscribe((projectName) => {
+        this.projectName = projectName;
+        this.isLoading = true;
+
+        this._timer = AppUtils.createTimer(0, this.initialDelayMillis)
```
Kirdock, line 34: If the project change would not redirect to `/settings/project`, we would have multiple polling intervals here. Nevertheless, this should be changed that this does not happen on param change.

bridge/server/services/data-service.ts
```diff
@@ -411,6 +411,20 @@ export class DataService {
     return tasks;
   }
 
+  public async getServiceNames(projectName: string): Promise<string[]> {
+    const resp = await this.apiService.getStages(projectName);
+    const stages = resp.data.stages;
+    const services: Map<string, string> = new Map();
```
Kirdock, line 417: Because we don't need the value here, only the keys, I would use a Set. But if the insertion order is not used, I would prefer a plain object for unique values, because Map and Set are much slower.

Kirdock (changes requested):

bridge/server/api/index.ts
```diff
@@ -114,6 +114,15 @@ const apiRouter = (params: {
     }
   });
 
+  router.get('/project/:projectName/services', async (req: Request, res: Response, next: NextFunction) => {
+    try {
+      const services = await dataService.getServiceNames(req.params.projectName);
```
Kirdock, line 119: Could you add some unit tests for it?

Kirdock (changes requested):

Kirdock (changes requested):

Kirdock (approved):

Issue #5243, opened 2021-09-15
Core: Refactor the Shipyard controller


thisthat:
## Action Item

The shipyard controller is complex to understand and maintain. The code should be refactored moving logical code units into their own components.

### Technical details
Keep the sequence aggregation in the Shipyard controller

## DoD

- [ ] Move towards a stateless shipyard controller
- [ ] Have a formal definition of the algorithm it implements
- [ ] Document the data-flow of events with MongoDB (how they are moved between collections)
- [ ] Reduce the code complexity (see [Cyclomatic](https://sonarcloud.io/component_measures?id=keptn_keptn&metric=complexity&selected=keptn_keptn%3Ashipyard-controller) and [Cognitive](https://sonarcloud.io/component_measures?id=keptn_keptn&metric=cognitive_complexity&selected=keptn_keptn%3Ashipyard-controller) complexity on SonarCloud - 1500 and 1368 respectively)

PR #5937, opened 2021-11-08
chore(shipyard-controller): cleanups & refactorings 2

warber:
part of #5243
integration test run: https://github.com/keptn/keptn/actions/runs/1436224211


shipyard-controller/db/repos.go
```diff
@@ -29,10 +29,10 @@ type UniformRepo interface {
 
 //go:generate moq --skip-ensure -pkg db_mock -out ./mock/tasksequencerepo_mock.go . TaskSequenceRepo
 type TaskSequenceRepo interface {
```
bacherfl, line 31: let's rename the interface to `TaskExecutionRepo` as well

bacherfl (approved):

Issue #4522, opened 2021-06-28
Graceful Shutdown of Keptn components


warber:
## Action Item
Services shall be able to shut down gracefully.

### Details
When shutting down a service (a.k.a., receiving a `SIGTERM`) a keptn service shall stop accepting/processing new messages.
In general, each service shall handle the termination signal *gracefully* by cleaning up its resources and finishing its current activity, so that it can gracefully shut down without being killed by K8S.

## Acceptance Criteria
- [ ] Keptn services do not process any messages/requests once they are scheduled for termination
- [ ] Keptn services are able to finish long-running tasks, such as executing a load test (jmeter-service) before they get terminated.

## Definition of Done
- [ ] feature implemented and reviewed
- [ ] unit tests
- [ ] manually tested

warber:
### Proposal

#### Zero Downtime
During shutdown/deletion of a pod this is what happens:
1) pod is removed from the endpoints of all services
2) preStop Hook is executed
3) `SIGTERM` signal is sent to containers in pod.
4) termination grace period is started (this defines the time in seconds K8S waits before killing the pod) (30 per default)
5) `SIGKILL` is sent after the termination grace period is over.

Note, that 1-4 is performed **simultaneously**. Hence, it is possible that the pod has not been removed from the endpoints after it received the `SIGTERM` signal which will result in a partial downtime of the service, since the pod already stopped working at this time.

The solution for that is to configure a `preStop` hook for the pod, so that the pod sleeps for some time after it receives `SIGTERM` (see `keptn-nats-cluster` for an example).


#### Long-running tasks
The majority of Keptn Integrations are task executors that are triggered by a `*.triggered*` event and ar then executing their task. This could be something that is immediately done like sending out a notification event, or it could be something that is potentially long-running like executing some tests against a system (e.g. `jmeter-service`)

Hence, the proposal is to make use of k8s `preStop` lifecycle hook. This is called immediately before the container is terminated. The lifecycle of the pod is *blocked* until the `preStop` handler completes or the pod's grace period expires. By introducing an endpoint in the service it can be (repeatedly) checked whether the service is currently active executing a (long-running) task or not. If not the script is exited, and the container is shutting down. Note that the `terminationGracePeriodSeconds` configuration parameter needs to be set to an appropriate (high) value.

johannes-b:
To allow estimating this issue, a breakdown in its component it necessary.

PR #5973, opened 2021-11-10
feat: graceful shutdown for jmeter and helm service (#4522)

RealAnna:
## This PR

- adds sigterm handling for jmeter and helm service
- helm service pod stays alive until it finishes
- jmeter service shutsdown immediatly after sending a finished event

### Related Issues

#4522 

### How to test
Kill Jmeter or helm pod while they are running. 
 
image


bacherfl (approved):

helm-service/controller/deployment_handler.go
```diff
@@ -36,8 +38,8 @@ func NewDeploymentHandler(keptnHandler Handler, mesh mesh.Mesh, onboarder Onboar
 
 // HandleEvent handles deployment.triggered events by first changing the new configuration and
 // afterwards applying the configuration in the cluster
-func (h *DeploymentHandler) HandleEvent(ce cloudevents.Event) {
-
+func (h *DeploymentHandler) HandleEvent(ctx context.Context, ce cloudevents.Event) {
+	defer ctx.Value(GracefulShutdownKey).(*sync.WaitGroup).Done()
```
bacherfl, line 42: let's add a `nil` check for `ctx.Value()` and for the cast to `*sync.WaitGroup`, i.e.

```
val := ctx.Value(GracefulShutdownKey)
if val == nil {
	return 
}
if wg, ok := val.(*sync.WaitGroup); ok {
	wg.Done()
}
```

bacherfl (approved):

bacherfl (changes requested):
the helm service unit tests need to be fixed, afterwards we can merge

jmeter-service/testrunner.go
```diff
@@ -29,63 +36,99 @@ func NewTestRunner(eventSender *keptnv2.HTTPEventSender) *TestRunner {
...
 
 	if testStrategy == "" {
 		logger.Info("No testStrategy specified therefore skipping further test execution and sending back success")
-		return true, nil
 	}
 
 	testStrategyWorkload, err := getWorkloadForStrategy(jmeterConf, testStrategy)
 	if err != nil {
-		return false, err
+		logger.Error(err.Error())
```
bacherfl, line 120: please add a message in front of the error. Also, was the removal of the `return` at this point intended (see also the following lines where previously an error was returned)?

RealAnna: yes since I propagate the error in the res channel line 131

bacherfl (approved):

PR #5986, opened 2021-11-11
fix(bridge): Exclude webhook-service results from sequence result (#5954)

### LINKED ISSUE #5954: Bridge: Failed Webhook call on a `.finished` event results in the Bridge as failed task
**Environment**

* Client OS (e.g., Linux, macOS, Windows):
* Keptn Version (`keptn version`): 0.10.0
* Kubernetes Cloud Provider (e.g., GKE, AKS):
* Kubernetes version (`kubectl version`):

**Affected Component**

* [ ] Approval
* [x] Bridge
* [ ] CLI
* [ ] Datastore
* [ ] Docs
* [ ] Helm
* [ ] Istio
* [ ] Jmeter
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [ ] REST API
* [x] Webhook

**Describe the bug**

When I subscribe the webhook-service to a `task.finished` and the Webhook call fails, the Bridge displays a failed task even though the task itself was successful: 

image

image

**To Reproduce**

1. Register a webhook to a `deployment.finished` 
2. Modify the Webhook payload to contain a special character, e.g.: `|`
3. Trigger delivery

**Expected behavior**

- When subscribing to a `started` `finished` event, a failed Webhook call should not determine the coloring of the task in the Bridge:

image

laneli:
Fixes #5954

By excluding the source webhook-service from isFaulty, isSuccess, is Warning, traces should now show the proper state.

### DIFF
--- a/bridge/client/app/_models/trace.mock.ts
+++ b/bridge/client/app/_models/trace.mock.ts
@@ -3020,6 +3020,186 @@ const multipleEvaluationsTraces: any = Trace.fromJSON({
   finished: false,
 });
 
+const webhookTrace = Trace.fromJSON({
+  traces: [
+    Trace.fromJSON({
+      traces: [],
+      data: {
+        project: 'sockshop',
+        service: 'carts',
+        stage: 'dev',
+        status: 'succeeded',
+      },
+      id: 'cdd49036-e590-4da6-876e-dcdc75bf8850',
+      source: 'helm-service',
+      specversion: '1.0',
+      time: '2021-11-11T13:05:29.701Z',
+      type: 'sh.keptn.event.deployment.started',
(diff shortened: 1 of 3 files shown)


christian-kreuzberger-dtx:
Does this also remove webhook-service `.started` and `.finished` events in case of listening to `.triggered` events?

laneli:
@christian-kreuzberger-dtx as it filters for the source, the event type should not matter.

laneli:
@christian-kreuzberger-dtx When subscribing to started / finished events the webhook shouldn't send an event at all. For triggered the behavior is as expected and it is allowed to fail the sequence.
As why the events are still sent, someone from core team is taking a look at it.

PR #5988, opened 2021-11-11
fix(bridge): Fixed wrong weight of SLI (#5987)

### LINKED ISSUE #5987: Wrong SLI weight if weight is not provided in SLO file
The bridge shows the wrong weight if the weight is not specified in the SLO file.

## Expected behavior
If the weight property is not provided there should be the default value `1`.

Kirdock:
Fixes #5987 


ermin-muratovic (approved):
lgtm

PR #6019, opened 2021-11-16
fix(bridge): Remove inline script for base url and upgrade-insecure-requests header part

laneli:
I removed the script that caused the page to load entirely and replaced it with the `--base-url` argument for the `yarn build` script. Also I removed the `upgrade-insecure-requests` header part of the CSP header in order to not cause switching requests from http to https - which causes `ERR_CONNECTION_REFUSED` errors on the client.

### DIFF
--- a/bridge/client/index.html
+++ b/bridge/client/index.html
@@ -4,16 +4,6 @@
     <meta charset="utf-8" />
     <title>keptn</title>
     <base id="base-href" href="/" />
-    <script>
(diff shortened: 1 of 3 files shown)


Kirdock (changes requested):
If I deploy the docker image I receive a lot of errors when I try to refresh the page
image

Kirdock (approved):

PR #6020, opened 2021-11-16
feat: Update pod config to be more strict w.r.t. security

thisthat:
Integration test run: https://github.com/keptn/keptn/actions/runs/1738995423

### DIFF
--- a/installer/manifests/keptn/charts/control-plane/templates/core.yaml
+++ b/installer/manifests/keptn/charts/control-plane/templates/core.yaml
@@ -77,6 +77,8 @@ spec:
     spec:
       securityContext:
         fsGroup: 65532
+        seccompProfile:
+          type: RuntimeDefault
       containers:
         - name: api-service
           image: {{ .Values.apiService.image.repository }}:{{ .Values.apiService.image.tag | default .Chart.AppVersion }}
@@ -205,6 +207,8 @@ spec:
     spec:
       securityContext:
         fsGroup: 65532
+        seccompProfile:
+          type: RuntimeDefault
       containers:
         - name: bridge
           image: {{ .Values.bridge.image.repository }}:{{ .Values.bridge.image.tag | default .Chart.AppVersion }}
@@ -330,6 +334,8 @@ spec:
     spec:
       securityContext:
         fsGroup: 65532
+        seccompProfile:
+          type: RuntimeDefault
       serviceAccountName: keptn-shipyard-controller
       containers:
         - name: shipyard-controller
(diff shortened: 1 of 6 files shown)


installer/manifests/keptn/charts/control-plane/templates/core.yaml
```diff
@@ -643,6 +671,12 @@ spec:
           volumeMounts:
             - mountPath: /data/config
               name: resource-volume
+          securityContext:
+            runAsNonRoot: true
+            runAsUser: {{ .Values.configurationService.fsGroup | default 1001 }}
```
odubajDT, line 676: I do not think this is correct, we are in resource-service and using values from configuration-service. Also, I am not sure, if we need a security context for the resource-service, as we are using `emptyDir` as a volume instead of `PersistentVolumeClaim` as when using configuration-service. @bacherfl can you maybe support my thoughts ?

installer/manifests/keptn/charts/control-plane/templates/core.yaml
```diff
@@ -604,6 +626,12 @@ spec:
         app.kubernetes.io/version: {{ .Values.resourceService.image.tag | default .Chart.AppVersion }}
         helm.sh/chart: {{ include "control-plane.chart" . }}
     spec:
+      securityContext:
+        fsGroup: {{ .Values.configurationService.fsGroup | default 1001 }}
```
odubajDT, line 630: Also, values from configuration-service should not be used in here as well

installer/manifests/keptn/charts/control-plane/templates/core.yaml
mowies: cross-usage of values is probably not the best idea, but having a security context never hurts IMHO :)

odubajDT: I am ok with that, but when adding it, we should add it also here https://github.com/keptn/keptn/blob/master/resource-service/deploy/service.yaml to keep the code consistent :)

thisthat: I am using cross-usage since we want to make sure the switch from configuration- to resource-service works with only the single FF. In the future, when we do the transition and mark the resource-service as production ready and the default, these properties will be renamed together with the other existing ones.

odubajDT: Thanks for the explanation!

odubajDT (approved):

mowies (approved):

Issue #5450, opened 2021-10-01
Data retrieval for environment screen does not scale well


Kirdock:
WIP: add the bridge ticket

At the moment there could be timeouts if there are a lot of services inside a project. The data retrieval does not scale well.
The APIs should be adjusted (new ones or refactor existing ones).

The current information needed and fetched is listed below:

- GET Project (`controlPlane/v1/project/projectName`). To get an overview of which services (including the latest events of them) we have
- GET open remediations (`controlPlane/v1/sequences/projectName `with name is "remediation"). To show the number of  remediations
- for each remediation
  - GET all events of it (`/mongodb-datastore/event`). In order to aggregate all performed actions and show them on the right in the service details		
- for each stage
  - for each service
    - get the latest sequence (`controlPlane/v1/sequences/projectName`). The latest sequence (name and evaluation result) should be shown in the service tile
    - get the latest deployment information (`/mongodb-datastore/event/type/${eventType}`). To get the latet deployment URL. There is a link in the service details on the right, next to the service name
get open approvals (`/mongodb-datastore/event/type/${eventType}`). On the right side (service details), we can approve/decline approvals
    - get latest evaluation (`/mongodb-datastore/event`). In the approval tile, we show the evaluation result (score and status) next to "accept", "decline"
	
API calls:		
1x project
1x remediations + 1 per open remediations
X stages * Y services * (3 + Z open approvals for this service)

For example, we have 3 stages, 30 services, and no open approvals or remediations:
`1 + 1 + 3x30x3 = 272 Calls`

Note that, the new endpoint(s) will be called regularly. We should look for an API that can deal with a lot of data and DB calls (potentially). Something like diff changes might be used to reduce the load. 

## DoD

- [ ] Propose a new (or multiple) API endpoint

PR #6040, opened 2021-11-17
refactor(bridge): Reduce number of API calls for project endpoint (#5450)

Kirdock:
Part of #5450 

## This PR reduces the number of API calls on the bridge server for the project endpoint

- open approvals are now fetched at the beginning instead for each service
- evaluations (for a sequence) are now fetched at the beginning instead of multiple calls in a loop
- the deployment URL for a service is now fetched at the beginning instead of multiple calls in a loop

**Number** of API calls now:
1x project
1x remediations + 1 per open remediations
1x open approvals
1x sequence evaluations
1x deployment URLs
X stages x Y services x 1 (latest sequence of a service in a stage)

For example, we have 3 stages, 30 services, and no open approvals or remediations:
`5 + 3x30x1 = 95 Calls`. Compared to 272 calls (before) we have 182 calls less (-65% calls).

**Speedup locally (7 services):**
before: ~8.74 seconds
after: 3.49 seconds
It's about 2.5 times faster than the old implementation


bridge/server/services/data-service.ts
```diff
@@ -53,87 +60,218 @@ export class DataService {
...
+          if (latestSequence) {
+            cachedSequences[latestEvent.keptnContext] = latestSequence;
           }
         }
       }
     }
     return project;
   }
 
+  private async getSequencesForServices(projectName: string, services: Service[]): Promise<Sequence[]> {
```
laneli, line 123: Rename: `getLatestSequenceForServices`, naming and implementation confused me, I expected it to load all sequences for all services

bridge/server/services/data-service.ts
```diff
@@ -53,87 +60,218 @@ export class DataService {
...
+          undefined,
+          keptnContexts.join(',')
+        );
+        sequences = [...sequences, ...response.data.states];
+      }
+    }
+    return sequences;
+  }
+
+  private async getDeploymentFinishedForServices(
```
laneli, line 154: I would also rename that to some sort of `latest` to not need to know implementation details

bridge/server/services/data-service.ts
```diff
@@ -53,87 +60,218 @@ export class DataService {
...
+      projectName,
+      services,
+      EventTypes.DEPLOYMENT_FINISHED,
+      (service: Service) => service.lastEventTypes[EventTypes.DEPLOYMENT_FINISHED],
+      undefined,
+      resultType
+    );
+  }
+
+  private async getEvaluationResultsForServices(
```
laneli, line 169: I would also rename them to some sort of latest to not need to know implementation details

bridge/server/services/data-service.ts
```diff
@@ -53,87 +60,218 @@ export class DataService {
...
+        const latestEvent = service.getLatestEvent();
+        const latestEvaluation = service.lastEventTypes[EventTypes.EVALUATION_FINISHED];
+        return latestEvent?.keptnContext === latestEvaluation?.keptnContext ? latestEvaluation : undefined;
+      },
+      KeptnService.LIGHTHOUSE_SERVICE,
+      resultType
+    );
+  }
+
+  private async getTracesOfMultipleServices(
```
laneli, line 188: I would also rename them to some sort of latest to not need to know implementation details

bridge/server/services/data-service.ts
```diff
@@ -53,87 +60,218 @@ export class DataService {
...
-              lastSequences[service.serviceName]
-            );
-            if (latestSequence) {
-              lastSequences[service.serviceName] = latestSequence;
-            }
-          } catch (error) {
-            console.error(error);
+        const latestEvent = service.getLatestEvent();
+        if (latestEvent) {
+          const latestSequence = await this.fetchServiceDetails(
```
laneli, line 105: Maybe naming, but by calling just this it's a bit hard to say, that this call mutates the service object in the project

PR #6058, opened 2021-11-18
fix(configuration-service): Completely replace previous helm chart directory when updating (#6050)

### LINKED ISSUE #6050: Configuration service: Updates to helm charts do not completely replace the previous chart
Let's say we have a service `foo` with the following helm chart:

```
foo
  -> chart
    -> templates
      -> svc.yml
      -> deployment.yaml
```

Then we update the chart of the service with a newer version, which has a different structure

```
foo
  -> chart
    -> templates
      -> service.yml
      -> deployment.yaml
```

In that case, the configuration service will currently not completely replace the previous chart with the new one, but actually merge those two. The resulting structure will look as follows:

```
foo
  -> chart
    -> templates
      -> service.yml
      -> svc.yaml // leftover of the previous version
      -> deployment.yaml // this will be correctly replaced
```

*Note:* Currently the workaround for this kind of scenario is to recreate the service - the event history will still be kept since the event collections are only deleted once the project containing the service is deleted. Configurations such as `slo.yaml` files will have to be re-uploaded as well

bacherfl:
Closes #6050 

Integration test run: https://github.com/keptn/keptn/actions/runs/1477137808

### DIFF
--- a/configuration-service/common/git.go
+++ b/configuration-service/common/git.go
@@ -341,6 +341,15 @@ func (g *Git) GetDefaultBranch(project string) (string, error) {
 	return masterBranch, nil
 }
 
+func (g *Git) Reset(project string) error {
+	projectConfigPath := config.ConfigDir + "/" + project
+	_, err := g.Executor.ExecuteCommand("git", []string{"reset", "--hard"}, projectConfigPath)
+	if err != nil {
+		return err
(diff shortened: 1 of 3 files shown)


RealAnna (approved):
LGTM

PR #6062, opened 2021-11-19
feat: Enable Nats queue groups in helm chart (#4519)

### LINKED ISSUE #4519: Multiple Replicas: Process messages at most once
## Action Item
Currently, we run all of our Deployments of the stateless service in Keptn Core with a single replica.
However, during a rolling update, there is a timespan where K8S is spawning up the new pod before terminating the current one, hence two pods will run at the same time.

We need to make sure that we do not process messages multiple times because multiple replicas of a keptn integration are running.

### Implementation Details

The implementation details will be based on the result of the research done in #4689

### Acceptance Criteria
- [ ] Message sent to a replicated keptn service is processed at most once.
- [ ] Messages can still be processed by multiple distinct Keptn services reacting on the same event

### Definition of Done
- [x] Feature implemented and reviewed
- [ ] Integration test in place
- [x] Unit tests
- [ ] Enable Queue groups

## Dependencies
- [x] https://github.com/keptn/keptn/issues/4689

bacherfl:
Closes #4519

Integration test run: https://github.com/keptn/keptn/runs/4296832461?check_suite_focus=true
Note: The failed test on minishift seemed to be a temporary problem with minishift itself


RealAnna (approved):
LGTM

PR #6064, opened 2021-11-19
feat: Added unit tests for webhook-service and improved error messages

### LINKED ISSUE #5692: Core: Improve error message of webhook-service
## Action Item

The `webhook-service`  should provide a more descriptive message for when a variable cannot be parsed. 

Currently, the message is generic; e.g.:
```
could not parse request 'curl --header 'Content-type: application/json' --request POST --data '{"text":"Hello, World! {{.env.data.project}}"}' https://hooks.slack.com/services/yada_yada_yada'
```

It should report something like:
```
could not parse {{.env.data.project}} for 'curl --header 'Content-type: application/json' --request POST --data '{"text":"Hello, World! {{.env.data.project}}"}' https://hooks.slack.com/services/yada_yada_yada' 
```

## DoD
- [ ] Feature implemented
- [ ] Unit tests

odubajDT:

## This PR
- adds unit tests for webhook-config and improved error messages

### Related Issues

Fixes #5692

### DIFF
--- a/webhook-service/handler/handler.go
+++ b/webhook-service/handler/handler.go
@@ -3,12 +3,13 @@ package handler
 import (
 	"errors"
 	"fmt"
+	"strings"
+
 	"github.com/keptn/go-utils/pkg/api/models"
 	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
 	"github.com/keptn/keptn/go-sdk/pkg/sdk"
 	"github.com/keptn/keptn/webhook-service/lib"
 	logger "github.com/sirupsen/logrus"
-	"strings"
 )
 
 const webhookConfigFileName = "webhook/webhook.yaml"
@@ -174,9 +175,9 @@ func (th *TaskHandler) onStartedWebhookExecution(keptnHandler sdk.IKeptn, event 
 		}
 	} else {
 		// if sendFinished is set to false, we need to send a .started event for each webhook request to be executed
-		for range webhook.Requests {
+		for _, req := range webhook.Requests {
 			if err := keptnHandler.SendStartedEvent(event); err != nil {
-				return sdkError(fmt.Sprintf("could not send .started event: %s", err.Error()), err)
+				return sdkError(fmt.Sprintf("could not send .started event: request '%s': %s", req, err.Error()), err)
 			}
 		}
 	}
--- a/webhook-service/lib/template_engine_test.go
+++ b/webhook-service/lib/template_engine_test.go
@@ -1,8 +1,9 @@
 package lib_test
 
 import (
-	"github.com/keptn/keptn/webhook-service/lib"
 	"testing"
+
+	"github.com/keptn/keptn/webhook-service/lib"
 )
 
 func TestTemplateEngine_ParseTemplate(t1 *testing.T) {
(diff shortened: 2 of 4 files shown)


webhook-service/handler/handler.go
```diff
@@ -84,7 +85,7 @@ func (th *TaskHandler) Execute(keptnHandler sdk.IKeptn, event sdk.KeptnEvent) (i
 		}
 		err = keptnHandler.SendFinishedEvent(event, result)
 		if err != nil {
-			return nil, sdkError(fmt.Sprintf("could not send finished event: %s", err.Error()), err)
+			return nil, sdkError(fmt.Sprintf("could not send finished event: request '%s' %s", webhook.Requests[0], err.Error()), err)
```
bacherfl, line 88: Can it happen that `webhook.Requests` is empty at this point? Also, is the first of the webhook requests actually tied to this particular error (sending the finished event failed)?

webhook-service/handler/handler.go
```diff
@@ -170,13 +171,13 @@ func (th *TaskHandler) onStartedWebhookExecution(keptnHandler sdk.IKeptn, event
 		// if sendFinished is set, we only need to send one started event
 		// the webhook service will then send a correlating .finished event with the aggregated response payloads
 		if err := keptnHandler.SendStartedEvent(event); err != nil {
-			return sdkError(fmt.Sprintf("could not send .started event: %s", err.Error()), err)
+			return sdkError(fmt.Sprintf("could not send .started event: request '%s': %s", webhook.Requests[0], err.Error()), err)
```
bacherfl, line 174: Can it happen that `webhook.Requests` is empty at this point? Also, is the first of the webhook requests actually tied to this particular error (sending the started event failed)?

webhook-service/handler/handler.go
odubajDT: Actually it shouldn't happen, we are now checking that webhook.Requests are not empty, see this PR below

webhook-service/lib/webhook_config.go
```diff
@@ -44,5 +48,24 @@ func DecodeWebHookConfigYAML(webhookConfigYaml []byte) (*WebHookConfig, error) {
...
+	for _, webhook := range webHookConfig.Spec.Webhooks {
+		if webhook.Type == "" {
+			return nil, errors.New("Webhook configuration invalid: missing 'webhooks[].Type' part")
+		}
+
+		if webhook.SubscriptionID == "" {
+			return nil, errors.New("Webhook configuration invalid: missing 'webhooks[].SubscriptionID' part")
+		}
+
+		if len(webhook.Requests) == 0 {
```
odubajDT, line 65: @bacherfl in this place

webhook-service/handler/handler.go
odubajDT: Actually it shouldn't happen, we are now checking that webhook.Requests are not empty, see this PR below

webhook-service/handler/handler.go
bacherfl: ok thank you for clarifying :) But still, adding the first of the requests to this error message might be misleading, as the error in this case does not see, to be tied to that particular request

webhook-service/lib/template_engine_test.go
```diff
@@ -42,6 +43,46 @@ func TestTemplateEngine_ParseTemplate(t1 *testing.T) {
 			want:    "",
 			wantErr: true,
 		},
+		{
```
bacherfl, line 46: Is there a way to get any information on what field of a template string was causing the error? I.e. like described in the issue (https://github.com/keptn/keptn/issues/5692)

odubajDT: will look at it :)

odubajDT:
@bacherfl please re-review? I removed the request[0] as they really do not make sense there and added the required errMsg to the unit tests, hope now it will be good :)

webhook-service/lib/template_engine_test.go
thisthat: Maybe this could help here: https://pkg.go.dev/text/template#Template.Option

odubajDT: will add, thanks!

odubajDT: we already have this Option() function used in ParseTemplate(), see https://github.com/keptn/keptn/blob/master/webhook-service/lib/template_engine.go#L16

bacherfl (approved):

PR #6065, opened 2021-11-19
fix(shipyard-controller): migrate and avoid mongodb keys containing dots

### LINKED ISSUE #6033: Circumvent field name restrictions of MongoDB
Amazon DocumentDB does **not** support dots (`.`) in a document field name. However, dots are used in the `keptnProjectsMV` collection in the `lastEventTypes` field:

image

This field is heavily used by the Keptn bridge, thus, cannot be removed.

### Proposal
In the end, we need to change the format of this key to use something different than a dot (`.`).
The two most promising options would be:
* encoding the key as base64 or
* replacing the `.` with its unicode equivalent (i.e., u002e)

Note, that there also other problematic characters such as a prefixed `$` (see Field Name Restrictions [here](https://docs.aws.amazon.com/documentdb/latest/developerguide/functional-differences.html))

The collection is populated by the `ShipyardController`, thus we need to convert data inserted into this collection to the correct format. Provide a `encodeKey` as  well as a `decodeKey` function that is doing the necessary translation

Moreover, existing data might need to be migrated to the new format. This shall be done in a similar way like it was already done with the `SequenceMigrator` in the ShipyardController. Please also check if the current `SequenceMigrator` is still needed, if not, delete it. The migration step shall take place during the startup of the Shipyard controller.

## DoD

- [ ] [mongoexport](https://docs.mongodb.com/database-tools/mongoexport/#mongodb-binary-bin.mongoexport) and [mongoimport](https://docs.mongodb.com/database-tools/mongoimport/#mongodb-binary-bin.mongoimport) work
- [ ] Unit tests

warber:
This PR includes the necessary changes to intercept interactions with the mongoDB project repository.
Dots (`.`) are replaced using the `~` as escape character

closes #6033


thisthat:
would it be possible to create an integration test that uses [mongoexport](https://docs.mongodb.com/database-tools/mongoexport/#mongodb-binary-bin.mongoexport) and [mongoimport](https://docs.mongodb.com/database-tools/mongoimport/#mongodb-binary-bin.mongoimport) to test our backup/restore strategy?

warber:
> would it be possible to create an integration test that uses [mongoexport](https://docs.mongodb.com/database-tools/mongoexport/#mongodb-binary-bin.mongoexport) and [mongoimport](https://docs.mongodb.com/database-tools/mongoimport/#mongodb-binary-bin.mongoimport) to test our backup/restore strategy?

yeah, let's do this in a separate ticket :) --> https://github.com/keptn/keptn/issues/6072

odubajDT:
When merging this PR, please merge also this one https://github.com/keptn/keptn/pull/6071

shipyard-controller/main.go
```diff
@@ -214,8 +215,11 @@ func main() {
 	logController := controller.NewLogController(logHandler)
 	logController.Inject(apiV1)
 
-	sequenceMigrator := handler.NewSequenceMigrator(createEventsRepo(), createStateRepo(), createProjectRepo())
-	sequenceMigrator.Run()
+	projectsMigrator := migration.NewProjectMVMigrator(db.GetMongoDBConnectionInstance())
+	err = projectsMigrator.MigrateKeys()
+	if err != nil {
+		log.Fatalf("Unable to run projects migrator: %v", err)
```
bacherfl, line 221: This seems a bit risky - If for some reason the `lastEventTypes` property  a project could not be migrated, this would mean that Keptn can not be used at all

bacherfl (dismissed):

thisthat (dismissed):

bacherfl (approved):

PR #6080, opened 2021-11-22
fix(bridge): Correctly show warning state (#6003)

### LINKED ISSUE #6003: Sequence in dashboard does not show warning result
If an evaluation results in `warning`, the dashboard still shows the evaluation for this sequence in green (`success`).
image

## Expected behavior
The bridge shows the evaluation icon in amber instead of green.

Kirdock:
Fixes #6003 


laneli (approved):
We have somehow multiple ways to check if we can show a result icon - I would not do this in scope of this issue, but we should clean that up at some point.

PR #6093, opened 2021-11-23
feat(shipyard-controller): Allow to filter sequence states by multiple keptnContext IDs (#6056)

### LINKED ISSUE #6056: Add the ability for the sequence endpoint to accept multiple keptn contexts
The `controlPlane/v1/sequence` endpoint should be adapted to accept multiple keptn contexts (then the sequences are queried by `context1 OR context2 ...`).
Suggestion: The `keptnContext` parameter could then be a concatenation of contexts separated by a comma.

## Pain
In the environment screen of the bridge, we have one call per service per stage to retrieve all the latest sequences.  This leads to a lot of API calls (30 services and 3 stages would be 90 calls). If the sequence endpoint supports several keptn contexts, we could reduce it to one call.

## Definition of Done
- [ ] The `keptnContext` parameter of the sequence endpoint can be used to provide several keptn contexts.
- [ ] Unit tests

bacherfl:
Closes #6056

Integration test run: https://github.com/keptn/keptn/actions/runs/1495544214

### DIFF
--- a/shipyard-controller/db/mongodb_state_repo.go
+++ b/shipyard-controller/db/mongodb_state_repo.go
@@ -9,6 +9,7 @@ import (
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
+	"strings"
 	"time"
 )
 
@@ -116,7 +117,11 @@ func (mdbrepo *MongoDBStateRepo) getSearchOptions(filter models.StateFilter) bso
 	}
 
 	if filter.KeptnContext != "" {
-		searchOptions["shkeptncontext"] = filter.KeptnContext
+		splitContexts := strings.Split(filter.KeptnContext, ",")
+		//searchOptions["shkeptncontext"] = filter.KeptnContext
+		searchOptions["shkeptncontext"] = bson.M{
+			"$in": splitContexts,
+		}
 	}
 
 	if filter.Name != "" {
--- a/shipyard-controller/db/mongodb_state_repo.go
+++ b/shipyard-controller/db/mongodb_state_repo.go
@@ -118,7 +118,6 @@ func (mdbrepo *MongoDBStateRepo) getSearchOptions(filter models.StateFilter) bso
 
 	if filter.KeptnContext != "" {
 		splitContexts := strings.Split(strings.ReplaceAll(filter.KeptnContext, " ", ""), ",")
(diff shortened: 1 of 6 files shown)


RealAnna (approved):
LGTM

shipyard-controller/db/mongodb_state_repo.go
```diff
@@ -116,7 +117,11 @@ func (mdbrepo *MongoDBStateRepo) getSearchOptions(filter models.StateFilter) bso
 	}
 
 	if filter.KeptnContext != "" {
-		searchOptions["shkeptncontext"] = filter.KeptnContext
+		splitContexts := strings.Split(strings.ReplaceAll(filter.KeptnContext, " ", ""), ",")
```
warber, line 120: how can `filter.KeptnContext` contain spaces? or did you just want to trim away leading/trailing spaces? in that case there's `string.TrimSpace()`

shipyard-controller/db/mongodb_state_repo.go
```diff
@@ -116,7 +117,11 @@ func (mdbrepo *MongoDBStateRepo) getSearchOptions(filter models.StateFilter) bso
...
+		//searchOptions["shkeptncontext"] = filter.KeptnContext
```
warber, line 121: remove commented line

shipyard-controller/db/mongodb_state_repo.go
RealAnna: this was added purely to make me happy 🍡

bacherfl: in theory it shouldn't but it does not harm as an additional safety measure

shipyard-controller/db/mongodb_state_repo.go
bacherfl: done

shipyard-controller/db/mongodb_state_repo.go
bacherfl: although I still think we should keep it, because the filter criteria might look like the following:

```
&keptnContext="context-id1, context-id2, context-id3"
```

in that case removing all whitespaces (TrimSpace won't suffice here) is necessary since the query likely won't work otherwise. I'll add a unit test to check that

warber: yeah, you're right, let's keep it.
+1 for the unit test

warber (approved):

PR #6099, opened 2021-11-23
refactor(bridge): Move secret picker in own component (#5733)

### LINKED ISSUE #5566: Typo: selectSecret in secret popup
add-webhook-secret

In the popup the headline is `selectSecret`, it should be `Select secret`.

### LINKED ISSUE #5733: Bridge: Move the secret pickup into its own component
## Action Item

The secret pickup should be moved into its own component and accept a generic tree-like data structure. 

## Techincal details 

Currently, it only supports two layers. This should be extended to be general.

## DoD

- [x] Unit tests with custom tree-like data
- [x] The new component is implemented

ermin-muratovic:
## This PR
- moves the secret picker in its own component

### Related Issues
Fixes #5733 
Fixes #5566


bridge/client/app/_components/ktb-variable-selector/ktb-variable-selector.component.html
```diff
@@ -0,0 +1,11 @@
+<button
+  dt-icon-button
+  variant="nested"
+  aria-label="Add secret"
```
laneli, line 4: aria-label should be configurable, as we will set other things than secrets with it

bridge/client/app/_components/ktb-variable-selector/ktb-variable-selector.component.html
```diff
@@ -0,0 +1,11 @@
...
+  ktbTreeListSelect
+  [data]="data"
+  [options]="options"
+  (selected)="setVariable($event)"
+>
+  <dt-icon name="resetpassword"></dt-icon>
```
laneli, line 10: icon name should be configurable, as we will set other things than secrets with it

laneli (changes requested):

bridge/client/app/_components/ktb-webhook-settings/ktb-webhook-settings.component.html
```diff
@@ -21,14 +21,17 @@ <h2>Webhook configuration</h2>
...
       <div dtSuffix>
-        <ng-container
-          [ngTemplateOutlet]="secretButton"
-          [ngTemplateOutletContext]="{
-            $implicit: undefined,
-            controlName: 'url',
-            selectionStart: urlInput.selectionStart
-          }"
-        ></ng-container>
+        <ktb-variable-selector
```
laneli, line 24: For this component, I think it would make sense to put this into an ng-template and reference it at needed places to avoid duplications. Also if we need to change text etc. we don't have to touch 3 places.

laneli (changes requested):

laneli (approved):

PR #6157, opened 2021-11-24
feat(bridge): Add event picker to webhook configuration (#5260)

### LINKED ISSUE #5260: Bridge: Add event payload variables to webhook configuration
## User Story
As I user I want to select an event payload variable for the custom payload from a predefined set, to not have to remember it, and I have not to search for the variables in an event payload.

## Mockup
Add Subscription3

## Technical Details

The implementation depends on https://github.com/keptn/keptn/issues/5733
It should only fetch one previous event for a given type. 

### Acceptance criteria
* When the user clicks on the button in the text area header of the "Custom payload" field, a popup overlay opens.
* The overlay provides a list of variables from the event payload that can be selected.
    * This list provides the base properties as defined in the spec.
    * When the last event is available the properties are enhanced with these.
* When selected, the variable is inserted at the point where the cursor stands currently.
* The variable is inserted with the helm variable syntax: e.g., `{{ .varName }}`
* If the field loses focus, and the user clicks on the button, the field has to be at least focused again.
Nice to have: focus on the last point where the cursor was, or at the beginning if not focused yet.
* The list is expandable for nested properties. Clicking on such a nested property fills the whole path of the property.

❓ What list is provided when the user did not select a trigger yet? Or when there is no prev event?
 - The team decided to provide a default event:
 ```
{
  id, 
  shkeptncontext, 
  type, 
  data: {
    project,
    service,
    stage
  },
  time
}
```

❓ What list is provided when the user selects a wildcard?
  - The intersection of the events: look for an algo out-of-the-box to do this. 

## DoD
- [ ] Open points are discussed within the Bridge team and the final decision is documented here as a comment
- [ ] All fields where secret can be defined, they can also be enriched with variables
- [ ] UI Tests
- [ ] Unit tests 

## Dependencies

- https://github.com/keptn/keptn/issues/5733

### PR DESCRIPTION
**Pull request by Kirdock**
Closes #5260 

### CODE REVIEW

**Comment by ermin-muratovic** [comment]
When the list in the selector is very long, the scrollbar is on the outer div. Can we fix this, to have the scrollbar in the inner div, so that the title stays on top.

image

File: bridge/client/app/_components/ktb-modify-uniform-subscription/ktb-modify-uniform-subscription.component.ts [hunk 1]
```diff
@@ -195,6 +202,35 @@ export class KtbModifyUniformSubscriptionComponent implements OnDestroy {
...
+                data: {
+                  project: undefined,
+                  service: undefined,
+                  stage: undefined,
+                },
+                id: undefined,
+                type: undefined,
+                time: undefined,
+                shkeptncontext: undefined,
+              };
```
ermin-muratovic, line 229: Can we move this to a model file?

ermin-muratovic (changes requested):

ermin-muratovic (approved):

PR #6194, opened 2021-11-26
feat(lighthouse-service): Add compared value to payload (#5496)

### LINKED ISSUE #5496: Core: Lighthouse-service should return the evaluation result as part of the event payload
## Action Item

The `lighthouse-service` shall be changed to return the evaluation values used for calculation as a part of the event payload.

## Technical Details

Currently, the `lighthouse-service` only returns the list of evaluation events used for the comparison under the `comparedEvents` property. To understand why an evaluation passes/fails, a user has to manually compute the comparison values which is an error-prone process. 
The service should report the past value used for the SLO assessment as a new field with the name `comparedValue`.

## DoD

- [ ] The service annotates the finished event with the past evaluation value
- [ ] Unit tests
- [ ] Adapt integration tests
- [ ] Spec update

RealAnna:
## This PR

When receiving an SLI result each metric is compared with previous values, those are aggregated following a certain aggregation strategy defined in the SLO. In case of comparison criterias this values stays the same for all comparisons and influence the target values used to decide whether the metric passes or fails. 

In this PR we store the aggregated value in each evaluation result as ComparedValue, to facilitate understanding of why the metric passes or fails a comparison criteria. 

For criterias with fixed targets the pass or fail is dependent only on the TargetValue, anyhow ComparedValue is calculated.

- unit tests modified
- integration test checked

### Related Issues

Fixes #5496

### DIFF
--- a/lighthouse-service/deploy/service.yaml
+++ b/lighthouse-service/deploy/service.yaml
@@ -28,6 +28,18 @@ spec:
       containers:
         - name: lighthouse-service
           image: keptn/lighthouse-service:latest
+          livenessProbe:
+            httpGet:
+              path: /health
+              port: 8080
+            initialDelaySeconds: 0
+            periodSeconds: 5
+          readinessProbe:
+              httpGet:
+                path: /health
+                port: 8080
+              initialDelaySeconds: 5
+              periodSeconds: 5
           ports:
             - containerPort: 8080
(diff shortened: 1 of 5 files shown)


odubajDT:
LGTM!

odubajDT (approved):

PR #6212, opened 2021-11-29
test: Update integration test pipeline

### LINKED ISSUE #4819: Pipeline should fail if tests fail
## Action Item

The [Integration Tests](https://github.com/keptn/keptn/blob/master/.github/workflows/integration_tests.yml) pipeline should fail if a single test fails. Currently, it fails only if a configuration step does not succeed.

Additionall, the test execution and creation of a test report shall be done using the `go` builtin tools.

Example: https://github.com/keptn/keptn/runs/3203020688?check_suite_focus=true#step:50:42

mowies:
### This PR
- re-writes the integration test pipeline
- makes use of [gotestsum](https://github.com/gotestyourself/gotestsum) to run go integration tests
- sets up a new way to configure platforms to run specific tests (through the `testsuite_<platform-name>.go` files)
- re-writes the report generation for the GitHub issue that gets created for failing tests
- removes all code related to GH draft release creation
- updates meta keptn CLI that is used to 0.11.4
- removes the `examples_branch` pipeline input since it's not used anymore (even before this PR)
- makes the pipeline actually red if any tests fail

I left lots of GH comments to make reviews easier, please use them during the code review.

Integration test run making use of the new pipeline: https://github.com/keptn/keptn/actions/runs/1653506044

Fixes #4819


.github/workflows/integration_tests.yml
```diff
@@ -771,56 +625,132 @@ jobs:
...
-          echo "| ---------------------- | ----------------- | ------------ | ---------------- | --------------- | ---------------- | ----------------- | ------------------ | ---------------- | ------------------ | ------------- | --------------- | -------------- | ------------ | ---------------- | ------------------ | ----------------- |----------- | ------------------------ | -------------- | ------------------- | ----------------- | ---------------------- |" >> final-test-report.txt
-          cat test-report*.txt >> final-test-report.txt
-
-          # print test report
-          REPORT=$(cat final-test-report.txt)
-          echo $REPORT
-
-      - name: Upload final test report as an artifact
+          npx tablemark-cli@v2.0.0 "${{ env.FINAL_TEST_REPORT_FILEPATH_JSON }}" > "${{ env.FINAL_TEST_REPORT_FILEPATH_MARKDOWN }}" -N -c Test
```
mowies, line 731: This step uses the `tablemark` CLI to generate a markdown table from the final JSON test report

.github/workflows/integration_tests.yml
```diff
@@ -829,8 +759,9 @@ jobs:
           github_token: ${{secrets.GITHUB_TOKEN}}
           workflow: CI.yml
           run_id: ${{ needs.integration-test.outputs.ARTIFACT_RUN_ID }}
+          name: build-config
```
mowies, line 762: Only download the `build-config` artifact instead of all artifact, since none of the other artifacts are actually needed here.

.github/workflows/integration_tests.yml
```diff
@@ -853,11 +784,18 @@ jobs:
...
         run: |
-          REPORT=$(cat final-test-report.txt)
+          REPORT=$(cat "$FINAL_TEST_REPORT_FILEPATH_MARKDOWN")
 
-          if [[ "$REPORT" == *"failure"* ]]; then
+          if [[ "$REPORT" == *"fail"* ]]; then
             echo "INTEGRATION TESTS FAILED! - Creating GH issue description now..."
+          
+            # adding emojis to test report
+            REPORT=$(echo "${REPORT//pass/:white_check_mark: pass}")
```
mowies, line 796: Add emoji to the test report :)

.github/workflows/integration_tests.yml
```diff
@@ -898,162 +836,6 @@ jobs:
         with:
           filename: integration-tests-failed.md
 
-      - name: Find PR number for current build
```
mowies, line 901: All the code related to creating a draft release was remove since this now lives in the `release.yml` and `prerelease.yml` pipelines.

.github/workflows/integration_tests.yml
```diff
@@ -771,56 +625,132 @@ jobs:
...
+                platformTestReportRaw.forEach(platformTestResult => {
+                    if(platformTestResult.test.includes("/")) {
+                        jsonReportData.push(platformTestResult);
+                    }
+                });
+            });
+            
+            const testResultMap = {};
+            
+            jsonReportData.forEach(testResult => {
```
mowies, line 684: Group the test results by test name.
Result: map: key=testname, value=array of test results for the different platforms

.github/workflows/integration_tests.yml
```diff
@@ -771,56 +625,132 @@ jobs:
...
+                
+                // Add to result list
+                testResultMap[trimmedTestResult.test].push(trimmedTestResult)
+            });
+            
+            const finalTestReport = [];
+            const allPlatforms = [];
+            
+            // Create properly formatted final test report
+            Object.entries(testResultMap).forEach(([test, testResults]) => {
```
mowies, line 700: Create final JSON object with the following form:
```
[
  {
    "test": "<test-name>",
    "platform1": "pass",
    "platform2": "fail",
    "platform3": "skip"
  },
  .
  .
  .
]
```

.github/workflows/integration_tests.yml
```diff
@@ -771,56 +625,132 @@ jobs:
...
+                    const currentPlatform = Object.keys(results)[0];
+                    if(!allPlatforms.includes(currentPlatform)) {
+                        allPlatforms.push(currentPlatform);
+                    }
+                });
+                finalTestReport.push(finalResult);
+            });
+            
+            // Fill in platforms where a test was not executed
+            finalTestReport.forEach(testResult => {
```
mowies, line 716: Add "skip" test result entries to the testcases where not all platforms were configured to run those testcases

test/utils/combine_and_filter_go_test_json_report.sh
```diff
@@ -0,0 +1,37 @@
...
+echo "Combining test report files from source folder..."
+# shellcheck disable=SC2066
+for file in "$SOURCE_TEST_REPORT_FOLDER"/*; do
+  echo "Appending $file ..."
+  cat "$file" >> "$DESTINATION_TEST_REPORT_FILE"
+done
+
+echo "Filtering combined test report file..."
+# shellcheck disable=SC2002
+filtered_report=$(cat "$DESTINATION_TEST_REPORT_FILE" | jq -c "select( (.Action == \"pass\" or .Action == \"fail\") and .Test != null ) | {\"test\": .Test, \"$REWRITE_TEST_TARGET\": .Action}")
```
mowies, line 36: This `jq` command takes the go test JSON report file generated by `gotestsum` and only takes those lines where the `Test` variable was defined and the `Action` was either `pass` or `fail`. Then, it prints those lines in the following format: `{"test": "<test-name>", :"<test-platform-name>": "<test-result>"}`

The whole result is the written back into a variable for output to a file.

.github/workflows/integration_tests.yml
```diff
@@ -99,18 +77,22 @@ jobs:
...
+      AIRGAPPED_REGISTRY_URL: "k3d-container-registry.localhost:12345"
       REMOTE_EXECUTION_PLANE: ${{ matrix.REMOTE_EXECUTION_PLANE }}
-      BACKUP_RESTORE: ${{ matrix.BACKUP_RESTORE }}
-      RESOURCE_SERVICE: ${{ matrix.RESOURCE_SERVICE }}
-      RUN_QUALITY_GATES_TEST: ${{ matrix.RUN_QUALITY_GATES_TEST }}
-      KEPTN_EXAMPLES_BRANCH: ${{ github.event.inputs.examples_branch }}
       COLLECT_RESOURCE_LIMITS: ${{ matrix.COLLECT_RESOURCE_LIMITS }}
       GO_VERSION: ^1.16
+      TEST_REPORT_FOLDER: test-reports-${{ matrix.CLOUD_PROVIDER}}-${{ matrix.PLATFORM_VERSION }}
+      FINAL_PLATFORM_TEST_REPORT_FOLDER: test-reports
```
thisthat, line 85: why this was has `PLATFORM` in its name?

.github/workflows/integration_tests.yml
```diff
@@ -492,214 +474,62 @@ jobs:
...
-        if: env.BACKUP_RESTORE == 'true'
-        timeout-minutes: 7
-        continue-on-error: true
-        env:
-          KEPTN_ENDPOINT: ${{ steps.authenticate_keptn_cli.outputs.KEPTN_ENDPOINT }}
-        run: cd test/go-tests && go test -run Test_BackupRestore -v
+      - name: Prepare test run
+        id: prepare_test_run
+        run: |
+          go install gotest.tools/gotestsum@latest
```
thisthat, line 480: I would pin it to a specific version to avoid flakiness due to a breaking change

.github/workflows/integration_tests.yml
```diff
@@ -749,20 +579,44 @@ jobs:
...
+          for ((i=0; i<${#test_names[@]}; i++)); do
+            test_name="${test_names[$i]}"
+            test_outcome="${test_outcomes[$i]}"
+          
+            if [[ "$test_outcome" == "" ]]; then
+              test_outcome="skip"
+            elif [[ "$test_outcome" == "success" ]]; then
+              test_outcome="pass"
+            elif [[ "$test_outcome" == "failure" ]]; then
+              test_outcome="fail"
```
thisthat, line 606: maybe also add
```
else
   test_outcome="something went wrong"
```

.github/workflows/integration_tests.yml
```diff
@@ -771,56 +625,132 @@ jobs:
...
     steps:
       - name: Check out code
         uses: actions/checkout@v2.4.0
 
-      - name: Debug - Output Branch
-        run: echo $BRANCH
+      - name: Set up Node
+        uses: actions/setup-node@v2.4.1
+        with:
+          node-version: 14
```
thisthat, line 646: I'd go with 16 if possible.

.github/workflows/integration_tests.yml
```diff
@@ -771,56 +625,132 @@ jobs:
...
+            const jsonReportData = [];
+            
+            const fileList = fs.readdirSync(TEST_REPORTS_PATH);
+            fileList.forEach(fileName => {1
+                console.log(`Reading file: ${fileName}`);
+                const platformReportFile = fs.readFileSync(TEST_REPORTS_PATH + fileName, {encoding:'utf8', flag:'r'});
+                const platformTestReportRaw = ndJsonParser(platformReportFile);
+                
+                platformTestReportRaw.forEach(platformTestResult => {
+                    if(platformTestResult.test.includes("/")) {
```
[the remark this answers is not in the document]
thisthat: I would suggest adding this explanation as a comment to the script here.

test/go-tests/testsuite_gke_test.go
```diff
@@ -0,0 +1,45 @@
+package go_tests
+
+import (
+	"testing"
+)
+
+func Test_GKE(t *testing.T) {
+	// Common Tests
```
thisthat, line 8: would make sense to create `testsuite_common_test.go` which is run for each platform?

test/utils/combine_and_filter_go_test_json_report.sh
thisthat: I would also add this nice explanation as a comment

.github/workflows/integration_tests.yml
mowies: I named it this way because it will contain test reports with this file format: `test-reports/test-report-final-GKE-1.12.log`.
But it will actually also contain the final test report in the end so the name is a little misleading, I will change it to `FINAL_TEST_REPORT_FOLDER`

.github/workflows/integration_tests.yml
mowies: good catch!

.github/workflows/integration_tests.yml
mowies: The strings I check against in the if/elif are actually constant and coming from GitHub actions ([reference](https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context)) but I will add a panic `else` block just in case :)

.github/workflows/integration_tests.yml
mowies: done

test/utils/combine_and_filter_go_test_json_report.sh
mowies: done

.github/workflows/integration_tests.yml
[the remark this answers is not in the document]
mowies: done

test/go-tests/testsuite_gke_test.go
mowies: I am not sure if this works but I can try. It's a bit tricky with those test suites since they are actually the go test functions but if I add a grouping function below I don't know how the test results will behave.

thisthat: We can iterate in a follow up if we see the need for such a thing :)

thisthat (dismissed):

.github/workflows/integration_tests.yml
```diff
@@ -492,214 +474,62 @@ jobs:
...
+      #######################################################################
+      # COMMON TESTS
+      ##############
 
-      - name: Test Continuous Delivery with parallel stages on Remote Exec Plane (with sockshop)
-        id: test_continuous_delivery_remote_exec
-        if: env.RUN_AIRGAPPED_TEST != 'true' && (env.RUN_CONTINUOUS_DELIVERY_TEST == 'true') && (env.REMOTE_EXECUTION_PLANE == 'true')
+      - name: Integration Tests
+        id: test_aggregated
         timeout-minutes: 45
```
bacherfl: I think we need to increase the timeout value since https://github.com/keptn/keptn/runs/4701677577?check_suite_focus=true timed out on the GKE platform tests

mowies: i already increased the timeout limit to 60 minutes 👍🏼

bacherfl (approved):

PR #6222, opened 2021-11-29
feat: validate shipyard.yaml when updating project

### LINKED ISSUE #5102: Shipyard validation for PUT `/project`
## Action Item

In the control plane API, the `PUT /project` accepts a new shipyard configuration. 
This endpoint should validate the shipyard and return a 400 HTTP error if:
- [ ] the format does not respect the specifications https://github.com/keptn/spec/blob/master/shipyard.md
- [ ] the new shipyard configuration add/remove/rename a stage 

Otherwise, it should return the HTTP 200 status code.

--- 
## DoD

- [ ] Unit tests
- [ ] Validation implemented
- [ ] Document this validation step in the swagger API

odubajDT:

## This PR

- validate shipyard.yaml format when updating project
- adds possibility to update shipyard file via CLI
- check stages names are not modified during update

### Related Issues

Fixes #5102


cli/cmd/update_project.go
```diff
@@ -47,29 +46,29 @@ For more information about updating projects or upstream repositories, please go
...
-			errorMsg += "Keptn relies on the following conventions: "
-			errorMsg += "start with a lower case letter, then lower case letters, numbers, and hyphens are allowed.\n"
-			errorMsg += "Please update project name and try again."
-			return errors.New(errorMsg)
-		}
 		return nil
 	},
 	RunE: func(cmd *cobra.Command, args []string) error {
+
+		shipyard, err := retrieveShipyard(*updateProjectParams.Shipyard)
```
thisthat, line 53: Can this code be extracted and shared with the create project function?

odubajDT: can be, sure

odubajDT:
Results of integration tests: https://github.com/keptn/keptn/actions/runs/1534374690

shipyard-controller/handler/projecthandler_test.go
```diff
@@ -348,8 +348,8 @@ func TestUpdateProject(t *testing.T) {
 		ProjectManager IProjectManager
 		EventSender    common.EventSender
 	}
-	examplePayload := `{"gitRemoteURL":"http://remote-url.com","gitToken":"99c4c193-4813-43c5-864f-ad6f12ac1d82","gitUser":"gituser","name":"myproject"}`
```
thisthat, line 351: why this change?

shipyard-controller/handler/projectmanager.go
```diff
@@ -234,9 +265,9 @@ func (pm *ProjectManager) Update(params *models.UpdateProjectParams) (error, com
 	}
 
 	// try to update shipyard project resource
-	if params.Shipyard != "" {
+	if *params.Shipyard != "" && params.Shipyard != nil {
```
thisthat, line 268:
```suggestion
	if params.Shipyard != nil && *params.Shipyard != "" {
```

shipyard-controller/models/project.go
```diff
@@ -117,7 +118,32 @@ func (updateProjectParams *UpdateProjectParams) Validate() error {
 		return errors.New("project name missing")
 	}
 	if !keptncommon.ValidateKeptnEntityName(*updateProjectParams.Name) {
-		return errors.New("provided project name is not a valid Keptn entity name")
+		errorMsg := "Project name contains upper case letter(s) or special character(s).\n"
```
thisthat, line 121: I would not tell what is wrong but rather only say what we accept. This way we need to maintain only one place 😉 
Furthermore, error messages should start with a lower case and don't be verbose. These details should be covered by documentation and not by error messages.

shipyard-controller/models/project.go
```diff
@@ -117,7 +118,32 @@ func (updateProjectParams *UpdateProjectParams) Validate() error {
...
+		errorMsg += "start with a lower case letter, then lower case letters, numbers, and hyphens are allowed.\n"
+		errorMsg += "Please update project name and try again."
+		return errors.New(errorMsg)
+	}
+
+	if updateProjectParams.Shipyard != nil && *updateProjectParams.Shipyard != "" {
+		shipyard := &keptnv2.Shipyard{}
+		decodeString, err := base64.StdEncoding.DecodeString(*updateProjectParams.Shipyard)
+		if err != nil {
+			return errors.New("could not decode shipyard content using base64 decoder: " + err.Error())
```
thisthat, line 132:
```suggestion
			return errors.New("could not decode shipyard content")
```

thisthat (changes requested):

shipyard-controller/handler/error.go
```diff
@@ -10,6 +10,8 @@ var ErrServiceNotFound = errors.New("service not found")
 
 var ErrProjectNotFound = errors.New("project not found")
 
+var ErrInvalidStageChange = errors.New("stage name cannot be changed/removed/updated")
```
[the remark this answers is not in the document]
odubajDT: ah, should be added/removed/updated :D

shipyard-controller/handler/projectmanager_test.go
```diff
@@ -676,12 +676,13 @@ func TestUpdate_UpdateProjectShipyardResourceFails(t *testing.T) {
 	}
 
 	instance := NewProjectManager(configStore, secretStore, projectMVRepo, taskSequenceRepo, eventRepo, sequenceQueueRepo, eventQueueRepo)
+	myShipyard := "my-shipyard"
```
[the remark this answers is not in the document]
odubajDT: actually, I assume we want to be consistent, in createProjectParams, Shipyard is type of *string, in updateProjectParams, shipyard is type of string. I think it is easier for us to be consistent due to possible future changes/updates, so we can handle both structs the "same way".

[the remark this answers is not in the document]
thisthat: Good catch :)

shipyard-controller/handler/projectmanager.go
```diff
@@ -184,6 +185,36 @@ func (pm *ProjectManager) Update(params *models.UpdateProjectParams) (error, com
...
+			ProjectName:     *params.Name,
+			Shipyard:        string(decodedShipyard),
+			ShipyardVersion: shipyardVersion,
+			Stages:          expandedStages,
+		}
+
+		err = pm.validateShipyardStagesUnchaged(oldProject, newProject)
+		if err != nil {
+			return ErrInvalidStageChange, nilRollback
+		}
```
thisthat, line 215: I would move this code into a function (for example `validateShipyardUpdate(params *models.UpdateProjectParams)`) and call this code later at line 269
This way, we have the logic of handling a change in the shipyard definition in a single place.

shipyard-controller/handler/projectmanager.go
```diff
@@ -234,9 +265,9 @@ func (pm *ProjectManager) Update(params *models.UpdateProjectParams) (error, com
 	}
 
 	// try to update shipyard project resource
-	if params.Shipyard != "" {
+	if params.Shipyard != nil && *params.Shipyard != "" {
```
thisthat, line 268: This check is done also a few lines later. You can consider creating a variable to store this value only once.

shipyard-controller/handler/projectmanager.go
```diff
@@ -472,6 +503,31 @@ func toModelProject(project models.ExpandedProject) apimodels.Project {
...
 }
 
+func (pm *ProjectManager) validateShipyardStagesUnchaged(oldProject *models.ExpandedProject, newProject *models.ExpandedProject) error {
+	if len(newProject.Stages) != len(oldProject.Stages) {
+		return fmt.Errorf("unallowed addition/removal of project stages")
+	}
+
+	for i, oldStage := range oldProject.Stages {
+		if oldStage.StageName != newProject.Stages[i].StageName {
+			if !stageInArrayOfStages(oldStage.StageName, newProject.Stages) {
```
thisthat, line 513: This check covers also the one at line 512.

shipyard-controller/handler/projectmanager.go
```diff
@@ -472,6 +479,62 @@ func toModelProject(project models.ExpandedProject) apimodels.Project {
 	}
 }
 
+func (pm *ProjectManager) validateShipyardStagesUnchaged(oldProject *models.ExpandedProject, newProject *models.ExpandedProject) error {
```
warber, line 482: this method should be a function (maybe in keptn_helpers package? )

shipyard-controller/handler/projectmanager.go
```diff
@@ -472,6 +479,62 @@ func toModelProject(project models.ExpandedProject) apimodels.Project {
...
+func stageInArrayOfStages(comparedStage string, stages []*models.ExpandedStage) bool {
+	for _, arrayStage := range stages {
+		if arrayStage.StageName == comparedStage {
+			return true
+		}
+	}
+	return false
+}
+
+func (pm *ProjectManager) validateShipyardUpdate(params *models.UpdateProjectParams, oldProject *models.ExpandedProject) error {
```
warber, line 507: this method should be a function (maybe in keptn_helpers package? )

shipyard-controller/models/project.go
```diff
@@ -120,5 +117,26 @@ func (updateProjectParams *UpdateProjectParams) Validate() error {
 		return errors.New("provided project name is not a valid Keptn entity name")
 	}
 
+	if updateProjectParams.Shipyard != nil && *updateProjectParams.Shipyard != "" {
```
warber, line 120: pls add a unit test

warber (approved):

odubajDT (commented):
Review finished

shipyard-controller/handler/projectmanager.go
thisthat: Thank you for the nice explanation, I would add a comment to explain this briefly otherwise while scheming through the code it is not immediate why this check is done :)

odubajDT: I will create a PR with a comment describing the reason for the double check. Thanks for the review :)

PR #6253, opened 2021-12-01
fix(bridge): Update services on project change (#6252)

### LINKED ISSUE #6252: Services in services view are not updated if the project changes
If you change the project in the service screen, the services don't change and are not updated.

Currently when switching from sockshop to empty project:
image

Expected:
image

Kirdock:
Fixes #6252 


thisthat (commented):
would it be possible to create a regression test for this problem?

Kirdock:
> would it be possible to create a regression test for this problem?

I just added a UI test for this behavior

thisthat (approved):

Issue #4206, opened 2021-05-31
Add problem title of remediation to sequence endpoint


Kirdock:
The sequence endpoint should also return the problem title of remediations in order to reduce the amount of api-calls in the bridge. 

Currently the open remediations in the service screen of the bridge need additional information and therefore there is another api-call to receive the `ProblemTitle` (e.g. "Failure rate increase") of it.
![image](https://user-images.githubusercontent.com/11599148/120198081-072d0680-c222-11eb-9e75-42e0abeefcba.png)

The sequence screen will also need this kind of information when the endpoint is changed to the new one.
![image](https://user-images.githubusercontent.com/11599148/120197871-cb923c80-c221-11eb-886e-7d6f412b4437.png)

New model:
```json
{
  "nextPageKey": 0,
  "pageSize": 0,
  "state": "...",
  "stages": [...],
  "problemTitle": "",
  "totalCount": 0
}
```

## Acceptance Criteria
- [x] The sequence endpoint returns the problem title of the remediation
- [x] State Sequence migration logic is adapted 

## Definition of done
 - [ ] Integration tests are updated

johannes-b:
I did some investigation around this topic by executing the [Quickstart](https://keptn.sh/docs/quickstart/), which contains a remediation scenario. My findings are summarized below.

Before, I would like to show the `remediation.triggered` event, which kicks off the remediation scenario: 

```
{
   "type":"sh.keptn.event.production.remediation.triggered",
   "specversion":"1.0",
   "source":"https:\/\/github.com\/keptn\/keptn\/prometheus-service",
   "id":"f2b878d3-03c0-4e8f-bc3f-454bc1b3d79d",
   "time":"2019-06-07T07:02:15.64489Z",
   "contenttype":"application\/json",
   "data":{
      "project":"podtatohead",
      "stage":"production",
      "service":"helloservice",
      "problem":{
         "problemTitle":"out_of_memory",
         "rootCause":"Response time degradation"
      }
   }
}
``` 

### Environment screen shows running remediation, but missing details:

![image](https://user-images.githubusercontent.com/729071/141940833-ef7b2bd5-2b76-4691-b3da-e5694df2f829.png)

* Which data field is used by this label? 

### Service screen does not show the remediation sequence: 

![image](https://user-images.githubusercontent.com/729071/141940987-42919e9b-1588-4dc3-9239-eee6471f81e7.png)

[...]

PR #6258, opened 2021-12-01
Feat/4206/problem title in sequence endpoint

RealAnna:
## This PR

- adds problemTitle to remediation sequences so that it can be shown just by calling the sequence endpoint
- changes data stracture of remediation service events

### Related Issues

#4206 

### Notes

if you are using 0.11.2  there is a bug in the remediation service that does not return the title, this has been fixed in master


shipyard-controller/handler/sequencehooks/sequencestate.go
```diff
@@ -49,6 +49,13 @@ func (smv *SequenceStateMaterializedView) OnSequenceTriggered(event models.Event
 		State:          models.SequenceTriggeredState,
 		Stages:         []models.SequenceStateStage{},
 	}
+
+	//if the next event in sequence is an action we get the problem title form it
+	getActionTriggeredData := &keptnv2.GetActionTriggeredEventData{}
```
bacherfl, line 54: please add a unit test for that case in `shipyard-controller/handler/sequencehooks/sequencestate_test.go` as well :)

RealAnna: sure thing I'll make one!

bacherfl (approved):

PR #6290, opened 2021-12-06
fix: Remove config-service init container in favor of better k8s security settings

mowies:


mowies:
Closed because it's obsolete

PR #6298, opened 2021-12-07
feat(bridge): Add hint that secrets are shared among all projects (#5832)

### LINKED ISSUE #5832: Bridge: Clarify with a message that secrets are shared across project.
## Action Item

In the secret screen, the message `secrets are shared across projects` should be added.

laneli:
Adds hint to secret list and create secret form that secrets are shared among all projects.

Closes #5832

### DIFF
--- a/bridge/client/styles.scss
+++ b/bridge/client/styles.scss
@@ -358,6 +358,32 @@ pre.code {
   border-radius: 6px;
 }
 
+ktb-root {
+  .dt-icon {
+    height: 20px;
+    width: 20px;
+  }
+}
+
+.inline-text-icon {
+  position: relative;
+  top: 1px;
+}
+
+.icon-inline-left {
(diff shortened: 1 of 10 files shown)


bridge/client/styles.scss
```diff
@@ -38,6 +38,11 @@ ktb-root {
   display: flex;
   flex-direction: column;
   place-content: stretch stretch;
+
+  .dt-icon {
+    height: 20px;
+    width: 20px;
```
Kirdock, line 44: This change affects all icons. Be careful here. It also affects menu icons on the left:
image
and icons like in the service screen
before:
image

now:
image

maybe there are more places

Kirdock (changes requested):
In the user panel, the warning and the duplicate icons have the wrong size.
image

Kirdock (approved):

PR #6300, opened 2021-12-07
feat: Deploy multiple replicas of configuration-service

### LINKED ISSUE #6229: Configuration Service: PoC of having multiple replicas with mandatory upstream
Currently, the configuration service is deployed as a K8s deployment with one replica with a Persistent Volume attached to it. 
The goal of this task is to evaluate wether we can run multiplpe replicas of the configuration service, where each replica has its own volume - Operating Keptn with this configuration was not possible before because we did not have the limitation of having a mandatory upstream repository: When only the internal git repo hosted by the configuration service is used, requests for adding/retrieving resources will reach different configuration service pods which are out of sync with each other. If however those instances are synced by an upstream repository, having multiple replicas should work.

For this purpose, the following tasks need to be done:

Tasks:

- [ ] Deploy configuration-service as a StatefulSet where each replica has its own PV
- [x] Verify if everything works as expected with one replica
- [x] Verify if everything works as expected with multiple replicas of the configuration service and a mandatory upstream git repository
- [x] Document results and roadblocks that are encountered
- [x] Document where we need to adapt to these changes (docs, integration tests, etc.)

odubajDT:
## This PR

- adds this new feature

### Related Issues

Fixes #6229


odubajDT:
Code already merged as a part of other PR

PR #6312, opened 2021-12-08
feat(installer): Allow API token to be pulled from pre-defined secret

Im5tu:
## This PR

- [Helm Charts] Adds the ability to use an existing secret instead of always generating a secret

### Notes

- How do you usually test the charts? I used `helm template .` as shown in the how to test section, but I'm not sure if there is something a little better to do?
- My helm knowledge is pretty basic, so please thoroughly review - some additional testing with someone with proper helm knowledge would be great if possible

[Slack discussion](https://keptn.slack.com/archives/CNRCGFU3U/p1638899153337500):

image

### How to test

- Main chart: `helm template . --set control-plane.apiService.existingSecret=test` && `helm template .`
- Helm-service/jmeter-service charts: ` helm template . --set remoteControlPlane.enabled=true --set remoteControlPlane.api.existingSecret=test --set remoteControlPlane.api.hostname=test --debug` && ` helm template . --set remoteControlPlane.enabled=true --set remoteControlPlane.api.token=123456 --set remoteControlPlane.api.hostname=test --debug`


oleg-nenashev (commented):
Generally it looks good but I am not sure about the variable name. I would suggest `tokenName` or `tokenSecretName` or `tokenId`. Would appreciate feedback from other maintainers

Also, a downstream pull request to https://keptn.sh/docs/0.11.x/operate/api_token/ is needed IMHO

Im5tu:
As per your comment I've updated to use `tokenSecretName`. I wonder whether it's worth splitting out the helm charts into a separate repository so they can be iterated on a bit easier and not tied to the main keptn releases.

Agreed on the docs, that said, I think the installation via helm instructions need a bit of work anyway because it says that you need to download the helm charts, when that's not quite true, you can use it directly from argo (for example), like so:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: keptn-core
  namespace: argocd
  finalizers:
    - resources-finalizer.argocd.argoproj.io
spec:
  project: keptn
  source:
    chart: keptn
    repoURL: https://storage.googleapis.com/keptn-installer
    targetRevision: 0.11.2
    helm:
      values: |
        control-plane:
          apiService:
            tokenSecretName: keptn-api-token # When this PR is merged
          bridge:
            secret:
              enabled: true
            installationType: "QUALITY_GATES,CONTINUOUS_OPERATIONS"
          ingress:
            enabled: false
          mongodb:
            global:
              namespaceOverride: keptn
          webhookService:
            enabled: false
       
  destination:
    server: https://kubernetes.default.svc
    namespace: keptn
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
      allowEmpty: true
    syncOptions:
      - Validate=false
      - CreateNamespace=true
      - PrunePropagationPolicy=foreground
      - PruneLast=true
      - ApplyOutOfSyncOnly=true
    retry:
      limit: 5
      backoff:
        duration: 5s
        factor: 2
        maxDuration: 3m
```

Im5tu:
Note to reviewers: For transparency, I haven't managed to get a chance to deploy a chart museum instance yet to my cluster, but appears to template correctly with `helm template . --set XXX` as described in the OP.

oleg-nenashev (approved):
Looks good to me, thanks for the patch!

bacherfl (approved):

jmeter-service/chart/values.schema.json
```diff
@@ -39,9 +38,6 @@
               },
               "apiValidateTls": {
                 "type": "boolean"
-              },
-              "token": {
-                "pattern": "^[A-Za-z0-9-.]{2,63}$"
```
mowies, line 44: It would be great if we could keep that constraint in place. I assume you removed it because the token has the option to not be defined now, right?
I think we can fix that by introducing a [oneOf](https://json-schema.org/understanding-json-schema/reference/combining.html#oneof) constraint in the JSON Schema here and also for the `helm-service`.

helm-service/chart/values.schema.json
```diff
@@ -39,9 +38,6 @@
               },
               "apiValidateTls": {
                 "type": "boolean"
-              },
-              "token": {
-                "pattern": "^[A-Za-z0-9-.]{2,63}$"
```
mowies, line 44: Same as the other JSON Schema file :)

mowies (changes requested):
Thanks for the contribution @Im5tu !
Generally, LGTM but I left one comment. Also, I think this is more a feature than a bugfix tbh :)

jmeter-service/chart/values.schema.json
Im5tu: Yeah exactly this and I couldn't figure out the syntax to make it conditional. I'll have another crack at this using oneOf

Im5tu:
Just to update, managed to figure out how to host a version of the chart on our cluster and noticed that the helm-service connection failed to lookup the known secret. I think I've got the syntax right but if anyone has any ideas about what could be wrong, that'd be great to a new helmer :)

mowies:
Hey @Im5tu !
I think I found the proper solution for our JSON schema problem in the PR.
```
"oneOf": [
  {
    "required": [
      "token"
    ],
    "pattern": "^[A-Za-z0-9-.]{2,63}$"
  },
  {
    "required": [
      "tokenSecretName"
    ]
  }
]
```

I will commit the code onto your branch. Is there anything else still unclear or unresolved here?

Im5tu:
Legend, thank you!

Nothing more from my perspective - unless reviews come back with anything :)

mowies:
Thanks @Im5tu 
Do you want to create a PR in the docs repo to update the new options over there?

mowies (approved):

oleg-nenashev (approved):
The change looks ready to go to me. +1 for creating a downstream documentation patch towards  0.12.0 in https://github.com/keptn/keptn.github.io

bacherfl (approved):

PR #6317, opened 2021-12-09
fix(configuration-service): Fix permission issues for configuration service (#6315)

### LINKED ISSUE #6315: Configuration service: Insufficient permissions on OpenShift
In some Openshift environments we discovered problems regarding the permissions of the configuration service when it tries to configure the git user and email during startup. The reason for that was apparently that it had insufficient permissions for setting the global git config because the global git lock file could not be obtained. By moving the git config setup to the individual project directories, this global lock is not needed anymore.

Also, the previously used `fsGroup: 65532` used in the `securityContext` was not enabled by the Security context constraints in some environments. When changing this value to `1001` (which is the same value used by e.g. the bitnami mongodb chart) the service was able to run properly.

bacherfl:
Closes #6315 

This PR changes the behavior of the configuration service by moving the git config setup to the individual project repositores instead of doing it globally during the startup. This way, we also do not need the `initContainer`, which required enhanced privileges, anymore.
Also, the `fsGroup` used by the configuration service has been changed to `1001` per default (see #6315 for further details).  This `fsGroup` value can be overridden using the `configurationService.fsGroup` option in the installer helm chart.

### DIFF
--- a/configuration-service/common/git.go
+++ b/configuration-service/common/git.go
@@ -24,6 +24,9 @@ var namespace = os.Getenv("POD_NAMESPACE")
 const masterBranch = "master"
 const mainBranch = "main"
 
+const gitKeptnUser = "keptn"
+const gitKeptnEmail = "keptn@keptn.sh"
+
 //go:generate moq -pkg common_mock -skip-ensure -out ./fake/command_executor_mock.go . CommandExecutor
 type CommandExecutor interface {
 	ExecuteCommand(command string, args []string, directory string) (string, error)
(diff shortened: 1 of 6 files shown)


configuration-service/common/git.go
```diff
@@ -24,6 +24,9 @@ var namespace = os.Getenv("POD_NAMESPACE")
 const masterBranch = "master"
 const mainBranch = "main"
 
+const gitKeptnUser = "keptn"
```
thisthat, line 27: would make sense to let these be configurable via helm-charts? 🤔

installer/manifests/keptn/charts/control-plane/templates/core.yaml
```diff
@@ -678,8 +666,7 @@ spec:
               name: configuration-volume
           securityContext:
             runAsNonRoot: true
-            runAsUser: 65532
-            readOnlyRootFilesystem: false # False is necessary because we have to write the git config file
```
thisthat, line 682: I would remove the comment and mark it as `true`

thisthat (approved):
LGTM

PR #6329, opened 2021-12-10
fix(configuration-service): Create tmpdir for unarchiving in /data/config

bacherfl:
During the unarchive operation the configuration service tried to create a tmp directory in the file system root - since it does not have the permissions to write anything there, the temp directory is now created within /data/config


thisthat (approved):

thisthat:
Integration tests are passing: https://github.com/keptn/keptn/runs/4482133787?check_suite_focus=true

PR #6341, opened 2021-12-13
feat(bridge): Login via OpenID (#6076) (#6077)

### LINKED ISSUE #6076: Bridge: Implement OIDC flow
## User Story

As a User, I would like to login in Bridge via my SSO

## Technical details

The discovery URL is set via the `OAUTH_DISCOVERY` env var.  The scope should be set to `openid`. 
The token will be stored in the bridge-backend in the express layer.

## DoD

- [ ] Unit tests
- [ ] Integration tests
- [ ] The Bridge stores the access and id tokens

## Dependencies 

- https://github.com/keptn/keptn/issues/6073

### LINKED ISSUE #6077: Bridge: Token refresh
## Action Item

The Bridge should handle the renewal of tokens.

## Technical details

The Bridge should check for the `exp` claim to verify if it needs to renew the access token. Based on the result of this check, the Bridge decides to refresh its tokens.

This check is done at each request in the Bridge-client.  

Mind that there is the possibility that the user has multiple tabs opens and this could trigger multiple calls of refresh tokens in parallel. 

## DoD

- [ ] Bridge handles the refresh of tokens
- [ ] Unit tests

## Dependencies 

- https://github.com/keptn/keptn/issues/6076

Kirdock:
Closes #6076
Closes #6077 

## This PR
- adds support to login via a custom OpenID provider

### How to test
The following environment variables have to be set
- `OAUTH_ENABLED`: true
- `OAUTH_BASE_URL`: e.g. http://localhost:3000 or http://myBridgeInstallation.com
- `OAUTH_DISCOVERY`: https://api.login.yahoo.com/.well-known/openid-configuration
- `OAUTH_CLIENT_ID`: myClientID
- `OAUTH_CLIENT_SECRET` (optional): myClientSecret
- `OAUTH_ID_TOKEN_ALG` (optional): myAlg (e.g. `ES256`). Default is `RS256`
- `OAUTH_NAME_PROPERTY` (optional): The property of the ID token that identifies the user (e.g. `username`). Default is `name`
image

### DIFF
--- a/bridge/README.md
+++ b/bridge/README.md
@@ -79,12 +79,19 @@ Per default login attempts are throttled to 10 requests within 60 minutes. This 
 To set up a login via OpenID you have to register an application by the identity provider you want in order to get an ID (`CLIENT_ID`) and a secret (`CLIENT_SECRET`).
 After this is done the following environment variables have to be set:
 - `OAUTH_ENABLED` - Flag to enable login via OpenID. To enable it set it to `true`.
-- `OAUTH_BASE_URL` - URL of the bridge (e.g. http://localhost:3000 or https://myBridgeInstallation.com).
+- `OAUTH_BASE_URL` - URL of the bridge (e.g. `http://localhost:3000` or `https://myBridgeInstallation.com`).
 - `OAUTH_DISCOVERY` - Discovery URL of the identity provider (e.g. https://api.login.yahoo.com/.well-known/openid-configuration).
 - `OAUTH_CLIENT_ID` - Client ID.
 - `OAUTH_CLIENT_SECRET` (optional) - Client secret. Some identity providers require using the client secret.
 - `OAUTH_ID_TOKEN_ALG` (optional) - Algorithm that is used to verify the ID token (e.g. `ES256`). Default is `RS256`.
 - `OAUTH_NAME_PROPERTY` (optional) - The property of the ID token that identifies the user. Default is `name` and fallback to `nickname`, `preferred_username` and `email`.
+
+#### Additional information:
+- Make sure you add the redirect URI `https://${yourDomain}/${pathToBridge}/oauth/redirect` to your identity provider.
+- The identity provider has to support the grant types `authorization_code` and `refresh_token` and provide the endpoints `authorization_endpoint`, `token_endpoint` and `jwks_uri`.
+- The refresh of the token is done by the bridge server on demand. 
+- If the identity provider provides the endpoint `end_session_endpoint`, it will be used for the logout.
+- The bridge server itself is a confidential client.
 
 ### Custom Look And Feel
 
--- a/bridge/README.md
+++ b/bridge/README.md
@@ -75,6 +75,17 @@ Per default login attempts are throttled to 10 requests within 60 minutes. This 
 - `REQUESTS_WITHIN_TIME` - how many login attempts in which timespan `REQUEST_TIME_LIMIT` (in minutes) are allowed per IP.
 - `CLEAN_BUCKET_INTERVAL` - the interval (in minutes) the saved login attempts should be checked and deleted if the last request of an IP is older than `REQUEST_TIME_LIMIT` minutes. Default is 60 minutes.
 
+### Setting up login via OpenID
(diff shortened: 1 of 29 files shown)


bridge/server/user/oauth-routes.ts
```diff
@@ -0,0 +1,149 @@
...
+    if (!params.code || !params.state) {
+      return res.redirect(getRootLocation());
+    }
+
+    const verifiers = codeVerifiers[params.state];
+    if (verifiers) {
+      delete codeVerifiers[params.state];
+      params.state = undefined;
+    } else {
+      return res.render('error', {
```
corneliarauch, line 62: In case ao an invalid state, I think it should be 401 response. e.g.: https://developers.google.com/identity/protocols/oauth2/openid-connect#confirmxsrftoken

bridge/server/user/oauth-routes.ts
```diff
@@ -0,0 +1,149 @@
...
+        title: 'Error',
+        message: 'Error while handling request. State does not exist.',
+        location: getRootLocation(),
+      });
+    }
+
+    try {
+      const tokenSet = await client.callback(redirectUri, params, {
+        code_verifier: verifiers.codeVerifier,
+        nonce: verifiers.nonce,
```
corneliarauch, line 72: shouldn't state also be checked

bridge/server/user/oauth.ts
```diff
@@ -1,225 +1,86 @@
...
+  // Initializing OAuth middleware.
+  app.use(oauthRouter(client, redirectUri, reduceRefreshDateSeconds));
+  // Authentication filter for API requests
+  app.use('/api', async (req: Request, resp: Response, next: NextFunction) => {
+    if (!isAuthenticated(req.session)) {
+      return next({ response: { status: 401 } });
+    } else {
+      const tokenSet = new TokenSet(req.session.tokenSet);
+      if (tokenSet.expired()) {
+        refreshPromises[req.session.id] ??= client.refresh(tokenSet).then((token) => {
```
corneliarauch, line 61: This is a very elegant solution. Only downside req.session.save can happen multiple times but this should not be a big issue. I think this is working fine if you have a single instance - one you have multiple instances(load balance) you might run into issues here.

bridge/server/user/oauth-routes.ts
Kirdock: `state` is checked before to get the `codeVerifier` and `nonce`. If the state is invalid the object for the identifiers is `undefined` and fails before the callback is being executed

bridge/server/user/oauth-routes.ts
Kirdock: We decided to stay with the error page. Text is adjusted

bridge/server/user/session.ts
```diff
@@ -72,15 +72,21 @@ function authenticateSession(req: Request, tokenSet: TokenSet): Promise<void> {
   return new Promise((resolve) => {
     // Regenerate session for the successful login
     req.session.regenerate(() => {
+      const userIdentifier = process.env.OAUTH_NAME_PROPERTY || 'name';
```
corneliarauch, line 82: || 'name' is also in getUserIdentifier

bridge/README.md
```diff
@@ -75,6 +75,17 @@ Per default login attempts are throttled to 10 requests within 60 minutes. This
 - `REQUESTS_WITHIN_TIME` - how many login attempts in which timespan `REQUEST_TIME_LIMIT` (in minutes) are allowed per IP.
 - `CLEAN_BUCKET_INTERVAL` - the interval (in minutes) the saved login attempts should be checked and deleted if the last request of an IP is older than `REQUEST_TIME_LIMIT` minutes. Default is 60 minutes.
 
+### Setting up login via OpenID
+To set up a login via OpenID you have to register an application by the identity provider you want in order to get an ID (`CLIENT_ID`) and a secret (`CLIENT_SECRET`).
```
corneliarauch, line 79: I am not sure but I think it would be helpful to add some additional information here. E.g.: for which grant types/flows is this working. Refresh token enabled/disabled. Is this a public or confidential client a.s.o.

thisthat (approved):

PR #6350, opened 2021-12-14
feat(bridge): Use textarea for webhook url config and adapt styles (#5706)

### LINKED ISSUE #5706: Bridge: Webhook Integration: Increase text-field for entering/editing URLs
Right now with Keptn 0.10.0-next.1 when I enter a URL, the text field is rather small. As URLs are usually a lot longer, it's hard to use it.

image

For reference, this is my URL: 
```
http://jenkins.<REDACTED>.nip.io/job/Keptn_test_finished_pipeline/buildWithParameters?token=<REDACTED>&shkeptncontext={{.shkeptncontext}}&triggeredid={{.triggeredid}}&Project={{.data.project}}&Service={{.data.service}}&Stage={{.data.stage}}
```

Same is true for slack integration, e.g.:

I'm aware that passing all those params via query parameters is not the best way (we should really make use of the payload field below and use a POST request), but please consider that certain 3rd-party services will need this.

**To be considered**

* Try to use as much real-estate as possible for this field (the field could dynamically grow to the right, if there is space)
* Allow the field to be resized (e.g., by using `<textarea>`)

laneli:
Closes #5706 

* Subscription and webhook forms are now a bit cleaner and more aligned
* Webhook URL is a textarea with standard row size 1 and can be expanded
* The buttons for payload and secret variables are now in the header of both textareas because it would otherwise overlap the content and the expand handle

image

### DIFF
--- a/bridge/client/app/_components/ktb-modify-uniform-subscription/ktb-modify-uniform-subscription.component.html
+++ b/bridge/client/app/_components/ktb-modify-uniform-subscription/ktb-modify-uniform-subscription.component.html
@@ -3,24 +3,26 @@
   <ng-container *ngIf="data$ | async as data; else loading">
     <h2>{{ editMode ? 'Edit' : 'Create' }} subscription</h2>
     <form [formGroup]="subscriptionForm">
-      <dt-form-field *ngIf="!isWebhookService" uitestid="edit-subscription-field-isGlobal">
-        <ng-container [ngTemplateOutlet]="data.subscription.hasFilter() ? showOverlay : isGlobalCheckbox">
-        </ng-container>
+      <div class="settings-section">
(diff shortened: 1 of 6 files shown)


Kirdock (changes requested):
Could you add some padding here for the "Show example payload"-button?
image

Should we show the event- and secret picker for the custom headers above the input field like the textareas (to align it with the style of the others)?
image

laneli:
Adaptions:

image

image

Kirdock (changes requested):
One additional minor change:
Could you add a min-height to the textareas? The scrollbar should not show up if the height is set to minimum.
image

Kirdock (approved):

PR #6351, opened 2021-12-14
fix(cli): Print error message if service does not exist with `trigger delivery`

### LINKED ISSUE #6288: Able to `trigger delivery` with incorrect/non-existing service name
When triggering a sequence e.g. `keptn trigger delivery --project my-project --service my-misspelled-service ...` keptn validates the existence of project `my-project` but is **not** checking whether service `my-misspelled-service` was created.

**To Reproduce**
Trigger a sequence (e.g. delivery) with a service name for which no service was created.

**Expected behavior**
The sequence is not launched.
The user is presented with an error message like it is the case for non-existing projects.

**Current behavior**
The sequence is launched but fails later on, e.g. when the helm service wants to deploy the artifact for the service.

odubajDT:

## This PR

- fixed triggered delivery when using non-existing service

### Related Issues

Fixes #6288

### DIFF
--- a/cli/cmd/helpers.go
+++ b/cli/cmd/helpers.go
@@ -130,9 +130,9 @@ func CheckEndpointStatus(endPoint string) error {
 	return nil
 }
 
-func ServiceInSlice(a string, list []*apimodels.Service) bool {
-	for _, b := range list {
-		if b.ServiceName == a {
+func ServiceInSlice(service string, serviceList []*apimodels.Service) bool {
+	for _, s := range serviceList {
+		if s.ServiceName == service {
 			return true
 		}
 	}
--- a/cli/cmd/helpers.go
+++ b/cli/cmd/helpers.go
@@ -4,16 +4,18 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	models "github.com/keptn/go-utils/pkg/api/models"
-	apiutils "github.com/keptn/go-utils/pkg/api/utils"
-	"github.com/keptn/keptn/cli/pkg/logging"
-	"github.com/spf13/cobra"
-	"gopkg.in/yaml.v3"
 	"io"
 	"math"
 	"net/http"
 	"strings"
 	"time"
+
+	apimodels "github.com/keptn/go-utils/pkg/api/models"
+	models "github.com/keptn/go-utils/pkg/api/models"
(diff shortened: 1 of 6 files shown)


cli/cmd/helpers.go
```diff
@@ -127,3 +129,12 @@ func CheckEndpointStatus(endPoint string) error {
 
 	return nil
 }
+
+func ServiceInSlice(a string, list []*apimodels.Service) bool {
```
thisthat, line 133:
```suggestion
func ServiceInSlice(service string, serviceList []*apimodels.Service) bool {
```

cli/cmd/helpers.go
```diff
@@ -127,3 +129,12 @@ func CheckEndpointStatus(endPoint string) error {
...
+	for _, b := range list {
```
thisthat, line 134:
```suggestion
	for _, s := range serviceList {
```

cli/cmd/trigger_delivery_test.go
```diff
@@ -206,3 +206,65 @@ func TestCheckImageNonAvailabilityD(t *testing.T) {
...
+				defer r.Body.Close()
+				w.Write([]byte(metadataMockResponse))
+				return
+			}
+			return
+		}),
+	)
+	defer ts.Close()
+
+	os.Setenv("MOCK_SERVER", ts.URL)
```
thisthat, line 261: is this necessary?

odubajDT: yes, we get errors without it.

thisthat: what errors do you get?

odubajDT: I will probably re-write the test to be more general :) Hopefully it will be better afterwards :)

odubajDT: Yeah and actually for a heads-up: I get segmentation fault regarding some missig variable/nil variable. Did not debug it until now but I can look at it afterwards.

odubajDT: I have generalized the test, but I am receiving failures for this test on github, locally works fine. Seems to me like some kind of infrastructure problem, debugging it.

bacherfl (approved):

Issue #5526, opened 2021-10-06
Introduce waiting status in /sequence endpoint


ermin-muratovic:
The `/sequence` endpoint returns a `state` property and each stage has also a `state` property. Currently the overall state might be `triggered`, `paused` or `finished`. However, the stage state is either `triggered` or `finished`, see example below.
The `state` value should also return `started` for started stages. 

```
{
  "name": "delivery",
  "service": "carts",
  "project": "sockshop",
  "time": "2021-10-06T09:28:50.796Z",
  "shkeptncontext": "ae02c298-e6ab-4453-87c6-492eea4fdca9",
  "state": "paused",
  "stages": [
    {
      "name": "dev",
      "image": "docker.io/keptnexamples/carts:0.12.1",
      "state": "triggered",
      "latestEvent": {
        "type": "sh.keptn.event.test.triggered",
        "id": "ba87752d-3a99-4388-84b0-08f988322a9a",
        "time": "2021-10-06T09:30:19.744Z"
      }
    }
  ]
}'
```

If the user paused a sequence, I would expect to see the state `paused` instead of the state `triggered`.
If the sequence is waiting for another sequence to be finished, I would expect to see the state `waiting` instead of the state `triggered`.

This is a breaking change. 

## DoD

- [ ] Unit tests
- [ ] Integration tests

thisthat:
We should define the list of possible states before we can tackle this issue

odubajDT:
From discussion we know that states of sequences can have these values: 

- triggered
- started
- finished
- paused
- timedOut
- aborted

Stages have also the same states as sequences + succeeded

It should be enough to introduce a new state `waiting` to cover when the sequence is waiting for a condition (f. e. another sequence). 

#Use Case
When a sequence (in state `started`) needs to wait, it changes its state to `waiting`. The same happens to each stage with state `started`

PR #6388, opened 2021-12-16
feat(bridge): Read problem title from GET /sequence endpoint

### LINKED ISSUE #4206: Add problem title of remediation to sequence endpoint

ermin-muratovic:
## This PR
- makes sure that the Bridge reads the problem title from GET /sequence endpoint and therefore it doesn't need to load the traces of the sequence anymore, only if also the actions are needed (e.g. environments screen)

### Related Issues
Fixes #4206 

### Note
A remediation should not end up in `waiting` state when the action is finished, but it is waiting for the evaluation task. Therefore the `isWaiting` check ignores now remediation sequences. This will be improved with #5526

### DIFF
--- a/bridge/client/app/_models/sequence.ts
+++ b/bridge/client/app/_models/sequence.ts
@@ -139,6 +139,10 @@ export class Sequence extends sq {
   }
 
   public isWaiting(): boolean {
+    if (this.isRemediation()) {
+      return false;
+    }
+
     const lastStageName = this.getLastStage();
     if (lastStageName && this.state === SequenceState.STARTED) {
       const lastStage = this.getStage(lastStageName);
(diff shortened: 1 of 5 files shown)


Kirdock (changes requested):
The bridge-server tests are currently failing (wrong exit code for bridge-server tests will be fixed in another PR)

Kirdock (approved):

PR #6457, opened 2022-01-03
fix(configuration-service): Make check for helm chart archives more strict (#6447)

### LINKED ISSUE #6447: Weird behaviour when adding resources in helm directory that contains "tgz" in filename
I ran into a weird behavior while trying out some stuff with job-executor-service. 

Essentially adding a single file with the right name (e.g., `foo-helm-bar/.tgz.txt`) can break the API / existing integrations that use the API.

# To Reproduce 

I added files like this:

```
keptn add-resource --project=$PROJECT --service=helloserver --stage=production --resource=./helm/helloserver.tgz --resourceUri=helm/helloserver.tgza
```

First, I query the Keptn API to get a list of all services resources:

```
	// get it from KeptnBase
	requestedResources, err := k.resourceHandler.GetAllServiceResources(k.project, k.stage, k.service)
	if err != nil {
		return nil, fmt.Errorf("resources not found: %s", err)
	}

```

Then I iterate over the list and fetch each file:

```
	keptnResources := make(map[string][]byte)
	for _, serviceResource := range requestedResources {
		log.Printf("Trying to fetch %s\n", *serviceResource.ResourceURI)
		// match against with and without starting slash
		resourceURIWithoutSlash := strings.Replace(*serviceResource.ResourceURI, "/", "", 1)
		if strings.HasPrefix(*serviceResource.ResourceURI, resource) || strings.HasPrefix(resourceURIWithoutSlash, resource) {
			keptnResourceContent, err := k.GetKeptnResource(fs, *serviceResource.ResourceURI)
			if err != nil {
                                log.Printf("Error: %s\n", err.Error())
				return nil, fmt.Errorf("could not find file %s: %s", *serviceResource.ResourceURI, err.Error())
			}
			keptnResources[*serviceResource.ResourceURI] = keptnResourceContent
		}
	}
```

# Current behavior

Fetching the file `helm/helloserver.tgza` fails.

The log output looks like this:
```
Managed to get a list of 11 resources
Trying to fetch /helm/endpoints.yaml
Trying to fetch /helm/helloserver/Chart.yaml
Trying to fetch /helm/helloserver/templates/deployment.yaml
Trying to fetch /helm/helloserver/templates/service.yaml
Trying to fetch /helm/helloserver/values.yaml
Trying to fetch /helm/helloserver.tgza
Error: resource not found: /helm/helloserver.tgza - Resource not found
```

This is how the respective git repo looks like:
image

Here is the log output of configuration-service:
```
{"timestamp":"2021-12-29T12:21:06.799363022Z","logLevel":"DEBUG","message":"Checking out production branch"}
{"timestamp":"2021-12-29T12:21:06.897785592Z","logLevel":"DEBUG","message":"Archive the Helm chart: %2Fhelm%2Fhelloserver.tgza"}
```

I agree that my filename is weird, but Keptn should really just treat it as a binary file that needs to be delivered.

# Expected behavior

Keptn configuration-service should just return the binary file content, and not try to archive/unarchive it.

# Analysis

Looking at

https://github.com/keptn/keptn/blob/220f1edbf86fd2a0a22654bcc507d709c51bca9f/configuration-service/handlers/service_resource.go#L79-L93

It seems that we automatically assume that anything with `helm` in its path (could also be `foohelmbar`) and `tgz` in its filename is a .tgz archive that should be archived/unarchived.

This check is rather broad, and I would like to see a more specific check (e.g., the actual file ending needs to be .tgz *and* it is in the `helm` folder).

bacherfl:
Closes #6447

### DIFF
--- a/configuration-service/handlers/service_resource.go
+++ b/configuration-service/handlers/service_resource.go
@@ -77,7 +77,7 @@ func GetProjectProjectNameStageStageNameServiceServiceNameResourceResourceURIHan
 	}
 
 	// archive the Helm chart
-	if strings.Contains(resourcePath, "helm") && strings.Contains(params.ResourceURI, ".tgz") {
+	if isHelmChart(resourcePath) {
 		logger.Debug("Archive the Helm chart: " + params.ResourceURI)
 
 		chartDir := strings.Replace(resourcePath, ".tgz", "", -1)
@@ -105,8 +105,8 @@ func GetProjectProjectNameStageStageNameServiceServiceNameResourceResourceURIHan
 			WithPayload(&models.Error{Code: 500, Message: swag.String("Could not read file")})
 	}
 
-	// remove Helch chart .tgz file
-	if strings.Contains(resourcePath, "helm") && strings.HasSuffix(params.ResourceURI, ".tgz") {
+	// remove Helm chart .tgz file
+	if isHelmChart(resourcePath) {
 		logger.Debug("Remove the Helm chart: " + params.ResourceURI)
 
 		if err := os.Remove(resourcePath); err != nil {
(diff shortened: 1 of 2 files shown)


odubajDT (approved):
LGTM

Issue #6346, opened 2021-12-14
Configuration Service: Implement new version 


bacherfl:
The goal of this ticket is to create a new implementation of the configuration service, based on the results of the following research tickets:

- https://github.com/keptn/keptn/issues/6229
- https://github.com/keptn/keptn/issues/6230
- https://github.com/keptn/keptn/issues/6231
- https://github.com/keptn/keptn/issues/6232

The following aspects should be considered when implementing the new service:

- Implementation in gin-gonic
- Interaction with git via the `go-git` library https://github.com/go-git/go-git - At least as much as possible, one limitation of the library is that it cannot handle non-fast-forward updates during a pull. In this case we need to fall back to the CLI
- Prepare the implementation to support multiple data models (See #6397)
- The service needs to be able to run as non-root 
- Instead of using a persistent volume, the service should use an emptyDir - the content of the directory will not remain after a restart, but since we now have a mandatory upstream, it can always clone the latest state of the repo if the project's directory is not there yet - See https://github.com/keptn/keptn/pull/6301 for a first implementation.
- The service needs to be able to run with multiple replicas without causing concurrency issues when multiple instances are pushing to the same upstream - https://github.com/keptn/keptn/pull/6301 contains a retry loop for pushing changes that tries to pull the latest state again if the previous push was rejected due to changes pushed by other instances.
- For requesting resources, the service should provide the possibility to fetch resources for a certain commit ID - this way, if the commit ID is already in the local history, it does not need to bother the upstream repository. Only if the commit is not available yet, the latest state from the upstream will need to be pulled. A first version of that approach is implemented in the draft PR https://github.com/keptn/keptn/pull/6330

Note: This new implementation of the configuration-service will live next to the current implementation of the configuration-service within our repository, because we will provide the option to decide which implementation should be deployed - See https://github.com/keptn/keptn/issues/6336.

## DoD

- [ ] Service implemented
- [ ] Unit test coverage ~75%
- [ ] Integration tests

## Resources

- https://github.com/libgit2/git2go (Go bindings to libgit2)

PR #6517, opened 2022-01-05
feat: Resource service first working version (#6346)

bacherfl:
Part of #6346
This PR introduces a first functional version of the resource service. Currently, some checks in the integration tests are not passing, but that is mainly due to http response code mismatches, which should be easy to fix in smaller follow up PRs


RealAnna (approved):
LGTM

PR #6526, opened 2022-01-10
feat(bridge): Show dialog to prevent data loss on unsaved form

### LINKED ISSUE #4932: Bridge: Add dialog in order to not loose data on create project
## User Story
As a user, I want to get informed when navigating back or closing the window when creating a project, to not lose any data, especially the shipyard configuration.

### Description
While creating a project the user should be able to configure a shipyard file with https://github.com/keptn/keptn/issues/4508. As the user has a lot of data to lose, if he misclicks or closes the window, it makes sense to prevent this.
We should therefore show a dialogue, that informs him that he will lose data in case of leaving the page. If he has already filled in the required fields he should also be able to save the data at this point. Otherwise, he/she can just confirm that he/she wants to leave.

### Acceptance criteria
- [ ] Show a notification/dialogue with information about losing data when the user has configured the project and tries to leave
(leaving: navigating to another page per click, navigating back, closing the tab/window)
- [ ] Provide a means to save the data from the notification when all required fields are filled in
- [ ] When not all required fields are filled in, just inform about losing data
- [ ] When the user confirms that data will be lost, execute expected action (navigate away, back or closing the window)

### Definition of done
- [ ]  Everything in acceptance criteria implemented and manually tested
- [ ]  Added UI Test

ermin-muratovic:
## This PR
- makes sure that a user is notified when leaving the project settings form

image

### Related Issues
Fixes #4932


bridge/client/app/_components/ktb-project-settings/ktb-project-settings.component.ts
```diff
@@ -248,11 +252,53 @@ export class KtbProjectSettingsComponent implements OnInit, OnDestroy {
...
+  }
+
+  canDeactivate(): boolean | Observable<boolean> {
+    return this.isProjectFormUntouched;
+  }
+
+  // @HostListener allows us to also guard against browser refresh, close, etc.
+  @HostListener('window:beforeunload', ['$event'])
+  showNotification($event?: BeforeUnloadEvent): void {
+    if (this.canDeactivate() !== true) {
```
Kirdock, line 293: If it's an `Observable` it's always `false`. The content of the observable is not checked here.
In my opinion, I would not even use `Observables` here

bridge/client/app/_guards/pending-changes.guard.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    expect(guard.canDeactivate(mockComponent)).toBeTruthy();
+  });
+
+  it('will route if guarded and user accepted the dialog', () => {
+    // Mock the behavior of the GuardedComponent:
+    mockComponent.formTouched();
+    const canDeactivate$ = <Observable<boolean>>guard.canDeactivate(mockComponent);
+    canDeactivate$.subscribe((deactivate) => {
+      // This is the real test!
+      expect(deactivate).toBeTruthy();
```
Kirdock, line 63: Is this even called? Because this is asynchronous I would expect that the test runs until `mockComponent.saveForm()` and finishes without waiting for the subscribe without the use of `waitForAsync`

bridge/client/app/_guards/pending-changes.guard.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    mockComponent.saveForm();
+  });
+
+  it('will not route if guarded and user rejected the dialog', () => {
+    // Mock the behavior of the GuardedComponent:
+    mockComponent.formTouched();
+    const canDeactivate$ = <Observable<boolean>>guard.canDeactivate(mockComponent);
+    canDeactivate$.subscribe((deactivate) => {
+      // This is the real test!
+      expect(deactivate).toBeFalsy();
```
Kirdock, line 75: here same as above

bridge/client/app/_guards/pending-changes.guard.ts
```diff
@@ -0,0 +1,24 @@
...
+  canDeactivate: () => boolean | Observable<boolean>;
+  showNotification: ($event?: BeforeUnloadEvent) => void;
+}
+
+@Injectable({
+  providedIn: 'root',
+})
+export class PendingChangesGuard implements CanDeactivate<PendingChangesComponent> {
+  canDeactivate(component: PendingChangesComponent): boolean | Observable<boolean> {
+    const hasPendingChanges = component.canDeactivate() !== true;
```
Kirdock, line 15: If it's an `Observable` it's always `false`.

Kirdock (changes requested):
- When a project is created, the notification about unsaved changes pops up and interrupts the redirect to the settings page of the created project.
- The unsaved changes dialog should have a pulse or shaking effect when the user tries to leave the page. To make it clearer what he has to do or has to look for.
- The project in the header can still be changed when there are unsaved changes. This leads to a wrong redirect after discard or save.
- "Discard changes" also redirects. Either change it to "Leave anyway" or "Discard changes and leave" or remove the redirect.

bridge/client/app/_components/ktb-project-settings/ktb-project-settings.component.ts
ermin-muratovic: The `Observable` allows the user to decide to stay or leave the page.

bridge/client/app/_guards/pending-changes.guard.ts
ermin-muratovic: But if no changes are applied to the Form it is set to `true`.

bridge/client/app/_guards/pending-changes.guard.spec.ts
ermin-muratovic: This simulates the user deciding to leave the page.

bridge/client/app/_guards/pending-changes.guard.spec.ts
ermin-muratovic: This simulates the user deciding to stay on the page, which results in deactivate to be no longer an `Observable` but `false`.

bridge/client/app/_guards/pending-changes.guard.ts
Kirdock: Yes, because the value is `boolean` then, but on change it is set to `Observable<boolean>` and `of(true) !== true` and `of(false) !== true`. It would be the same if `false` is set instead of assigning an observable to it

bridge/client/app/_guards/pending-changes.guard.spec.ts
Kirdock: I'm talking about the asynchronous `subscribe`. `subscribe` would be called right after `mockComponent.saveForm();`, but after this, Jest will tell that the test succeeded, no matter what's inside the `subscribe`, because it does not wait for ongoing asynchronous tasks.
No matter if `expect(deactivate).toBeTruthy();` or `expect(deactivate).toBeFalsy();` is called inside the subscribe, the test will pass.

ermin-muratovic:
> * When a project is created, the notification about unsaved changes pops up and interrupts the redirect to the settings page of the created project.
I fixed this.
> * The unsaved changes dialog should have a pulse or shaking effect when the user tries to leave the page. To make it clearer what he has to do or has to look for.
The only way how we can add a shake effect is using following "ugly" code:
`document.querySelector('div[aria-label="Dialog for notifying about unsaved data"]').classList.add('shake');`
I added it, but do you have any other ideas?
> * The project in the header can still be changed when there are unsaved changes. This leads to a wrong redirect after discard or save.
When changing the project and choosing to leave the page the redirect works fine, but if you choose to stay the dropdown will still have the changed value. Not sure how to "revert" the click. Do you have any idea?
> * "Discard changes" also redirects. Either change it to "Leave anyway" or "Discard changes and leave" or remove the redirect.
Actually I wanted to allow the user 3 options, "Stay", "Discard changes and leave" and "Save changes". I updated this.

bridge/client/app/_guards/pending-changes.guard.spec.ts
ermin-muratovic: You are right. I added `waitForAsync`

bridge/client/app/_guards/pending-changes.guard.ts
ermin-muratovic: Yes, but here the `hasPendingChanges` is just used to show the Notification. But the `Observable<boolean>` is also returned by the Guard, so the Route change can be triggered after e.g. saving the form or declined by the user. Without `Observable` this will not work.

Kirdock (changes requested):
There are still two bugs
- If you try to leave the page during the project creation and click "save changes", nothing is done.
- If you edit the project settings (without saving) and then delete the project, it is redirected to the wrong URL. For example, deleting the "test" project redirects then to  `/project/test/settings/project`

Kirdock (approved):

Issue #6085, opened 2021-11-22
CLI: Token refresh


thisthat:
## Action Item

If the CLI detects there are access and refresh tokens, it should handle the renewal of them. 

## Technical details

The CLI should check for the `exp` claim to verify if it needs to renew the access token. Based on the result of this check, the CLI decides to refresh its tokens.

This check is done at each request in the CLI.  

## DoD

- [ ] CLI handles the refresh of tokens
- [ ] Unit tests

## Dependencies 

- https://github.com/keptn/keptn/issues/6083

Issue #6084, opened 2021-11-22
CLI: Send the access token as Bearer header


thisthat:
## Action Item

If the CLI detects there are access and refresh tokens, it shall send the access token as an HTTP header to every Keptn API request.

## Technical details
The token is available in the local filesystem #6083
The HTTP header will be of the following format: `Authorization: Bearer <access-token>`.
The `x-token` header is not necessary if we logged in via SSO. 

## DoD

- [ ] Unit tests

Kavindu-Dodan:
Given that CLI is installed locally and is public, the CLI need a mechanism to obtain required parameters to build OAuth requests. In general this is authorization server URLs and OAuth client credentials.

Given that we follow the bearer token specification ([RFC6750](https://datatracker.ietf.org/doc/html/rfc6750)), CLI can obtain the Authorization server and Public client details as claims in WWW-Authenticate Response Header [1]

For example, Keptn core must respond unauthorized API requests with following payload,

`HTTP/1.1 401 Unauthorized
WWW-Authenticate: error="no_token", issuer=<issuer_discovery>, client =<client_id>` 

Which will trigger a reauthentication in CLI. 

Further, CLI should be able to utilize a `localhost` redirect url if authorization code flow [2] is used to obtain the access token. 

[1] https://datatracker.ietf.org/doc/html/rfc6750#section-3
[2] https://datatracker.ietf.org/doc/html/rfc6749#section-4.1

PR #6549, opened 2022-01-11
feat(cli): SSO integration

### LINKED ISSUE #6083: CLI: Add SSO flag to Auth command
## Action Item

The CLI Auth command needs a new `--sso` flag to allow a user to log in a user via SSO.

## Technical details

The CLI shall open a local HTTP server to receive tokens from the OIDC server via callback.  
Then, it will open the browser to redirect the user to the SSO link provided so it can log in.
The scope is set to `openid`.
When the callback is fired, the CLI will retrieve the access and refresh tokens and store them locally. 

Please also consider the following suggestion: https://github.com/keptn/keptn/issues/6084#issuecomment-979321777

## DoD

- [ ] Unit tests

warber:
This PR adds the possibility to use SSO with the Keptn CLI.
The auth command support multiple new flags:
To authenticate the CLI using SSO/OAuth2 the CLI needs to be called like this:

`keptn auth --sso --sso-client-id UserClientID --sso-discovery http://discovery.com/.well-known`

Oauth information such as tokens, sso discovery information or user information is stored under the keptn configuration directory (`~/.keptn`)

In order to disable SSO authentication, the auth command the `sso-logout` command can be used: 

`keptn auth --sso-logout`

The access token will automatically be renewed once it is expired using the refresh token retrieved from the oauth provider.
Once the refresh token is invalid, the user will need to login again using SSO by execution the `keptn auth --sso ....` command mentioned above.

Closes #6083 #6084  #6085 

**TODO** items (follow up tickets will be created):
-  [ ] add a timeout which causes the cli to exit when the user takes too long to enter credentials in the browser
- [ ] cleanup version calls which might be executed (and fail) because of expired tokens
- [ ] exlude x-token when sso is used


cli/cmd/get_event_approvalTriggered_test.go
```diff
@@ -60,7 +60,7 @@ const eventsResponse = `{
 
 func Test_getApprovalTriggeredEvents(t *testing.T) {
 
-	mocking = true
+	//mocking = true
```
bacherfl, line 63: let's remove that commented line

cli/cmd/get_event_test.go
```diff
@@ -14,7 +14,7 @@ func init() {
 }
 
 func TestGetTriggeredEvent(t *testing.T) {
-	mocking = true
+	//mocking = true
```
bacherfl, line 17: let's remove that commented line

bacherfl (approved):

Issue #5101, opened 2021-08-31
Update sequence screen data model


thisthat:
## Action Item

The polling result of the project is not used in every view and some views only require one information of the project instead of the whole one.

### Implementation details

Improvements suggested by @laneli:

- Load all sequences once and remove polling - finished sequences do not have to be updated
- Poll only unfinished sequences
- Return `Observable<Sequence[]>` instead of writing it into project
- For approval - endpoint or method to get evaluation result - this is also stored in project
- Sequences filter - is currently filtered from all sequences, we can keep it or just return the default sequences 

Improvements suggested by @Kirdock:
- Don't send another call to the metadata when the sequence list is updated. Just retrieving that in an interval is fine. On sequences update, we just need to update the current metadata with the sequences we have.

--- 

### DoD
- [ ] Unit tests
- [ ] Add some UI tests for the screen
- [ ] Introduce a new endpoint in the bridge-server that returns only the information needed.

PR #6558, opened 2022-01-12
feat(bridge): Poll sequence metadata for filters and deployed artifacts (#5246)

### LINKED ISSUE #5246: Bridge: Improve polling mechanism for sequences
## Action Item

Improve the polling mechanism for sequences. 

### Technical Details

Sequence Screen: Only requires the latest deployment of a service in a stage. This information is used for the approval-trace. There the currently deployed artifact is shown. 

## DoD 

- [ ] Improved polling logic

laneli:
Closes #5246 

Instead of getting the info about a deployment of the project it is now put into a separate call in bridge-server where metadata like last deployed image and information for filters are returned.

Additional improvements for the sequence screen are proposed in: #5101


Kirdock (changes requested):

bridge/cypress/integration/sequences.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      });
+    });
+
+    cy.visit('/project/sockshop/sequence');
+
+    cy.byTestId('keptn-loadingSequences').should('exist');
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    sendResponse();
```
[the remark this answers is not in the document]
laneli: https://github.com/cypress-io/cypress/issues/3931

It seems to be recommended to just use wait

Kirdock (changes requested):
Currently, the filter is reset every time the metadata is fetched/updated.

Kirdock (changes requested):
- Now deleted services are added to the filter if they appear in the sequence list, but if "Show older sequences" is clicked and loads older sequences, the filter is not adjusted immediately, only after polling new sequences.
- Under some conditions, if a service in the filter is selected, the filter still resets. Can't really determine the root cause. I'm loading older sequences, there is a deleted service in the service filter and the one that is not deleted is selected.

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.html
```diff
@@ -35,6 +35,7 @@
             [events]="getFilteredSequences(project.sequences)"
             [selectedEvent]="currentSequence"
             (selectedEventChange)="selectSequence($event)"
+            (loadOldSequencesClicked)="loadSequenceMetadata(project.projectName)"
```
Kirdock, line 38: Actually, we don't need this. The component listens to `this.dataService.sequencesUpdated` and so does `ktb-sequence-view.component`. Additionally, I wouldn't even send another call to the metadata. Just retrieving that in an interval is fine. We just need to update the current metadata with the sequences we have.

bridge/server/api/index.ts
```diff
@@ -171,6 +171,15 @@ const apiRouter = (params: {
     }
   });
 
+  router.get('/project/:projectName/sequences/metadata', async (req: Request, res: Response, next: NextFunction) => {
```
Kirdock, line 174: Could you add some tests for this endpoint?

Kirdock (changes requested):

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.html
laneli: I keep it as it is

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.spec.ts
```diff
@@ -31,142 +77,130 @@ describe('KtbEventsListComponent', () => {
...
+    // when
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore // Ignore private property
+    component.mapServiceFilters(metadata);
 
     // then
-    expect(loadingIndicator).toBeFalsy();
-    expect(emptyStateContainer).toBeFalsy();
-    expect(sequenceList).toBeTruthy();
+    expect(component._seqFilters).toEqual(activeFilter);
```
Kirdock, line 148: Both are referencing the same array.
`filter[0][0].autocomplete = filter[0][0].autocomplete.slice(0, 1);` also touches the original one.
No matter how the filter will be changed, it will be as the same as `activeFilter`

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.spec.ts
```diff
@@ -31,142 +77,130 @@ describe('KtbEventsListComponent', () => {
...
+    expect(component._seqFilters).toEqual(filter);
+
+    // when
+    metadata.filter.services.pop();
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore // Ignore private property
+    component.mapServiceFilters(metadata);
+
+    // then
+    expect(component._seqFilters).toEqual(activeFilter);
```
Kirdock, line 180: Same here

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.spec.ts
```diff
@@ -31,142 +77,130 @@ describe('KtbEventsListComponent', () => {
...
-    fixture.detectChanges();
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore // Ignore private property
+    component.mapServiceFilters(metadata);
 
     // then
-    const loadingIndicator = fixture.nativeElement.querySelector('[uitestid=keptn-loadingSequences]');
-    const emptyStateFilteredContainer = fixture.nativeElement.querySelector('[uitestid=keptn-noSequencesFiltered]');
-    const sequenceList = fixture.nativeElement.querySelector('[uitestid=keptn-sequence-view-roots]');
+    expect(component._seqFilters).toEqual(filter);
```
Kirdock, line 171: Same here

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.spec.ts
```diff
@@ -31,142 +77,130 @@ describe('KtbEventsListComponent', () => {
...
+    // when
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore // Ignore private property
+    component.mapServiceFilters(SequenceMetadataMock);
 
-    // then
-    expect(loadingIndicator).toBeFalsy();
-    expect(emptyStateContainer).toBeTruthy();
-    expect(sequenceList).toBeFalsy();
+    expect(component._seqFilters).toEqual(activeFilter);
```
Kirdock, line 125: Both are referencing the same array.
No matter how the filter will be changed (except a new array is assigned), it will be as the same as `activeFilter`

Kirdock (changes requested):

Kirdock (approved):

PR #6576, opened 2022-01-13
test: Move Keptn CLI authentication to go tests (#6573)

bacherfl:
This PR removes the unintential execution of the configuration service integration test on K3D. Additionally, the keptn CLI authentication is moved to the go-tests. This should improve the stability especially on minishift where sometimes the keptn auth command fails on the first try due to the api server not being ready yet, and the retry mechanism written in bash was not functioning correctly (my guess is that the `authenticate_keptn_cli` step was failing as soon as `keptn auth` was returning an error code, without proceeding with the retry logic)

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -585,7 +585,6 @@ jobs:
 
           test_outcomes=(
             "${{ steps.keptn_install.outcome }}"
-            "${{ steps.determine_keptn_endpoint.outcome }}"
           )
 
           for ((i=0; i<${#test_names[@]}; i++)); do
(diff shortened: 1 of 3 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -591,7 +585,7 @@ jobs:
 
           test_outcomes=(
             "${{ steps.keptn_install.outcome }}"
-            "${{ steps.authenticate_keptn_cli.outcome }}"
+            "${{ steps.determine_keptn_endpoint.outcome }}"
```
mowies, line 588: Does it have any special value to have the `determine_keptn_endpoint` in the integration tests report on GH?
If no, you could remove it here

bacherfl: good catch, I think does does not provide any meaningful value. I have removed it

mowies (approved):

PR #6577, opened 2022-01-13
feat(bridge): Introduce keptn color scheme #5081

### LINKED ISSUE #5081: Bridge: Use Keptn color scheme for interaction elements
As branding is now possible with the custom look-and-feel feature, I would suggest streamlining the appearance of the Keptn Bridge. Bridge uses the Barista component library and the build-in color scheme for interaction elements. This looks a bit out-of-place, as the turquoise color does not match the Keptn color scheme.

I would suggest using Keptn colors for:
* Buttons (filled, outlined)
* Icon buttons
* Drag and drop component
* Expandable tile indicators
* Overlay
* Filter elements (checkboxes, filtered search, autocompletion)

The main color of Keptn is `#006bb8`.
What we also need is a palette with darker variants of this color for hover and active states of buttons (see scss styling for the "danger" button).
Suggestions (use 2 of them): `#0063B1`, `#0058A8`, `#004EA0`

What we also have to consider is to make the color scheme of all these elements configurable, to be used with the custom look-and-feel.

## DoD 
- [ ] Introduce color palette
- [ ] Make it configurable via the custom look and feel

ermin-muratovic:
## This PR
- introduces a keptn color scheme for Buttons, Icons, etc.

### Related Issues
Fixes #5081 

image

### DIFF
--- a/bridge/client/_barista.scss
+++ b/bridge/client/_barista.scss
@@ -1,6 +1,0 @@
-@import '~@dynatrace/barista-components/style/main';
-@import '~@dynatrace/barista-components/style/index';
-
-.dt-button.dt-button-nested .dt-loading-spinner-svg {
-  stroke: $gray-400 !important;
-}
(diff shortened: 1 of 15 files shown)


Kirdock (changes requested):
The dropdown icon for the project header now has the same color as the header itself, which makes it invisible
image

before:
image

ermin-muratovic:
> The dropdown icon for the project header now has the same color as the header itself, which makes it invisible image
> 
> before: image

Fixed it

Kirdock (approved):

PR #6578, opened 2022-01-13
fix: Support Openshift 3.11

### LINKED ISSUE #4407: Deprecate Kubernetes Versions below 1.19
**For the next upcoming "major" release of Keptn**

Starting with 1.19, Kubernetes changed its implementation of networking, e.g.
`apiVersion: networking.k8s.io/v1beta1` became `apiVersion: networking.k8s.io/v1`. This is a breaking change announced for 1.22 (due in August 2021) - see [Deprecation Guide for 1.22](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22)

We painfully realized that those two versions are incompatible (see PR https://github.com/keptn/examples/pull/180 ), which requires us to maintain both for our docs and tutorials.

As Kubernetes 1.19 is already available at GKE (checked on June 18th, 2021), I recommend that we can raise the minimum required Kubernetes version from 1.14 to 1.19 for the upcoming release.

# Definition of Done

- [ ] Every occurrance of networking.k8s.io/v1beta1 has been changed to networking.k8s.io/v1 (cli, examples, tutorials) and the manifests have been adapted accordingly
- [ ] Minimum required Kubernetes version set to 1.19 in CLI and CI Jobs
- [ ] Maximum supported Kubernetes version set to 1.22 in CLI and CI Jobs
- [ ] Keptn manually tested for AKS 1.19 and EKS 1.19
- [ ] Integration Tests: Keptn automatically tested on Kubernetes 1.19 and 1.22 (e.g. using K3s)
- [ ] Docs https://keptn.sh/docs/0.8.x/operate/k8s_support/ (replace 0.8.x with 0.9.x if available) updated

odubajDT:

Fixes #4407 

Integration Tests: https://github.com/keptn/keptn/actions/runs/1707575219

### DIFF
--- a/installer/manifests/keptn/charts/control-plane/templates/ingress.yaml
+++ b/installer/manifests/keptn/charts/control-plane/templates/ingress.yaml
@@ -1,5 +1,9 @@
 {{- if .Values.ingress.enabled -}}
+{{- if lt .Capabilities.KubeVersion.Minor "19" }}
+apiVersion: networking.k8s.io/v1beta
+{{- else }}
 apiVersion: networking.k8s.io/v1
+{{- end -}}
 kind: Ingress
 metadata:
 {{- if .Values.ingress.annotations }}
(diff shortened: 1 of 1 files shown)


installer/manifests/keptn/charts/control-plane/templates/ingress.yaml
```diff
@@ -1,3 +1,47 @@
+{{- if lt .Capabilities.KubeVersion.Minor "19" }}
```
mowies, line 1: Why did you create a complete separate ingress manifest here, instead of putting the if/elses in the manifest below?

mowies (changes requested):

installer/manifests/keptn/charts/control-plane/templates/ingress.yaml
odubajDT: It's a matter of choise... If I will add if/else to the existing manifest, it will require changes on two places (2x if/else) -> `apiVersion` and `spec`. The current change requires only one time if/else, but the manifest is longer. It's actually upon us, so if you like, I can easily change it. It actually does not matter that much.

mowies: I'm always in favour of reducing lines of code but I respect your decision of course if you want to leave it as it is :)

odubajDT: fixed, re-review? :)

installer/manifests/keptn/charts/control-plane/templates/ingress.yaml
```diff
@@ -1,5 +1,9 @@
 {{- if .Values.ingress.enabled -}}
+{{- if lt .Capabilities.KubeVersion.Minor "19" }}
+apiVersion: networking.k8s.io/v1beta
```
mowies, line 3: I am pretty sure it's `networking.k8s.io/v1beta1` :)

installer/manifests/keptn/charts/control-plane/templates/ingress.yaml
```diff
@@ -1,5 +1,9 @@
...
+{{- else }}
 apiVersion: networking.k8s.io/v1
+{{- end -}}
```
mowies, line 6:
```suggestion
{{- end }}
```
If both dashes are there, a newline will be missing in the rendered chart

mowies (changes requested):

odubajDT:
@mowies fixed, thanks you :)

mowies (approved):

PR #6598, opened 2022-01-17
ci: Add pipeline to nightly check docker image digests

### LINKED ISSUE #5738: Store Hashes (e.g., SHA) of Release Images
As an admin/dev I want to be able to verify the integrity of Keptn containers/images published on docker.io, quay.io and GHCR.

# Problem

It is possible that a malicious person gains access to one of the aforementioned container registries and overwrites one or multiple images with malicious content (e.g., viruses, bitcoin miners, ...).

In order to detect this, we need to store the hash of the docker image (e.g., SHA-256).
The data is available in the UI to verify:
**GHCR**
https://github.com/keptn/keptn/pkgs/container/lighthouse-service

image

**Docker Hub**
https://hub.docker.com/r/keptn/lighthouse-service/tags?page=1&name=0.10.0-next
image

**Quay.io**
https://quay.io/repository/keptn/lighthouse-service?tab=tags
image

However, **manual verification if the hashes are correct is tedious and errorprone** (just consider that we have 10+ microservices that we need to check!) - therefore, some sort of automation is needed.

# Definition of Done

- [ ] When publishing Keptn containers/images to docker.io, quay.io and ghcr with a release tag (e.g., 0.10.0), the SHA checksum should be stored (e.g., in an artifact on the GH release) which allows users to manually (and automatically) verify the checksums of the image
- [ ] We should fetch checksums from all Keptn release images on all container registries once per night and compare the checksums against the ones from the corresponding github release

mowies:
## This PR

- adds a new pipeline that checks the docker image digests from dockerhub against the digests that are attached to the releases of keptn
- adds the upload of a `digests.csv` file to the (pre)release pipeline
- streamlines image upload to quay.io and ghcr.io (now, the images are not first released and then copied to the other registries, instead, they are pushed directly at the same time to all 3 registries)

### Related Issues

Fixes #5738

### DIFF
--- a/.github/workflows/check-docker-image-digests.yml
+++ b/.github/workflows/check-docker-image-digests.yml
@@ -1,0 +1,17 @@
+name: Check Docker Image Digests
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 5 * * 1-5'
+defaults:
+  run:
+    shell: bash
+jobs:
+  check:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2.4.0
+
+      - name: Check docker image digests
+        run: ./gh-actions-scripts/check-image-digests.sh
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -194,6 +194,20 @@ jobs:
           username: ${{ secrets.REGISTRY_USER }}
           password: ${{ secrets.REGISTRY_PASSWORD }}
 
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: "ghcr.io"
+          username: "keptn-bot"
+          password: ${{ secrets.KEPTN_BOT_TOKEN }}
+
+      - name: Login to Quay.io
+        uses: docker/login-action@v1
+        with:
+          registry: "quay.io"
+          username: ${{ secrets.QUAY_USER }}
+          password: ${{ secrets.QUAY_TOKEN }}
+
       - id: docker_build_image
         name: "Docker Build keptn/${{ matrix.config.artifact }}"
         uses: docker/build-push-action@v2
@@ -201,6 +215,8 @@ jobs:
           context: ${{ matrix.config.working-dir }}
           tags: |
             keptn/${{ matrix.config.artifact }}:${{ env.VERSION }}
+            quay.io/keptn/${{ matrix.config.artifact }}:${{ env.VERSION }}
+            ghcr.io/keptn/${{ matrix.config.artifact }}:${{ env.VERSION }}
           build-args: |
             version=${{ env.VERSION }}
           push: ${{ matrix.config.should-push-image }}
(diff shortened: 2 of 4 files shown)


odubajDT (approved):
LGTM

.github/workflows/check-docker-image-digests.yml
```diff
@@ -0,0 +1,17 @@
+name: Check Docker Image Digests
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 5 * * 1-5'
```
christian-kreuzberger-dtx, line 5: What is the reasoning for this? Should we not check this every day?

.github/workflows/pre-release.yml
```diff
@@ -194,18 +194,51 @@ jobs:
           username: ${{ secrets.REGISTRY_USER }}
           password: ${{ secrets.REGISTRY_PASSWORD }}
 
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: "ghcr.io"
+          username: "keptn-bot"
+          password: ${{ secrets.KEPTN_BOT_TOKEN }}
```
christian-kreuzberger-dtx, line 202: can we not use the included `GITHUB_TOKEN` here?

.github/workflows/pre-release.yml
```diff
@@ -355,42 +388,13 @@ jobs:
...
-          src: docker.io/keptn/${{ matrix.config.artifact }}:${{ env.VERSION }}
-          dst: |
-            quay.io/keptn/${{ matrix.config.artifact }}:${{ env.VERSION }}
-            ghcr.io/keptn/${{ matrix.config.artifact }}:${{ env.VERSION }}
+          DIGEST_FILE: "digests.csv"
+          RELEASE_TAG: ${{ steps.create-release-package.outputs.tag-name }}
+        run: |
+          for digest_file in ./dist/image-digests/digest-*.txt; do
+            cat "$digest_file" >> $DIGEST_FILE
+          done
```
christian-kreuzberger-dtx, line 398: This could be simplified:
```bash
cat ./dist/image-digests/digest-*.txt >> $DIGEST_FILE
```

.github/workflows/release.yml
```diff
@@ -343,50 +376,21 @@ jobs:
...
-          src: docker.io/keptn/${{ matrix.config.artifact }}:${{ env.VERSION }}
-          dst: |
-            quay.io/keptn/${{ matrix.config.artifact }}:${{ env.VERSION }}
-            ghcr.io/keptn/${{ matrix.config.artifact }}:${{ env.VERSION }}
+          DIGEST_FILE: "digests.csv"
+          RELEASE_TAG: ${{ steps.create-release-package.outputs.tag-name }}
+        run: |
+          for digest_file in ./dist/image-digests/digest-*.txt; do
+            cat "$digest_file" >> $DIGEST_FILE
+          done
```
christian-kreuzberger-dtx, line 394: same here, this could be simplified

.github/workflows/check-docker-image-digests.yml
mowies: basically this was copied from the integration tests pipeline to only run on working days, but of course with such a short running pipeline it should also be fine to just run it every night

.github/workflows/pre-release.yml
mowies: I want the docker images to be pushed by the keptn-bot user and not the user that runs the release pipeline which would be the result if we use the `GITHUB_TOKEN` i think

.github/workflows/release.yml
mowies: ahh nice, I will change that, thanks!

.github/workflows/pre-release.yml
mowies: ahh nice, I will change that, thanks!

PR #6603, opened 2022-01-17
feat: Introduce 'waiting' status to sequences

### LINKED ISSUE #5526: Introduce waiting status in /sequence endpoint

odubajDT:

## This PR

Adds a new `waiting` status in to the sequence state. If a sequence is executed and another sequence is running on the given stage, the sequence waits until the first sequence is finished in the stage.

Typical workflow:
- `sequence1` is started by user -> state `triggered`
- no sequence is running in stage `dev`, so started event is sent and the state of `sequence1` is changed to state `started`  -> the `sequence1` is running in stage `dev`
- `sequence2` is started by user -> state `triggered` 
- `sequence1` is running in stage `dev`, so `sequence2` needs to wait -> set `sequence2` to state `waiting` until `sequence1` is done in stage `dev` (for example `sequence1` is `finished` or `aborted`)

### Related Issues

Fixes #5526 

Integration tests: https://github.com/keptn/keptn/actions/runs/1728571338 and https://github.com/keptn/keptn/actions/runs/1728895942


shipyard-controller/go.mod
```diff
@@ -17,6 +17,9 @@ require (
 	github.com/swaggo/swag v1.7.8
 	github.com/tryvium-travels/memongo v0.3.2
 	go.mongodb.org/mongo-driver v1.7.5
+	go.uber.org/zap v1.13.0 // indirect
```
bacherfl, line 20: not sure if these additional entries are intended - can you try to run `go mod tidy` and see if that removes them?

odubajDT: executed, nothing changed, I have not added anything manually, just executed `go mod tidy`

shipyard-controller/handler/sequencedispatcher.go
```diff
@@ -63,6 +64,12 @@ func (sd *SequenceDispatcher) Add(queueItem models.QueueItem) error {
 			if err2 := sd.sequenceQueue.QueueSequence(queueItem); err2 != nil {
 				return err2
 			}
+		} else if err == ErrSequenceBlockedWaiting {
```
bacherfl, line 67: I believe we can simplify this as follows:

```
if err := sd.dispatchSequence(queueItem); err != nil {
		if errors.Is(err,ErrSequenceBlocked) || errors.Is(ErrSequenceBlockedWaiting) {
			// if the sequence is currently blocked, insert it into the queue
			if err2 := sd.sequenceQueue.QueueSequence(queueItem); err2 != nil {
				return err2
			}
		} else {
			return err
		}
	}
```

It should not make a difference if the sequence is currently blocked because it is paused, or if another sequence has to be finished first - what we put into the queue should always just be the .triggered event that triggered the sequence, so we don't need to change the event type with th `triggeredToWaitingQueueItem()` function imho.

odubajDT: yes, but when the sequence should be waiting, the ErrSequenceBlockedWaiting is returned to the shipyardcontroller.go, so that is the main difference. This is actually needed to discover that onSequenceWaiting() should be executed from the shipyardcontroller.go

bacherfl: ah that's true, so we still need to return the appropriate error. However, i do still think we don't need the `triggeredToWaitingQueueItem()` here

odubajDT: removed, thanks :)

bacherfl (approved):

PR #6613, opened 2022-01-18
ci: Minor pipeline improvements

mowies:
## This PR

- runs the docker digest check pipeline every night instead of just on workday nights
- simplifies some bash for loop to just wildcard expansions


odubajDT (approved):
LGTM

PR #6620, opened 2022-01-19
feat(bridge): Unify notifications (#5087)

### LINKED ISSUE #5087: Bridge: Unify error / success notifications
In the bridge we have currently 2 different ways of showing errors to users:
* With a grey popup (toast) on the bottom of the page that vanishes after a few seconds
    * This popup is used by the HTTP interceptor to push errors to the frontend
* With a red bar that shows error information on form submit

For success we use the same bar as for errors on form submit but in green.

Suggestion:
* Get rid of the interceptor error messages, as they are not very visible and do not provide useful information
    * For page loading / polling errors we already have a error message in place. Maybe we can reuse this instead of the toast messages.
* Use https://barista.dynatrace.com/components/alert instead of the `ktb-notification-bar` for form submit feedback
    * Extend the component to also have a success state
    * Make it closable
    * Make it configurable for it to vanish after n seconds
    * Needs discussion: Show it on top of the related element, e.g. a form or the whole page that is affected - this will move the content a bit lower, but this is a common usage pattern for such notification elements, so I have no objections against it
    * Do not overlay it over the content, as content gets unreadable and also may get unusable because it overlays some interaction elements
    
## Acceptance criteria
- [ ] Shows a success message for form interactions that are successful
- [ ] Shows an error message for form interactions that fail
- [x] Shows an error message for every other interaction that fails
- [x] Shows an error for severe problems that occur without interaction
- [x] Message can be closed
- [x] Message can be configured to vanish after n seconds

Kirdock:
Closes #5087 
image
image
image


ermin-muratovic (dismissed):
lgtm

bridge/client/app/_components/ktb-notification-bar/ktb-notification-bar.component.scss
```diff
@@ -1,65 +1,11 @@
-@import '../../../variables';
-@import '~@dynatrace/barista-components/core/src/style/variables';
-
-.page-note {
-  z-index: 101;
+.notification-container {
+  pointer-events: none;
+  position: absolute;
```
laneli, line 3: Idk maybe `position: fixed;` makes more sense if it may be provided in other components than root.

bridge/client/app/_components/ktb-notification-bar/ktb-notification-bar.component.spec.ts
```diff
@@ -30,7 +30,7 @@ describe('KtbNotificationBarComponent', () => {
   });
 
   it('should add and remove notifications', () => {
-    let notifications = fixture.debugElement.queryAll(By.css('.page-note'));
+    let notifications = fixture.debugElement.queryAll(By.css('ktb-notification'));
```
laneli, line 33: Any chance that these tests go into UI tests to avoid accessing the component? Unit test can look differently, inject the notification service and check if notifications set, the observable contains the number of elements or so? Also for hide, if notification service was called and contains proper number?

bridge/client/app/_components/ktb-notification/ktb-notification.component.scss
```diff
@@ -0,0 +1,32 @@
+@import '../../../variables';
+@import '~@dynatrace/barista-components/core/src/style/variables';
+
+dt-alert {
+  max-width: 600px;
```
laneli, line 5: I am just thinking, if we should give it a fixed size. It could look weird if we have 2 stacked that differ in size.

bridge/client/app/_services/notifications.service.ts
```diff
@@ -12,39 +13,24 @@ export class NotificationsService {
     return this._notifications.asObservable();
   }
 
-  addNotification(
+  public addNotification(
```
laneli, line 16: I think this service is also a good candidate for unit tests :)

bridge/client/app/_components/ktb-notification-bar/ktb-notification-bar.component.spec.ts
Kirdock: I can only UI test the success message so far.
`info`: We disabled the version check for Cypress
`error`: Weird behavior. If an HTTP error is thrown, it is logged two times in the console (don't know where the root cause is so far) and this lets Cypress fail the test. Disabling the `uncaught:exception` event does not have any effect

bridge/cypress/integration/notification.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    const basePage = new BasePage();
+    cy.visit('/project/dynatrace/settings/services/create');
+    showSuccess();
+
+    const notification = basePage.notificationSuccessVisible();
+    cy.get('button[title="close"]').click();
+    notification.should('not.exist');
+  });
+
+  it('should show the same notifications', () => {
```
laneli, line 70: should _not_ show the same notifications

laneli (approved):

laneli (approved):

PR #6625, opened 2022-01-19
feat: Block external traffic to internal API endpoints

### LINKED ISSUE #6302: Core: Don't expose internal API endpoints
## Action Item 

Remove via Nginx the access to internal endpoints

## Technical details

Currently, the endpoints marked as internal are accessible from outside of Keptn. 
*Only external calls shall be forbiden*.
For example, the following call returns 204 assuming `podtatohead` is a valid project:
```
curl -vvv -X PUT "http://$KEPTN_URL/api/configuration-service/v1/project/podtatohead" -H "accept: application/json" -H "x-token: $KEPTN_API_TOKEN" -H "Content-Type: application/cloudevents+json" -d "{ \"projectName\": \"podtatohead\"}"
```

The Nginx configuration should be updated to mark these APIs as [internal](http://nginx.org/en/docs/http/ngx_http_core_module.html#internal).

This is a breaking change.

- [ ] Check that the CLI does not use any internal endpoint

warber:
closes #6302

This PR contains Nginx configuration changes that actually block external access to endpoints mentioned in #6302
Further, it contains changes to the integration tests in order to let them run against the internal APIs.

Integration test run with log insgestion and resourceservice tests: https://github.com/keptn/keptn/runs/4894470445?check_suite_focus=true

### DIFF
--- a/installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
+++ b/installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
@@ -64,12 +64,12 @@ data:
     rewrite ^{{ .Values.prefixPath }}/api$ {{ .Values.prefixPath }}/api/swagger-ui/ permanent;
 
     # block POST calls to /api/mongodb-datastore/event
-    location = /api/mongodb-datastore/event {
+    location = {{ .Values.prefixPath }}/api/mongodb-datastore/event {
       limit_except GET {
         deny all;
       }
       auth_request               /api/v1/auth;
-      rewrite /api/mongodb-datastore/(.*) /$1  break;
+      rewrite {{ .Values.prefixPath }}/api/mongodb-datastore/(.*) /$1  break;
       proxy_pass         http://mongodb-datastore:8080;
       proxy_redirect     off;
       proxy_set_header   Host $host;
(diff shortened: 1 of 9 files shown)


installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
```diff
@@ -77,7 +77,7 @@ data:
       proxy_set_header X-Forwarded-Proto $scheme;
     }
 
-    location {{ .Values.prefixPath }}/bridge {
+    location ~* {{ .Values.prefixPath }}/bridge/.* {
```
mowies, line 80: this would change the previous behaviour a little bit in that requests that end with `*bridge/` won't be proxied anymore since the `/` is required in the URL now. don't know if those requests are something we need or not though...

warber: thb, i don't know either. This change was brought up by @thisthat. What's your opinion on that?

thisthat: I would say that this is a nice side-effect bug fix 😉
This way, we are more strict and only URLs with `/bridge/` are redirected to it.

mowies (approved):

PR #6641, opened 2022-01-20
fix: Adapt http status code for not found upstream repositories

### LINKED ISSUE #6630: Creating or updating the git credentials with an invalid repository returns 500 instead of 404
At the moment when updating or creating a project with a repository that does not exist, a 500 status code instead of 404 is returned.
Next to that, it looks like the swagger-UI has an invalid response (403 instead of 404) for PUT /project
image

bacherfl:
Closes #6630


RealAnna (approved):
LGTM

PR #6655, opened 2022-01-21
chore: adapted CLI to newest state of `APISet` in go-utils

warber:
This PR contains the code adaptations for CLI to be compatible with the newest codebase in `go-utils` see: https://github.com/keptn/go-utils/pull/378


odubajDT (approved):
LGTM

PR #6670, opened 2022-01-25
ci: Fix intergation tests taking master artifacts from bot commits

mowies:
## This PR

- enhances the `jq` command that filters the GH actions runs on master to find the latest successful build.
- Now, the filter takes bot builds into account, since those don't push any images which in turn makes the integration tests fail during keptn installation

Integration test run where the correct Run ID was extracted from previous master runs:
https://github.com/keptn/keptn/runs/4934107939?check_suite_focus=true#step:8:36


oleg-nenashev (approved):

odubajDT (approved):
LGTM

PR #6696, opened 2022-01-27
fix: Fixed wrong nginx location for bridge urls

warber:
#6302 

This PR fixes the nginx location configuration for bridge so that it works without a trailing `/` after

### DIFF
--- a/installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
+++ b/installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
@@ -77,7 +77,7 @@ data:
       proxy_set_header X-Forwarded-Proto $scheme;
     }
 
-    location ~* {{ .Values.prefixPath }}/bridge/.* {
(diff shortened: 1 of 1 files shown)


thisthat (approved):

installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
```diff
@@ -77,7 +77,7 @@ data:
       proxy_set_header X-Forwarded-Proto $scheme;
     }
 
-    location ~* {{ .Values.prefixPath }}/bridge/.* {
+    location ~* {{ .Values.prefixPath }}/bridge {
```
mowies, line 80: are you sure this works? if it's a direct regex matcher then i would assume that with this, no requests like `/bridge/somethingsomething` are now served correctly

thisthat: AFAIK nginx has a greedy regex parser. Since it matches the beginning of the URL it accepts it. We manually tested and it was working.

PR #6718, opened 2022-01-31
chore: Add k8s resource stats to release notes

### LINKED ISSUE #6178: Add CI run info and k8s resource info to (pre-)releases
Before the new release automation, releases for GH were generated during integration test runs. Those also added additional information about the CI run, k8s resource limit info and other things as metadata to the release notes.
With the new release automation, this behaviour was not carried over but is still needed.

Carry over the steps to generate additional info and metadata and attach that info to the release notes in the (pre-)release pipelines.

For the k8s resource limits, a new way of getting the resource data needs to be found. Before, the data was fetched from an active cluster but since the release process is not included in the integration test pipeline anymore, there is no option to get that data from an active cluster.

**Additional context**
The relevant code in the integration test pipeline: https://github.com/keptn/keptn/blob/master/.github/workflows/integration_tests.yml#L728-L1006

mowies:
### This PR
- adds a new shell script to generate k8s resource consumption data from the Keptn Helm Chart
- doesn't use an actual installation for the gathering of data but rather parses the rendered helm chart to get the data
- attaches the resource data to (pre-)releases

Fixes #6178 

### Notes
Markdown result for the release notes can be checked here: https://github.com/keptn/keptn/runs/5018553305?check_suite_focus=true

### DIFF
--- a/gh-actions-scripts/generate-k8s-resource-stats.sh
+++ b/gh-actions-scripts/generate-k8s-resource-stats.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+helm template ../dist/keptn-installer/keptn-*.tgz --dry-run > "$HELM_TEMPLATE"
+yq eval-all '[
+  .spec.template.spec.containers.[] |
+  {
+    "name": .name,
+    "resources": .resources
+  }
+]' "$HELM_TEMPLATE" -o=json | \
+jq '[
+  .[] | {
+    name: .name,
+    cpu_request: .resources.requests.cpu,
+    cpu_limit: .resources.limits.cpu,
+    mem_request: .resources.requests.memory,
+    mem_limit: .resources.limits.memory
+  }
+] |
+unique_by(.name)' > "$RESOURCE_JSON"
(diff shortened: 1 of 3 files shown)


gh-actions-scripts/generate-k8s-resource-stats.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+# Print complete yaml of the keptn installer
+helm template ../dist/keptn-installer/keptn-*.tgz --dry-run > "$HELM_TEMPLATE"
+
+# yq: Extract container name and resource limits/requests for all resources
+# jq: Bring yq result into nice format that can be read by the tablemark-cli
+# unique_by: Remove duplicated distributor resource stats
```
christian-kreuzberger-dtx, line 8: AFAIK every deployment that has a distributor will also consume memory and CPU resources - not sure why you want to remove it

mowies: fixed

christian-kreuzberger-dtx (approved):
LGTM

PR #6729, opened 2022-02-01
feat: Introduce Oauth integration for distributor and Oauth enhancements for CLI

### LINKED ISSUE #6728: Distributor: add support for client credential oauth flow for remote execution plane
The distributor shall support configurable support for performing the client credential (machine to machine) OAuth flow when calling the public Keptn API.

The way to configure the relevant information shall be possible via environment variable:

(1) `OAUTH_CLIENT_ID` - the Oauth client ID, `string default=""`
(2) `OAUTH_CLIENT_SECRET` - the Oauth client secret  `string default=""`
(3) `OAUTH_SCOPES` - the Oauth Scopes used during the oauth flow  `string default=""`
(4) `OAUTH_TOKEN_URL` - the URL to obtain the token from `string default=""`

All of these must be set in order to activate the OAuth flow.

### Acceptance criteria
```
Given the distributor is not running
When the distributor is started on a remote execution plane with (1), (2), (3). (4) set, and (5) not set
Then the distributor uses a HTTP client that supports Oauth Client Credentials flow and does not contain the Keptn API token
```

```
Given the distributor is not running
When the distributor is started on a remote execution plane with (1), (2), (3). (4) and (5) set
Then the distributor uses a HTTP client that supports Oauth Client Credentials flow and also contain the Keptn API token
```

```
Given the distributor is not running
When the distributor is started on a remote execution plane with one of (1), (2), (3) or (4) not set
Then the distributor shall use a HTTP client which is not configured for Oauth and does contain the KEPTN API token.
```

```
Given the distributor is not running
When the distributor is started on the control plane
Then the distributor shall ignore the values of (1), (2), (3), (4), and (5)
```

warber:
closes #6728 

PR contains: 
* changes according to #6728 
* rename of cli oauth flags to have prefix `oauth` instead of `sso`
* bug fixes in cli
* ability to setup oauth for distributor in helm installer

Integration Tests Run: https://github.com/keptn/keptn/actions/runs/1851816917


bacherfl (approved):

jmeter-service/chart/templates/deployment.yaml
```diff
@@ -147,6 +147,16 @@ spec:
                 fieldRef:
                   fieldPath: metadata.labels['app.kubernetes.io/name']
             {{- end }}
+            - name: OAUTH_CLIENT_ID
+              value: "{{ (((.Values.distributor).config).oauth).clientID }}"
+            - name: OAUTH_CLIENT_SECRET
```
mowies, line 152: I think the same applies here as for the bridge: secret values should not be used as env variables. The k8s secret should be mounted as a volume and then read from there...

installer/manifests/keptn/charts/control-plane/templates/_helpers.tpl
```diff
@@ -137,4 +137,14 @@ lifecycle:
     fieldRef:
       fieldPath: metadata.labels['app.kubernetes.io/name']
 {{- end }}
+- name: OAUTH_CLIENT_ID
+  value: "{{ (((.Values.distributor).config).oauth).clientID }}"
+- name: OAUTH_CLIENT_SECRET
```
mowies, line 142: same comment as for the jmeter service

bacherfl (approved):

jmeter-service/chart/templates/deployment.yaml
warber: Thanks @mowies for this remark.. I think we agreed with @thisthat and @agrimmer that we will keep it like env vars for now. correct?

agrimmer: Correct. Currently, we prefer to consistently use ENV vars.

PR #6747, opened 2022-02-02
test: Remove obsolete check of deployments

bacherfl:
The checks performed in the step that has been removed with this PR are covered by helm. Therefore this step has become obsolete


mowies (approved):
LGTM

Issue #6686, opened 2022-01-26
Research: High Availability of Keptn


thisthat:
## Action Item

Currently, we cannot deploy Keptn to be resilient against pod or node crashes. As a first step, we should scale up the replica count of the keptn core services. 

## Acceptance Criteria

This research should verify that: 

- [ ] Only one replica process a message (avoiding duplicate) - which should be guaranteed by the NATs queue groups
- [ ] How to do that for the shipyard-controller and what changes are required
- [ ] Resource quotas impact - what services can we merge? A good candidate is the ligth-house service. Mind that, backward compatibility must be a priority. 
- [ ] What do we need to change in the API gateway Nginx configuration? E.g., load balancing among different replicas, sticky sessions, etc

## DoD

- [x] Create a POC with deployments that have multiple replicas
- [x] Create a POC for the shipyard-controller to run with multiple replicas
- [ ] Create a POC for merging services together
- [ ] Do manual testing (and run the integration tests)

## Notes
- MongoDB is out of scope
- Helm- and Jmeter-service are out of scope

odubajDT:
*Notes from the meeting:*

- Api Gateway
-Core service:
-- shipyard controller
-- resource service
- Api Service:
-- secret service
-- api service
- Default Usecase service:
-- approval service
-- lighthouse service
-- remediation service
- Webhook Service 
- MongoDB Datastore
- MongoDB
- NATS

odubajDT:
picking the "Api Service"

RealAnna:
**Approval service** is now merged with **remediation service**, **ligthouse configuration handler** is also merged,  the other two handlers have dependancies on https://github.com/keptn/enhancement-proposals/pull/56 

[PoC](https://github.com/keptn/keptn/pull/6835)

odubajDT:
*Api-service update:*
Api-service and secret-service are merged, manually tested and working. Secret service was implemented using gin-gonic, but api-service not, therefore there was a need to reimplement it using the mentioned framework. The reimplementation is not completely done in terms of clean code (no managers for handlers) and without unit tests. Also there was a need to adapt charts (keptn-endpoints.conf) and roleBindings, as currently we have only 1 service instead of 2. Poc implementation available here:

https://github.com/keptn/keptn/pull/6838

thisthat:
Notes on api- and secret-service. I would not go ahead with merging these two because:
- semantically serve a very different purpose
- security concerns:
	- auth and secrets are in the same place 
[...]

PR #6749, opened 2022-02-02
feat: Trying out multiple replicas (#6686)

RealAnna:
POC for #6686

### DIFF
--- a/installer/manifests/keptn/charts/control-plane/templates/continuous-operations.yaml
+++ b/installer/manifests/keptn/charts/control-plane/templates/continuous-operations.yaml
@@ -17,7 +17,7 @@ spec:
     matchLabels:
       app.kubernetes.io/name: remediation-service
       app.kubernetes.io/instance: {{ .Release.Name }}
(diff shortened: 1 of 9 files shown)


RealAnna:
I run integration tests on my GKE cluster with keptn 0.13.0-dev  

**Services that needs some work on/ changes**
Configuration-service, as we already knew, cannot handle multiple replicas but it will be substituted anyway.
Resource-service is very close to being able to do so, (see TODOs)

**Services that worked with no need for any change**
- api-service
- approval-service
- helm-service
- jmeter-service
- lighthouse-service
- mongodb-datastore
- remediation-service
- secret-service
- statistics-service
- webhook-service


**TODOs:**

1. Make sure that there is an appropriate retry logic for every part in the service/stage manager where we write something to the repo to avoid failing for no fast forward updates when multiple replicas try to change a repo. This is for sure missing at least in writeAndCommitResource . Falures still happens at create service or project and during most calls that write on the repos.

PR #6760, opened 2022-02-03
fix(installer): External connection string not used while helm upgrade

heckelmann:
When upgrading keptn and setting a new external connection string it will use the existing one from the `mongodb-credentials`.
This PR just changes the order to use the provided value if there is one.


mowies (approved):
LGTM, thanks for the PR :)

PR #6764, opened 2022-02-03
perf: Directly return Bridge version

thisthat:

### DIFF
--- a/api/handlers/metadata.go
+++ b/api/handlers/metadata.go
@@ -72,16 +72,16 @@ func (h *metadataHandler) getMetadata() middleware.Responder {
 	payload.Bridgeversion = defaultVersion
 	payload.Shipyardversion = "0.2.0"
 
-	if bridgeVersion, err := h.getBridgeVersion(namespace); err == nil {
-		payload.Bridgeversion = bridgeVersion
-	} else {
+	if bridgeVersion, err := h.getBridgeVersion(namespace); err != nil {
 		logger.WithError(err).Error("Error getting brdige version")
+	} else {
+		payload.Bridgeversion = bridgeVersion
 	}
 
-	if keptnVersion, err := h.getSwaggerKeptnVersion(); err == nil {
-		payload.Keptnversion = keptnVersion
-	} else {
+	if keptnVersion, err := h.getSwaggerKeptnVersion(); err != nil {
 		logger.WithError(err).Error("Error getting swagger info")
+	} else {
+		payload.Keptnversion = keptnVersion
 	}
 
 	return metadata.NewMetadataOK().WithPayload(&payload)
--- a/api/handlers/metadata.go
+++ b/api/handlers/metadata.go
@@ -73,36 +73,42 @@ func (h *metadataHandler) getMetadata() middleware.Responder {
 
 	if h.k8sClient != nil {
 		deploymentsClient := h.k8sClient.AppsV1().Deployments(namespace)
-		bridgeDeployment, err := deploymentsClient.Get(context.TODO(), "bridge", metav1.GetOptions{})
-		if err != nil {
-			// log the error, but continue
-			logger.WithError(err).Error("Error getting deployment info")
-		} else {
+
+		if bridgeDeployment, err := deploymentsClient.Get(context.TODO(), "bridge", metav1.GetOptions{}); err == nil {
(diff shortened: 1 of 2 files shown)


api/handlers/metadata.go
```diff
@@ -60,45 +63,54 @@ type metadataHandler struct {
...
+	payload.Bridgeversion = defaultVersion
 	payload.Shipyardversion = "0.2.0"
 
 	if h.k8sClient != nil {
 		deploymentsClient := h.k8sClient.AppsV1().Deployments(namespace)
-		bridgeDeployment, err := deploymentsClient.Get(context.TODO(), "bridge", metav1.GetOptions{})
-		if err != nil {
-			// log the error, but continue
+
+		if bridgeDeployment, err := deploymentsClient.Get(context.TODO(), "bridge", metav1.GetOptions{}); err == nil {
```
warber, line 78: since the retrieval of getting the swagger keptn version is extracted into a method, it should also be done for getting the bridge version, to keep the level of detail equal...
maybe something like this:
```go
func (h *metadataHandler) getBridgeVersion(namespace string) (string, error) {
	if h.k8sClient == nil {
		return "", fmt.Errorf("unable to get bridge version")
	}
	bridgeDeployment, err := h.k8sClient.AppsV1().Deployments(namespace).Get(context.TODO(), "bridge", metav1.GetOptions{})
	if err != nil {
		return "", fmt.Errorf("unable to get bridge version %w", err)
	}

	v := strings.Split(bridgeDeployment.Spec.Template.Spec.Containers[0].Image, ":")
	if len(v) >= 2 {
		return v[1], nil
	}

	return "", fmt.Errorf("unable to get bridge version")
}
```

api/handlers/metadata.go
```diff
@@ -3,13 +3,14 @@ package handlers
...
-	logger "github.com/sirupsen/logrus"
 	"io/ioutil"
 	"os"
 	"strings"
 
+	logger "github.com/sirupsen/logrus"
+
 	"github.com/go-openapi/runtime/middleware"
-	"gopkg.in/yaml.v2"
+	yaml "gopkg.in/yaml.v2"
```
warber, line 13: the alias is redundant and can be removed

api/handlers/metadata.go
```diff
@@ -60,45 +63,65 @@ type metadataHandler struct {
...
-			// log the error, but continue
-			logger.WithError(err).Error("Error getting deployment info")
-		} else {
-			payload.Bridgeversion = strings.TrimPrefix(bridgeDeployment.Spec.Template.Spec.Containers[0].Image, "keptn/")
-		}
+	if bridgeVersion, err := h.getBridgeVersion(namespace); err == nil {
+		payload.Bridgeversion = bridgeVersion
+	} else {
+		logger.WithError(err).Error("Error getting brdige version")
+	}
```
warber, line 79:
```suggestion
	if bridgeVersion, err := h.getBridgeVersion(namespace); err != nil {
		logger.WithError(err).Error("Error getting brdige version")
	} else {
		payload.Bridgeversion = bridgeVersion
	}
```

api/handlers/metadata.go
```diff
@@ -60,45 +63,65 @@ type metadataHandler struct {
...
 
-		// Load swagger.yaml from /swagger-ui/swagger.yaml
-		mapSwagger := make(map[interface{}]interface{})
-		yamlFile, err := ioutil.ReadFile(h.swaggerFilePath)
+	if keptnVersion, err := h.getSwaggerKeptnVersion(); err == nil {
+		payload.Keptnversion = keptnVersion
+	} else {
+		logger.WithError(err).Error("Error getting swagger info")
+	}
```
warber, line 85:
```suggestion
	if keptnVersion, err := h.getSwaggerKeptnVersion(); err != nil {
		logger.WithError(err).Error("Error getting swagger info")
	} else {
		payload.Keptnversion = keptnVersion
	}
```

api/handlers/metadata.go
warber: just a minor thing, usually we check for the existence of an error, and not for the existence of no error. It's just a bit weird to see it like this. :)

warber (approved):

warber (approved):

PR #6779, opened 2022-02-04
style(bridge): dark mode

Kirdock:
WIP

### DIFF
--- a/bridge/client/app/_services/style-manager.service.spec.ts
+++ b/bridge/client/app/_services/style-manager.service.spec.ts
@@ -1,0 +1,16 @@
+import { TestBed } from '@angular/core/testing';
+
+import { StyleManagerService } from './style-manager.service';
+
+describe('StyleManagerService', () => {
+  let service: StyleManagerService;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({});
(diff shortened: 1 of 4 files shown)


oleg-nenashev (commented):
It would be great to have a few screenshots attached for the references and the discussion.

A few questions:
- Do we have a web browser support matrix? `var(...)` do not work in IE 11 from the Jenkins Dark Mode experiences
- What will be the user experiences? IIUC the current code, the style won't be preserved when navigating links without the stylesheet specification, e.g. when navigating from external link. Maybe it's better to use cookies here.

Kirdock:
> It would be great to have a few screenshots attached for the references and the discussion.
> 
> A few questions:
> 
> * Do we have a web browser support matrix? `var(...)` do not work in IE 11 from the Jenkins Dark Mode experiences
> * What will be the user experiences? IIUC the current code, the style won't be preserved when navigating links without the stylesheet specification, e.g. when navigating from external link. Maybe it's better to use cookies here.

At the moment it's just meant as an easter egg.
The IE 11 support ended and will not be considered when developing Keptn.
I haven't considered saving the theme option yet and currently, there are just about 2 pages supported for the dark theme.
But here is a preview:
image

PR #6785, opened 2022-02-07
fix(bridge): Fix style content security policy (#6750)

### LINKED ISSUE #6750: Not supported stylesheet MIME type error
Since version 0.11.1 an error is shown in the console and tells that the global stylesheet can't be applied. It still looks like it is applied but nevertheless, this error should not occur in the console.
image

Kirdock:
Fixes #6750 
`style-src` `unsafe-inline` has to be used for Angular applications. https://github.com/angular/angular/issues/37631


ermin-muratovic (approved):
lgtm

PR #6804, opened 2022-02-08
chore: Removed makefile and all usages of it

mowies:
### This PR
- removes the Makefile from keptn
- changes all usages of the Makefile to use the scripts underneath directly
- removes lots of unneeded shell scripts
- removes the option to set the `KUBE_CONSTRAINTS` on compile time in the CLI since they are mostly static anyways (this also reduces the places where they need to be adjusted to just 1)
- makes use of `gotestsum` in the CLI unit tests

(builds are red because github still tries to use the pipeline from master which doesn't work with this branch)


odubajDT (approved):
LGTM

PR #6813, opened 2022-02-09
feat(bridge): Use new evaluation finished payload and UI adoptions in SLI breakdown

### LINKED ISSUE #5498: Bridge: Make use of new evaluation.finished payload
## Action Item

Adapt the Bridge to use the `comparedValue` property of the `evaluation.finished` event.

## Technical Details

In #5496, the `evaluation.finished` event contains the values used for comparison inside the `comparedValue` property. The bridge should prefer this field. If missing, it should fall back to use `comparedEvents`.

## DoD

- [x] The Bridge should prefer the  `comparedValue` property to show data in the UI, if present
- [x] Unit tests

## Dependencies

- https://github.com/keptn/keptn/issues/5496

ermin-muratovic:
## This PR
- takes the `comparedValue` from the `evaluation.finished` event payload if available and calculates it only if necessary
- enhances the readability of the SLI breakdown
- fixes the absolute and relative change calculation

### Related Issues
Fixes #5498 

image


bridge/shared/interfaces/indicator-result.ts
```diff
@@ -5,6 +5,7 @@ export type Target = { criteria: string; targetValue: number; violated: boolean
 export interface IndicatorResult {
   value: {
     value: number;
+    comparedValue: number;
```
Kirdock, line 8: Does the core have a migration step for `comparedValue`? Else old indicator results will lead to incorrect results or null reference errors. Then I would mark it as optional (`comparedValue?: number`)

bridge/client/app/_components/ktb-sli-breakdown/ktb-sli-breakdown.component.ts
```diff
@@ -95,22 +95,13 @@ export class KtbSliBreakdownComponent implements OnInit {
...
-        if (result) {
-          accSum += result.value.value;
-          accCount++;
-        }
-      }
-      const comparedValue = accSum / accCount;
+      const comparedValue = this.calculateComparedValue(indicatorResult);
       const compared: Partial<SliResult> = {};
-      if (comparedValue) {
+      if (!!comparedValue) {
```
Kirdock, line 100: This will not work if `comparedValue` is `0`

bridge/client/app/_components/ktb-sli-breakdown/ktb-sli-breakdown.component.spec.ts
```diff
@@ -45,16 +45,16 @@ describe('KtbSliBreakdownComponent', () => {
     expect(rows.length).toBe(1);
     expect(fixture.nativeElement.querySelector('dt-table')).toBeTruthy();
     expect(cells[0].querySelector('button')).toBeTruthy();
-    expect(firstCell.textContent).not.toContain('compared with');
+    expect(firstCell.textContent).not.toContain('Compared with:');
```
Kirdock, line 48: Could you move all UI tests to Cypress?

bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.ts
```diff
@@ -785,13 +785,20 @@ export class KtbEvaluationDetailsComponent implements OnInit, OnDestroy {
...
   highlightHeatmap(): void {
     if (this._selectedEvaluationData && !this.isInvalidated) {
+      this.comparedIndicatorResults = [];
+      const secondaryHighlightIndexes = this._selectedEvaluationData?.data.evaluation?.comparedEvents?.map(
+        (eventId) => {
+          const eventIndex = this._heatmapSeries[0]?.data.findIndex((e) => e.evaluation?.id === eventId);
+          this.comparedIndicatorResults.push(
+            this._heatmapSeries[0]?.data[eventIndex].evaluation?.data.evaluation?.indicatorResults ?? []
+          );
+          return eventIndex;
```
Kirdock, line 795: Currently, we are loading all evaluations, right? So the index can't technically be `-1` or `undefined`. Nevertheless, I would consider this case and exclude not found evaluations. This would also be future proof then if we ever decide to not load all evaluations.

Kirdock (changes requested):
- What's the reason, the logo is changed from `png` to `svg`? I also haven't seen any change for the path to the logo, so I guess it will be missing if the bridge is newly built.
- Could you add some UI tests for with and without `comparedValue`?
- Could you set the expand-icon to always be on top? Currently it's a bit weird if it changes the position depending if the row is expanded or not.
expanded:
image
collapsed: 
image

bridge/client/app/_components/ktb-sli-breakdown/ktb-sli-breakdown.component.ts
```diff
@@ -131,6 +122,23 @@ export class KtbSliBreakdownComponent implements OnInit {
     });
   }
 
+  private calculateComparedValue(indicatorResult: IndicatorResult): number {
+    if (indicatorResult.value.comparedValue == undefined) {
```
Kirdock, line 126: Could you change this to `===`?

Kirdock (changes requested):

ermin-muratovic:
@Kirdock 

> > 
> 
> * [x]  * What's the reason, the logo is changed from `png` to `svg`? I also haven't seen any change for the path to the logo, so I guess it will be missing if the bridge is newly built.
>  Is removed now (again). The svg was a test, and now it is coming back again after every rebase.
> > 
> 
> * [ ]  * Could you add some UI tests for with and without `comparedValue`?
> 
> > 
> 
> * [ ]  * Could you set the expand-icon to always be on top? Currently it's a bit weird if it changes the position depending if the row is expanded or not.
> I made actually all cell contents to be always on top, to prevent things jumping around when expanding a row.
> > 
> 
> * [ ]  * Should we provide a way to see the whole value instead of just two decimals? There are some differences that may confuse the user. E.g.: value shows 39.53 and compared with 25.72 but the relative change is 53.66% (all decimals) instead of 53.69 (two decimals)

If somebody needs the exact value, he can check the payload. Other then that we could also provide a tooltip, but not sure if that is really necessary here?

bridge/client/app/_components/ktb-sli-breakdown/ktb-sli-breakdown.component.ts
ermin-muratovic: Actually I thought we should not calculate the change if the `comparedValue` is `0`, but of course `0` could also be a regular value.

bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.ts
ermin-muratovic: Yea, true. Actually we had that check, but since we are loading all evaluations, I thought that would be good without the check, but still I think it might happen that some evaluations are not loaded (e.g. we do have a heatmap pageSize). Restructured this a little bit and moved all related code to the `setSecondaryHighlight` method.

bridge/shared/interfaces/indicator-result.ts
ermin-muratovic: There is no migration planned, that's why we keep the backwards-compatibility. Made it optional.

bridge/client/app/_utils/app.utils.ts
```diff
@@ -19,12 +19,13 @@ export class AppUtils {
   }
 
   public static formatNumber(value: number): number {
+    const abs = Math.abs(value);
```
Kirdock, line 22: This would change the return value from a negative to a positive number.
Suggestion:
```ts
return n * (value < 0 ? -1 : 1)
```

Kirdock (changes requested):

bridge/client/app/_utils/app.utils.ts
ermin-muratovic: In this case it is not, since the `abs` value is only used to decide how many decimals are displayed, but for the calculation itself the original `value` is used. I also added some unit tests to show this and also changed from `Math.floor` to `Math.trunc`, as we have seen with the score, that this makes more sense.

Kirdock (changes requested):

bridge/cypress/support/pageobjects/ServicesPage.ts
```diff
@@ -1,8 +1,122 @@
...
+    cy.byTestId(`keptn-sli-breakdown-row-${result.name}`)
+      .find('dt-cell')
+      .eq(2)
+      .find('.error')
+      .should('have.length', result.result == 'pass' ? 0 : 1);
+
+    return this;
+  }
+
+  assertSliColumn(name: string, index: number, assertion: string, value: string): this {
```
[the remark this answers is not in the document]
ermin-muratovic: I will introduce proper uitestid's on the cells and use the sli name instead of index. Also if we add or change the order of columns, the index might always change, you are right.

bridge/client/app/_components/ktb-sli-breakdown/ktb-sli-breakdown.component.ts
```diff
@@ -95,30 +95,21 @@ export class KtbSliBreakdownComponent implements OnInit {
...
+          relative: AppUtils.formatNumber((indicatorResult.value.value / (comparedValue || 1)) * 100 - 100),
         };
       }
 
       return {
         name: indicatorResult.displayName || indicatorResult.value.metric,
         value: indicatorResult.value.message || AppUtils.formatNumber(indicatorResult.value.value),
         result: indicatorResult.status,
-        score: totalscore === 0 ? 0 : AppUtils.round((indicatorResult.score / totalscore) * this.score, 2),
+        score: totalscore === 0 ? 0 : (indicatorResult.score / totalscore) * this.score,
```
Kirdock, line 112: With this change, there may be different values for the score (SLI-breakdown and overlay of an evaluation). It should be the same for both of them. Also, the value is affected.
https://github.com/keptn/keptn/blob/246753a3e7e63342482378e7e1b5eef206dbbcc0/bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.ts#L589-L615
https://github.com/keptn/keptn/blob/246753a3e7e63342482378e7e1b5eef206dbbcc0/bridge/client/app/_components/ktb-evaluation-details/ktb-evaluation-details.component.html#L69

Kirdock (changes requested):

bridge/client/app/_components/ktb-sli-breakdown/ktb-sli-breakdown.component.ts
ermin-muratovic: Now we use the same truncateNumber for all places were the score is displayed.
image

ermin-muratovic: As discussed, we will truncate the score for the badges to 0 decimals.

image

bridge/client/app/_components/ktb-stage-badge/ktb-stage-badge.component.scss
```diff
@@ -6,7 +6,7 @@
 
   .dt-tag {
     background-color: $gray-200;
-    padding-right: 45px;
+    padding-right: 54px;
```
Kirdock, line 9: Because the decimals are removed again in the bubbles, could you also revert the increased padding here?

Kirdock (changes requested):

Kirdock (approved):

PR #6818, opened 2022-02-09
fix: made integration tests always runnable ( #6814)

### LINKED ISSUE #6814: Integration Tests: Continuous Delivery Test is not executable locally
The CD integration test is not executable locally:
image

The triggering of the test can be fixed by changing the test name to match the test method.
However, then the problem appears that local assets like helm charts cannot be located, e.g.:

```
test_continuous_delivery.go:145
File /home/luser/git/keptn/test/helm-charts/helloservice.tgz not found on local file system\n"
```
ALL integration tests should be isolated and be able to run locally

RealAnna:
### Related Issues
Fixes #6814

### DIFF
--- a/test/go-tests/test_backuprestore.go
+++ b/test/go-tests/test_backuprestore.go
@@ -145,6 +145,11 @@ func Test_BackupRestore(t *testing.T) {
 	err = os.MkdirAll("config-svc-backup", os.ModePerm)
 	require.Nil(t, err)
 
+	defer func() {
+		err = resetTestPath("../../go-tests")
+		require.Nil(t, err)
+	}()
+
 	t.Logf("Executing backup of configuration-service")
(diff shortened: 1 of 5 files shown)


test/go-tests/test_backuprestore.go
```diff
@@ -145,6 +144,11 @@ func Test_BackupRestore(t *testing.T) {
 	err = os.MkdirAll("config-svc-backup", os.ModePerm)
 	require.Nil(t, err)
 
+	defer func() {
+		err = resetTestPath("../../../go-tests")
+		require.Nil(t, err)
+	}()
```
mowies, line 150: Is it not possible to directly pass a utils function to defer instead of having to define one again only to call the utils function inside?

mowies (approved):

PR #6834, opened 2022-02-10
feat(shipyard-controller): Subscribe to events using Jetstream

### LINKED ISSUE #6685: Subscribe shipyard controller to NATS via Jetstream
The goal of this issue is to enable the shipyard controller to receive incoming CloudEvents using a pull based Nats Jetstream subscription, which will have the advantage of achieving an exactly-once message delivery guarantee for the shipyard controller.

A proof of concept of this is already done in the following draft PR: https://github.com/keptn/keptn/pull/6631 

In this PR, the subscription to Nats is happening in a component of the shipyard controller, instead of the distributor. The reason for that is to not add yet another way of subscribing to events to the distributor. The Jetstream subscription handling should be provided as a component of the go-utils to have the possibility of re-using it in other services and integrations. However, using Jetstream is for now only planned for the shipyard controller. 

**Tasks:**

- [ ] Create component for connecting to Jetstream in go-utils (#6631 can be used as a reference)
- [ ] Use that component in the shipyard controller to subscribe to Jetstream
- [ ] Upgrade helm-chart of NATs
- [ ] The feature should be implemented directly inside the shipyard-controller

**Definition of Done:**

- [ ] The shipyard controller receives messages via Jetstream
- [ ] When a message is sent to Keptn via the API and the shipyard controller is currently not available, the message will be received and handled by the shipyard controller as soon as it is up again
- [ ] Integration test (The graceful shutdown integration test potentially covers this, but a dedicated test that guarantees that the shipyard controller is not there when a message is sent would be even better)

bacherfl:
Closes #6685


installer/manifests/keptn/charts/control-plane/values.yaml
```diff
@@ -20,8 +20,27 @@ prefixPath: ""
...
-  nats.cluster.replicas: 3
+  nameOverride: keptn-nats
+  fullnameOverride: keptn-nats
+  cluster:
+    replicas: 3
+    name: nats
+  nats:
+    healthcheck:
+      startup:
+        enabled: false
```
bacherfl, line 31: Needs to be disabled because `startupProbe` is not supported by Kubernetes version used in OpenShift

odubajDT (approved):
LGTM!

shipyard-controller/nats/pullsubscription.go
```diff
@@ -0,0 +1,101 @@
...
+	for {
+		select {
+		case <-ps.ctx.Done():
+			ps.isActive = false
+			return
+		default:
+		}
+		msgs, err := ps.subscription.Fetch(10)
+		if err != nil {
+			// timeout is not a problem since that simple means that no event for that topic has been sent
```
mowies, line 67:
```suggestion
			// timeout is not a problem since that simply means that no event for that topic has been sent
```

shipyard-controller/nats/receiver.go
```diff
@@ -0,0 +1,145 @@
...
+		nch.natsConnection, err = nats.Connect(nch.natsURL)
+
+		if err != nil {
+			return errors.New("failed to create NATS connection: " + err.Error())
+		}
+
+		err = nch.setupJetStreamContext(topics)
+		if err != nil {
+			return err
+		}
```
mowies, line 81: This seems like something that would be fitting for a new function

shipyard-controller/nats/receiver.go
```diff
@@ -0,0 +1,145 @@
...
+		for i, v := range a1 {
+			if v != a2[i] {
+				return false
+			}
+		}
+	} else {
+		return false
+	}
+	return true
+}
```
mowies, line 145: Why not use something like https://pkg.go.dev/reflect#DeepEqual ?

shipyard-controller/nats/mock/keptn_nats_handler_mock.go
```diff
@@ -0,0 +1,76 @@
...
+// 		mockedIKeptnNatsMessageHandler := &IKeptnNatsMessageHandlerMock{
+// 			ProcessFunc: func(event models.Event, sync bool) error {
+// 				panic("mock out the Process method")
+// 			},
+// 		}
+//
+// 		// use mockedIKeptnNatsMessageHandler in code that requires nats.IKeptnNatsMessageHandler
+// 		// and then make assertions.
+//
+// 	}
```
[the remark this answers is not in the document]
bacherfl: yep, this is commented out automatically by the `moq` code generate when using the `skip-check` flag - which is often needed to avoid import cycles

installer/manifests/keptn/charts/control-plane/templates/shipyard-controller.yaml
```diff
@@ -113,16 +113,7 @@ spec:
...
-            - name: PUBSUB_URL
-              value: 'nats://keptn-nats-cluster'
-            - name: PUBSUB_TOPIC
-              value: 'sh.keptn.>'
-            - name: PUBSUB_RECIPIENT
-              value: '127.0.0.1'
-            - name: PUBSUB_RECIPIENT_PATH
-              value: '/v1/event'
-            - name: DISABLE_REGISTRATION
-              value: "true"
```
mowies, line 125: Those variables are not all in the `dist.common.env.vars`. Is there no need for those anymore?

bacherfl: yes, since the shipyard controller is using the distributor only for sending events with this PR.

shipyard-controller/nats/receiver.go
bacherfl: true, will use that

test/go-tests/test_selfhealing.go
```diff
@@ -3,14 +3,103 @@ package go_tests
...
+metadata:
+  name: unleash-service
+  labels:
+    run: unleash-service
+spec:
+  ports:
+    - port: 8080
+      protocol: TCP
+  selector:
+    run: unleash-service`
```
mowies, line 101: This would be fitting for a yml files under te `go-tests/assets` folder

bacherfl: as discussed, I will create an issue for refactoring this integration test to not be dependent on the unleash service anymore

mowies (approved):
LGTM! Awesome work @bacherfl👌

warber:
I think this potentially introduces a thread leak here: https://github.com/keptn/keptn/blob/0a991c1eff48b505f092d7ea66afef9ecd099ad3/shipyard-controller/handler/shipyardcontroller.go#L175

You can circumvent this one using something like:
```golang
	if waitForCompletion {
		select {
		case err := <-done:
			return err
		case <-time.After(5 * time.Minute):
			return fmt.Errorf("timed out waiting for completion")
		}
	}
``` 

However, in order to fix this problem we would need to make use of `context` in the event handling code.

mowies (dismissed):

mowies (approved):

PR #6837, opened 2022-02-10
fix(shipyard-controller): Reflect cancellation in sequence state even when no triggered event is there anymore

### LINKED ISSUE #6836: Shipyard controller returns error when cancelling sequences where no corresponding sequence.triggered event is available anymore
The logic of cancelling sequences tries to ensure the availability of the originating sequence.triggered event in the `<project>-triggeredEvents` collection. If this event is not available anymore for some reason, it will return a 404 error without setting the corresponding sequence state to `aborted`. This in turn leads to the sequence being displayed as `started` in the bridge,  even though it may have been completed already.

Note: In this scenario, new sequences can still be started and are not blocked, but the sequence state of the seemingly stalling sequence is not represented correctly

bacherfl:
Closes #6836

### DIFF
--- a/shipyard-controller/handler/eventdispatcher.go
+++ b/shipyard-controller/handler/eventdispatcher.go
@@ -86,16 +86,16 @@ func (e *EventDispatcher) Add(event models.DispatcherEvent, skipQueue bool) erro
 	})
 }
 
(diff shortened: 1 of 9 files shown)


odubajDT (approved):
LGTM

PR #6839, opened 2022-02-10
feat(bridge): Show history of quality gates in environment details

### LINKED ISSUE #3547: Show history of quality gates results for each service
## User story

Status quo: The environment screen shows the actual quality gate result for each service. #3491 

image

**Enhancement:** As a user, I would like to see the history of the quality gate results for a service, by taking a look at the last 5 evaluations. 

**Details:**
- Add the last 5 evaluations to a service similar to the approach shown here: 
- The score on the right (e.g., 55) is the actual score. Hence the user gets the same behavior like on the Heatmap on the service screen.
- Attention: No extra API calls for fetching the previous evaluations should be sent to avoid overloading the API with this enhancement. 

image

### Acceptance criteria

- [x] I see the previous evaluations to derive a trend 
- [x] Proper handling of corner cases:
  - No previous evaluation is available
  - Just 1-4 previous evaluations are available 

### Definition of Done
- [x]  Everything in acceptance criteria implemented and manually tested
- [x]  Added unit tests
- [x]  Put attention to test corner cases
- [x]  Documentation updated

ermin-muratovic:
## This PR
- enhances the details on a service in the environment details screen by showing the history of quality gates

### Related Issues
Fixes #3547 

image


Kirdock (dismissed):
- The latest 5 evaluations should be the latest 5 evaluations past the current one. Currently, the first evaluation to the left of the current one is the same as the current one.
- The past evaluations are lost after the project polling happened. The evaluations should remain and should be updated (they should not disappear, instead they should be replaced)
- The details are now incorrectly formatted.
Before 
image
After: 
image
- In my opinion, the past evaluations should have 2px or more space between the next one
- The SLI breakdown is not correctly displayed for elements too near to the right side 
image
- It would be nice to see some sort of loading indicator or a skeleton for the evaluations of each service
- Could you add (at least two) UI tests here, that a service is showing 5 or fewer evaluation results?

ermin-muratovic:
When having 5 results in the history, at first I see this "1 More ..." link, but also see the 
image

When c
image

Kirdock:
@ermin-muratovic 
> When having 5 results in the history, at first I see this "1 More ..." link, but also see the fifth element cut off below:

This seems to be a problem in resolutions less than FullHD. The `dt-tag-list` component itself is also a bit tricky. The `min-width` here would be `190px` but for some reason, the list shows the `1 more` message until a width of `250px`, if manually set and this would break the style.

> For the historic results, the tooltip opens to the right side:

This is due to a bug in the overlay component or in our evaluation-details component. The problem is that the initial height and width of the evaluation details are just about 10px. Based on that size the overlay is rendered. Either we have to re-think our implementation of the evaluation details or the overlay component (which is of the barista components) needs to be re-rendered or be aware of content changes.

> The loading indicators are also showing this "2 More ...". Maybe it would be good enough to show here just 1 loading item, since we don't know if it will be 1 or 5 historic results.

I would say we place five loading indicators here. The reason is, that we are trying to load five evaluations and not one.

PR #6875, opened 2022-02-16
ci: Ensure that build matrix runs as far as possible

mowies:
## This PR

- sets the build matrix in both release pipelines to `fail-fast: false`. This has the effect that even if one matrix job fails, the others continue running instead of being cancelled immediately.


christian-kreuzberger-dtx (approved):
LGTM

PR #6938, opened 2022-02-21
feat: Add prometheus-service scope to secret-service

christian-kreuzberger-dtx:
## This PR

- Adds `keptn-prometheus-service` as a scope to secret-service, such that it also appears in Keptn Bridge

Part of https://github.com/keptn-contrib/prometheus-service/issues/274


johannes-b (approved):
lgtm

PR #6940, opened 2022-02-21
feat(bridge): Make secret selection dynamic

### LINKED ISSUE #6653: Bridge: Make secret scope selection dynamic
## Action Item

The secret scope for creating a secret in Bridge should be dynamically loaded. 
Currently, the scopes are hardcoded in [scopes.yaml](https://github.com/keptn/keptn/blob/master/secret-service/scopes.yaml) and fetch from the secret-service. However, this list of scopes is not returned via any API and it is only used internally. 

## Acceptance Criteria

- [x] The secret-service is extended with a new public API endpoint that provides the current list of available scopes
- [x] Bridge is updated to make use of this new API

## DoD 

- [x] Unit tests for Bridge and secret-service

Kirdock:
Closes #6653 


secret-service/pkg/backend/secretbackend.go
```diff
@@ -12,4 +12,6 @@ type SecretBackend interface {
 	UpdateSecret(model.Secret) error
 	DeleteSecret(model.Secret) error
 	GetSecrets() ([]model.GetSecretResponseItem, error)
+
+	ScopeBackend
```
christian-kreuzberger-dtx, line 16: This looks misplaced, but I guess it's valid Go Code...

I would however favour something like
```
type Backend interface {
    SecretBackend
    ScopeBackend
}
```

secret-service/pkg/backend/scopebackend_k8s.go
```diff
@@ -0,0 +1,19 @@
...
+		return nil, err
+	}
+	scopeArray := make([]string, len(scopes.Scopes))
+
+	i := 0
+	for scope := range scopes.Scopes {
+		scopeArray[i] = scope
+		i++
+	}
+	sort.Strings(scopeArray)
```
christian-kreuzberger-dtx, line 17: This can be simplified... will comment with a better solution in a second

christian-kreuzberger-dtx:
```suggestion
	for scope := range scopes.Scopes {
	        scopeArray = append(scopeArray, scope)
	}
	sort.Strings(scopeArray)
```

secret-service/pkg/backend/scopebackend_k8s.go
```diff
@@ -0,0 +1,19 @@
+package backend
+
+import "sort"
+
+func (k K8sSecretBackend) GetScopes() ([]string, error) {
+	scopes, err := k.ScopesRepository.Read()
+	if err != nil {
+		return nil, err
+	}
+	scopeArray := make([]string, len(scopes.Scopes))
```
christian-kreuzberger-dtx, line 10: With my suggestion below, you don't need to pre-create the array.
```suggestion
	var scopeArray []string
```

See https://go.dev/tour/moretypes/15 for more details

bridge/shared/interfaces/secret.ts
```diff
@@ -7,7 +7,7 @@ export interface SecretKeyValuePair {
 
 export interface Secret {
   name: string;
-  scope: SecretScope;
+  scope: SecretScope | string;
```
christian-kreuzberger-dtx, line 10: I wonder why this is needed, can you explain?

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.ts
```diff
@@ -1,66 +1,104 @@
...
+  private getSecretScopes(): void {
+    AppUtils.createTimer(0, this.initialDelayMillis) // interval if it fails to fetch scopes
+      .pipe(
+        takeWhile(() => !this.scopes),
+        tap(() => {
+          this.isLoading = true;
+        }),
+        switchMap(() => this.dataService.getSecretScopes()),
+        takeUntil(this.unsubscribe$)
+      )
```
christian-kreuzberger-dtx, line 76: How often will it retry in case of a failure? 
Some inline docs would be nice here

warber:
Thanks @Kirdock,
Here are my suggestions wrt. your design decisions:
I think you can delete the files `scopebackend.go`, `scopebackend_k8s.go` and `scopebackend_k8s_test.go`.
Since a `SecretBackend` is by definition something which can handle secrets **as well as** scopes, just put the implementation for `GetScopes(..)` into `secretbackend_k8s.go`. Also move the tests to `secretbackend_k8s_test.go`.

I like your idea of segregating the interface for `SecretBackend` so that you can pass just something that is capable of "getting scopes" into the scopes handler. Why not segregate the interface for managing secrets as well? :)
In `secretbackend.go` just add another interface, e.g. something like that (i renamed ScopeBackend with ScopeGetter):

```golang
type SecretBackend interface {
	SecretManager
	ScopeGetter
}

type ScopeGetter interface {
	GetScopes() ([]string, error)
}

type SecretManager interface {
	CreateSecret(model.Secret) error
	UpdateSecret(model.Secret) error
	DeleteSecret(model.Secret) error
	GetSecrets() ([]model.GetSecretResponseItem, error)
}
```

The `ScopeHandler` receives a `ScopeGetter` and the `SecretHandler` receives a `SecretManager`, both being fed with the `K8sSecretBackend` implementation like you already have it. You could even create separate interfaces for creating, updating, deleting and getting secrets (`SecretGetter`, `SecretUpdater`,....) if you want, but i leave this up to you.

Hope that made sense.

secret-service/pkg/backend/scopebackend_k8s.go
Kirdock: Pre-creating the slice is needed. Behind a slice, there is an array. If the array is full, a bigger array is created with a complexity of O(N).

bridge/shared/interfaces/secret.ts
Kirdock: Autocompletion in the IDE. When assigning a value you get the available default ones and can also provide everything else.
Useful for static assignments like Webhook secrets or tests

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.ts
Kirdock: At the moment there isn't a limit and the retry happens until the user leaves the page

secret-service/pkg/backend/scopebackend_k8s.go
Kirdock: If the slice is pre-created we can't use this syntax.
On create `scopeArray = ["", "", ""]`
With append, the result will be `scopeArray = ["", "", "", "keptn-default", "keptn-webhook service", "dynatrace-service"]`

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.ts
christian-kreuzberger-dtx: I think a limit would make sense, maybe max. 3-5 times, max. 120 seconds (if it doesn't resolve until then, it's probably broken and needs manual intervention by the admin)

bridge/shared/interfaces/secret.ts
christian-kreuzberger-dtx: `SecretScope` provides default values which makes auto completion and testing easier, now I see :) :+1:

secret-service/pkg/backend/scopebackend_k8s.go
christian-kreuzberger-dtx: > Pre-creating the slice is needed. Behind a slice, there is an array. If the array is full, a bigger array is created with a complexity of O(N).

You are correct about the negative performance implications of my approach.
However, we're talking about scopes here. There will be maybe 5 to 20 scopes, not 100k, therefore the complexity of O(N) is negligible.

Nevertheless, if you feel that you need to pre-create the slice, that's fine.

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.scss
```diff
@@ -0,0 +1,11 @@
+
+/* barista fixes */
+::ng-deep {
+  .dt-form-field-disabled .dt-form-field-prefix {
+    background-color: #f2f2f2;
```
christian-kreuzberger-dtx, line 5: Wondering if this is a color that should be fixed via a variable/in the theme?

bridge/client/app/_models/secret.ts
```diff
@@ -3,12 +3,12 @@ import { SecretScope } from '../../../shared/interfaces/secret-scope';
 
 export class Secret implements scrt {
   name!: string;
-  scope!: SecretScope;
+  scope: SecretScope | string;
```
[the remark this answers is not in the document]
christian-kreuzberger-dtx: I guess this is the same reason with auto completion and testing - resolving this comment.

secret-service/pkg/backend/secretbackend_k8s.go
```diff
@@ -295,6 +296,22 @@ func (k K8sSecretBackend) createK8sSecretObj(secret model.Secret, namespace stri
...
+	if err != nil {
+		return nil, err
+	}
+	scopeArray := make([]string, len(scopes.Scopes))
+
+	i := 0
+	for scope := range scopes.Scopes {
+		scopeArray[i] = scope
+		i++
+	}
```
pchila, line 310: the for loop for copying can be simplified like so: 

```suggestion
        copy(scopeArray, scopes.Scopes)
```

Kirdock: Actually `scopes.Scopes` is not `[]string`, it's ` map[string]Scope`. We have to iterate over the keys of it

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.scss
Kirdock: In Barista it's not a variable either, that's why I chose to assign it that way. Else I would have to look for every usage of the disabled field (input, select, etc.) and overwrite it

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.ts
Kirdock: I just remembered we actually have a retry mechanism already in place, but this one immediately tries again (up to 3 times). I guess we can remove the implementation here.

secret-service/pkg/backend/secretbackend_k8s.go
pchila: oups I thought the source was a string slice :)

christian-kreuzberger-dtx: Then this one should work:
```suggestion
        scopeArray := make([]string, 0, len(scopes.Scopes)) // pre-initialize slice with capacity len(scopes.Scopes)

	for scope := range scopes.Scopes {
		scopeArray = append(scopeArray, scope)
	}
```

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.scss
Kirdock: Ok, the color is not set via `var(...)` but there is an existing value for the color (`$gray-130`). Just adapted it

secret-service/pkg/backend/secretbackend_k8s.go
Kirdock: done :)

bridge/client/app/_components/ktb-create-secret-form/ktb-create-secret-form.component.ts
```diff
@@ -1,66 +1,95 @@
...
     private router: Router,
     private route: ActivatedRoute,
     private fb: FormBuilder,
-    private notificationService: NotificationsService
-  ) {}
-
-  get data(): FormArray | null {
-    return this.createSecretForm.get('data') as FormArray;
+    private notificationService: NotificationsService,
+    @Inject(POLLING_INTERVAL_MILLIS) private initialDelayMillis: number
```
ermin-muratovic, line 61: This is not used, so it can be removed?

ermin-muratovic (changes requested):

ermin-muratovic (approved):
lgtm

laneli (approved):

christian-kreuzberger-dtx (approved):
Looks good to me. I reviewed everything *but* `ktb-create-secret-form.component.ts`

PR #6955, opened 2022-02-22
test: Use gitea api for (re)creating projects instead of k8s jobs

bacherfl:
This PR uses the gitea API via port forward instead of deploying a job that interacts via the API via `curl`. This way we have more control over the project upstream (re)creation during the integration tests. Also gets rid of the pod corpses of the `recreate-upstream` job that pile up if integration tests are being executed frequently on a cluster.

Integration test run: https://github.com/keptn/keptn/runs/5317698464?check_suite_focus=true 
Note: The failing test on GKE seems to be due to a resource shortage because of the high number of leftover namespaces due to failed runs


odubajDT (approved):
LGTM

PR #6967, opened 2022-02-23
feat(shipyard-controller): Added leader election with UT

### LINKED ISSUE #6822: Shipyard controller: Leader election for running periodic background tasks
The shipyard controller contains the following background tasks are being executed periodically:

- *Sequence dispatcher:* Periodically checks for sequences in the queue to start their execution if they are not blocked by other sequences
- *Event dispatcher:* Periodically checks for events in the queue (e.g. events that should be delayed, or events of a sequence that has been paused previously)
- *Sequence watcher:* Periodically checks for timed out sequences to cancel them

When running multiple replicas of the shipyard controller, where each replica runs these jobs we have the potential of concurrency problems, since this might lead to sequences being triggered multiple times, or the same event being sent out multiple times.

In this case, leader election can help to avoid these issues: If only the current leader is able to run these periodic tasks, we can avoid those concurrency issues.

In the PoC created in https://github.com/keptn/keptn/pull/6777, the leader election has been implemented using the leaderelection package of the kubernetes library.

*Note:* `Leases` are not supported on OpenShift 3.11 - so either we keep this as a known limitation (i.e. on OpenShift 3.11 only one replica of the shipyard controller can run), or we search for an alternative solution for implementing leader election. The `leases` API was introduced in K8s v1.14 but a beta API was available before. Please verify if we could use the beta API in OC 3.11. 

## Definition of Done

- [x] Sequence/Event-Dispatchers and Sequence Watchers are only being executed after acquiring the leadership among the shipyard-controller's replica set. 
- [x] The leader election should only work if the K8s APIs are available. 
- [x] OpenShift 3.11 is still supported without any additional replica of the shipyard-controller.

RealAnna:
## This PR

- adds k8s leader election to shipyard-controller
- unit test
- integration-test

### Related Issues

Closes #6822 

### Notes

Enabling leader election trough k8s in openshift would require a  downgraded k8s.io from v0.22.6 to around v0.0.0-20180608153232-46b5520a878b this is not a good idea... So for now Leader election can be turned off via helm value

### How to test

with current graceful shutdown leader election is delayed between 30-60 sec due to graceperiod and prestophook in distributor. This means that when one leader receives a SIGTERM it will take that time before a new one is elected. You can reproduce with a single copy of the shipyard-controller, delete it and see that while the other is spawned the leadership is to a different copy, finally a new election happens.
image


thisthat:
Don't forget to mark this as a breaking change :)

installer/manifests/keptn/charts/control-plane/templates/rbac.yaml
```diff
@@ -212,6 +212,29 @@ rules:
...
+    app.kubernetes.io/name: keptn-acquire-lease
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/part-of: keptn-{{ .Release.Namespace }}
+    app.kubernetes.io/component: {{ include "control-plane.name" . }}
+    helm.sh/chart: {{ include "control-plane.chart" . }}
+rules:
+  - apiGroups:
+      - "coordination.k8s.io"
+    resources:
```
bacherfl, line 230: here we can limit the access to the specific resource name (shipyard-controller-dispatcher)

installer/manifests/keptn/charts/control-plane/values.yaml
```diff
@@ -146,6 +146,8 @@ shipyardController:
   config:
     taskStartedWaitDuration: "10m"
     uniformIntegrationTTL: "48h"
+    disableLeaderElection: true
+    replicas: 3
```
bacherfl, line 150: I think the number of replicas should stay at 1 per default until https://github.com/keptn/keptn/pull/6977 is merged

RealAnna: because of [this](https://github.com/keptn/keptn/pull/6967/files#diff-4287874700cb1d1f460fd185db5a5bed707a6c5868649ebe663b5380dbbb765aR19) replicas are set to 1 anyway, do you think it is better still to explicitly make them 1 there too?

shipyard-controller/handler/sequencedispatcher.go
```diff
@@ -57,24 +62,34 @@ func NewSequenceDispatcher(
...
-			// if the sequence is currently blocked, insert it into the queue
-			if err2 := sd.sequenceQueue.QueueSequence(queueItem); err2 != nil {
-				return err2
-			}
-		} else if err == ErrSequenceBlockedWaiting {
-			// if the sequence is currently blocked and should wait, insert it into the queue
-			if err2 := sd.sequenceQueue.QueueSequence(queueItem); err2 != nil {
-				return err2
+
+	if os.Getenv(EnvVarDisableLeaderElection) == "true" {
```
warber, line 66: Please do not read environment variables out of the blue right in the middle of the code....It is hard to test. Also, the `SequenceDispatcher` does not need to know anything about the concept of "leader election" it can just operate in two different operational "modes" as it seems. I would propose to introduce a kind of distinction using an enum:

```go
type SDMode int

const (
   // document what this mode is supposed to do
   SDModeRW SDMode = iota
   // document what that mode is supposed to do
   SDModeW 
)
```

And store the SD mode as a field of the `SequenceDispatcher`

installer/manifests/keptn/charts/control-plane/values.yaml
warber: maybe i did not understand it correctly but isn't disabling LE and setting replicas to more than 1 kind of contradicting?

warber (changes requested):

installer/manifests/keptn/charts/control-plane/values.yaml
RealAnna: yeah it can be confusing for the user but it works :D I will set it to 1 for now so that it does not sounds strange

warber (approved):

PR #6976, opened 2022-02-24
fix(bridge): Rounding evaluation score correctly

### LINKED ISSUE #6890: Rounding evaluation score causes confusion on evaluation board
### Problem

When having an evaluation score of 89.69... it is still below the warning threshold of, e.g. 90, but due to rounding this value it looks like being exactly 90 and this violates the expression 90 < 90 shown below:

issue

### Proposed fix

* Round the value to two decimals. Given the above example, this would be `89.68`

ermin-muratovic:
## This PR
- is rounding the evaluation score correctly

### Related Issues
Fixes #6890 

image


mowies (approved):
LGTM!

PR #6984, opened 2022-02-24
feat: Support auth via proxy

### LINKED ISSUE #6801: Core: Support HTTP(S) git authorization via proxy in resource-service
**User story**
As a user, I want to be able to proxy for Keptn git repository authorization like I use it when working with remote git repository

**Technical details**
There is a need to store the proxy credentials into secret and use it when:
- clone
- pull
- push
- fetch

The implementation of the feature (without tests and refactoring) is available as a part of this draft pull-requests: 
- https://github.com/keptn/keptn/pull/6746
- https://github.com/keptn/go-utils/pull/384

Research ticket about this feature: https://github.com/keptn/keptn/issues/6647

**Acceptance Criteria**
- [ ] Implement support of HTTP(S) git authorization via proxy in resource-service
- [ ] Implement acceptation of new parameters in CLI
- [ ] Add new parameters to API endpoints in shipyard-controller (⚠️ this is a breaking change⚠️)
- [ ] Update go-utils accordingly

**DoD**
- [ ] Acceptance Criteria are implemented
- [ ] Unit tests
- [ ] Manual tests
- [ ] Integration test

odubajDT:
## This PR

- adds support for authentication via proxy to remote git repository in resource-service
- adds additional unit tests for proxy and ssh auth
- adds integration tests for new auth methods
- refactor

### Related Issues

Fixes #6801 

#### Resource service
https://github.com/keptn/keptn/actions/runs/1945201942

https://github.com/keptn/keptn/actions/runs/1951608810

https://github.com/keptn/keptn/actions/runs/1957828876

Note: integration tests are failing due to execution plane installation problem -> helm did not installed helm-service and jmeter-service


odubajDT:
Required to merge before merging: https://github.com/keptn/go-utils/pull/409

cli/cmd/create_project.go
```diff
@@ -112,14 +127,38 @@ keptn create project PROJECTNAME --git-user=GIT_USER --git-remote-url=GIT_REMOTE
 			project.GitToken = *createProjectParams.GitToken
 			project.GitRemoteURL = *createProjectParams.RemoteURL
 
+			if *createProjectParams.GitProxyURL != "" && strings.HasPrefix(*createProjectParams.RemoteURL, "ssh://") {
+				return errors.New(gitErrMsg)
```
bacherfl, line 131: I think we need to be more specific with the error message here, telling the user why ssh:// cannot be used as a prefix

cli/cmd/create_project.go
```diff
@@ -112,14 +127,38 @@ keptn create project PROJECTNAME --git-user=GIT_USER --git-remote-url=GIT_REMOTE
...
+			}
+
+			if *createProjectParams.GitProxyURL != "" && *createProjectParams.GitProxyScheme == "" {
+				return errors.New(gitErrMsg)
```
bacherfl, line 135: here we also need a specific error message, since `gitErrMsg` only explains that git-token and private key cannot be used at the same time

cli/cmd/update_project_test.go
```diff
@@ -59,3 +59,28 @@ func TestUpdateProjectCmdTokenAndKey(t *testing.T) {
 		t.Errorf("missing expected error, but got %v", err)
 	}
 }
+
+// IMPORTANT NOTE: tests below are disabled due to broken cli, which is unrepairable adn needs to be rewritten
```
bacherfl, line 63: what exaclty is broken? maybe we can create a follow up issue on that :)

resource-service/common/git.go
```diff
@@ -84,6 +88,28 @@ func getAuthMethod(gitContext common_models.GitContext) (transport.AuthMethod, e
...
+				// 15 second timeout
+				Timeout: 15 * time.Second,
+
+				// don't follow redirect
+				CheckRedirect: func(req *nethttp.Request, via []*nethttp.Request) error {
+					return nethttp.ErrUseLastResponse
+				},
+			}
+
+			client.InstallProtocol("https", http.NewClient(customClient))
```
bacherfl, line 111: can you add a comment describing what `client.InstallProtocol()` does exactly and why it is needed? to me it seem like this will cause all https traffic through this proxy, which would mean that other projects would then also go through the proxy, or am I understanding it wrong?

resource-service/common/gitsuite_test.go
```diff
@@ -992,6 +994,110 @@ func (s *BaseSuite) Test_getGitKeptnEmail(c *C) {
...
+					GitProxyUser:      "",
+					GitProxyPassword:  "",
+				},
+				Project: "my-proj",
+			},
+			wantErr:        false,
+			expectedOutput: nil,
+		},
+		{
+			name: "invalid ssh credentials",
```
bacherfl, line 1048: please also add a test case for valid SSH credentials

test/go-tests/test_proxyauth.go
```diff
@@ -0,0 +1,125 @@
...
+	require.Nil(t, err)
+
+	t.Logf("Updating project credentials")
+	user := GetGiteaUser()
+	token, err := GetGiteaToken()
+	require.Nil(t, err)
+
+	namespace := osutils.GetOSEnvOrDefault(KeptnNamespaceEnvVar, DefaultKeptnNamespace)
+
+	squidIP, err := GetSquidExternalIP(namespace)
```
bacherfl, line 111: is it somehow possible (e.g. by inspecting the logs of the proxy) to verify that:
1. The resource service is indeed going through the proxy for requests regarding this project
2. The resource service does not go through the proxy for other projects where no proxy is configured

odubajDT: will check

resource-service/common/gitsuite_test.go
odubajDT: that's an integration test

resource-service/common/git.go
odubajDT: No, it actually uses https protocol to route all the traffic (it should work like that now as default), but when the user specifies, that he does not want to use a secure connection, he specifies --git-proxy-insecure and uses the https protocol without the SSL verification. I will add a comment there

test/go-tests/test_proxyauth.go
odubajDT: Actually we have no logs from proxy, so there is no way to test if the routing goes through it

PR #6993, opened 2022-02-25
fix(bridge): Truncate evaluation score

### LINKED ISSUE #6890: Rounding evaluation score causes confusion on evaluation board

ermin-muratovic:
## This PR
- is truncating the evaluation score correctly

### Related Issues
Fixes #6890 

image

### DIFF
--- a/bridge/client/app/_pipes/truncate-number.pipe.spec.ts
+++ b/bridge/client/app/_pipes/truncate-number.pipe.spec.ts
@@ -1,0 +1,25 @@
+import { TruncateNumberPipe } from './truncate-number';
+
+describe('TruncateNumberPipe', () => {
+  const pipe = new TruncateNumberPipe();
+
+  it('truncates 99.999999999 to 1.99', () => {
+    expect(pipe.transform(99.999999999, 2)).toBe(99.99);
+  });
+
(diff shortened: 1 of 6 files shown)


laneli (changes requested):
Can you please add Unit Tests for any other case that is not 1.99999, to be sure that this also is still working. Maybe for one for different decimals and one for a number without decimals?

bridge/client/app/_pipes/truncate-number.pipe.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { TruncateNumberPipe } from './truncate-number';
+
+describe('TruncateNumberPipe', () => {
```
Kirdock, line 3: Maybe also add a test for negative numbers and a case to truncate a number to 3 or 1 decimals (a parameter that isn't 2)

Kirdock (approved):

laneli (approved):

PR #7006, opened 2022-02-28
test: Use resource service during integration tests

### LINKED ISSUE #6947: Integration tests: Execute tests on new resource-service as well
Right now, we are still only using the current configuration-service during our integration tests, while the new resource-service is hidden in the shadows.

To detect any issues within this new service as early as possible, we should add another dimension to our test matrix where we can define whether the configuration-service or the resource service should be used when installing Keptn.

Suggestion: Since we test on two different versions of GKE and K3s respectively we could use one version for executing the tests with the configuration service and the other with the new resource service.

mowies:
### This PR
- enable the resource service one k3s 1.21 and GKE 1.21
- switches the installation of Keptn on those platforms to directly use the helm chart instead of `keptn install` to be able to customize the `values.yml`
- Fixes the `k3d` airgapped integration tests by updating to the correct images in the airgapped installer helper scripts
- Removes k8s resource data gathering since it was moved to the release pipelines
- Makes integration test resource cleanup more reliable and non-blocking
- Makes airgapped helper scripts more customizable by adding an option to define the keptn docker organization (for the new keptndev org)
- Makes test report generation more reliable by testing for existence of report files before trying to access them

Fixes #6947 

Integration test run: https://github.com/keptn/keptn/actions/runs/1927043343
(only zero downtime test fails)


.github/workflows/integration_tests.yml
```diff
@@ -396,34 +408,15 @@ jobs:
...
-            selector:
-              istio: ingressgateway # use Istio default gateway implementation
-            servers:
-            - port:
-                number: 80
-                name: http
-                protocol: HTTP
-              hosts:
-              - "*"
-          EOF
```
mowies, line 416: This is not needed anymore since it's namespace/testrun independent and therefore statically deployed on the both integration test clusters

installer/airgapped/pull_and_retag_images.sh
```diff
@@ -1,26 +1,54 @@
...
 
-IMAGES="$IMAGES_CONTROL_PLANE_THIRD_PARTY $IMAGES_CONTROL_PLANE $IMAGES_CONTINUOUS_DELIVERY"
+IMAGES_CONTROL_PLANE_THIRD_PARTY=(
+  "bitnami/mongodb:4.4.9-debian-10-r0"
+  "nats:2.7.2-alpine"
+  "natsio/nats-server-config-reloader:0.6.2"
+  "natsio/prometheus-nats-exporter:0.9.1"
+  "nginxinc/nginx-unprivileged:1.21.4-alpine"
+)
+IMAGES_CONTROL_PLANE=(
```
bacherfl, line 24: should we add `resource-service` here as well?

mowies: good catch, added it now :)

bacherfl (approved):

PR #7014, opened 2022-02-28
feat(cli): trigger authorization code flow when refresh token is expired

### LINKED ISSUE #6717: CLI: Run the auth command automatically if OAuth was enabled and the refresh token is expired
## User Story

As a user that uses an SSO integration, I don't want to retrigger the auth command again if the tokens are expired.

## Acceptance Criteria

This feature is available only if the SSO integration was used to authenticate the CLI with a Keptn installation. 
When the tokens are expired and a new login is required, the CLI should open the browser windows to trigger the login with the configured values stored locally. 

## DoD 

- [ ] Unit tests
- [ ] Manual tests

warber:
closes #6717

### DIFF
--- a/cli/internal/api.go
+++ b/cli/internal/api.go
@@ -44,7 +44,7 @@ func getAPISetWithOauthGetter(baseURL string, keptnXToken string, oauthAuthentic
 			return nil, err
 		}
 		// get the ready to use HTTP client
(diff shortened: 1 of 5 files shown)


odubajDT (approved):
LGTM

PR #7023, opened 2022-03-01
fix(bridge): Prevent spaces in URL

### LINKED ISSUE #7000: Bridge or Webhook Service Doesn't URL Encode Input
**Environment**
* Keptn Version (`keptn version`): 0.12.2

**Affected Component**

* [ ] Approval
* [x] Bridge
* [ ] CLI
* [ ] Datastore
* [ ] Docs
* [ ] Helm
* [ ] Istio
* [ ] Jmeter
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [ ] REST API
* [x] Webhook

**Describe the bug**
The bridge UI and / or webhook service does not URL encode input but rather truncates at the first whitespace.

```
kind: WebhookConfig
metadata:
  name: webhook-configuration
spec:
  webhooks:
    - type: sh.keptn.event.hello-world.finished
      requests:
        - curl --request GET https://example.com?foo=bar this
      subscriptionID: 09bbde98-f956-4afc-a15e-a1efeca2f859
      sendFinished: false
```

**To Reproduce**
Create a webhook subscription with the URL set to:

```
https://example.com?a=foo bar
```

Edit the subscription and it has been truncated to:

```
https://example.com?a=foo
```

**Expected behavior**
Bridge and / or webhook is able to take the raw string and URL encode it:

```
https://example.com?a=foo%20bar
```

Update: I think this is a bridge issue:

Given a request of: `curl --request GET https://example.com?foo=bar this`

1. It is stored correctly in YAML:
```
requests:
        - curl --request GET https://example.com?foo=bar this
```
2. It appears to be unmarshalled correctly:
```
yaml.Unmarshal(data, webHookConfig)
for _, webhook := range webHookConfig.Spec.Webhooks {
  for _, request := range webhook.Requests {
    fmt.Printf("Request is: %s", request)
  }
}
```
Output is `Request is:  curl --request GET https://example.com?foo=bar this`

Kirdock:
Fixes #7000 
- The user is now not allowed to enter spaces for the URL (it has to be encoded)
- If the webhook.yaml contains spaces inside the URL, the full URL is now returned instead of the first part before the space

### DIFF
--- a/bridge/client/app/_components/ktb-webhook-settings/ktb-webhook-settings.component.html
+++ b/bridge/client/app/_components/ktb-webhook-settings/ktb-webhook-settings.component.html
@@ -28,7 +28,9 @@
         class="resize-vertical"
       ></textarea>
       <dt-error *ngIf="getFormControl('url').errors?.required">URL must not be empty</dt-error>
-      <dt-error *ngIf="getFormControl('url').errors?.url">URL must start with http(s)://</dt-error>
+      <dt-error *ngIf="getFormControl('url').errors?.url">
+        URL must start with http(s):// and must not contain spaces
+      </dt-error>
       <div dtSuffix uitestid="event-url">
         <ng-container
           [ngTemplateOutlet]="payloadButton"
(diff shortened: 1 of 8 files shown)


ermin-muratovic (approved):
lgtm

PR #7024, opened 2022-03-01
feat(resource-service): Delete project via cloud events

### LINKED ISSUE #6928: Resource-service : delete a project with multiple replicas
## User Story

As a user, I'd like to delete a project while running Keptn with multiple replicas

## Technical details

The deletion request hits a single replica and we need to propagate this information. Research how to tackle this problem. A possible solution is to subscribe to `sh.keptn.project.delete` event which is used by the shipyard-controller. 

## DoD

- [ ] Integration tests

RealAnna:
-  resource service subscribes to JS, and reacts on `sh.keptn.event.project.delete.finished` events in order to trigger a deletion of a project
-  changed JS stream name from "shipyard-controller" to "keptn"

This PR also contains a fix for pontential NATS reconnection issue in resource service and shipyard controller.

Closes #6928 

### Notes
- no need to use distributor since the service only reacts to events without producing any
- the (internal) endpoint for deleting a project was **not* deleted.
- for now JS related code is mostly copypasted from Shipyard Controller (follow up: https://github.com/keptn/keptn/issues/7079)

Signed-off-by: RealAnna <anna.reale@dynatrace.com>


resource-service/handler/project_resource_handler_test.go
```diff
@@ -751,79 +751,3 @@ func TestProjectResourceHandler_UpdateProjectResource(t *testing.T) {
 		})
 	}
 }
-
```
bacherfl, line 754: was this removed accidentally?

resource-service/nats/connection.go
```diff
@@ -0,0 +1,152 @@
...
+	return NewPublisher(nch.natsConnection), nil
+}
+
+func (nch *NatsConnectionHandler) renewNatsConnection() error {
+	var err error
+	nch.RemoveAllSubscriptions()
+
+	nch.natsConnection.Close()
+	logger.Infof("Connecting to NATS server at %s ...", nch.natsURL)
+	nch.natsConnection, err = nats.Connect(nch.natsURL, nats.MaxReconnects(-1))
```
bacherfl, line 107: can you please add the `nats.MaxReconnects(-1)` to the nats connection setup in the shipyard controller as well?

resource-service/handler/project_resource_handler_test.go
warber: yep, (re)added it

resource-service/nats/connection.go
warber: done

bacherfl (approved):

PR #7103, opened 2022-03-09
test(bridge): Fix flaky UI test

Kirdock:
Sometimes some UI tests fail because of an empty page. To prevent this, we will wait until the metadata endpoint is called after the page is visited. Could be that this still does not work because of a bug in Cypress https://github.com/keptn/keptn/issues/7107


ermin-muratovic (approved):
lgtm

PR #7110, opened 2022-03-09
docs: Reference the code of conduct in the .github repository

iamrajiv:


oleg-nenashev (approved):
This content was moved. We should keep the file, because it is referenced from outsode

oleg-nenashev (approved):

PR #7114, opened 2022-03-10
fix(shipyard-controller): Make sure sequence queue is free again after aborting a sequence

bacherfl:
Backport for #7116 
This PR makes sure that the sequence queue is freed up for new sequences when a sequence is cancelled, even if there might be an error with retrieving certain entries from the database (e.g. the original sequence.triggered event)

Also, additional test cases for queueing/pausing/aborting sequences have been added

Integration test run: https://github.com/keptn/keptn/runs/5496316358?check_suite_focus=true

Note: the failing minishift test seems to be unrelated to the changes made in this PR

New integration test run: https://github.com/keptn/keptn/runs/5507244019?check_suite_focus=true

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -515,7 +515,7 @@ jobs:
 
       - name: Integration Tests
         id: test_aggregated
-        timeout-minutes: 60
+        timeout-minutes: 100
         env:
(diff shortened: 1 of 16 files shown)


odubajDT:
Looks good, but the integration tests are failing

odubajDT (approved):

PR #7121, opened 2022-03-11
fix(configuration-service): Return 404 when token is invalid

odubajDT:
## This PR

- modifies err response code

### Integration tests
https://github.com/keptn/keptn/actions/runs/1979505177


RealAnna (approved):
LGTM

PR #7126, opened 2022-03-11
fix(shipyard-controller): Make sure sequence queue is free again after aborting a sequence

### LINKED ISSUE #7116: Investigate why stalled sequences are blocking other sequences
Affected Versions: 0.12.x, 0.13.x

In some cases we encountered sequences that were in an inconsistent state, which could not be aborted and therefore blocked other sequences from being started.
Also, when aborting a sequence with parallel stages, the open tasks for the cancelled keptnContext were not cleaned up properly, leading to new sequences to being blocked in some stages.

We also need to verify if this is happening in the current master.

bacherfl:
Forward-port of #7114
Closes #7116


RealAnna (approved):
LGTM

PR #7175, opened 2022-03-16
fix(shipyard-controller): Abort multi-stage sequences

bacherfl:
This PR adds the integration test cases introduced in #7126 
In the shipyard controller, the logic for timing out/aborting sequences has been adapted to make it less error prone and remove the dependency on the original sequence.triggered event for cancelling/timing out a sequence execution


RealAnna (approved):
LGTM

RealAnna (dismissed):
LGTM

RealAnna (approved):
LATM

PR #7183, opened 2022-03-17
feat(bridge): Allow to configure sendStarted flag for webhook config

### LINKED ISSUE #6059: Bridge: Add checkbox to set the `sendStarted` flag
## User Story

As a user, I would like my integration via WebHook to return the `*.started` event.

## Technical Details
 
The Bridge should provide a checkbox to allow to configure this behavior for the WebHook integration.
The status of the checkbox should be saved in the webhook configuration, which looks like the following:

```
kind: WebhookConfig
metadata:
  name: webhook-configuration
spec:
  webhooks:
    - type: sh.keptn.event.deployment.triggered
      subscriptionid: my-sub-id
      sendFinished: false         
      sendStarted: false         # <- new field
      envFrom:
        - name: "secretKey"
          secretRef:
            name: "slack-secret"
            key: "endpoint"
      requests:
        - "curl -X POST -H 'Content-type: application/json' --data '{\"text\":\"Test triggered in {{.data.stage}}\"}' https://hooks.slack.com/services/{{.env.secretKey}}"
```
 
The flag has the following behavior:
- If `sendStarted` is true, the WebHook-service will send the `started` event (default)
- If `sendStarted` is false, the WebHook-service will **not** send the `started` event

This should be documented on hover. Mind to explain that if unchecked, the integration must provide the respective started event.

## Acceptance criteria

- [x] Enable this only if the subscription is on a `triggered` type
- [x] Feature implemented
- [x] Default value is true
 
## DoD

- [x] UI Test
- [x] On hover, present a small description of the feature

## Dependencies

- [x] #6524

ermin-muratovic:
## This PR
- allows the user to configure the sendStarted flag for a webhook configuration via Keptn Bridge

### Related Issues
Fixes #6059 

image

### DIFF
--- a/bridge/cypress/integration/uniform-integrations.spec.ts
+++ b/bridge/cypress/integration/uniform-integrations.spec.ts
@@ -290,7 +290,15 @@ describe('Add webhook subscriptions', () => {
     cy.byTestId(uniformPage.EDIT_SUBSCRIPTION_FIELD_GLOBAL_ID).should('not.exist');
   });
 
-  it('should have sendFinished and sendStarted checkbox disabled', () => {
+  it('should have sendFinished and sendStarted checkbox disabled for started events', () => {
+    uniformPage
+      .setTaskPrefix('deployment')
+      .setTaskSuffix('started')
+      .assertIsSendStartedButtonsEnabled(false)
+      .assertIsSendFinishedButtonsEnabled(false);
+  });
+
+  it('should have sendFinished and sendStarted checkbox disabled for finished events', () => {
     uniformPage
       .setTaskPrefix('deployment')
       .setTaskSuffix('finished')
@@ -298,7 +306,7 @@ describe('Add webhook subscriptions', () => {
       .assertIsSendFinishedButtonsEnabled(false);
   });
 
-  it('should have sendFinished and sendStarted checkbox enabled and true by default', () => {
+  it('should have sendFinished and sendStarted checkbox enabled and true by default for triggered events', () => {
     uniformPage
       .setTaskPrefix('deployment')
       .setTaskSuffix('triggered')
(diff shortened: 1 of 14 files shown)


bridge/server/models/webhook-config-yaml.ts
```diff
@@ -110,6 +113,7 @@ export class WebhookConfigYaml implements WebhookConfigYamlResult {
         const parsedConfig = this.parseConfig(curl);
         parsedConfig.secrets = webhook.envFrom;
         parsedConfig.sendFinished = webhook.sendFinished ?? false;
```
laneli, line 115: Shouldn't be `sendFinished` also be true? Because if I create a new subscription and select triggered, sendFinished is also set to `true` as default

ermin-muratovic: The behaviour of the `webhook-service` is as follows:
- if the `sendStarted` flag is not set in the `webhook.yaml` it sends the .started event by default
- if the `sendFinished` flag is not set in the `webhhook.yaml` it doesn't send the .finished event by default

Parsing the file, will behave the same, but the default when creating a new webhook subscription is true for both, `sendStarted` and `sendFinished`.

bridge/cypress/integration/uniform-integrations.spec.ts
```diff
@@ -289,6 +289,32 @@ describe('Add webhook subscriptions', () => {
...
+  it('should have sendFinished and sendStarted checkbox enabled and true by default for triggered events', () => {
+    uniformPage
+      .setTaskPrefix('deployment')
+      .setTaskSuffix('triggered')
+      .assertIsSendStartedButtonsEnabled(true)
+      .assertIsSendFinishedButtonsEnabled(true)
+      .assertIsSendStarted(true)
+      .assertIsSendFinished(true);
+  });
 });
```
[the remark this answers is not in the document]
ermin-muratovic: UI tests are added ;)

laneli (approved):

PR #7223, opened 2022-03-21
test: Add safe-to-evict annotation to enable frictionless autoscaling

mowies:
## This PR

- adds the `safe-to-evict` label to enable easy autoscaling of the helloservice pod. This is only needed because during integration tests, also blue/green deployments are done which add an envoy proxy to the pod, which in turn has a local volume that prevents evicting of this pod during cluster scaling.

Successful integration tests: https://github.com/keptn/keptn/actions/runs/2027010186


odubajDT (approved):

RealAnna (approved):
LGTM

odubajDT (approved):

PR #7240, opened 2022-03-22
chore: Use correct Keptn branding logo and spelling

thisthat:

### DIFF
--- a/bridge/client/assets/default-branding/app-config.json
+++ b/bridge/client/assets/default-branding/app-config.json
@@ -1,5 +1,5 @@
 {
-  "appTitle": "keptn",
+  "appTitle": "Keptn",
   "logoUrl": "assets/default-branding/logo.png",
   "logoInvertedUrl": "assets/default-branding/logo_inverted.png",
   "stylesheetUrl": "assets/default-branding/keptn-theme.css"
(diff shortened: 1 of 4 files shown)


ermin-muratovic (changes requested):

bridge/client/assets/default-branding/app-config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "appTitle": "keptn",
+  "appTitle": "Keptn",
```
ermin-muratovic, line 2: For this change, you also need to update the UI test here (which is failing right now): https://github.com/keptn/keptn/blob/a5dc071e72ce6d8d13e128bb1e48aa92fdd1ac2d/bridge/cypress/integration/app.spec.ts#L10

But this also changes our appearance and I'm not sure if this maybe should even stay `keptn`? Looping in @johannes-b

ermin-muratovic (changes requested):

bridge/client/assets/default-branding/app-config.json
johannes-b: When we are touching this piece, please go with `Keptn`.

ermin-muratovic (approved):
lgtm

Issue #7227, opened 2022-03-22
Distributor does not exit when it's not able to retrieve events from the control plane


warber:
This could happen e.g. when using an invalid API token and for several other reasons.
In order to make it more visible that there is a problem, the distributor shall exit with an appropriate error message when it isn't able to (re)register or fetching events to/from the control plane.

PR #7263, opened 2022-03-24
fix(distributor): shut down distributor when not able to send heartbeat to control plane

warber:
fixes: #7227

### DIFF
--- a/distributor/pkg/uniform/watch/uniformwatch.go
+++ b/distributor/pkg/uniform/watch/uniformwatch.go
@@ -1,7 +1,6 @@
 package watch
 
 import (
-	"fmt"
 	"github.com/keptn/go-utils/pkg/api/models"
 	"github.com/keptn/go-utils/pkg/common/retry"
 	"github.com/keptn/keptn/distributor/pkg/config"
@@ -41,7 +40,6 @@ type UniformWatch struct {
 //
 // It returns a pointer to a new UniformWatch without any subscription listeners
 func New(controlPlane controlplane.IControlPlane, env config.EnvConfig) *UniformWatch {
-	fmt.Println(env.HeartbeatIntervalDuration)
 	return &UniformWatch{
 		HeartbeatInterval:   env.HeartbeatIntervalDuration,
 		MaxHeartbeatRetries: env.MaxHeartBeatRetries,
--- a/distributor/pkg/uniform/watch/uniformwatch.go
+++ b/distributor/pkg/uniform/watch/uniformwatch.go
@@ -55,12 +55,15 @@ func New(controlPlane controlplane.IControlPlane) *UniformWatch {
 func (sw *UniformWatch) Start(ctx *utils.ExecutionContext) (string, bool) {
 	logger.Info("Registering Keptn Integration")
 	var id string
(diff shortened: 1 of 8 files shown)


distributor/pkg/uniform/watch/uniformwatch.go
```diff
@@ -1,82 +1,125 @@
...
-func (sw *UniformWatch) Start(ctx context.Context) string {
+// Start triggers starts the attempt(s) to do the initial
+// registration to the control plane.
+// If it was successful, it will start to send heartbeat messages in the background
+// This method does not block
+func (sw *UniformWatch) Start(ctx *utils.ExecutionContext) (string, bool) {
 	logger.Info("Registering Keptn Integration")
 	var id string
+	failRegisterCount := 0
 	err := retry.Retry(func() error {
```
thisthat, line 60: is the retry logic implemented with an exp backoff?

distributor/pkg/utils/utils.go
```diff
@@ -21,6 +21,30 @@ type EventMatcher struct {
...
+	CancelFn context.CancelFunc
+}
+
+// NewExecutionContext creates a new ExecutionContext with the given underlying go context.
+// waitGroupCount determines the number of components this ExecutionContext
+// is used in
+func NewExecutionContext(ctx context.Context, waitGroupCount int) *ExecutionContext {
+	wg := new(sync.WaitGroup)
+	wg.Add(waitGroupCount)
+	return &ExecutionContext{ctx, wg, func() {}}
```
thisthat, line 45: I would add to the `CancelFn` a log message at info level that it has been called

distributor/pkg/uniform/watch/uniformwatch.go
warber: no, and if we want that we should add this to the retry package.
i was also thinking about just using this for the future: https://github.com/avast/retry-go
it's another dependency, yes, but it has nice features and is bulletproof .... it's worth it i think

distributor/pkg/utils/utils.go
warber: i think there is no way to do that

distributor/pkg/uniform/watch/uniformwatch.go
```diff
@@ -1,83 +1,114 @@
...
+		listeners:           []SubscriptionListener{},
 	}
 }
 
-func (sw *UniformWatch) Start(ctx context.Context) string {
+// Start triggers starts the attempt(s) to do the initial
+// registration to the control plane.
+// If it was successful, it will start to send heartbeat messages in the background
+// This method does not block
+func (sw *UniformWatch) Start(ctx *utils.ExecutionContext) (string, bool) {
```
bacherfl, line 56: should we return an error instead of a bool value here? This would make it a little more explicit/easier for the caller to recognize that something went wrong. If we decide to keep the bool return, let's add a short description to the comment above when the returned value is set to false

warber: agree, lets do it the `err` way

bacherfl (approved):

PR #7275, opened 2022-03-25
fix(bridge): Add latestEvaluationTrace to every stage

### LINKED ISSUE #7128: Environment screen does not display evaluation result (for staging)
**Environment**

* Keptn Version (`keptn version`): `0.13.2`

**Affected Component**

* [x] Bridge

**Describe the bug**

I triggered a delivery sequence that has an evaluation task in `dev` and `staging`. Both evaluations were executed, but just the evaluation in `dev` is displayed here: 

image

See here for the executed evaluations: 

image

image

**Expected behavior**
* I see the evaluation result also in staging

laneli:
Fixes #7128 

The problem here is, that if we check if the `cachedSequence` exists, the property `latestEvaluationTrace` is set for the first stage but then will never be set for every following stage, as the sequence is already cached.
I removed the check for the `cachedSequence` and added one to check if the `latestEvaluationTrace` already exists for this particular stage. Then we don't have to set it again.

### DIFF
--- a/bridge/server/services/data-service.ts
+++ b/bridge/server/services/data-service.ts
@@ -319,14 +319,13 @@ export class DataService {
     service.latestSequence = latestSequence ? Sequence.fromJSON(latestSequence) : undefined;
     service.latestSequence?.reduceToStage(options.stageName);
 
-    if (!cachedSequence && service.latestSequence) {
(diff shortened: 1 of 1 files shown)


ermin-muratovic (approved):
lgtm

PR #7287, opened 2022-03-28
fix(bridge): Do not send a start date for evaluation if none is given by the user

### LINKED ISSUE #7268: Trigger sequence: Timeframe is not sent, instead start / end date is set
**Affected Component**

**Describe the bug**
For triggering an evaluation, when I put in 100h 50s and no start date. The evaluation fails with following error:
`no evaluation performed by lighthouse because SLI failed with message error validating time range: Supplied end-time 2022-03-24 13:29:05.867 +0000 UTC is too far (>120seconds) in the future (now: 2022-03-24 12:28:17.472012479 +0000 UTC m=+257312.515908793 - diff in sec: -3648.394987521)
`

This is what was sent in the event payload for the event:
```"evaluation": {
      "end": "2022-03-24T13:29:05.867Z",
      "start": "2022-03-24T12:28:15.867Z",
      "timeframe": ""
    },
```

It seems, that no timeframe was set at all, just the start and end date. There should be the timeframe 100h50s and the start date in the payload.

**To Reproduce**
* Go to Bridge
* Trigger a new sequence
* Trigger an evaluation
* Put in for timeframe 100h50s (or any other)
* Trigger sequence and observe started sequence + payload

**Expected behavior**
The timeframe is sent in the payload.

**Current behavior**
Start and end date are sent as payload.

laneli:
Fixes #7268 

I changed the endpoint to POST `/event` instead of POST `project/${project}/stage/${stage}/service/${service}/evaluation` to align the evaluation payload with what is used by the CLI.

The behavior is be the following:
* If a timeframe and no start date is given, we also do not send a start date. Lighthouse-service is then taking the current date and performs an evaluation to the past.
* If a timeframe and a start date is given, the backend calculates the end date from it. If the end date lies in the future, the error from dynatrace-service `no evaluation performed by lighthouse because SLI failed with message error validating time range` is expected. It was aligned, that no validation in the frontend is needed in this case.

### DIFF
--- a/bridge/client/app/_components/ktb-trigger-sequence/ktb-trigger-sequence.component.spec.ts
+++ b/bridge/client/app/_components/ktb-trigger-sequence/ktb-trigger-sequence.component.spec.ts
@@ -268,48 +268,6 @@ describe('KtbTriggerSequenceComponent', () => {
     });
   });
 
-  it('should trigger the evaluation sequence with a timeframe set', () => {
-    // given
-    const dataService = TestBed.inject(DataService);
-    const spy = jest.spyOn(dataService, 'triggerEvaluation');
-    const date = moment();
-    component.sequenceType = TRIGGER_SEQUENCE.EVALUATION;
-    component.selectedStage = 'hardening';
-    component.selectedService = 'helloservice';
-    component.evaluationFormData = {
-      evaluationType: TRIGGER_EVALUATION_TIME.TIMEFRAME,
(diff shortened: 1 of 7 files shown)


ermin-muratovic (approved):
lgtm

ermin-muratovic (changes requested):
please fix unit tests

ermin-muratovic (approved):
lgtm

PR #7288, opened 2022-03-28
fix: Correct passing of projectNameMaxSize helm value with quotes

warber:
Pass helm value for max project name size as string


installer/manifests/keptn/charts/control-plane/templates/shipyard-controller.yaml
```diff
@@ -94,7 +94,7 @@ spec:
               value: {{ .Values.shipyardController.config.disableLeaderElection | default false | quote }}
               {{- end}}
             - name: PROJECT_NAME_MAX_SIZE
-              value: {{ .Values.shipyardController.config.validation.projectNameMaxSize }}
+              value: {{ .Values.shipyardController.config.validation.projectNameMaxSize | quote }}
```
mowies, line 97: please also provide a default value here

PR #7305, opened 2022-03-30
ci: Remove master branch restriction from re-used workflows

mowies:
## This PR

- re-uses different workflows from the current branch instead of always using master


christian-kreuzberger-dtx (approved):
LGTM

PR #7306, opened 2022-03-30
ci: More pipeline metrics

mowies:
### This PR
- adds 2 more metrics to CI and integration test pipeline
- the 2 metrics are
  - total runtime of the workflow run
  - total job count that was executed in the workflow run


.github/workflows/CI.yml
```diff
@@ -463,3 +463,24 @@ jobs:
...
+    needs:
+      - calculate-queue-time
+      - prepare_ci_run
+      - store-output-in-build-config
+      - unit-tests-microservices
+      - unit-tests-cli
+      - build-cli
+      - helm_charts_build
+      - helm_charts_upload
+      - docker_build
```
christian-kreuzberger-dtx, line 479: this is probably hard to maintain :D 

But after some research, I also didn't find a better solution, so... it's the teams decision if you want that

christian-kreuzberger-dtx, line 479: this is probably hard to maintain :D 

But after some research, I also didn't find a better solution, so... it's the teams decision if you want that

.github/workflows/CI.yml
```diff
@@ -463,3 +463,24 @@ jobs:
             gitSha=${{ env.GIT_SHA }}
           push: ${{ matrix.config.should-push-image == 'true' && (( github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository ) || ( github.event_name == 'push' )) && github.actor != 'renovate[bot]' && github.actor != 'dependabot[bot]' }}
           pull: true
+
+  calculate-total-runtime:
+    name: End-of-Pipeline Metrics
+    if: always()
```
christian-kreuzberger-dtx, line 469: This step will probably fail for PRs from forks - should check that in the if.

christian-kreuzberger-dtx, line 469: This step will probably fail for PRs from forks - should check that in the if.

.github/workflows/integration_tests.yml
```diff
@@ -1093,3 +1093,20 @@ jobs:
...
+    needs:
+      - calculate-queue-time
+      - integration-test
+      - publish-final-test-report
+      - deploy_keptn_with_keptn
+      - trigger_keptn_delivery
+    uses: ./.github/workflows/end-of-pipeline-metrics.yml
+    with:
+      workflow_name: "Integration Tests"
+      branch: ${{ github.ref }}
```
christian-kreuzberger-dtx, line 1109: do we have `github.ref` when this is triggered manually?

mowies: yes, this should always be defined

.github/workflows/CI.yml
mowies: aha! good catch, i will fix that

.github/workflows/CI.yml
mowies: i did the same thing and realised that this is kind of the only way to ensure that this job always runs last

.github/workflows/CI.yml
mowies: done

christian-kreuzberger-dtx (approved):
LGTM

PR #7307, opened 2022-03-30
fix(cli): Cleaned up Oauth command  

### LINKED ISSUE #7222: CLI: Polishing OAuth implementation
## Action items

The following items only apply to the case where the CLI is used with OAuth 2.0. 

- [x] Given the user is not authenticated, when it runs `keptn get services --project=podtato`, then it gets the following message `You are not authenticated. Use the keptn auth command to authenticate based on the API token or an OAuth client.`. Currently, it receives a general 401 error.
- [x] Given the user is not authenticated, when it runs the `keptn auth --oauth ...` command, then the CLI should not perform the version checks.
- [x] Given the user is not authenticated, when it runs the `keptn auth --oauth ...` command the CLI prints the following message: 
```
Successfully authenticated against the Keptn cluster https://mykeptn.url/api
Bridge URL: https://mykeptn.url/bridge
WARNING: You are using the OAuth integration feature without a Keptn API Token. Please verify that your configuration allows it
```
Instead, it should print the following:
```
Successfully authenticated against: https://mykeptn.url/api
Bridge URL: https://mykeptn.url/bridge
```

- [x] When the CLI correctly authenticates a user with OAuth, the browser page/tab should automatically close.
- [x] When the CLI receives a 403 error, then it should print `You do not have enough permissions. This for all commands.
- [x] When the user runs `keptn auth --oauth-logout`, then the CLI should print `Successfully logged out`. 

##  DoD
- Manual tests

RealAnna:
## This PR

- [x] Given the user is not authenticated,  `keptn get services --project=podtato` returns `You are not authenticated. Use the keptn auth command to authenticate based on the API token or an OAuth client.`
- [x] Given the user is not authenticated, when it runs the `keptn auth --oauth ...` command, then the CLI does not perform the version checks.
- [x] Given the user is not authenticated, when it runs the `keptn auth --oauth ...` command the CLI prints the following:
```
Successfully authenticated against: https://mykeptn.url/api
Bridge URL: https://mykeptn.url/bridge
```

- [x] When the CLI correctly authenticates a user with OAuth, the browser page/tab automatically closes.
- [x] When the CLI receives a 403 error, CLI  prints `You do not have enough permissions. This for all commands.
- [x] When the user runs `keptn auth --oauth-logout`, then the CLI prints `Successfully logged out`. 

### Related Issues

Fixes #7222

### How to test

before  running auth run  keptn get services --project=podtato , run oauth command as described in https://keptn.sh/docs/0.13.x/operate/user_management/microsoft/ 

https://github.com/keptn/keptn/runs/5755512785?check_suite_focus=true

### DIFF
--- a/cli/cmd/auth.go
+++ b/cli/cmd/auth.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"errors"
 	"fmt"
+	"github.com/keptn/keptn/cli/internal"
 	"net/http"
 	"net/url"
 	"os"
@@ -91,10 +92,12 @@ keptn auth --skip-namespace-listing # To skip the listing of namespaces and use 
 			if err != nil {
 				return err
 			}
+			fmt.Println("Successfully logged out")
 			return nil
 		}
 
 		if *authParams.oauth {
+
 			if *authParams.oauthDiscovery == "" {
 				return fmt.Errorf("Unable to login: No OAuth Discovery URL provided")
 			}
@@ -113,14 +116,15 @@ keptn auth --skip-namespace-listing # To skip the listing of namespaces and use 
 				return err
 			}
 		}
+
 		return authenticator.Auth(AuthenticatorOptions{Endpoint: *authParams.endPoint, APIToken: *authParams.apiToken})
 	},
 }
 
 func init() {
+
 	rootCmd.AddCommand(authCmd)
 	authParams = &authCmdParams{}
-
 	authParams.endPoint = authCmd.Flags().StringP("endpoint", "e", "", "The endpoint exposed by the Keptn installation (e.g., api.keptn.127.0.0.1.xip.io)")
 	authParams.apiToken = authCmd.Flags().StringP("api-token", "a", "", "The API token to communicate with the Keptn installation")
 	authParams.exportConfig = authCmd.Flags().BoolP("export", "c", false, "To export the current cluster config i.e API token and Endpoint")
@@ -155,7 +159,7 @@ func verifyAuthParams(authParams *authCmdParams, smartKeptnAuth smartKeptnAuthPa
 	if !*authParams.skipNamespaceListing && (authParams.endPoint == nil || *authParams.endPoint == "") && (authParams.apiToken == nil || *authParams.apiToken == "") {
 		namespace, err = smartKeptnCLIAuth()
 		if err != nil {
-			return err
+			return internal.OnAPIError(err)
 		}
 	}
 
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -146,7 +146,6 @@ func runVersionCheck(vChecker *version.VersionChecker, flags []string, cliConfig
 // skipVersionCheck checks if the subcommand is `install` or `--oauth`
 // args here does not contain the main command
 // e.g., For `keptn -q install`, args would be just ['-q', 'install']
-//
 func skipVersionCheck(args []string) bool {
 	for _, arg := range args {
 		switch {
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -2,13 +2,12 @@ package cmd
 
 import (
 	"fmt"
-	"os"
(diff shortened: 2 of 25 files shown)


cli/cmd/root.go
```diff
@@ -88,10 +88,11 @@ func initConfig() {
 	logging.PrintLog(fmt.Sprintf("Using config file: %s", cfgMgr.CLIConfigPath), logging.VerboseLevel)
 
 	var err error
-	rootCLIConfig, err = cfgMgr.LoadCLIConfig()
+	configCLI, err := cfgMgr.LoadCLIConfig()
```
warber, line 91: why doing it this way?

cli/cmd/auth.go
```diff
@@ -91,10 +92,12 @@ keptn auth --skip-namespace-listing # To skip the listing of namespaces and use
 			if err != nil {
 				return err
 			}
+			fmt.Println("Successfully logged out")
```
warber, line 95: delete this or not?

cli/cmd/root.go
```diff
@@ -148,14 +145,16 @@ func runVersionCheck(vChecker *version.VersionChecker, flags []string, cliConfig
 	}
 }
 
-// isInstallSubCommand checks if the subcommand is `install`
+// skipVersionCheck checks if the subcommand is `install` or `--oauth`
 // args here does not contain the main command
 // e.g., For `keptn -q install`, args would be just ['-q', 'install']
-func isInstallSubCommand(args []string) bool {
+//
```
warber, line 151: pls delete this and empty line after

cli/cmd/auth.go
RealAnna: I was requested to add this log, do you mean since it does not work? 😆

cli/internal/api.go
```diff
@@ -67,3 +71,20 @@ func getAPISetWithOauthGetter(baseURL string, keptnXToken string, oauthAuthentic
 	}
 	return apiutils.New(baseURL, apiutils.WithAuthToken(keptnXToken), apiutils.WithHTTPClient(client))
 }
+
+func OnAPIError(err error) error {
+	switch 0 {
+	case compareError(err, ErrWithStatusCode, 401):
+		return fmt.Errorf("You are not authenticated. Use the keptn auth command to authenticate based on the API token or an OAuth client.")
+	case compareError(err, ErrWithStatusCode, 403):
+		return fmt.Errorf("You do not have enough permissions. This for all commands")
```
odubajDT, line 80: this sentence does not make sense :)

cli/cmd/root.go
```diff
@@ -148,14 +143,14 @@ func runVersionCheck(vChecker *version.VersionChecker, flags []string, cliConfig
 	}
 }
 
-// isInstallSubCommand checks if the subcommand is `install`
+// skipVersionCheck checks if the subcommand is `install` or `--oauth`
 // args here does not contain the main command
 // e.g., For `keptn -q install`, args would be just ['-q', 'install']
-func isInstallSubCommand(args []string) bool {
+func skipVersionCheck(args []string) bool {
```
odubajDT, line 149: this is a bad name for function checking if the subcommand is install or --oauth

cli/internal/api.go
RealAnna: well found :)

thisthat, line 80:
```suggestion
		return fmt.Errorf("You do not have enough permissions.")
```

cli/cmd/root.go
warber: ... on the other hand, the sole purpose of this func is to determine whether we should skip the version check or not ;)

cli/cmd/auth.go
warber: no, i just thought that might be a "debug fmt" you forgot to delete :D

odubajDT (approved):

odubajDT (approved):

PR #7310, opened 2022-03-30
chore(cli): Polish upgrade message when no upstream is present

thisthat:
Before:
pre

After the proposed changes:
after


mowies (approved):
LGTM

odubajDT (approved):

PR #7296, opened 2022-03-29
feat(distributor): Added preamble to distributor logs

warber:
closes #7027 

Distributor is printing the following during startup:
image

PR #7314, opened 2022-03-31
ci: Pass datetime and git sha to docker build in release pipelines

warber:
This PR passes datetime and git sha to the docker build also in the release workflow like it is already done for the ci workflow.
part of  #7296

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -128,6 +128,18 @@ jobs:
           fetch-depth: 0
           token: ${{ secrets.KEPTN_BOT_TOKEN }}
 
+      - name: Extract branch name
+        id: extract_branch
+        # see https://github.com/keptn/gh-action-extract-branch-name for details
+        uses: keptn/gh-action-extract-branch-name@main
+
+      - name: Get current date and time
+        id: get_datetime
+        run: |
+          echo "::set-output name=DATE::$(date +'%Y%m%d')"
+          echo "::set-output name=TIME::$(date +'%H%M')"
+          echo "::set-output name=DATETIME::$(date +'%Y%m%d')$(date +'%H%M')"
+
       - name: Find next version number
         id: version_number
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/pre-release.yml
```diff
@@ -217,6 +229,9 @@ jobs:
       - id: docker_build_image
         name: "Docker Build keptn/${{ matrix.config.artifact }}"
         uses: docker/build-push-action@v2
+        env:
+          DATETIME: ${{ steps.get_datetime.outputs.DATETIME }}
+          GIT_SHA: ${{ steps.extract_branch.outputs.GIT_SHA }}
```
mowies, line 234: When relying on data from a previous job you need to reference the outputs in a different way:

```suggestion
          DATETIME: ${{ needs.prepare.outputs.datetime }}
          GIT_SHA: ${{ needs.prepare.outputs.git_sha }}
```

mowies: and then you also need to add the outputs to the `prepare` job like so:
```
prepare:
  outputs:
    datetime: ${{ steps.get_datetime.outputs.DATETIME }}
    git_sha: ${{ steps.extract_branch.outputs.GIT_SHA }}
```

.github/workflows/release.yml
```diff
@@ -214,6 +226,9 @@ jobs:
       - id: docker_build_image
         name: "Docker Build keptn/${{ matrix.config.artifact }}"
         uses: docker/build-push-action@v2
+        env:
+          DATETIME: ${{ steps.get_datetime.outputs.DATETIME }}
+          GIT_SHA: ${{ steps.extract_branch.outputs.GIT_SHA }}
```
mowies, line 231: same as for the other pipeline :)

mowies (changes requested):

.github/workflows/pre-release.yml
```diff
@@ -225,6 +242,8 @@ jobs:
             ghcr.io/keptn/${{ matrix.config.artifact }}:${{ env.VERSION }}
           build-args: |
             version=${{ env.VERSION }}
+            buildTime=${{ env.DATETIME }}
+            gitSha=${{ env.GIT_SHA }}
```
mowies, line 246: you can directly use the outputs here without the need to first set them in the env above
```suggestion
            buildTime=${{ needs.prepare.outputs.datetime }}
            gitSha=${{ needs.prepare.outputs.git_sha }}
```

mowies (approved):
LGTM

Issue #6439, opened 2021-12-28
Refactor 'keptn upgrade' CLI command code


odubajDT:
The 'keptn upgrade' CLI command code needs to be refactored to make it more testable. For example `install.go` representing code for CLI command 'keptn install' has an `InstallCmdHandler{}` struct where components for retrieving information from the API can be injected and mocked during the unit tests - the changes in the `upgrade.go` would also benefit from that approach since the `addWarningNonExistingProjectUpstream()` function would not have to initialize the `credentialsManager` and `projectHandler` within the function anymore.

## DoD 

- [ ] Refactor the code to support dependency injection to enable easier mocking of API 
- [ ] Remove the `if mocking` checks
- [ ] Updating the existing tests to properly mock the API endpoints

eyskim:
Hi @odubajDT,

I wanted to attempt this refactoring and have come up with the following approach:
1. Refactor `upgrade.go` and `upgrade_test.go` with a new `UpgradeCmdHandler` struct with dependencies similar to what is used in `install.go`'s `InstallCmdHandler`.
2. Add dependencies to `UpgradeCmdHandler` as required by `addWarningNonExistingProjectUpstream()` and further refactor/add unit tests.

These steps might result in a struct that looks similar to:
```
type UpgradeCmdHandler struct {
	helmHelper        helm.IHelper
	namespaceHandler  kube.IKeptnNamespaceHandler
	userInput         common.IUserInput
        credentialManager credentialmanager.CredentialManager
        projectHandler    api.ProjectsV1Interface
}
```

With the above in mind, I was wondering how I should handle the fact that `ProjectsV1Interface.GetAllProjects()` ([link to code](https://github.com/keptn/keptn/blob/b3e9ceff7355ca8b18432a97b47d524cfee21e84/cli/cmd/upgrade.go#L265-L268)) depends on the prior execution of `CredentialManager.GetCreds(namespace)` ([link to code](https://github.com/keptn/keptn/blob/b3e9ceff7355ca8b18432a97b47d524cfee21e84/cli/cmd/upgrade.go#L253-L256)) to add the auth headers to the `APISet` ([link to code](https://github.com/keptn/keptn/blob/b3e9ceff7355ca8b18432a97b47d524cfee21e84/cli/cmd/upgrade.go#L258-L261)) in `addWarningNonExistingProjectUpstream()`?

In other words, the eventual `projectHandler` will depend on the prior execution of methods available to `credentialManager`.

[...]

PR #7316, opened 2022-03-31
fix(cli): Provide values needed for upgrading the nats dependency

### LINKED ISSUE #7315: CLI: Upgrade from 0.13.x to 0.14.x does not configure Nats properly
When upgrading Keptn using the CLI, which internally uses the `--reuse-values`, Jetstream is not enabled on the nats cluster, because in previous versions of Keptn, the value `nats.jetstream.enabled` (which is used by the nats dependency) was set to `false` by default. Also, the image tags for nats and the prometheus metrics exporter will still have the same value as in previous versions, which will lead to incompatibilities with the configuration of the recently upgraded nats helm chart.

Therefore, we need to ensure the following values have the correct values, even when upgrading with the `--reuse-values` flag:

`control-plane.nats.nats.jetstream.enabled`
`control-plane.nats.nats.image`
`control-plane.nats.exporter`

bacherfl:
Closes #7315
This PR makes sure the values provided to the nats dependency are compatible with the currently used nats helm chart. Unfortunately this meas that in this case we also need to set the image of nats and the nats metrics exporter, since those are defined via the values file of the nats helm chart, and the values that have been used by the previous version are incompatible with the current chart.

For adding unit tests, a bigger refactoring will be needed, as outlined in https://github.com/keptn/keptn/issues/6439

### DIFF
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -325,24 +325,12 @@ func doUpgrade() error {
 	// if yes, they need to be installed separately as they have moved to their own charts
 	helmHelper := helm.NewHelper()
 
-	// this object holds values that were not set in previous versions of Keptn and which need to be explicitly set if using the --reuse-values flag
-	newVals := map[string]interface{}{
-		"control-plane": map[string]interface{}{
-			"nats": map[string]interface{}{
-				"nats": map[string]interface{}{
-					"image": "nats:2.7.2-alpine",
-					"jetstream": map[string]interface{}{
-						"enabled": true,
-					},
-				},
-				"exporter": map[string]interface{}{
-					"image": "natsio/prometheus-nats-exporter:0.9.1",
-				},
-			},
-		},
+	previousValues, err := helmHelper.GetValues(keptnReleaseName, keptnNamespace)
+	if err != nil {
+		return fmt.Errorf("Could not complete Keptn upgrade: %s", err.Error())
 	}
 
-	if err := helmHelper.UpgradeChart(keptnUpgradeChart, keptnReleaseName, keptnNamespace, newVals); err != nil {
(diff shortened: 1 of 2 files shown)


cli/cmd/upgrade.go
```diff
@@ -325,7 +325,14 @@ func doUpgrade() error {
 	// if yes, they need to be installed separately as they have moved to their own charts
 	helmHelper := helm.NewHelper()
 
-	if err := helmHelper.UpgradeChart(keptnUpgradeChart, keptnReleaseName, keptnNamespace, nil); err != nil {
+	// fetch user-defined values of the previous Keptn installation and provide those to the upgrade command
+	// this will ensure that options such as 'apiGatewayNginx.type' will stay the same, but newly introduced values will correctly be set to their default
+	previousValues, err := helmHelper.GetValues(keptnReleaseName, keptnNamespace)
```
odubajDT, line 330: do we have a unit test for this ? If no, it is possible to add it ?

cli/cmd/upgrade.go
```diff
@@ -325,7 +325,14 @@ func doUpgrade() error {
...
+	if err != nil {
+		return fmt.Errorf("Could not complete Keptn upgrade: %s", err.Error())
+	}
+
+	if err := helmHelper.UpgradeChart(keptnUpgradeChart, keptnReleaseName, keptnNamespace, previousValues); err != nil {
```
odubajDT, line 335: also a unit test for this ?

cli/cmd/upgrade.go
bacherfl: in both cases a bigger refactoring would be needed, so we decided to create a dedicated issue for that back then: https://github.com/keptn/keptn/issues/6439

odubajDT: fine by me :)

odubajDT (approved):

thisthat (approved):
Tested locally and it works

PR #7318, opened 2022-03-31
fix(bridge): Do not send a start date for evaluation if none is given by the user

laneli:
Backport from #7268 to 0.14.x
Details in PR: #7287


ermin-muratovic (approved):
lgtm

PR #7325, opened 2022-03-31
feat(secret-service): provide HTTP 400 when scope is not found

thisthat:
## This PR

- When the scope is not found, the APIs return an HTTP 400 status code

Integration test run: https://github.com/keptn/keptn/actions/runs/2073164360

### DIFF
--- a/secret-service/pkg/backend/secretbackend_k8s.go
+++ b/secret-service/pkg/backend/secretbackend_k8s.go
@@ -27,27 +27,27 @@ var ErrSecretNotFound = errors.New("secret not found")
 var ErrTooBigKeySize = errors.New("name and key values must be no more than 253 characters")
 var ErrScopeNotFound = errors.New("")
 
-// ScopeErrorNotFound provides a way to better describe the error when a scope is not found.
-type ScopeErrorNotFound struct {
+// scopeErrorNotFound provides a way to better describe the error when a scope is not found.
+type scopeErrorNotFound struct {
 	Scope  string
 	Secret string
 	Err    error
 }
 
-func NewScopeErrorNotFound(scope, secret string) *ScopeErrorNotFound {
-	return &ScopeErrorNotFound{
+func NewScopeErrorNotFound(scope, secret string) *scopeErrorNotFound {
+	return &scopeErrorNotFound{
 		Scope:  scope,
 		Secret: secret,
 		Err:    ErrScopeNotFound,
 	}
 }
 
-func (ve ScopeErrorNotFound) Error() string {
-	return fmt.Sprintf("scope %s not available for creation of Secret %s", ve.Scope, ve.Secret)
+func (se scopeErrorNotFound) Error() string {
+	return fmt.Sprintf("scope %s not available for creation of Secret %s", se.Scope, se.Secret)
 }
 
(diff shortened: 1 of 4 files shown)


secret-service/pkg/backend/secretbackend_k8s_test.go
```diff
@@ -422,6 +447,23 @@ func TestUpdateSecret_SecretNotFound(t *testing.T) {
 	assert.Equal(t, ErrSecretNotFound, err)
 }
 
+func TestUpdateSecret_ScopeNotFound(t *testing.T) {
```
odubajDT, line 450: Does it makes sense to to create a similar unit test also for creating a secret ?

odubajDT: ah sorry, It already exists, that's why I was missing it :)

odubajDT (approved):

secret-service/pkg/backend/secretbackend_k8s.go
```diff
@@ -24,6 +25,30 @@ const SecretServiceName = "keptn-secret-service"
 var ErrSecretAlreadyExists = errors.New("secret already exists")
 var ErrSecretNotFound = errors.New("secret not found")
 var ErrTooBigKeySize = errors.New("name and key values must be no more than 253 characters")
+var ErrScopeNotFound = errors.New("")
```
warber, line 28:
```suggestion
var ErrScopeNotFound = errors.New("scope not found")
```

secret-service/pkg/backend/secretbackend_k8s.go
```diff
@@ -24,6 +25,30 @@ const SecretServiceName = "keptn-secret-service"
...
+	}
+}
+
+func (se scopeErrorNotFound) Error() string {
+	return fmt.Sprintf("scope %s not available for creation of Secret %s", se.Scope, se.Secret)
+}
+
+func (se scopeErrorNotFound) Unwrap() error {
+	return se.Err
+}
```
warber, line 51: you can delete all of this

secret-service/pkg/backend/secretbackend_k8s.go
```diff
@@ -46,7 +71,7 @@ func (k K8sSecretBackend) checkScopeDefined(secret model.Secret) (model.Scopes,
 	}
 	if _, ok := scopes.Scopes[secret.Scope]; !ok {
 		log.Errorf("Unable to find scope %s for secret %s", secret.Scope, secret.Name)
-		return model.Scopes{}, fmt.Errorf("scope %s not available for creation of Secret %s", secret.Scope, secret.Name)
+		return model.Scopes{}, NewScopeErrorNotFound(secret.Scope, secret.Name)
```
warber, line 74:
```suggestion
		return model.Scopes{}, fmt.Errorf("unable to check defined scope %s for secret %s: %w", secret.Scope, secret.Name, ErrScopeNotFound)
```

warber (commented):
I think introducing a `scopeErrorNotFound` struct is not really necessary. The same can be accomplished with regular sentinel errors much more neatly, imo. see comments. :)

warber (approved):

PR #7329, opened 2022-04-01
fix(webhook-service): Implement v1beta1 webhook version support

### LINKED ISSUE #7238: Core: webhook-service improve configuration
## User Story

As a user, I would like to have more guidance in defining a webhook configuration.

## Acceptance Criteria

- The webhook configuration should be updated at version `v1alpha1` to `v1beta1`
- The webhook configuration should have the following parameters:
   - `URL`: string
   - `headers`: list of key-value pairs
   - `payload`: string
   - `options`: string

## DoD

- The webhook can deal with both configuration versions
- Unit tests

odubajDT:
## This PR

- implements new version of webhook configuration

### Related Issues

Fixes #7238


webhook-service/lib/webhook_config.go
```diff
@@ -23,24 +25,42 @@ type WebHookConfigSpec struct {
...
-	SendFinished   bool      `yaml:"sendFinished"`
-	SendStarted    *bool     `yaml:"sendStarted,omitempty"`
-	EnvFrom        []EnvFrom `yaml:"envFrom"`
-	Requests       []string  `yaml:"requests"`
+	Type           string        `yaml:"type"`
+	SubscriptionID string        `yaml:"subscriptionID"`
+	SendFinished   bool          `yaml:"sendFinished"`
+	SendStarted    *bool         `yaml:"sendStarted,omitempty"`
+	EnvFrom        []EnvFrom     `yaml:"envFrom"`
+	Requests       []interface{} `yaml:"requests"`
```
warber, line 33: why interface{}

odubajDT: we currently need to support both versions of webhook-configuration (`alpha1` and `beta1`). In `alpha1` the parameter is a `[]string`, in `beta1` the parameter is a `[]Request`. If we do not want to create 2 separate structures for this, it needs to be done this way (at least I don't know about any cleaner way). Do you have some suggestions how to make it better ?

warber: I'd suggest to introduce different packages (alpha1, beta1) each just containing the exact model that is used for that version. Like it's done in e.g. kubernetes source: https://github.com/kubernetes/api/tree/master/apps

odubajDT: After a discussion with @warber we came to a conclusion that there is no simple way to implement this feature with a different packages. It will make more problems than advantages. From this reason we are sticking for the current implementation for now.

RealAnna (approved):
LGTM

PR #7336, opened 2022-04-03
feat(installer): Add option to define nodeSelector globally or per service

### LINKED ISSUE #2887: Add nodeSelector attribute to linux-based container manifests
It is always frustrating when I am hosting a mixed-pool cluster containing both Windows and Linux nodes, and a helm install (such as the [keptn quickstart](https://keptn.sh/docs/quickstart/#3-install-keptn-and-authenticate-keptn-cli)) does not target appropriate nodes. This results in a broken deployment as the scheduler allocates linux containers to windows node pools, causing the quickstart installer to time out/hang indefinitely.

Please consider adding a [nodeSelector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) targeting the appropriate os (`kubernetes.io/os: linux`)

It is possible to run `helm template` to extract the helm chart and then manually patch using kustomize, but this is both time consuming and fragile when the underlying helm template(s) are updated

**Additional context**
N/A

markuslackner:
## This PR

Adds the possibility to set nodeSelectors for all keptn pods

### Related Issues

Fixes #2887

### Notes

Although the DoD in issue #2887 states that the default nodeSelector `kubernetes.io/os: linux` should be set, i rather would prefer to set no default, because it seems to be best practice in most of the charts (for example also nats and mongodb)

After this PR you could set nodeSelectors quite easily. For example to set `kubernetes.io/os: linux` for control-plane use:

```yaml
control-plane:
  common:
    nodeSelector:
      kubernetes.io/os: linux
  nats:
    nodeSelector:
      kubernetes.io/os: linux
  mongo:
    nodeSelector:
      kubernetes.io/os: linux
```

Additionally you could also overwrite nodeSelectors per service.

### How to test

Without setting any values for nodeSelectors nothing should change (the output of `helm template .` should be the same before and after merging this PR). To see nodeSelectors use `helm template --set control-plane.common.nodeSelector.testkey=testvalue`


installer/manifests/keptn/templates/_helpers.tpl
```diff
@@ -71,3 +71,47 @@ resources:
     memory: "32Mi"
     cpu: "100m"
 {{- end }}
+
+{{/*
+Renders nodeSelector if either "value" or "default" is not empty. the needed indentation must be set
+with "indent". Templates may be used in values and keys!
```
mowies, line 77: Is there a reason why you put the indentation into the helper function instead of using it without it and then using the `nindent` or `indent` function outside?

then instead of this:
```
{{- include "keptn.nodeSelector" (dict "value" .Values.statisticsService.nodeSelector "default" .Values.common.nodeSelector "indent" 6 "context" . )}}
```
you would use:
```
{{- include "keptn.nodeSelector" (dict "value" .Values.statisticsService.nodeSelector "default" .Values.common.nodeSelector "context" . ) | indent 6 }}
```

I am not fully sure if this would actually work though...

markuslackner: The reason behind is, that i wanted to avoid all additional/senseless whitespace. 

The template
```
      terminationGracePeriodSeconds: 30
      {{- include "keptn.nodeSelector" (dict "value" .Values.statisticsService.nodeSelector "default" .Values.common.nodeSelector "context" . ) | indent 6 }}
```
with unset nodeSelector renders to
```
      terminationGracePeriodSeconds: 30< 6 Spaces>   
```

You could also avoid that with an if around the include (but i also wanted to avoid duplicate code, because that "if" would be needed for each deployment)

I think that all 3 solutions are completely valid! 

1. current solution (avoids whitespace but introduces `indent` parameter for template function)
2. applying `indent` outside template function (renders 6 space chars when no nodeSelector is set)
3. solution 2 + additional if around template function call (no whitespace but more code)

I would be happy to adapt the code to the preferred solution.

mowies: In that I would say you already chose the best solution :)
Let's leave it the way it is. That will result in the cleanest helm template and also less code to maintain

mowies (approved):
LGTM! Great work :)

PR #7349, opened 2022-04-04
fix(api): Metadata model update

### LINKED ISSUE #7346: Duplication of metadata models property
We have a metadata model present in api-service and in go-utils as well. There is no reason, why these properties shouldn't be merged together to be easily maintainable.

odubajDT:
## This PR

- Remove duplicated metadata model

### Related Issues

Fixes #7346


RealAnna (approved):
lgtm

PR #7361, opened 2022-04-05
test(bridge): Fix flaky sequence UI test

Kirdock:
Some UI tests for the sequence screen may fail if a sequence is selected because Angular throws an error `expression changed after it has been checked`. This is because the time for `Last time fetched` is `undefined` at the start until the traces are loaded and then the current date is returned, which changes on every call.
Therefore if the date is undefined `Last time fetched: never` is displayed instead of the current date.


ermin-muratovic (approved):
lgtm

Issue #3409, opened 2021-03-01
Support for multi-arch images of keptn components (ARM)


imrajdas:
**Is your feature request related to a problem? Please describe.**
Currently, keptn docker images are not built for multi-arch support.

**Describe the solution you'd like**
Add arm64 and amd64 support for the docker images using `docker buildx`. Currently, amd64 is supported, but not arm64

**Describe alternatives you've considered**

**Additional context**

rajdas98:
cc: @jetzlstorfer

christian-kreuzberger-dtx:
I've already looked at this at
* https://github.com/keptn/gh-action-build-docker-image/issues/1
* and PR https://github.com/keptn/keptn/pull/3055

At this point it's not feasible to build ARM containers/images using `docker buildx`, as the build process takes multiple times longer using `buildx` (we are talking several hours instead of 30 minutes).

If we get an ARM/ARM64 based GH Action Runner the story might be different.

rajdas98:
> If we get an ARM/ARM64 based GH Action Runner the story might be different.

Is it taking the same time after making the pipelines parallel for AMD and ARM?

christian-kreuzberger-dtx:
Due to our mono-repo our CI already consists of many parallel steps (unit tests, building docker images, building CLI for multiple architectures, ...). 
Adding parallel steps for Docker Builds will block other steps like unit tests and building the CLI.

In addition, just building images for ARM/ARM64 is not the whole story. Those images also need to be e2e/integration tested, etc... 
If the use-case is that someone needs a one-time build, then it's absolutely possible to do that by cloning the repo, checking out the respective branch/tag, and running `make build-docker` locally.

rajdas98:
@christian-kreuzberger-dtx, I thought the ARM support would be good for the community. Currently, I don't have this requirement. You may close this issue if you want. Thanks.

rajdas98:
One more way is to use more than one CI tool(circleCI + GitHub workflow) to reduce pipeline time. But this will create an overhead to manage two CI tool.

christian-kreuzberger-dtx:
> I thought the ARM support would be good for the community

Long-term, yes. We can leave the issue open, such that people can comment on it or upvote it if they have this requirement.

nlamirault:
Hi, any news on this feature ? It could be nice to try Keptn on K3S running on RPi 4

oleg-nenashev:
[...]

PR #7365, opened 2022-04-05
feat: [PoC] ARM64 Docker images

mowies:
Research results:
- This PR produces working ARM64 based container images for Keptn 🎉
- Keptn on ARM only works with an external MongoDB, since the Bitnami MongoDB does not have ARM support
- Cross platform docker builds are enabled through the docker-build-push GH action and QEMU
  - The docker build CI step now builds both the x86 and the ARM image in parallel and pushes both
- Dockerfiles are adjusted slightly to cross-compile from x86 to ARM. The final images then have the copied over version for the correct architecture inside
- Dockerfiles were also refactored slightly
- The installation of the `git` package was removed from most container images since it's not needed

References:
https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/
https://www.digitalocean.com/community/tutorials/how-to-build-go-executables-for-multiple-platforms-on-ubuntu-16-04
https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

PoC for #3409
Integration test run (on x86 images): https://github.com/keptn/keptn/actions/runs/2746706711

### DIFF
--- a/.github/workflows/test-and-build-docker-images.yml
+++ b/.github/workflows/test-and-build-docker-images.yml
@@ -166,6 +166,11 @@ jobs:
     steps:
     - name: Check out code
       uses: actions/checkout@v3
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v2
+      with:
+        platforms: arm64
       
     - name: Set up Docker Buildx
       id: buildx
@@ -215,6 +220,7 @@ jobs:
       uses: docker/build-push-action@v3
       with:
         context: ${{ inputs.working-dir }}
+        platforms: linux/amd64,linux/arm64
         tags: ${{ inputs.tags }}
         build-args: |
           version=${{ inputs.version }}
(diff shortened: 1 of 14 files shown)


oleg-nenashev (commented):
It looks good to me. We would need to also update some orchestration plane services to make them available for demo purposes.

Do the packages remain the same, e.g. git?

mowies:
Yes! some integrations would be great. They can probably follow the same approach the I took in this PR here.
What do you mean by packages and git? In this PR I only removed git from the builder docker stage because we don't actually need it there.

nlamirault:
Why not migrate yo official MongoDB docker image ? It's supporting arm64 

https://hub.docker.com/_/mongo

mowies:
I don't really wanna migrate away because bitnami provides good helm charts that we can use

mowies:
Retracting this, since we can't provide a full set of images with MongoDB right now.

PR #7369, opened 2022-04-05
refactor(bridge): Move static server pages to client

### LINKED ISSUE #6744: Move static OAuth pages from bridge server to bridge client
With the new OAuth flow, there's 2 pages that are served directly from the bridge server instead of having the static content in the bridge directly.
Those pages are:
- Login Error page 
- Logout page

In order to prepare for the architecture change done here #6216 those pages need to be moved into the bridge.
For this to work, we need to ensure that on those pages, no API calls happen. Otherwise we have a potential to run into a redirect loop with user clicks "logout" -> logout success page opens -> HTTP 401 from API call in bridge -> redirect to login -> login happens because user is probably already logged into their SSO provider.

## Acceptance criteria

- [x] The aforementioned pages are served by the Angular static pages
- [x] No polling or other background API calls are done on these pages

## DoD
- UI Test
- Unit Test(s)

Kirdock:
Closes #6744 
Moves `app.component.spec` UI tests to Cypress

**UI test depends on a fix in #7394**


ermin-muratovic (dismissed):
lgtm

PR #7377, opened 2022-04-06
fix:  Added correct error and test to jmeter exec

### LINKED ISSUE #7359: jmeter-service: Misleading error log and problem with v0.14.1
- [x] jmeter-service v 0.14.1

### Bug

* [ ] Jmeter-service has a misleading error log: `"time="2022-04-05T07:46:16Z" level=error msg="Successfully executed JMeter test: Project: sockshop, Service: carts, Stage: dev, TestStrategy: functional, Context: a09a2033-6ef3-4151-998f-02fc6b910466"`
due do a copy-paste from the previous line: https://github.com/keptn/keptn/blob/master/jmeter-service/jmeterUtils.go#L213

* [x] jmeter-service:0.14.1 returns a failed evaluation: `time="2022-04-05T07:52:28Z" level=info msg="Sending result Project: sockshop, Service: carts, Stage: staging, TestStrategy: performance, Context: a09a2033-6ef3-4151-998f-02fc6b910466 false"` < (`false` = test result) When changing back to 0.13.4 - everything works fine.
  * Will dig deeper to get more insights here.

### Expected behavior

- [ ] Logline: Starting with "error ... " and showing the error message
- [x] jmeter-service working again.

RealAnna:
### Related Issues

Fixes #7359

### DIFF
--- a/jmeter-service/jmeterUtils.go
+++ b/jmeter-service/jmeterUtils.go
@@ -216,7 +216,7 @@ func execute(testInfo TestInfo, workload *Workload, err error, jMeterCommandLine
 	} else {
 		logger.Errorf("Could not parse JMeter test (%v). Error: %s  ", testInfo, err.Error())
 	}
-	return result, nil
+	return result, err
 }
 
 func createDir(dir string) error {
--- a/jmeter-service/jmeterUtils.go
+++ b/jmeter-service/jmeterUtils.go
@@ -194,6 +194,10 @@ func executeJMeter(testInfo TestInfo, workload *Workload, resultsDir string, url
 	}
 	// Step 2: Lets execute the script - but be aware that we launch jmeter from the localTempDir as a working directory!
 	jMeterCommandLineArgs := addJMeterCommandLineArguments(testInfo, createJMeterCLIArguments(workload, url, resultsDir, loadTestName))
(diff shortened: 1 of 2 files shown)


jmeter-service/jmeterUtils.go
```diff
@@ -210,9 +214,9 @@ func executeJMeter(testInfo TestInfo, workload *Workload, resultsDir string, url
 	if result && err != nil {
 		logger.Debugf("Successfully executed JMeter test: %v", testInfo)
 	} else {
-		logger.Errorf("Successfully executed JMeter test: %v", testInfo)
+		logger.Errorf("Could not parse JMeter test (%v). Error: %s  ", testInfo, err.Error())
 	}
-	return result, err
+	return result, nil
```
mowies, line 219: Why do you return `nil` here? If an error actually happened, shouldn't it also be returned?
And if no error happened, `err` is `nil` anyways.

mowies (changes requested):

jmeter-service/jmeterUtils.go
RealAnna: at that point in code err is nill but if you want I can substitute it with err

mowies: But you could have an error, print the error message and then still return `nil` instead of an error object. Couldn't that be a problem?

mowies (approved):

PR #7399, opened 2022-04-07
feat(bridge): Consider real waiting state for sequences

### LINKED ISSUE #4625: Sequence endpoint: State for each stage
Because of parallel stages, we may have several states (e.g. production-A is `waiting` and production-B is `failed`) and therefore the `state` property in the sequence state is not enough to determine the status. An additional `state` property in each stage is necessary.
In addition to this the bridge needs to adapt these changes and should show the state of each stage in the list of sequences.

# Definition of done
- [ ] The `/sequence` endpoint also returns the state of each stage
- [ ] The bridge is able to show this information in the list of sequences

Kirdock:
Fixes #4625 (the stage state is already implemented, so we agreed on tackling the waiting state here)
This also introduces the waiting status for the filter.


ermin-muratovic (approved):
lgtm

Issue #7363, opened 2022-04-05
Git upstream config failed: SSL certificate problem: unable to get local issuer certificate


melcsanad:
Your issue may already be reported! Please search on the [issue tracker](https://github.com/keptn/keptn/issues) before creating a new one.

**Environment**

* Client OS (e.g., Linux, macOS, Windows): Red Hat Enterprise Linux release 8.5 (Ootpa)
* Keptn Version (`keptn version`): 0.13.4
* Kubernetes Cloud Provider (e.g., GKE, AKS): On Premise
* Kubernetes version (`kubectl version`): 1.18.15

Chrome Version 85.0.4183.83 (Official Build) (32-bit)
**Affected Component**

* [ ] Approval
* [X] Bridge
* [ ] CLI
* [ ] Datastore
* [ ] Docs
* [ ] Distributor / Event Broker
* [ ] Helm
* [ ] Istio
* [ ] Jmeter
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [ ] REST API
* [ ] Webhook

**Describe the bug**
 Try to add Git upstream repo to an exiting project. Get error message: "The Git upstream could not be canged"
Config Service pod log:
configuration-service-bb46cbdcc-x6hcs_log.txt:time="2022-04-01T11:07:00Z" level=error msg="Could not add upstream repository while updating project smarttest" error="failed to show remote origin for project 'smarttest': Error executing command git remote show origin: exit status 128\nfatal: unable to access '[https://***.***.at/scm/pult/keptn-smarttest.git/':](https://***.***.at/scm/pult/keptn-smarttest.git/%27:)
SSL certificate problem: unable to get local issuer certificate\n"

It is not possible to turn off SSL cert check or import CA certs.

**To Reproduce**

1. Go to Main screen
2. Select a project
3. Go to project configuration
4. Try to setup a Git upstream repo

**Expected behavior**

Possible to ignore SSL Cert check or import CA cert.
Add Git upsteam successfuly than!

**Current behavior**

Cannot add Git upstream to exiting project.

**Screenshots**

![keptn git](https://user-images.githubusercontent.com/98130057/161743465-f12a05e2-2d8d-4791-ba23-863a6eecdafb.jpg)

**Additional context**

thisthat:
This will be fixed by enabling the resource-service

thisthat:
With Keptn 0.16 out, we can close this issue

PR #7410, opened 2022-04-08
fix: Rename GitProxyInsecure to InsecureSkpTLS and pass it to upstream interactions

bacherfl:
Part of #7363

Manually tested as follows:

1. Enable HTTPS on gitea, according to the following documentation: https://blog.differentpla.net/blog/2022/02/04/gitea-https/ (Note: a self signed certificate and a key can easily be generated on https://certificatetools.com)
2. Create project with https upstream URL pointing to gitea and `insecureSkipTLS` set to `false`: i.e. the credentials stored in the secret should look as follows:

```
{"user":"<user>","token":"<token>","remoteURI":"https://gitea-http:3000/<user>/<project>","insecureSkipTLS":false}
```

--> This will cause the resource service to fail when trying to communicate with the upstream (e.g. when cloning the repo)

3. Create project with https upstream and `insecureSkipTLS` set to true:

```
{"user":"<user>","token":"<token>","remoteURI":"https://gitea-http:3000/<user>/<project>","insecureSkipTLS":true}
```

--> With these settings, the resource service can communicate with the upstream over https even when gitea uses a self signed certificate

Note: This is a breaking change, since the `GitProxyInsecure` parameter has been renamed to `InsecureSkipTLS`, as this flag is not intended only for proxy communication, but for https upstream urls in general

Integration test run: https://github.com/keptn/keptn/runs/5972898882?check_suite_focus=true (the failing log forwarding test on one platform is not affected by the changes made in this PR and is probably a one-off, but we need to keep an eye on it in case it fails again)


RealAnna (dismissed):
LGTM

odubajDT (dismissed):

odubajDT (approved):

Issue #7079, opened 2022-03-07
Chore: Create common module in `keptn/keptn` that contains shared core related code


warber:
## Action Item

Create a separate module inside `keptn/keptn` that contains shared code used by components of the control plane (keptn core).

## Acceptance criteria

- [x] Find the coolest name for the module
- [x] ~Create a new module with a Dockerfile for compiling the code~
- [x] Create the entry in our pipeline to run the unit tests as part of our pipeline
- [x] Provide the [APISet](https://github.com/keptn/go-utils/blob/master/pkg/api/utils/apiUtils.go#L14) ❓ 

## DoD 

- Acceptance criteria are met

PR #7418, opened 2022-04-11
feat: Initial implementation of `cp-connector` library

warber:
THis PR contains the first implementation of `cp-connector` a library that can be used instead of the distributor to implement keptn services. The first POC of using this lib with the lighthouse service can be found here #7334

Note that `internal.go`, `log.go` as well as `eventmatcher.go` was copied from existing modules, hence the code duplication warning from sonar. Eventually, this code can be moved into a "common" module, as it planned with https://github.com/keptn/keptn/issues/7079


RealAnna (commented):
now LGTM but I think @bacherfl should also check it 😄

mowies:
Maybe it would be good to already introduce a new scope for our semantic PR checking in this PR so that further features for the library are grouped in the changelog correctly.

lib-cp-connector/pkg/api/internal.go
```diff
@@ -0,0 +1,220 @@
...
+var DefaultInClusterAPIMappings = InClusterAPIMappings{
+	ConfigurationService: "configuration-service:8080",
+	ShipyardController:   "shipyard-controller:8080",
+	ApiService:           "api-service:8080",
+	SecretService:        "secret-service:8080",
+	MongoDBDatastore:     "mongodb-datastore:8080",
+}
+
+// NewInternal creates a new InternalAPISet usable for calling keptn services from within the control plane
+func NewInternal(client *http.Client, apiMappings ...InClusterAPIMappings) (*InternalAPISet, error) {
```
bacherfl, line 56: given that the package only contains this component, should we rename this to just `New()`? Or will this eventually also include an `ExternalAPISet`?

lib-cp-connector/pkg/controlplane/controlplane.go
```diff
@@ -0,0 +1,96 @@
...
+	for _, subscription := range cp.currentSubscriptions {
+		if subscription.Event == eventUpdate.MetaData.Subject {
+			matcher := NewEventMatcherFromSubscription(subscription)
+			if matcher.Matches(eventUpdate.KeptnEvent) {
+				if err := integration.OnEvent(context.WithValue(ctx, EventSenderKey, cp.eventSource.Sender()), eventUpdate.KeptnEvent); err != nil {
+					if errors.Is(err, ErrEventHandleFatal) {
+						cp.logger.Errorf("Fatal error during handling of event: %v", err)
+						return err
+					}
+					if errors.Is(err, ErrEventHandleIgnore) {
```
bacherfl, line 80: should we add a default case here as well, to also log errors other than `EventHandleFatal` and `ErrEventHandleIgnore`?

lib-cp-connector/pkg/api/internal.go
warber: It *could* be that there will be a `NewExternal` function besides this one which will just use the regular `APISet` provided by `go-utils` https://github.com/keptn/go-utils/blob/24667368a1594c17cfa725b88a97d19ddfef109e/pkg/api/utils/client.go#L146

warber: Update: agreed with @bacherfl  to keep `NewInternal` for now

lib-cp-connector/pkg/controlplane/controlplane.go
warber: also the question is, what should be the default behavior?
should the user actively need to decide that it is a fatal error or should she actively decide that it is an error that can be ignored?
i.e., **any** error will be treated to be ignoreable, while `ErrEventHandleFatal` will be treated as fatal
or
**any** error will be treated to be fatal, while `ErrEventHandleIgnore` will be treated as ignore

bacherfl:
also LGTM from my side, nice work!

mowies:
Please add the new scope for the library here: https://github.com/keptn/keptn/blob/master/.github/workflows/validate-semantic-pr.yml

thisthat (approved):

thisthat (approved):

PR #7445, opened 2022-04-12
fix(shipyard-controller): Added option to configure maximum service name length, adapted returned http status code

### LINKED ISSUE #7425: Core(shipyard-controller): Incorrect status code for POST /service
If a long service name is provided, the API responds with a 500 status code. This should be changed to a 400 status code.

TBD: How should the bridge validate the length? As for now it looks like that lengthOfProject + lengthOfService has to be lower than 56 characters.

bacherfl:
Closes #7425


mowies (commented):
Helm charts LGTM!

PR #7452, opened 2022-04-12
fix(cli): Hide warning message during upgrade

### LINKED ISSUE #7419: `keptn upgrade` should not display the warning about CLI and cluster mismatch
### Enhancement

The `keptn upgrade` command should not display the warning: *Your Keptn CLI version (0.13.1) and Keptn cluster (0.12.0) don`t match ... * because at this point of time the CLI will always be ahead of the cluster:

image (3)

### Defintion of Done: 
- [ ] `keptn upgrade` does not display the warning message.

Lucifergene:
## This PR

- Hides the warning message during `keptn upgrade`

### Related Issues

Fixes #7419

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -2,12 +2,13 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	"strings"
+
 	"github.com/keptn/keptn/cli/pkg/config"
 	"github.com/keptn/keptn/cli/pkg/logging"
 	"github.com/keptn/keptn/cli/pkg/version"
 	"github.com/spf13/cobra"
-	"os"
-	"strings"
 )
 
 var cfgFile string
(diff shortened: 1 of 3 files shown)


johannes-b:
@Lucifergene many thanks! Can you please sign your commits to make the DCO happy? Please see: https://github.com/keptn/keptn/pull/7452/checks?check_run_id=5994778217

Lucifergene:
@johannes-b Thanks! I have signed the commits!

johannes-b:
Hi @Lucifergene,

can you please provide a unit test as we discussed on Slack?

johannes-b:
Hi @Lucifergene,

can you please provide a unit test as we discussed on Slack?

Lucifergene:
> Hi @Lucifergene,
> 
> can you please provide a unit test as we discussed on Slack?

Sure. I will do it by this weekend.

cli/cmd/root.go
```diff
@@ -130,7 +131,9 @@ func runVersionCheck(vChecker *version.VersionChecker, flags []string, cliConfig
 		}
 
 		if clusterVersion != Version {
-			fmt.Fprintf(os.Stderr, "* Warning: Your Keptn CLI version (%s) and Keptn cluster version (%s) don't match. This can lead to problems. Please make sure to use the same versions.\n", Version, clusterVersion)
+			if len(flags) == 0 || flags[0] != "upgrade" {
```
odubajDT, line 134: Hi, thank you for your contribution! There is actually no need to test the length of the `flags` to 0. According to the implementation, this can never happen in this piece of code. I suggest this:

`if flags[0] != "upgrade"`

I understand that afterwards you will receive a SIGSEGV error during running the unit tests, but that's only due to missing `args` parameter in the input struct, adding this parameter, for example like `args: []string{"status"},` should solve the problem on tests side

PR #7463, opened 2022-04-13
feat(bridge): Custom sequence depends on selected stage

### LINKED ISSUE #6983: Trigger sequence: Move custom sequences stage selection next to sequence selection
Custom sequences are configured per stage and not globally in the shipyard.
Currently, the stage for a custom sequence is selected in the overview, like for delivery and evaluation. Then the sequence is selected in the next screen and the user can potentially select a sequence that is not available for the selected stage. The user can still trigger the sequence, and will get an error for the sequence execution in the sequences screen:

image

Proposal:
image

- [x] Return with the `/customSequences` endpoint the association between stage and sequence
- [x] Instead of the word custom show a select with the sequences for the selected stage.
- [x] If no stage is selected yet, disable the select.
- [ ] If the user selects a custom sequence, the stage dropdown is configured with the stages in which the selected sequence is defined. 
- [ ] If the user has both already selected and selects a different sequence, clear the stage dropdown menu and reset it to "Select..." 
- [ ] The preselected stage does not make any sense here - we cannot predict what the user will pick. Remove this functionality.

## DoD

- [x] UI tests need to be updated
- [x] Add and update existing unit tests

Kirdock:
Closes #6983 

Custom sequences are now in the sequence selection:
image

image


ermin-muratovic (dismissed):
lgtm

laneli (changes requested):

laneli (approved):

PR #7464, opened 2022-04-13
feat(shipyard-controller): stop pulling messages after receiving sigterm

### LINKED ISSUE #7304: Shipyard Controller: Events not being processed completely during upgrade
When upgrading Keptn, it can happen that sequenceExecution states end up in an inconsistent state, if the pod currently handling an event is terminated. E.g., in this case, a sequence Execution was not properly cleaned up, since the overall state is already set to `finished`, while there is still a `currentTask` property, instead of the relevant information being added to the `previousTasks` property:

```
{"_id":"7359c200-fc35-4ea1-96e4-926b7e6b64be","inputProperties":{"project":"keptn-upgrade-zero-downtime5","service":"my-service","stage":"dev-13"},"scope":{"project":"keptn-upgrade-zero-downtime5","stage":"dev-13","service":"my-service","labels":null,"status":"","result":"","message":"","keptnContext":"31b447d8-0ef9-4cca-a7d0-d81b5b76e044","triggeredId":"98a1961f-132e-47ac-84c7-a6613e087d36","gitcommitid":"a4c690cd1d2815dad08f892d476b8581d58d0129","eventType":"sh.keptn.event.dev-13.evaluation.triggered"},"sequence":{"name":"evaluation","triggeredon":null,"tasks":[{"name":"evaluation","triggeredafter":"","properties":null}]},"status":{"state":"finished","stateBeforePause":"","previousTasks":[],"currentTask":{"name":"evaluation","triggeredID":"6b222da0-e84c-4727-9368-e962b7e99aec","events":[{"eventType":"sh.keptn.event.evaluation.finished","source":"lighthouse-service","result":"fail","status":"errored","time":"2022-03-30T09:13:12.291Z","properties":{"result":"fail","service":"my-service","stage":"dev-13","status":"errored","evaluation":{"timeEnd":"","timeStart":"","gitCommit":"","indicatorResults":null,"result":"fail","score":0,"sloFileContent":""},"message":"evaluation.triggered event does not contain evaluation timeframe","project":"keptn-upgrade-zero-downtime5"}},{"eventType":"sh.keptn.event.evaluation.started","source":"lighthouse-service","result":"","status":"succeeded","time":"2022-03-30T09:13:12.290Z","properties":null}]}}}
```

Also, in some cases it happens that the reception of a valid sequence.triggered event does not lead to a `sequenceExecution` item being persisted in the database, if the pod is terminated.

My theory is that the component responsible for pulling events from jetstream is pulling events right until the pod is finally killed. So in that case we would need to stop fetching events as soon as we have received the `SIGTERM` signal, i.e. at the beginning of the grace period.

## DoD 

- [ ] Stop ack events as soon as `SIGTERM` is received
- [ ] Unit test (look at distributor code for a reference) 
- [ ] Manual test (document how and what is tested as part of this ticket)

bacherfl:
Closes #7304

How to test:

1. Make sure the termination grace period of the shipyard controller is set
2. Shut down the shipyard controller pod
3. While shutting down, continue to send events to keptn (e.g. sequence.triggered events)
4. Make sure that during the grace period no events are processed by the old replica of the shipyard controller
5. Make sure that every event sent during the termination of the old pod is eventually picked up by the new replica of the shipyard controller

### DIFF
--- a/shipyard-controller/nats/connection_test.go
+++ b/shipyard-controller/nats/connection_test.go
@@ -77,6 +77,54 @@ func TestNatsConnectionHandler(t *testing.T) {
 	}, 15*time.Second, 5*time.Second)
 }
 
+func TestNatsConnectionHandler_ShutdownSubscriber(t *testing.T) {
+	mockNatsEventHandler := &natsmock.IKeptnNatsMessageHandlerMock{
+		ProcessFunc: func(event apimodels.KeptnContextExtendedCE, sync bool) error {
+			return nil
+		},
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+
+	nh := NewNatsConnectionHandler(ctx, natsURL())
+
+	err := nh.SubscribeToTopics([]string{"sh.keptn.>"}, NewKeptnNatsMessageHandler(mockNatsEventHandler.Process))
+
+	require.Nil(t, err)
+
+	publisher, err := nh.GetPublisher()
+
+	require.Nil(t, err)
+
+	event := cloudevents.NewEvent()
+	event.SetType(keptnv2.GetTriggeredEventType("test"))
+	_ = event.SetData(cloudevents.ApplicationJSON, map[string]interface{}{
+		"project": "my-project",
+	})
+
+	// send with Send method
+	err = publisher.Send(context.TODO(), event)
+	require.Nil(t, err)
+
+	// send with SendEvent method
+	err = publisher.SendEvent(event)
+	require.Nil(t, err)
+
+	require.Eventually(t, func() bool {
+		return len(mockNatsEventHandler.ProcessCalls()) == 2
+	}, 15*time.Second, 5*time.Second)
+
+	// call cancel() and wait for the consumer to shut down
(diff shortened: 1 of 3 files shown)


shipyard-controller/nats/connection_test.go
```diff
@@ -77,6 +77,54 @@ func TestNatsConnectionHandler(t *testing.T) {
...
+	// send with Send method
+	err = publisher.Send(context.TODO(), event)
+	require.Nil(t, err)
+
+	// send with SendEvent method
+	err = publisher.SendEvent(event)
+	require.Nil(t, err)
+
+	require.Eventually(t, func() bool {
+		return len(mockNatsEventHandler.ProcessCalls()) == 2
```
odubajDT, line 113: why didn't we use `require.Equal(t, 2, len(mockNatsEventHandler.ProcessCalls()))` as on line 125 ?

odubajDT: exaplained, looks good then

odubajDT (approved):

PR #7465, opened 2022-04-13
style(bridge): Adapt button position for triggering a sequence

### LINKED ISSUE #7157: Trigger sequence: Improve buttons on trigger sequence component
## Acceptance Criteria 

- verify if the button alignment is the best one or it is better to move "cancel" on the left and "next" & "trigger" on the right
- verify if the text "cancel and close" is the best or just "close" or "cancel" suffices 

## DoD

- UI/Unit tests are adapted

laneli:
Closes #7157 

* Button alignment is improved to have Next and Back button on a more intuitive place
* "Cancel and close" was renamed to "Cancel" to be more aligned with the rest of the Bridge

### DIFF
--- a/bridge/client/app/_components/ktb-trigger-sequence/ktb-trigger-sequence.component.html
+++ b/bridge/client/app/_components/ktb-trigger-sequence/ktb-trigger-sequence.component.html
@@ -1,7 +1,9 @@
 <div class="ktb-trigger-sequence pl-3 pr-3 pb-3">
   <div fxFlex="66">
     <div fxLayout="column" *ngIf="state === 'ENTRY'">
-      <h2 uitestid="keptn-trigger-entry-h2">Trigger a new sequence for project {{ projectName }}</h2>
+      <div fxLayout="row" fxFlexAlign="space-between">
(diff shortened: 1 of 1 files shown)


bridge/client/app/_components/ktb-trigger-sequence/ktb-trigger-sequence.component.html
```diff
@@ -1,7 +1,9 @@
 <div class="ktb-trigger-sequence pl-3 pr-3 pb-3">
   <div fxFlex="66">
     <div fxLayout="column" *ngIf="state === 'ENTRY'">
-      <h2 uitestid="keptn-trigger-entry-h2">Trigger a new sequence for project {{ projectName }}</h2>
+      <div fxLayout="row" fxFlexAlign="space-between">
```
Kirdock, line 4: What's the reason for adding a fxLayout here? I don't see any difference

Kirdock (changes requested):

bridge/client/app/_components/ktb-trigger-sequence/ktb-trigger-sequence.component.html
laneli: You are right, it is a leftover from one layout iteration I tried. Thx for noticing.

Issue #7453, opened 2022-04-12
Configuration Service Storage Error for pvc


suhasln:
**Environment**

* Client OS (e.g., Linux, macOS, Windows): Ubuntu
* Keptn Version (`keptn version`): 0.11.4
* Kubernetes Cloud Provider (e.g., GKE, AKS): DigitalOcean
* Kubernetes version (`kubectl version`): .1.21.7

**Affected Component**
Configuration-service

* [x] Datastore

**Describe the bug**
/tree/master/installer/manifests/keptn/charts/control-plane)/values.yaml file has pv claim for 100Mi, when we install keptn from cli to k8s cluster configuration service gives error as its'nt enough for bounding to pv. So we updated the value for 1Gi to resolve this issue.

**To Reproduce**

**Expected behavior**

**Current behavior**

**Screenshots**

**Additional context**

thisthat:
Hi @suhasln, for such a use case we recommend using `helm` directly for installing Keptn. More information are available here: https://keptn.sh/docs/0.14.x/operate/advanced_install_options/

PR #7468, opened 2022-04-14
fix(installer): Use 1Gi as default PVC size

thisthat:
Closes https://github.com/keptn/keptn/issues/7453


odubajDT (approved):

mowies:
It would be good to check how this PR influences upgrades of Keptn. I would assume that the PVC will be re-created which would mean that we could potentially lose data.

PR #7479, opened 2022-04-14
feat: Introducing ZeroDowntime tests

### LINKED ISSUE #7373: Test Keptn Upgrade w.r.t. Zero Downtime
To ensure our Zero downtime upgrade capabilities are met, we need an automated test that verifies if our APIs are reachable during upgrading Keptn, and sequences can be triggered and eventually finished during an upgrade.
A first draft of how this could look like has been implemented in https://github.com/keptn/keptn/pull/7266 - for now, this has been added to the integration tests, but I believe this test should rather be executed in a different step, e.g. after the integration tests have been passed.

In essence, the test implemented in the draft PR performs a number of helm upgrades between two different Keptn versions, while continuously sending HTTP requests to the API, as well as triggering sequences. At the end, it is verified whether all requests have been processed by the API, and if all triggered sequences have been finished.

## DoD 

- [x] Setting up a keptn instance during the test
- [x] Have an API load test
- [x] The test is executed automatically at the end of the integration tests
- [x] Testing that sequences always finish (either w/ success or error status)

RealAnna:
## This PR

- adds API test suite
- adds webhook and sequences test suite
- adds  a suite to run Zero Downtime tests: rolling upgrade starts in parallel to API probes and one test (in the first round sequences are triggered then webhook tests)

### Related Issues

Fixes #7373 

### Notes

Example Run https://github.com/keptn/keptn/actions/runs/2274375103

### Follow-up Tasks

add more tests for other services


test/go-tests/test_webhook.go
```diff
@@ -404,28 +404,10 @@ spec:
         - url: http://shipyard-controller:8080/v1/project/{{.data.project}}/stage/{{.data.stage}}
           method: GET`
 
-func Test_Webhook_Alpha(t *testing.T) {
```
odubajDT, line 407: Is there a functional reason for these changes in `test_webhook.go` ? Or is it just refactoring to make it nicer ?

test/zero-downtime/testsuite_zerodowntime_test.go
```diff
@@ -0,0 +1,176 @@
...
+	t.Log("-----------------------------------------------")
+	t.Log("TOTAL SEQUENCES: ", env.FiredSequences)
+	t.Log("TOTAL SUCCESS ", env.PassedSequences)
+	t.Log("TOTAL FAILURES ", env.FailedSequences)
+	t.Log("-----------------------------------------------")
+
+}
+
+//Returns current test helm charts for the rolling upgrade
+func GetCharts(t *testing.T) (string, string) {
```
bacherfl, line 165: please add the names of the required environment variables to the comment

test/zero-downtime/testsuite_zerodowntime_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+}
+
+func RollingUpgrade(t *testing.T, env *ZeroDowntimeEnv) {
+	defer func() {
+		close(env.quit)
+		t.Log("Rolling upgrade terminated")
+		env.Wg.Wait()
+
+		PrintSequencesResults(t, env)
```
bacherfl, line 126: would it make sense to call the `PrintResults()` methods of the other tests in `ZDTestTemplate()`? This might be a bit cleaner, as the `RollingUpgrade()` would then be independent from the tests that are executed during the upgrades

test/go-tests/test_webhook.go
RealAnna: I am recycling the create and the webhook test in the zd tests 😄

test/zero-downtime/testsuite_zerodowntime_test.go
RealAnna: the problem is that since the template is parallel the code would be run immediately and sometimes there is a race condition for which the sequence test has not yet added to the wait group, so we would print before the end of the test... I could try to add a sleep there maybe ...

ex. run 
image

RealAnna: adding  wait in the templates is blocking 😢

test/zero-downtime/sequences_test.go
```diff
@@ -0,0 +1,205 @@
...
+	if sequenceFinishedEvent == nil || err != nil {
+		suite.env.failSequence()
+		suite.T().Errorf("sequence %s with keptnContext %s in project %s has NOT been finished", sequence.sequenceName, sequence.keptnContext, sequence.projectName)
+
+	} else {
+		suite.T().Logf("sequence %s with keptnContext %s in project %s has been finished", sequence.sequenceName, sequence.keptnContext, sequence.projectName)
+	}
+}
+
+func GetShipyard() (string, error) {
```
odubajDT, line 176: Is there a reason why we don't add a simple shipyard file string instead of creating the structure, converting it to a string and creating tmp file afterwards ? Seems like it is a very simple one, so it shouldn't be a problem...

odubajDT: also I would expect this method to be present in `test/zero-downtime/testsuite_zerodowntime_test.go`

test/zero-downtime/sequences_test.go
```diff
@@ -0,0 +1,204 @@
...
+func Test_Sequences(t *testing.T) {
+	Env := setEnv(t)
+
+	s := &TestSuiteSequences{
+		env: Env,
+	}
+	suite.Run(t, s)
+}
+
+func setEnv(t *testing.T) *ZeroDowntimeEnv {
```
odubajDT, line 73:
```suggestion
func setSequencesTestEnv(t *testing.T) *ZeroDowntimeEnv {
```

It can be misunderstood, because `os.SetEnv()` is a `os` library method and we shouldn't use the same naming

test/zero-downtime/webhook_test.go
```diff
@@ -0,0 +1,81 @@
...
+	suite.T().Log("Starting test for webhook")
+
+}
+
+func (suite *TestSuiteWebhook) BeforeTest(suiteName, testName string) {
+	atomic.AddUint64(&suite.env.FiredSequences, 1)
+	suite.T().Log("Running one more test, tot ", suite.env.FiredSequences)
+}
+
+//Test_Sequences can be used to test a single run of the sequence test suite
```
odubajDT, line 28: Copy pasta error :)

test/zero-downtime/sequences_test.go
RealAnna: sure it can be done like that

test/zero-downtime/sequences_test.go
```diff
@@ -0,0 +1,204 @@
...
+	suite.Nil(err)
+	output, err := testutils.ExecuteCommand(fmt.Sprintf("keptn create service %s --project=%s", "myservice", suite.project))
+	suite.Nil(err)
+	suite.Contains(output, "created successfully")
+	suite.T().Logf("Starting test for project %s ", suite.project)
+}
+
+func (suite *TestSuiteSequences) BeforeTest(suiteName, testName string) {
+	atomic.AddUint64(&suite.env.FiredSequences, 1)
+	suite.T().Log("Running one more test, tot ", suite.env.FiredSequences)
```
odubajDT, line 60: probably total, or?

test/zero-downtime/API_test.go
```diff
@@ -0,0 +1,209 @@
...
+	suite.T().Logf("Response Body: %v", res.Body)
+	suite.T().Logf("Duration: %s", finished.Sub(started))
+	if res.StatusCode != expected {
+		atomic.AddUint64(&suite.env.FailedAPICalls, 1)
+	} else {
+		atomic.AddUint64(&suite.env.PassedAPICalls, 1)
+	}
+}
+
+func getClient(sec time.Duration) *http.Client {
```
odubajDT, line 197: maybe also moving this to `test/zero-downtime/testsuite_zerodowntime_test.go` ? To make it reusable also in other ZD tests in the future if needed

RealAnna: not sure if it is needed this is specifically for HTTP probes and I would leave them in the API suite

odubajDT: ok

bacherfl (approved):

.github/workflows/zero-downtime-tests.yml
```diff
@@ -47,6 +47,7 @@ jobs:
       KUBECONFIG: ${{ matrix.KUBECONFIG }}
       KEPTN_NAMESPACE: "keptn-zd-test-${{ github.run_number }}-${{ github.run_attempt }}"
       GO_VERSION: 1.17
+      GOMAXPROCS: 10
```
mowies, line 50: i think you could also move this variable to the step where it is needed (`test_zero_downtime`)

.github/workflows/zero-downtime-tests.yml
```diff
@@ -324,30 +324,29 @@ jobs:
       - name: Prepare test run
         id: prepare_test_run
         run: |
-          go install gotest.tools/gotestsum@v1.7.0
+          go install github.com/haveyoudebuggedit/gotestfmt/v2/cmd/gotestfmt@latest
```
mowies, line 327: please use a pinned version instead of latest

test/zero-downtime/go.mod
```diff
@@ -0,0 +1,104 @@
+module github.com/keptn/keptn/test/zero_downtime
+
+go 1.17
+
+require (
+	github.com/benbjohnson/clock v1.3.0
+	github.com/keptn/go-utils v0.14.1-0.20220414081235-2e23eb712e3d
```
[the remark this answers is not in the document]
RealAnna: I will try to see whether 0.14.0 is enough, also integration test run currently not on a released version 😢

.github/workflows/zero-downtime-tests.yml
RealAnna: I tought to keep the go env setup togheter, but sure I can move it

mowies (approved):
pipeline LGTM :)

bacherfl (approved):

PR #7519, opened 2022-04-19
fix(bridge): Fix update of git upstream without a user

### LINKED ISSUE #7496: Git set upstream without `Git user` does not work
**Environment**

* Keptn Version (`keptn version`): `0.14.1`

**Affected Component**

* [x] Bridge

**Describe the bug**

* When trying to set an upstream without `Git user` (since it is not required anymore), the **Set Git upstream** button does not work. Please see the network trace that stays empty: 
image

**To Reproduce**

* Create a project without upstream repo
* Follow the link: 
image
* Enter `Git URL` and `Git Token`
* Click **Set Git upstream** button

**Expected behavior**

* Since the `Git user` is not mandatory anymore, I would expect that setting upstream without user works. 

**Current behavior**

When providing the `Git user`, the operation works: 
image

Kirdock:
Fixes bridge part of #7496 

### DIFF
--- a/bridge/client/app/_components/ktb-project-settings/ktb-project-settings.component.spec.ts
+++ b/bridge/client/app/_components/ktb-project-settings/ktb-project-settings.component.spec.ts
@@ -343,4 +343,25 @@ describe('KtbProjectSettingsComponent', () => {
       },
     });
   });
+
+  it('should update git upstream without user', () => {
+    // given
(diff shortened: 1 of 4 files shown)


johannes-b:
Thanks! I can confirm that the bug fix is working, from an end-user point of view: 

image

ermin-muratovic (approved):
lgtm

PR #7533, opened 2022-04-21
fix: Do not require git user being set when updating project upstream credentials

### LINKED ISSUE #7496: Git set upstream without `Git user` does not work

bacherfl:
Closes #7496 

A couple of adjustments had to be made in the shipyard-controller, as well as the configuration service in order to support setting the upstream without a user.

### DIFF
--- a/configuration-service/common/git.go
+++ b/configuration-service/common/git.go
@@ -72,7 +72,7 @@ func (K8sCredentialReader) GetCredentials(project string) (*common_models.GitCre
 	if err != nil {
 		return nil, fmt.Errorf(unmarshalGitCredentialsFail)
 	}
-	if credentials.User != "" && credentials.Token != "" && credentials.RemoteURI != "" {
+	if credentials.Token != "" && credentials.RemoteURI != "" {
 		return &credentials, nil
 	}
 	return nil, nil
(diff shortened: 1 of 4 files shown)


configuration-service/common/git.go
```diff
@@ -519,32 +525,8 @@ func ServiceExists(project string, stage string, service string, disableUpstream
 
 // GetCredentials returns the git upstream credentials for a given project (stored as a secret), if available
 func GetCredentials(project string) (*common_models.GitCredentials, error) {
```
odubajDT, line 527: We probably do not have a reason to have this function, as we can use `credentialsReader.GetCredentials(project)`, but as it is configuration-service, there is probably no reason to refactor it and spent time on improving the code

configuration-service/common/git_test.go
```diff
@@ -786,3 +795,55 @@ func TestGit_CloneRepo(t *testing.T) {
...
+		{
+			name: "malformed credentials",
+			args: args{
+				data: []byte(`invalid`),
+			},
+			want:    nil,
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
```
odubajDT, line 838: please add test for input: `{"token":"token","remoteURI":"http://remote-url.git"}`

bacherfl: done

odubajDT (approved):
LGTM

PR #7534, opened 2022-04-21
fix(api): Allow to enable/disable rate limit 

thisthat:

## This PR

- add a helm value to enable/disable the rate limit

### Notes

I also observed that `X-Forwarded-For` is more reliable to detect the request IP address

### DIFF
--- a/api/restapi/configure_keptn.go
+++ b/api/restapi/configure_keptn.go
@@ -90,8 +90,11 @@ func configureAPI(api *operations.KeptnAPI) http.Handler {
 
 	//api.EvaluationTriggerEvaluationHandler = evaluation.TriggerEvaluationHandlerFunc(handlers.TriggerEvaluationHandlerFunc)
 
-	rateLimiter := custommiddleware.NewRateLimiter(env.AuthEnabled, env.AuthRequestsPerSecond, env.AuthRequestMaxBurst, tokenValidator, clock.New())
-	api.AddMiddlewareFor(http.MethodPost, "/auth", rateLimiter.Handle)
+	if env.AuthEnabled {
+		rateLimiter := custommiddleware.NewRateLimiter(env.AuthRequestsPerSecond, env.AuthRequestMaxBurst, tokenValidator, clock.New())
+		api.AddMiddlewareFor(http.MethodPost, "/auth", rateLimiter.Handle)
+	}
+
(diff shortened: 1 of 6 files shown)


api/restapi/configure_keptn.go
```diff
@@ -31,6 +32,7 @@ import (
 const envVarLogLevel = "LOG_LEVEL"
 
 type EnvConfig struct {
+	AuthEnabled           bool    `envconfig:"MAX_AUTH_ENABLED" default:"true"`
```
bacherfl, line 35: let's also change the name of the variable to `MaxAuthEnabled`, otherwise the name of the variable would suggest that it's used to disable authentication in general

api/restapi/configure_keptn.go
```diff
@@ -88,7 +90,7 @@ func configureAPI(api *operations.KeptnAPI) http.Handler {
 
 	//api.EvaluationTriggerEvaluationHandler = evaluation.TriggerEvaluationHandlerFunc(handlers.TriggerEvaluationHandlerFunc)
 
-	rateLimiter := custommiddleware.NewRateLimiter(env.AuthRequestsPerSecond, env.AuthRequestMaxBurst, tokenValidator, clock.New())
+	rateLimiter := custommiddleware.NewRateLimiter(env.AuthEnabled, env.AuthRequestsPerSecond, env.AuthRequestMaxBurst, tokenValidator, clock.New())
```
bacherfl, line 93: i think it would be easier to not call `api.AddMiddlewareFor(http.MethodPost, "/auth", rateLimiter.Handle)` if `env.AuthEnabled` is se to false

installer/manifests/keptn/charts/control-plane/values.yaml
```diff
@@ -86,6 +86,7 @@ apiService:
     repository: docker.io/keptn/api
     tag: ""
   config:
+    maxAuthEnabled: true
```
mowies, line 89: Maybe it would give us a better structure if we had a separate block called `maxAuth` and then have `enabled`, `requestsPerSecond` and `requestsBurst` inside?

api/restapi/configure_keptn.go
thisthat: I prefer keeping the same naming like the other env vars

bacherfl (approved):

PR #7548, opened 2022-04-22
feat(webhook-service): Introduce keptn-webhook-config ConfigMap with denyList

### LINKED ISSUE #7239: Core: webhook-service improve permissions
## User Story

As an operator, I want to configure Keptn to forbid specific URLs via webhooks.

## Acceptance criteria

The following applies only to the webhook config file in version `v1beta1`.

- Allow a list `denyList` to be read from ConfigMap mounted as a volume.
- At each webhook, read the `denyList` from the file. 

- Given a webhook has to be fired, if the URL `startWith` matches an entry in `denyList`, the webhook fails.
- Given a webhook has to be fired, if the IP resolution for the URL matches (`startWith` an entry in `denyList`, the webhook fails.
- Otherwise, the webhook can be fired.

- Previous denied hardcoded URLs will be the default value of the configMap. 

## DoD

- Unit tests
- Integrations tests

## Resources 

- Resolve IP for a given URL example: 
```go
import (
	"fmt"
	"net"
	"net/url"
)

func main() {
	url, _ := url.Parse("https://mydomain.org/pathpar?quey=myquery#")
	ips, _ := net.LookupIP(url.Hostname())
	for _, ip := range ips {
		if addr := ip.To4(); addr != nil {
			fmt.Println(addr)
		}
	}
}
```

odubajDT:
## This PR

- introduces keptn-webhook-config ConfigMap with list of denied URLs for beta1 configurations
- refactors existing validation of alpha1 configurations to be easily removable

### Related Issues

Fixes #7239

### Integration tests
https://github.com/keptn/keptn/actions/runs/2245478709
https://github.com/keptn/keptn/actions/runs/2256117177


webhook-service/lib/curl_validator.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+	"strings"
+)
+
+type CurlValidator struct {
+	denyListProvider IDenyListProvider
+	ipResolver       IIPResolver
+}
+
+type ICurlValidator interface {
```
warber, line 13: Let's call the interface `RequestValidator` and the implemenataion `requestValidator`

webhook-service/go.mod
```diff
@@ -16,45 +17,78 @@ require (
 )
 
 require (
+	cloud.google.com/go v0.54.0 // indirect
+	github.com/Azure/go-autorest v14.2.0+incompatible // indirect
```
bacherfl, line 21: that's quite a lot of new dependencies - are they really required, or can this be tidied up using go mod tidy?

webhook-service/lib/curl_validator.go
```diff
@@ -0,0 +1,44 @@
...
+	validator := CurlValidator{
+		denyListProvider: denyListProvider,
+		ipResolver:       ipResolver,
+	}
+	return validator
+}
+
+func (c CurlValidator) Validate(request Request) error {
+	if request.URL == "" {
+		return fmt.Errorf("Invalid curl URL: '%s'", request.URL)
```
warber, line 27: pls start the phrase with lowercase

webhook-service/handler/handler.go
```diff
@@ -196,10 +198,10 @@ func (th *TaskHandler) performWebhookRequests(webhook lib.Webhook, eventAdapter
 	executedRequests := 0
 	logger.Infof("executing webhooks for subscriptionID %s", webhook.SubscriptionID)
 	for _, req := range webhook.Requests {
-		request, err := CreateRequest(req)
+		request, err := th.CreateRequest(req)
 		if err != nil {
-			logger.Infof("creating CURL request failed: %s", err.Error())
-			return nil, err
+			logger.Errorf("creating CURL request failed: %s", err.Error())
```
bacherfl, line 203: here we might to check for specific error types, as e.g. the detection of a disallowed URL is not an error in the execution of the webhook service, so that should probably be logged on another error level

odubajDT: So do you suggest have Info there ?

webhook-service/lib/denylist_provider.go
```diff
@@ -0,0 +1,77 @@
...
+	"strings"
+
+	keptnkubeutils "github.com/keptn/kubernetes-utils/pkg"
+	logger "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+type IDenyListProvider interface {
```
warber, line 14: Lets call this `DenyListProvider` and its implementation `denyListProvider`

installer/manifests/keptn/charts/control-plane/templates/rbac.yaml
```diff
@@ -192,6 +192,24 @@ rules:
...
+  name: keptn-get-configmaps
+  labels:
+    {{ include "control-plane.labels" . | nindent 4 }}
+    app.kubernetes.io/name: keptn-get-configmaps
+    app.kubernetes.io/part-of: keptn-{{ .Release.Namespace }}
+    app.kubernetes.io/component: {{ include "control-plane.name" . }}
+rules:
+  - apiGroups:
+      - ""
+    resources:
```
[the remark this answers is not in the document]
odubajDT: fixed

webhook-service/lib/denylist_provider.go
```diff
@@ -0,0 +1,77 @@
...
+	GetDenyList() []string
```
warber, line 15: would just `Get []string` be enough to make it less verbose (`denyListProvider.getDenyList()` vs `denyListProvider.Get()`

webhook-service/lib/ip_resolver.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"net"
+	"net/url"
+
+	logger "github.com/sirupsen/logrus"
+)
+
+type IIPResolver interface {
+	ResolveIPAdresses(curlURL string) []string
```
warber, line 11: should we just name this method `Resolve(URL)` ?

webhook-service/lib/denylist_provider.go
```diff
@@ -0,0 +1,77 @@
...
+	GetDeniedURLs               GetDeniedURLsFunc
+	GetWebhookDenyListConfigMap GetWebhookDenyListConfigMapFunc
+}
+
+type GetKubeAPIFunc func(useInClusterConfig bool) (v1.CoreV1Interface, error)
+type GetDeniedURLsFunc func(env map[string]string) []string
+type GetWebhookDenyListConfigMapFunc func(kubeAPI v1.CoreV1Interface) (*corev1.ConfigMap, error)
+
+func NewDenyListProvider() DenyListProvider {
+	return DenyListProvider{
```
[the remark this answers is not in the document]
bacherfl: See https://github.com/keptn/keptn/blob/b1b3d11c6d0ed6dea1016b0757ce4a1d0bddbc85/resource-service/common/credentials_test.go#L46 for an example of how to use the fake client

installer/manifests/keptn/charts/control-plane/templates/rbac.yaml
```diff
@@ -192,6 +192,24 @@ rules:
...
+metadata:
+  name: keptn-get-webhook-config
+  labels:
+    {{ include "control-plane.labels" . | nindent 4 }}
+    app.kubernetes.io/name: keptn-get-webhook-config
+    app.kubernetes.io/part-of: keptn-{{ .Release.Namespace }}
+    app.kubernetes.io/component: {{ include "control-plane.name" . }}
+rules:
+  - apiGroups:
+      - keptn-webhook-config
```
bacherfl, line 207: the name of the configMap should rather be listed in the `resourceNames` property - see https://github.com/keptn/keptn/blob/f4d5242055f9fd3aa398a4b6bef8ae4b1f902893/installer/manifests/keptn/charts/control-plane/templates/rbac.yaml#L232 for an example :)

webhook-service/lib/ip_resolver.go
```diff
@@ -0,0 +1,42 @@
+package lib
+
+import (
+	"net"
+	"net/url"
+
+	logger "github.com/sirupsen/logrus"
+)
+
+type IIPResolver interface {
```
bacherfl, line 10: to be consistent with other interfaces, I suggest to also use `moq` to generate the mock implementations here

webhook-service/lib/curl_validator.go
bacherfl: also, the URL will be always empty at this point, so I suggest to change the message to something like `request contains empty URL`

webhook-service/lib/common.go
```diff
@@ -1,7 +1,52 @@
...
+	"strings"
+)
+
+const WebhookConfigMap = "keptn-webhook-config"
 
 func GetNamespaceFromEnvVar() string {
 	return os.Getenv("POD_NAMESPACE")
 }
+
+func GetEnv() map[string]string {
```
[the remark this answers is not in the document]
odubajDT: it should be there, but maybe that's a action for a webhook refactoring ticket? There are so many things to refactor there, that it does not really makes sense to move one method to go-utils now...

webhook-service/lib/ip_resolver.go
warber: Maybe we can have a discussion about that somewhen :)
For me, I've discovered that I am less and less making use of `moq` and co. But just try to have small interfaces (which this one clearly is) and implement the mock on the fly. I only use `moq` for bigger third party interfaces, actually

odubajDT: same opinion on my side, for a few line mock, we do not need to generate it

webhook-service/lib/denylist_provider.go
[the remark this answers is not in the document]
odubajDT: done, thank you for help!

webhook-service/handler/handler_test.go
```diff
@@ -973,6 +1007,16 @@ func TestTaskHandler_Execute_WebhookConfigInProject(t *testing.T) {
...
+	secretReaderMock := &fake.ISecretReaderMock{}
+	curlExecutorMock := &fake.ICurlExecutorMock{}
+	curlValidatorMock := &fake.ICurlValidatorMock{}
+	curlValidatorMock.ValidateFunc = func(request lib.Request) error {
+		return nil
+	}
+
+	taskHandler := handler.NewTaskHandler(templateEngineMock, curlExecutorMock, curlValidatorMock, secretReaderMock)
+
 	tests := []struct {
```
bacherfl, line 1020: let's also add a test case where the validator returns an error

odubajDT: good catch, added :)

bacherfl (approved):
lgtm,, but let's also wait for Bernd's review and the integration tests

webhook-service/lib/ip_resolver.go
```diff
@@ -0,0 +1,42 @@
...
+	logger "github.com/sirupsen/logrus"
+)
+
+type IPResolver interface {
+	Resolve(curlURL string) []string
+}
+
+type LookupFunc func(host string) ([]net.IP, error)
+
+type IpResolver struct {
```
warber, line 16: you can also just use the unexported name `ipResolver` for this one. Why does it need to be an exported type?
In the tests either change the test package to `lib` or use your constructor function below to create an ipResolver value.

webhook-service/lib/denylist_provider.go
```diff
@@ -0,0 +1,62 @@
...
+	logger "github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+type DenyListProvider interface {
+	Get() []string
+}
+
+type DenyListProviderStruct struct {
```
warber, line 16: Same here, change to unexported `denyListProvider`

webhook-service/lib/ip_resolver.go
```diff
@@ -0,0 +1,42 @@
...
+	LookupIP LookupFunc
+}
+
+func NewIPResolver(lookUpIPFunc ...LookupFunc) IPResolver {
```
warber, line 20: `lookUpIPFunc` is not used + why is it a slice?

webhook-service/lib/common.go
```diff
@@ -1,7 +1,56 @@
...
+	WebhookConfigMap        = "keptn-webhook-config"
+	KubernetesSvcHostEnvVar = "KUBERNETES_SERVICE_HOST"
+	KubernetesAPIPortEnvVar = "KUBERNETES_SERVICE_PORT"
+)
 
 func GetNamespaceFromEnvVar() string {
 	return os.Getenv("POD_NAMESPACE")
 }
+
+func GetEnv() map[string]string {
```
warber, line 18: let's add a unit test for this guy

webhook-service/lib/ip_resolver.go
```diff
@@ -0,0 +1,42 @@
...
+	return IpResolver{
+		LookupIP: net.LookupIP,
+	}
+}
+
+func (i IpResolver) Resolve(curlURL string) []string {
```
warber, line 26:
```suggestion
func (i IpResolver) Resolve(url string) []string {
```

warber: since the url is given as a raw string, you should also test correct behavior of the method when feeding it with an invalid url....

webhook-service/lib/request_validator.go
```diff
@@ -0,0 +1,44 @@
...
+	validator := requestValidator{
+		denyListProvider: denyListProvider,
+		ipResolver:       ipResolver,
+	}
+	return validator
+}
+
+func (c requestValidator) Validate(request Request) error {
+	if request.URL == "" {
+		return fmt.Errorf("invalid curl URL: '%s'", request.URL)
```
warber, line 27: since `request.URL` will always be just empty:
```suggestion
		return fmt.Errorf("curl command contains empty URL")
```

webhook-service/lib/common.go
odubajDT: the method is there for a longer time without tests, just moved it to common.go and to be honest, do not really know how to test this kind of behaviour... It may need some test, but will definitely need a refactoring as well as the rest of webhook-service, but that is not the point of this issue and PR. If you agree I would rather skip adding tests to functionality which was not modified, as that is another story...

webhook-service/lib/ip_resolver.go
```diff
@@ -0,0 +1,45 @@
...
+
+type IPResolver interface {
+	Resolve(url string) []string
+}
+
+type LookupFunc func(host string) ([]net.IP, error)
+type ParseFunc func(rawURL string) (*neturl.URL, error)
+
+type ipResolver struct {
+	LookupIP LookupFunc
```
warber, line 18:
```suggestion
	lookupIP LookupFunc
```

webhook-service/lib/ip_resolver.go
```diff
@@ -0,0 +1,45 @@
...
+	Parse    ParseFunc
```
warber, line 19:
```suggestion
	parse    ParseFunc
```

webhook-service/lib/denylist_provider.go
```diff
@@ -0,0 +1,62 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+type DenyListProvider interface {
+	Get() []string
+}
+
+type denyListProvider struct {
+	GetDeniedURLs GetDeniedURLsFunc
```
warber, line 17:
```suggestion
	getDeniedURLs GetDeniedURLsFunc
```

webhook-service/lib/denylist_provider.go
```diff
@@ -0,0 +1,62 @@
...
+	KubeClient    kubernetes.Interface
```
warber, line 18:
```suggestion
	kubeClient    kubernetes.Interface
```

webhook-service/lib/ip_resolver_test.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"fmt"
+	"net"
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCurlValidator_ResolveIPAddresses(t *testing.T) {
```
warber, line 12: Test for a failing `lookupIP()` call is missing

warber (approved):

PR #7553, opened 2022-04-22
test(bridge): Fix flaky sequence UI test

Kirdock:
The test `should filter waiting sequences` in `sequence.spec.ts` was flaky (detached from DOM error). This was most likely related to a re-render after the sequence metadata has been fetched. The fix is to wait until the metadata is fetched.


renepanzar (approved):
LGTM, this should do the trick.

ermin-muratovic (approved):
lgtm

PR #7605, opened 2022-05-02
feat(shipyard-controller): Store sequence executions in new format without potential dots (.) in property names

### LINKED ISSUE #7382: Core(shipyard-controller): Change data model to avoid having `.` (dots) as part of keys stored in mongodb
There are some functional differences between MongoDB and e.g. Amazon DocumentDB. One of them, having stricter field name restrictions on Amazon DocumentDB. More specifically, it's not allowed to store keys containing dots (`.`) and other characters. [1]

With #6065 #6266 we've implemented mitigations to circumvent that restriction by encoding/decoding fields known to contain dots in the ShipyardController and the StatisticsService.

However, as a long-term solution, the data model shall be changed to not require storing data in MongoDB in this way.

## Technical details

Convert `task` data into a JSON string which is stored in the database. `task` name would be the key and the JSON string the value.

## Acceptance Criteria

- [ ] The ShipyardController does not store data with unsupported characters in MongoDB
- [ ] Old data is automatically migrated to the new model
- [ ] Backward compatibility is maintained 

## DoD

- [ ] Units tests
- [ ] Integration tests

---
[1] [Amahon DocumentDB -Functional Differences](https://docs.aws.amazon.com/documentdb/latest/developerguide/functional-differences.html)

bacherfl:
Closes #7382 

Integration test run: https://github.com/keptn/keptn/runs/6270718947?check_suite_focus=true


shipyard-controller/db/models/sequence_execution/v1/transformer.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+type ModelTransformer struct{}
+
+// TransformToDBModel transforms an instance of models.SequenceExecution to the db specific schema defined in this package
+func (ModelTransformer) TransformToDBModel(execution models.SequenceExecution) interface{} {
+	return fromSequenceExecution(execution)
+}
+
+func (ModelTransformer) TransformEventToDBModel(event models.TaskEvent) interface{} {
```
odubajDT, line 16: Can we unit test this method ?

shipyard-controller/db/models/sequence_execution/v1/transformer.go
```diff
@@ -0,0 +1,154 @@
...
+	return transformTaskEvent(event)
+}
+
+func (ModelTransformer) TransformToSequenceExecution(dbItem interface{}) (*models.SequenceExecution, error) {
```
odubajDT, line 20: as well as this ?

shipyard-controller/db/models/sequence_execution/v1/transformer.go
bacherfl: good catch - added a unit test just now

shipyard-controller/db/models/sequence_execution/v1/transformer.go
bacherfl: good catch - added a unit test just now

odubajDT (approved):
lgtm

odubajDT (approved):

PR #7610, opened 2022-05-02
feat(bridge): Introduce automaticprovisioning flag to metadata

laneli:
Adds the automaticprovisioning to the metadata interface.
To be backwards compatible to configuration-service, the property has to be optional. This will only be set if resource-service is already configured in the Keptn installation.


renepanzar (approved):
LGTM!

PR #7658, opened 2022-05-05
feat(bridge): Implement heatmap with d3

### LINKED ISSUE #6162: Replace highcharts heatmap with d3
As part of #6001 the highcharts heatmap has to be replaced by a heatmap made with d3.

## Acceptance Criteria:
- Tooltip for each cell on hover
  - score: value, total passed SLIs, total warnings, total failed SLIs and the threshold (pass and warning)
  - SLIs: value, score, is keySLI, pass criterias, warning criterias
- Selection. If a cell is clicked the whole column should be selected.
- Secondary selection: It should be possible to add several selections, independent of the current selection (compared events)

## Definition of Done
- [x] Highcharts heatmap can be replaced with the new component
- [ ] Unit tests

Kirdock:
Closes #6162 

## How to test:
- Add a new route to `app.routing.ts`. e.g. `{ path: 'd3', component: KtbTestHeatmapComponent },`
- In `ktb-heatmap.module` add `KtbTestHeatmapComponent` to declarations and exports
- In `KtbTestHeatmapComponent` the generated amount of SLIs and evaluations can be configured.

image


bridge/client/app/_components/ktb-heatmap/ktb-heatmap-tooltip.component.html
```diff
@@ -0,0 +1,89 @@
+<div class="tooltip-container" *ngIf="tooltip">
+  <dt-key-value-list>
+    <dt-key-value-list-item>
+      <dt-key-value-list-key> Value </dt-key-value-list-key>
+      <dt-key-value-list-value>
+        {{ tooltip.value | truncateNumber: 2 }}
```
ermin-muratovic, line 6: If the `tooltip.type` is `IHeatmapTooltipType.SLI` shouldn't we use here `formatNumber(value)` as well instead of the `truncateNumber` pipe?

ermin-muratovic (changes requested):

bridge/client/app/_components/ktb-heatmap/ktb-heatmap-tooltip.component.html
Kirdock: Done. For the `score` this would not change anything, so it does not matter that `formatNumber` is applied to both

ermin-muratovic (dismissed):
lgtm

ermin-muratovic (approved):
lgtm

PR #7661, opened 2022-05-05
feat: Make deployment resources customizable in helm charts

### LINKED ISSUE #6981: Keptn Installer: introduce deployment parameters to keptn helm chart (control-plane) to make resource limits/requests customisable
# Problem:
We experience scalability issues with the api-gateway-nginx (control-plane chart) when running Keptn on K8s with the default Keptn Helm Charts. From time to time the liveness-probes fail resulting in triggered restarts, furthermore we experienced pod kills by OOM.

Our plan to overcome these issues would be to customise the resource limits/requests of pods to our individual demands. However, to change resource-definitions it currently demands manual intervention in the Keptn Helm Chart and leads to the problem that we can no longer relate/use the charts from the official Keptn Helm Repo (https://charts.keptn.sh/).

Keptn Version: 0.10.0

# Proposed fix:
The introduction of deployment parameters (with default values) would provide a generic way to make resource definitions customisable. As a reference implementation I suggest the bitnami charts that provide best-practices for helm chart templating:

* https://github.com/bitnami/charts/tree/master/bitnami/nginx/#nginx-deployment-parameters
* https://github.com/bitnami/charts/blob/master/bitnami/nginx/templates/deployment.yaml#L197

Currently we encounter this scalability issues on api-gateway-nginx, but I think it would make sense to introduce these deployment parameters (with current default values) to all containers.

I am open for feedback as well as for other solutions that help us to solve our issue!

## DoD

- Implement the suggestion with having the current values as default

albertogdd:
## This PR treats with #6981 

Makes possible to inject your own custom resources values using the Helm Charts in the following pods:

from /helm-service
- `helm-service`

from /jmeter-service
- `jmeter-service`

from /installer
- `api-gateway-nginx`
- `approval-service`
- `resource-service`
- `configuration-service`
- `remediation-service`
- `bridge`
- `mongodb-datastore`
- `secret-service`
- `shipyard-controller`
- `statistics-service`
- `webhook-service`

This resources values were hardcoded before in the templates.

### Related Issues
Closes #6981 

### Notes
Init Containers aren't included in the changes, as they had no specified resources. Hence, they use the [default K8s values](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#resources).

All `control-plane` distributor pods use the same resources properties (they all use the same `keptn.distributor.resources` function). Now the value of this function is customizable (see `keptn/installer/manifests/keptn/values.yaml`)

### Follow-up Tasks
No further changes are needed, as the template resource values now are in values.yaml, the templating engine should generate the exact same output as before.

### How to test
Comparing the templates generated in this feature branch with the ones from `charts-features`. 
[diff in keptn installer generated templates](https://www.diffchecker.com/yweJ6VZM)
[diff in jmeter-service generated templates](https://www.diffchecker.com/aYYrlStf)
[diff in helm-service generated templates](https://www.diffchecker.com/rWb2UI0r)

There's a difference in the order of the lines inside the generated resources, my guess is that `toYaml` orders them alphabetically. Also, the quotes `""` are removed in some cases.
```
// keptn/installer/manifests/keptn
helm template . --output-dir ../../../temp/test-templates/installer
```
```
// before
          resources:	
            requests:	
              memory: "32Mi"	
              cpu: "25m"	
            limits:	
              memory: "64Mi"	
              cpu: "100m"
// after:
          resources:
            limits:
              cpu: 100m
              memory: 64Mi
            requests:
              cpu: 25m
              memory: 32Mi
```
Injecting a custom value for resources and seeing it reflected in the generated template
```
// keptn/helm-service/chart
helm template . --output-dir ../../temp/test-templates/with-values/helm-service --set helmservice.resources.limits.cpu=2000m
```

### DIFF
--- a/installer/manifests/keptn/templates/_helpers.tpl
+++ b/installer/manifests/keptn/templates/_helpers.tpl
@@ -79,16 +79,6 @@ readinessProbe:
   periodSeconds: 5
 {{- end }}
 
-{{- define "keptn.distributor.resources" -}}
-resources:
-  requests:
-    memory: "16Mi"
-    cpu: "25m"
-  limits:
-    memory: "32Mi"
-    cpu: "100m"
-{{- end }}
-
 {{/*
 preStop hook for keptn deployments
 */}}
--- a/installer/manifests/keptn/templates/remediation-service.yaml
+++ b/installer/manifests/keptn/templates/remediation-service.yaml
@@ -52,12 +52,7 @@ spec:
         ports:
         - containerPort: 8080
         resources:
-          requests:
-            memory: "64Mi"
-            cpu: "50m"
-          limits:
-            memory: "1Gi"
-            cpu: "200m"
+          {{- toYaml .Values.remediationService.resources | nindent 10 }}
         env:
           - name: EVENTBROKER
             value: 'http://localhost:8081/event'
@@ -73,7 +68,8 @@ spec:
         {{- include "keptn.dist.readinessProbe" . | nindent 8 }}
         ports:
           - containerPort: 8080
-        {{- include "keptn.distributor.resources" . | nindent 8 }}
+        resources:
+          {{- toYaml .Values.distributor.resources | nindent 10 }}
         env:
           - name: PUBSUB_TOPIC
             value: 'sh.keptn.event.get-action.triggered'
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -82,3 +82,4 @@ For a realtime overview of current contributors to the Keptn project, we refer t
 * [Eric Y. Kim](https://github.com/eyskim)
 * [Markus Lackner](https://github.com/markuslackner)
 * [René Panzar](https://github.com/renepanzar)
+* [Alberto Gil de Dios](https://github.com/albertogdd)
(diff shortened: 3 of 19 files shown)


installer/manifests/keptn/templates/_helpers.tpl
```diff
@@ -81,12 +81,7 @@ readinessProbe:
 
 {{- define "keptn.distributor.resources" -}}
 resources:
-  requests:
-    memory: "16Mi"
-    cpu: "25m"
-  limits:
-    memory: "32Mi"
-    cpu: "100m"
+  {{- toYaml .Values.distributor.resources | nindent 2 }}
```
mowies, line 84: Is this helper even needed anymore if we have the resources inside the values now anyways?

thisthat (approved):
LGTM :shipit: 
Thank you for your contribution 🙌

installer/manifests/keptn/templates/remediation-service.yaml
```diff
@@ -73,7 +68,8 @@ spec:
         {{- include "keptn.dist.readinessProbe" . | nindent 8 }}
         ports:
           - containerPort: 8080
-        {{- include "keptn.distributor.resources" . | nindent 8 }}
+        resources:
+            {{- toYaml .Values.distributor.resources | nindent 12 }}
```
mowies, line 72: pls fix the indentation here :)

mowies (approved):
LGTM!

PR #7663, opened 2022-05-05
fix(installer): Custom readiness probe for MongoDB to fix default one

mowies:
## This PR

- adds a custom readiness probe for mongodb since the default one has error in keptn
- the probe was copied from bitnami/mongodb and simplified since we don't have a secondary DB

References:
https://github.com/bitnami/charts/pull/9916
https://github.com/bitnami/charts/blob/master/bitnami/mongodb/templates/common-scripts-cm.yaml#L33-L35


RealAnna (approved):
LGTM

PR #7671, opened 2022-05-06
feat: Introduce EventSender to cp-connector

odubajDT:

### DIFF
--- a/cp-connector/pkg/controlplane/cp_eventsender.go
+++ b/cp-connector/pkg/controlplane/cp_eventsender.go
@@ -1,0 +1,28 @@
+package controlplane
+
+import (
+	"context"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/keptn/go-utils/pkg/lib/v0_2_0"
+)
+
+type CPEventSender struct {
+	Sender EventSender
+}
+
+func (e *CPEventSender) SendEvent(event cloudevents.Event) error {
+	keptnEvent, err := v0_2_0.ToKeptnEvent(event)
+	if err != nil {
+		return err
+	}
+	return e.Sender(keptnEvent)
+}
+
+func (e *CPEventSender) Send(ctx context.Context, event cloudevents.Event) error {
+	keptnEvent, err := v0_2_0.ToKeptnEvent(event)
+	if err != nil {
+		return err
+	}
+	return e.Sender(keptnEvent)
+}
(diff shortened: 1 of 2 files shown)


cp-connector/pkg/controlplane/cp_eventsender.go
```diff
@@ -0,0 +1,28 @@
...
+type CPEventSender struct {
+	Sender EventSender
+}
+
+func (e *CPEventSender) SendEvent(event cloudevents.Event) error {
+	keptnEvent, err := v0_2_0.ToKeptnEvent(event)
+	if err != nil {
+		return err
+	}
+	return e.Sender(keptnEvent)
```
thisthat, line 19: I would avoid code duplication here
```suggestion
	return Send(context.TODO(), keptnEvent)
```

warber (changes requested):
`cp-connector` was implemented without a direct dependency to the cloudevents sdk. It would be nice to keep it like this. Plus, I don't see a benefit of having this snippet in the lib.

odubajDT:
> `cp-connector` was implemented without a direct dependency to the cloudevents sdk. It would be nice to keep it like this. Plus, I don't see a benefit of having this snippet in the lib.

It is not necessarry to have it here, but if multiple services are using CPEventSender (e.g. lighthouse, approval, probably also mongodb-datastore,...) there is no need to duplicate this code. What are your reasons for not having it in this library ?

warber:
I would not share it anywhere. Given that we shall really avoid bringing in the cloudevents-sdk libary, we would need to pass in the event directly as a type of `KeptnContextExtendedCE`. What would be left is basically a tiny struct around the EventSender with the same method signature:
```
type CPEventSender struct {
	Sender EventSender
}

func (e *CPEventSender) SendEvent(event models.KeptnContextExtendedCE) error {
	return e.Sender(event)
}
```

odubajDT:
> I would not share it anywhere. Given that we shall really avoid bringing in the cloudevents-sdk libary, we would need to pass in the event directly as a type of `KeptnContextExtendedCE`. What would be left is basically a tiny struct around the EventSender with the same method signature:
> 
> ```
> type CPEventSender struct {
> 	Sender EventSender
> }
> 
> func (e *CPEventSender) SendEvent(event models.KeptnContextExtendedCE) error {
> 	return e.Sender(event)
> }
> ```

Ok, so you agree to have this in each service? If yes, then we're good :)

warber:
yeah if you're fine with it lets keep it like that :)

Issue #7435, opened 2022-04-11
core: adapt internal services to use `cp-connector`


warber:
With https://github.com/keptn/keptn/pull/7418 `cp-connector` will be introduced to connect Keptn integrations/services to the control plane. The following services need to be adapted to make use `cp-connector` instead of the `distributor`:

- [ ] lighthouse-svc - POC available (1)
- [x] approval-svc
- [x] api-svc (this service only sends events to NATS)
- [ ] mongodb-datastore

## Acceptance Criteria
- [ ] All control plane services above are able to run **without** distributor but using the `cp-connector` library
- [ ] Helm charts are updated to don't use the distributor

## DoD 

- [ ] Manual teast of install/upgrade via helm w/ and w/o `--reuse-values` 
- [ ] Adapt the integration tests 

## Resources

(1) https://github.com/keptn/keptn/pull/7334

PR #7672, opened 2022-05-06
feat(api): Send events directly to nats instead via distributor

bacherfl:
Part of #7435 

Integration test run: https://github.com/keptn/keptn/runs/6346288532?check_suite_focus=true


odubajDT:
Do we please have an integration tests run for this PR?

odubajDT (approved):

PR #7570, opened 2022-04-25
refactor(installer): Merge charts

mowies:
### This PR
- merges the 2 keptn sub-charts together:
  - control-plane
  - continuous-delivery
- this merge had the consequence that some helm values needed to be adjusted
- the keptn CLI was also changed corresponding to the changed helm values
- helm values that were changed in this PR:
  - `continuous-delivery` -> `continuousDelivery`
  - `control-plane`: removed because the keptn chart itself is now the control-plane
  - `control-plane.enabled`: removed for the above reason and because it's always enabled
  - all helper functions and values from `control-plane` and `continuous-delivery`subcharts were merged together
- updated all skaffold files to reflect the new k8s labels

#### Notes
There are some follow-up tasks noted in #7626 

⚠️this is a breaking change⚠️

PR #7529, opened 2022-04-20
refactor(installer): Make use of common chart

mowies:
### This PR
- makes use of the new common chart
- removes the `common` key from the helm values in favor of moving all contained values out of it. This is to prevent name collisions with the `common` helm chart
- removes a few unused helper functions

PR #7678, opened 2022-05-09
feat(installer): Helm Chart revamp

### LINKED ISSUE #6042: Use best practise helm chart patterns in installer
This is a task derived from #5456.

As a provider of a whole lot of different helm charts, Bitnami has come to be a sort of best practise provider in helm chart implementations. They make use of a [common library chart](https://github.com/bitnami/charts/tree/master/bitnami/common) which we could also use as inspiration in this repo.
It provides useful helper functions to streamline the different manifests inside our installer and make it overall more maintainable and readable.

### DoD
* implement a separate helm chart or make use of bitnamis chart (Apache-2.0 License)
* replace custom implementations in the installer with the helper from the library chart where possible

Lots of work for this was already done in #5456 so it should be used as inspiration and baseline for this ticket.

### LINKED ISSUE #6981: Keptn Installer: introduce deployment parameters to keptn helm chart (control-plane) to make resource limits/requests customisable

### LINKED ISSUE #6982: Keptn Installer: introduce deployment parameters to keptn helm chart (control-plane) to enable custom sidecar configurations
We have the necessity to configure a custom sidecar container additionally to the keptn api-gateway-nginx container. The idea is to run a reverseproxy as a sidecar container independently of the already existent nginx-container within the same deployment of api-gateway-nginx.

In order to make this possible the current api-gateway-nginx.yaml of the Keptn Helm Chart (control-plane) needs some adjustments.

The introduction of deployment parameters would enable such a custom sidecar injections in a generic way. Furthermore, to be fully flexible it is also necessary to introduce extra volumes (for mounting configmaps or secrets) that can be referenced by the sidecar. The bitnami nginx-charts provide a good reference implementation:

* sidecar: https://github.com/bitnami/charts/blob/master/bitnami/nginx/templates/deployment.yaml#L298
* extra volumes: https://github.com/bitnami/charts/blob/master/bitnami/nginx/templates/deployment.yaml#L323

This proposal has no incompatibilities within the current keptn helm charts but rather introduces a generic extension point to inject such sidecars or any other containers.

I am open for feedback as well as for other solutions!

mowies:
⚠️THIS IS A BREAKING CHANGE ⚠️

Fixes done in this PR of PRs:
### Merge Charts (#7570):
- merges the 2 keptn sub-charts together:
  - control-plane
  - continuous-delivery
- this merge had the consequence that some helm values needed to be adjusted
- the keptn CLI was also changed corresponding to the changed helm values
- helm values that were changed in this PR:
  - `continuous-delivery` -> `continuousDelivery`
  - `control-plane`: removed because the keptn chart itself is now the control-plane
  - `control-plane.enabled`: removed for the above reason and because it's always enabled
  - all helper functions and values from `control-plane` and `continuous-delivery`subcharts were merged together
- updated all skaffold files to reflect the new k8s labels

### Resource usage limits/requests customization (#7661):
- Makes possible to inject your own custom resources values using the Helm Charts in the following pods

### Introduce usage of Keptn Common Helm Chart (#7529):
- makes use of the new common chart
- removes the `common` key from the helm values in favor of moving all contained values out of it. This is to prevent name collisions with the `common` helm chart
- removes a few unused helper functions

Integration tests: https://github.com/keptn/keptn/actions/runs/2501384576

Diff views of the changes to the final helm template:
[JMeter-Service changes](https://www.diffchecker.com/3dYhFsYU)
[Helm-Service changes](https://www.diffchecker.com/fu2dwa6I)
[Keptn changes](https://www.diffchecker.com/p0ucQV1F)

Fixes #6981
Fixes #6982
Fixes #6042

### Follow-up tickets
https://github.com/keptn/keptn.github.io/issues/1267


distributor/pkg/uniform/controlplane/controlplane.go
```diff
@@ -66,7 +66,7 @@ func (c *ControlPlane) createRegistrationData() models.Integration {
 	if c.env.Location == "" {
 		location = config.ConnectionTypeToLocation[c.connectionType]
 	} else {
-		location = c.env.Location
+		location = c.env.K8sDeploymentComponent
```
bacherfl, line 69: i believe we can revert this back to `location = c.env.Location` - if `env.Location` is not set explicitly, the location will be set to either `control-plane` or `remote-execution-plane`, so I think using the name of the component, i.e. the service might not be the correct value for this field

docs/developer/install_master.md
```diff
@@ -18,7 +18,7 @@ helm repo add keptn-dev https://charts-dev.keptn.sh
 7. For current master use `<current_version>-dev` images, for images built in a PR add `-PR-<PR_#>` to the tag,
 8. To install version 0.13.0-dev-PR-1234 
 ``` 
-helm upgrade --install keptn keptn-dev/keptn -n keptn --create-namespace --set=continuous-delivery.enabled=true --wait --version 0.13.0-dev-PR-1234
+helm upgrade --install keptn keptn-dev/keptn -n keptn --create-namespace --set=continuousDelivery.enabled=true --wait --version 0.13.0-dev-PR-1234
```
bacherfl, line 21: let`s use a `0.16.1-dev-PR` build as an example, as the `continuousDelivery` values object was not available in previous versions

distributor/pkg/uniform/controlplane/controlplane.go
mowies: reverted 👍🏼

docs/developer/install_master.md
mowies: fixed

bacherfl (dismissed):

bacherfl (dismissed):

RealAnna (approved):
LGTM

Issue #7613, opened 2022-05-02
Core: Polish ZDU implementation


thisthat:
## Action Items

- [ ] Configure pre-stop and grace period to a default of 5 and 60 seconds, respectively
- [ ] Immediately unsubscribe from NATS when the pre-hook starts (stopping the cp-library & shipyard-controller JetStream)

## Technical details

Prestop and grace period can be configured through our helm charts. 
To unsubscribe from NATS as the pre-hook starts we could look for reference implementations. One option is to use the filesystem. A library that might be worth using is [fsnotify](https://github.com/fsnotify/fsnotify). 

## DoD

- [ ] Add a unit test 
- [ ] Action items are completed

## Progress:

- [x] shipyard-controller
- [x] approval service 
- [ ] webhook-service
- [ ] lighthouse-service (in progress)
- [ ] remediation-service

PR #7682, opened 2022-05-09
feat: Change default values of preStop hook time and grace period

bacherfl:
Part of #7613

This PR changes the default values of the preStop hook time and grace period to 5s/60s respectively.

For the shipyard-controller, an internal preStop hook endpoint is introduced, which will take care of unsubscribing from nats and shutting down background tasks, such as the sequence dispatcher/event dispatcher, as well as waiting a defined number of seconds before the sigterm signal is sent to the container

Integration test run: https://github.com/keptn/keptn/actions/runs/2304981277


odubajDT (approved):
lgtm

PR #7686, opened 2022-05-09
feat: Refactor `go-sdk` to use `cp-connector` internally

### LINKED ISSUE #7434: go-sdk: use `cp-connector` to connect to keptn
Currently the experimental [go-sdk](https://github.com/keptn/keptn/tree/master/go-sdk) is, like every other regular keptn service relying on the `distributor` to communicate with Keptn. With the introduction of `cp-connector` however the `distributor` is not required anymore. Hence, the `go-sdk` shall be adopted to use `cp-connector` internally.

## Acceptance Criteria
- `go-sdk` is using `cp-connector` internally
- Keptn Service written in using `go-sdk` can run without a `distributor`

## DoD

- [x] Webhook-, remediation-, approval-svc don't use the distributor
- [ ] Integration tests are adapted
- [x] Extend the unit tests in the SDK
- [ ] Manual test of installing/upgrading via helm w/ and w/o `--reuse-values`

warber:
closes #7434 

PR contains the following changes:
* integration of `cp-connector` to go sdk
* enhancement to sdk to host its health endpoint
* restructuring of Fake keptn for handler tests
* adaptations to code of webhook service and remediation service (only handler test code using fake keptn)
* adaptations to k8s manifests (local ones as well in installer) to not include the distributor in the deployments any more

we agreed to keep this changes in one single PR

latest integration test run: https://github.com/keptn/keptn/runs/6485611987?check_suite_focus=true


go-sdk/pkg/sdk/events.go
```diff
@@ -0,0 +1,141 @@
+package sdk
```
warber, line 1: I've refactored a bit and I've moved out these functions from `keptn.go` to have it in one place and covered by basic unit tests. This logic can/should be greatly improved in a future PR

go-sdk/pkg/sdk/keptn.go
```diff
@@ -258,179 +213,117 @@ func (k *Keptn) gotEvent(ctx context.Context, event cloudevents.Event) {
...
-	keptnv2.Decode(eventData, &genericEvent)
-	if genericEvent["status"] == nil || genericEvent["status"] == "" {
-		genericEvent["status"] = "succeeded"
+func (k *Keptn) Start() error {
+	if k.env.HealthEndpointEnabled {
+		k.healthEndpointRunner(k.env.HealthEndpointPort, k.controlPlane)
 	}
+	ctx := k.getContext(k.gracefulShutdown)
+	err := k.controlPlane.Register(ctx, k)
+	ctx.Value(gracefulShutdownKey).(wgInterface).Wait()
```
bacherfl, line 280: should we add a check whether the cast to `wgInterface` was succesful here? Alternatively, we could return the waitGroup as a result of `getContext()`

go-sdk/pkg/sdk/mock_eventsource.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"context"
+	"fmt"
+	"github.com/keptn/go-utils/pkg/api/models"
+	"github.com/keptn/keptn/cp-connector/pkg/controlplane"
+	"sync"
+)
+
+func NewTestEventSource() *TestEventSource {
```
bacherfl, line 11: should we move the mock implementations to `pkg/sdk/mock`?

go-sdk/pkg/sdk/keptn.go
```diff
@@ -258,179 +213,117 @@ func (k *Keptn) gotEvent(ctx context.Context, event cloudevents.Event) {
...
-	keptnContext, err := triggeredEvent.Context.GetExtension(KeptnContextCEExtension)
-	if err != nil {
-		return nil, fmt.Errorf("unable to get keptn context from '.triggered' event: %v", err)
+	return controlplane.RegistrationData{
+		Name: k.source,
+		MetaData: models.MetaData{
+			Hostname:           k.env.K8sNodeName,
+			IntegrationVersion: k.env.Version,
+			Location:           k.env.Location,
+			DistributorVersion: "0.15.0", // note: to be deleted when bridge stops requiring this info
```
bacherfl, line 263: should we just use `k.env.Version` here as well for now?

go-sdk/pkg/sdk/mock_eventsource.go
warber: I've tried it. unfortunately It results in a circular dependency for m̀ock_taskhandler.go` :(

warber: turned out that i don't need the mocks anymore at all :D

go-sdk/pkg/sdk/keptn.go
warber: good idea, will return the wg

go-sdk/pkg/sdk/keptn.go
warber: agreed that we keep the dummy value, as it is not visible to the user and will be deleted eventually

bacherfl (approved):
LGTM - later on we should refacor the go-sdk's `OnEvent()` method to reduce its complexity, but that would exceed the scope of this PR right now, so I suggest to do this seperately

PR #7696, opened 2022-05-10
perf(bridge): Use adapted sequence endpoint for project endpoint of bridge server

### LINKED ISSUE #6213: Use the adapted sequence endpoint to reduce the amount of API-calls for the data in the environment screen
With #6093 the sequence endpoint provides the possibility to accept multiple keptn contexts. With this change, the bridge server can remove again stages x services x 1 API calls. This will lead to a best-case of 6 API calls instead of 5 + (stages x services x 1).

## Definition of Done
- [x] The bridge server uses the adjusted endpoint for fetching the environment screen data
- [x] Unit test adapted

Kirdock:
Closes #6213 


ermin-muratovic (approved):
lgtm

PR #7699, opened 2022-05-10
test(bridge): Fixed flaky create project UI test

Kirdock:
There was a flaky test in `project-create.spec.ts`.
Fix: Added missing API mocks.

Cause: Missing API mocks.
On create, there were two calls that failed. Because nothing after the creation is validated the two calls sometimes came in and sometimes did not.


ermin-muratovic (approved):
lgtm

PR #7703, opened 2022-05-10
chore(bridge): Remove dev-dependency jest-fetch-mock

Kirdock:
- Removes dev dependency `jest-fetch-mock`
- Adapted fetching the config for the look and feel (with `HttpClient` instead of `fetch`)
- Removed data stored in `window.config` and instead correctly save it in the environment file
- Additional error check for fetching the look and feel zip file


ermin-muratovic (changes requested):

ermin-muratovic (approved):
lgtm

PR #7716, opened 2022-05-11
fix(webhook-service): Added denied curl in webhook beta based on host

### LINKED ISSUE #7697: Webhook inconsistent denied list
**Environment**

* Keptn Version (`keptn version`): keptn 0.14.2
* Kubernetes Cloud Provider (e.g., GKE, AKS): GKE 19-21  vs GKE22

**Affected Component**

* [ ] Approval
* [ ] Bridge
* [ ] CLI
* [ ] Datastore
* [ ] Docs
* [ ] Helm
* [ ] Istio
* [ ] Jmeter
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [ ] REST API
* [x] Webhook

**Describe the bug**

Webhook denies all services under the KUBERNETES_SERVICE_HOST mask. This variable is not defined by default up to gke 21 but it is in GKE 22 meaning webhook filters for e.g. 10.24.0.1 and thus keptn internal services only from GKE22 

**To Reproduce**

run webhook integration test removing the env var from webhook to see all of them pass, add it to make  a lot of failures
image

**Expected behavior**

Denied list should be consistent for all GKE versions

## Technical Details

Integration tests should be adapted to use the public Cluster IP instead of internal service names

### LINKED ISSUE #7792: Webhook cannot retrieve configuration if stored in project without --all-stages
Your issue may already be reported! Please search on the [issue tracker](https://github.com/keptn/keptn/issues) before creating a new one.

**Environment**

* Keptn Version (`keptn version`): 0.15.0 
*Resource service in branch mode

**Affected Component**

* [ ] Approval
* [ ] Bridge
* [ ] CLI
* [ ] Datastore
* [ ] Docs
* [ ] Helm
* [ ] Istio
* [ ] Jmeter
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [ ] REST API
* [X] Webhook

**Describe the bug**

Since the resource service in branching mode works with different commit id per branch if we trigger a delivery for webhook on a branch, e.g. dev  after saving the config in master branch (cli without --stage or --all-stages), shipyard may use a commitid that is not pointing at master brench causing the sequence to fail.

**Expected behavior**

If we fail to retrieve by commit id webhook should try to retrieve without it

**Screenshots**
webhook looks 
image
image
image

RealAnna:
## This PR

- adds default k8s service IP to the denied list
- adds check for host domain in webhook beta, so that all services under a forbidden domain are blocked
- refactors integration tests
- adds integration test for minishift where we ping the exposed ip of the API
-  removes integration tests from minishift that would fail due to www.keptn.sh being unreachable in that cluster
- adds comment about incompatibilities of adding resources only on project master branch, with resource service on branching mode

### Related Issues

Fixes #7697
Fixes #7792

integration test : https://github.com/keptn/keptn/runs/6523668460?check_suite_focus=true
new run for minishift https://github.com/keptn/keptn/runs/6549876435?check_suite_focus=true


test/go-tests/test_webhook.go
```diff
@@ -188,12 +188,12 @@ spec:
             name: "my-webhook-k8s-secret"
             key: "my-key"
       requests:
-        - url: http://shipyard-controller:8080/v1/project/{{.data.project}}
+        - url: http://google.com
```
thisthat, line 191: Let's do some self advertise 😄 
```suggestion
        - url: http://keptn.sh
```

webhook-service/handler/handler.go
```diff
@@ -199,6 +199,7 @@ func (th *TaskHandler) performWebhookRequests(webhook lib.Webhook, eventAdapter
 	logger.Infof("executing webhooks for subscriptionID %s", webhook.SubscriptionID)
 	for _, req := range webhook.Requests {
 		request, err := th.CreateRequest(req)
+		logger.Info("request is ", request, "error", err)
```
warber, line 202: leftover or intended?

webhook-service/lib/common.go
```diff
@@ -54,3 +64,8 @@ func GetDeniedAlphaURLs(env map[string]string) []string {
 	}
 	return urls
 }
+
+func GetDefaultHost() string {
+	res, _ := keptnutils.ExecuteCommand("kubectl", []string{"get", "svc", "kubernetes", "-o", "jsonpath='{.spec.clusterIP}'"})
```
warber, line 69: pls provide an abstraction for it so that this isn't going to be executing during unit tests

warber (dismissed):

webhook-service/lib/common.go
RealAnna: the whole thing is leftover that is not really needed

installer/manifests/keptn/charts/control-plane/templates/webhook-service.yaml
```diff
@@ -124,6 +124,8 @@ data:
     kubernetes.default
     kubernetes.default.svc
     kubernetes.default.svc.cluster.local
+    svc.cluster.local
+    cluster.local
```
mowies, line 128: doesn't this mean that you can't send a webhook to anything running on k8s now?
So all webhook traffic needs to leave the cluster and east-to-west traffic is basically forbidden, am I missing something?

RealAnna: arg that was for testing I revert it

mowies: having `svc.cluster.local` still has similar behaviour...

webhook-service/main.go
```diff
@@ -46,7 +46,7 @@ func main() {
 			taskHandler,
 		),
 		sdk.WithAutomaticResponse(false),
-		sdk.WithLogger(log.New()),
```
thisthat, line 49: is this intended?

RealAnna: put it back thankyou!

installer/manifests/keptn/charts/control-plane/templates/webhook-service.yaml
RealAnna: so at the end we discussed to go that route, and disable any connection by default, I revert the changes to include both denied

thisthat (approved):
LGTM

warber (approved):

PR #7723, opened 2022-05-12
feat(approval-service): Consider nats connection in readiness probe

bacherfl:
Part of #7613
This PR modifies the readiness probe of the approval service to consider the current state of the control plane connector. The service will now be considered ready, only when it's connected to the controlplane. Until that is the case, previous pods of the approval-service will not be terminated. This will ensure that there is always at least one replica available that is ready to receive events.

Integration test run: https://github.com/keptn/keptn/actions/runs/2313756860

### DIFF
--- a/approval-service/main.go
+++ b/approval-service/main.go
@@ -9,13 +9,14 @@ import (
 	"sync"
 	"syscall"
 
+	keptnapi "github.com/keptn/go-utils/pkg/api/utils"
+
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	"github.com/kelseyhightower/envconfig"
 	"github.com/pkg/errors"
 	logger "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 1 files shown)


warber (dismissed):
lgtm

approval-service/go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/keptn/go-utils v0.15.1-0.20220512082458-c5e1b7519bec
-	github.com/keptn/keptn/cp-connector v0.0.0-20220509104843-d040776782e2
+	github.com/keptn/keptn/cp-connector v0.0.0-20220512103206-c36faf0071dc
```
odubajDT, line 10: pls update dependency to the newest

odubajDT (approved):
lgtm

PR #7727, opened 2022-05-12
feat: Add ability to customize client_max_body_size in the nginx gateway

### LINKED ISSUE #7567: Cannot add big resources to Keptn
**Environment**

* Client OS (e.g., Linux, macOS, Windows): Linux
* Keptn Version (`keptn version`): 0.13.1
* Kubernetes Cloud Provider (e.g., GKE, AKS): k3d
* Kubernetes version (`kubectl version`): Client: v1.23.5; Server: v1.22.6+k3s1

**Affected Component**

* [ ] Approval
* [ ] Bridge
* [x] CLI
* [ ] Datastore
* [ ] Docs
* [ ] Helm
* [ ] Istio
* [ ] Jmeter
* [ ] Kubernetes Integration
* [ ] Openshift Integration
* [x] REST API
* [ ] Webhook

**Describe the bug**

It is not possible to add bigger resources to Keptn via the Keptn CLI or the go-client.
All resources seem to be limited to around 768KiB and will be rejected with the following message:
```
Adding resource random-ascii.txt to all stages of project test-project
Error: Resource random-ascii.txt could not be uploaded: invalid character '<' looking for beginning of value
-----DETAILS-----<html>
<head><title>413 Request Entity Too Large</title></head>
<body>
<center><h1>413 Request Entity Too Large</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

**To Reproduce**

1. Create a random file, I used the following command to generate a random ascii file:
`base64 /dev/urandom -w0 | head -c $((1024*1024)) > random-ascii.txt`
2. Add the generated file to a already existing keptn project: `keptn add-resource --project=test-project --service=test-service --all-stages --resource=random-ascii.txt --resourceUri=random-ascii.txt`
3. See error

**Expected behavior**

Adding bigger resources to a Keptn project should work and not display an error.

**Current behavior**

Adding big resources yields an error

**Additional context**

We have a use-case for the job-executor-service which requires us to upload multiple bigger (>1MiB) files to Keptn, such that they can be used in the defined jobs. With the current behavior this is not possible.

Raffy23:
## This PR
- adds the ability to adjust the `client_max_body_size` parameter of nginx at install time (via helm values) to allow upload of larger resources if desired

### Related Issues

Fixes #7567 

### Notes
I'm not sure if we should already increase the default value to 5 MB or if 1 MB is sufficient for most workloads.

### DIFF
--- a/installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
+++ b/installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
@@ -55,7 +55,9 @@ data:
           listen       8080;
           server_name  _;
 
-          client_max_body_size {{ .Values.apiGatewayNginx.clientMaxBodySize | default "1m" }};
+          {{- if .Values.apiGatewayNginx.clientMaxBodySize }}
+            client_max_body_size {{ .Values.apiGatewayNginx.clientMaxBodySize }};
+          {{- end }}
 
           include /etc/nginx/keptn-endpoints-pre-0-7.conf;
           include /etc/nginx/keptn-endpoints.conf;
--- a/installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
+++ b/installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
@@ -55,6 +55,8 @@ data:
           listen       8080;
           server_name  _;
 
+          client_max_body_size {{ .Values.apiGatewayNginx.clientMaxBodySize | default "1m" }};
+
           include /etc/nginx/keptn-endpoints-pre-0-7.conf;
           include /etc/nginx/keptn-endpoints.conf;
       }
--- a/installer/manifests/keptn/charts/control-plane/values.yaml
+++ b/installer/manifests/keptn/charts/control-plane/values.yaml
@@ -75,6 +75,7 @@ apiGatewayNginx:
   nodeSelector: {}
   gracePeriod: 60
   preStopHookTime: 5
+  clientMaxBodySize: "1m"
 
 remediationService:
   image:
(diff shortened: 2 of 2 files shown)


installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
```diff
@@ -55,6 +55,8 @@ data:
           listen       8080;
           server_name  _;
 
+          client_max_body_size {{ .Values.apiGatewayNginx.clientMaxBodySize | default "1m" }};
```
christian-kreuzberger-dtx, line 58: We should make sure this value exists (e.g., what if `.Values.apiGatewayNginx` is `nil`)

installer/manifests/keptn/charts/control-plane/values.yaml
```diff
@@ -75,6 +75,7 @@ apiGatewayNginx:
   nodeSelector: {}
   gracePeriod: 60
   preStopHookTime: 5
+  clientMaxBodySize: "1m"
```
christian-kreuzberger-dtx, line 78: I would suggest raising this either to a higher value, or `0`,w hich according to the [docs](http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size) means that there are no checks for this.
@bacherfl @mowies any concerns regarding that?

installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml
Raffy23: I refactored it such that the `client_max_body_size` directive will only be printed if the helm value is defined.

installer/manifests/keptn/charts/control-plane/values.yaml
```diff
@@ -75,6 +75,7 @@ apiGatewayNginx:
   nodeSelector: {}
   gracePeriod: 60
   preStopHookTime: 5
+  clientMaxBodySize: ""
```
christian-kreuzberger-dtx, line 78: Let's set this to "5m"

mowies (approved):
LGTM!

PR #7736, opened 2022-05-13
feat(bridge): Select project stage from project overview

### LINKED ISSUE #7015: Selection of stage
### Acceptance Criteria

When clicking on the stage in the project overview screen:

image

the corresponding stage is already selected:

image

#### Details

Right now, I just get to the environment screen and no stage is selected. 

## DoD 

- UI tests

TannerGilbert:
## This PR

- Allows the user to select a stage from the project overview page
- Adds a query parameter for the selected stage (e.g. `?stage=production`)

### Related Issues

Fixes #7015

https://user-images.githubusercontent.com/36239763/168252934-e165644c-90f1-428a-8aa9-2d5a2f1d7e94.mp4

### DIFF
--- a/bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
+++ b/bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
@@ -137,7 +137,7 @@ export class KtbStageOverviewComponent implements OnDestroy, OnInit, AfterConten
     return stage?.toString();
   }
 
-  public selectStage($event: MouseEvent, stage: Stage): void {
+  public selectStage($event: MouseEvent, stage: Stage, filterType: ServiceFilterType): void {
     this.router.navigate([], {
       queryParams: {
         stage: stage.stageName,
--- a/bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
+++ b/bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
@@ -58,15 +58,14 @@ export class KtbStageOverviewComponent implements OnDestroy, OnInit, AfterConten
   }
 
   public ngAfterContentInit(): void {
-    this.route.queryParams.subscribe((params) => {
-      if (params['stage'] && this.project) {
-        let stage = this.project.getStage(params['stage']);
-        if (stage) {
-          this.selectedStage = stage;
-          this.selectedStageChange.emit({ stage: stage, filterType: undefined });
-        }
+    let stageName = this.route.snapshot.queryParamMap.get('stage');
+    if (stageName && this.project) {
+      let stage = this.project.getStage(stageName);
+      if (stage) {
+        this.selectedStage = stage;
+        this.selectedStageChange.emit({ stage: stage, filterType: undefined });
       }
-    });
+    }
   }
 
   private setFilter(projectChanged: boolean): void {
@@ -138,6 +137,7 @@ export class KtbStageOverviewComponent implements OnDestroy, OnInit, AfterConten
   }
 
   public selectStage($event: MouseEvent, stage: Stage, filterType: ServiceFilterType): void {
+    this.selectedStage = stage;
     this.router.navigate([], {
       queryParams: {
         stage: stage.stageName,
@@ -145,6 +145,7 @@ export class KtbStageOverviewComponent implements OnDestroy, OnInit, AfterConten
       queryParamsHandling: 'merge',
     });
     $event.stopPropagation();
+    this.selectedStageChange.emit({ stage, filterType });
   }
 
   public ngOnDestroy(): void {
--- a/bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
+++ b/bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
@@ -1,4 +1,4 @@
-import { Component, EventEmitter, OnDestroy, OnInit, Output } from '@angular/core';
+import { Component, EventEmitter, OnDestroy, OnInit, AfterContentInit, Output } from '@angular/core';
 import { Project } from '../../_models/project';
 import { Stage } from '../../_models/stage';
 import { DataService } from '../../_services/data.service';
@@ -7,7 +7,7 @@ import { ApiService } from '../../_services/api.service';
 import { Service } from '../../_models/service';
 import { DtAutoComplete, DtFilter, DtFilterArray } from '../../_models/dt-filter';
 import { filter, map, switchMap, takeUntil, tap } from 'rxjs/operators';
-import { ActivatedRoute } from '@angular/router';
+import { Router, ActivatedRoute } from '@angular/router';
 import { Subject } from 'rxjs';
 import { DtFilterFieldDefaultDataSourceAutocomplete } from '@dynatrace/barista-components/filter-field/src/filter-field-default-data-source';
 import { ServiceFilterType } from '../ktb-stage-details/ktb-stage-details.component';
@@ -17,7 +17,7 @@ import { ServiceFilterType } from '../ktb-stage-details/ktb-stage-details.compon
   templateUrl: './ktb-stage-overview.component.html',
   styleUrls: ['./ktb-stage-overview.component.scss'],
 })
-export class KtbStageOverviewComponent implements OnDestroy, OnInit {
+export class KtbStageOverviewComponent implements OnDestroy, OnInit, AfterContentInit {
   public project?: Project;
   public selectedStage?: Stage;
   public _dataSource = new DtFilterFieldDefaultDataSource();
@@ -30,7 +30,12 @@ export class KtbStageOverviewComponent implements OnDestroy, OnInit {
   @Output() selectedStageChange: EventEmitter<{ stage: Stage; filterType: ServiceFilterType }> = new EventEmitter();
   @Output() filteredServicesChange: EventEmitter<string[]> = new EventEmitter<string[]>();
 
-  constructor(private dataService: DataService, private apiService: ApiService, private route: ActivatedRoute) {}
+  constructor(
+    private dataService: DataService,
+    private apiService: ApiService,
+    private router: Router,
+    private route: ActivatedRoute
+  ) {}
 
   public ngOnInit(): void {
     // needs to be in init because of emitter
@@ -49,6 +54,18 @@ export class KtbStageOverviewComponent implements OnDestroy, OnInit {
       const differentProject = project?.projectName !== this.project?.projectName;
       this.project = project;
       this.setFilter(differentProject);
+    });
+  }
+
+  public ngAfterContentInit(): void {
+    this.route.queryParams.subscribe((params) => {
+      if (params['stage'] && this.project) {
+        let stage = this.project.getStage(params['stage']);
+        if (stage) {
+          this.selectedStage = stage;
+          this.selectedStageChange.emit({ stage: stage, filterType: undefined });
+        }
+      }
     });
   }
 
@@ -120,10 +137,14 @@ export class KtbStageOverviewComponent implements OnDestroy, OnInit {
     return stage?.toString();
   }
 
-  public selectStage($event: MouseEvent, stage: Stage, filterType: ServiceFilterType): void {
-    this.selectedStage = stage;
+  public selectStage($event: MouseEvent, stage: Stage): void {
+    this.router.navigate([], {
+      queryParams: {
+        stage: stage.stageName,
+      },
+      queryParamsHandling: 'merge',
+    });
     $event.stopPropagation();
-    this.selectedStageChange.emit({ stage, filterType });
   }
 
   public ngOnDestroy(): void {
--- a/bridge/client/app/_components/ktb-project-tile/ktb-project-tile.component.html
+++ b/bridge/client/app/_components/ktb-project-tile/ktb-project-tile.component.html
@@ -58,7 +58,9 @@
     </div>
   </dt-tile-subtitle>
   <dt-tag-list aria-label="stages" uitestid="keptn-project-tile-stagesTags">
-    <dt-tag *ngFor="let stage of project.stages" [textContent]="stage.stageName"></dt-tag>
+    <dt-tag *ngFor="let stage of project.stages">
+      <a [routerLink]="['/project', project.projectName]" [queryParams]="{'stage': stage.stageName}">{{ stage.stageName }}</a>
+    </dt-tag>
   </dt-tag-list>
   <p class="small">Recent sequences:</p>
   <ktb-sequence-state-list [project]="project"></ktb-sequence-state-list>
--- a/bridge/client/app/_components/ktb-project-tile/ktb-project-tile.component.scss
+++ b/bridge/client/app/_components/ktb-project-tile/ktb-project-tile.component.scss
@@ -30,6 +30,6 @@ p {
   vertical-align: bottom;
 }
 
-dt-tile-title a {
+dt-tile-title a, dt-tag a {
   text-decoration: none;
 }
--- a/bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.html
+++ b/bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.html
@@ -41,7 +41,7 @@
     >
       <ktb-selectable-tile
         *ngFor="let stage of project.getStages(parentStage)"
-        (click)="selectStage($event, stage, undefined)"
+        (click)="selectStage($event, stage)"
         [selected]="selectedStage === stage"
       >
         <h2 class="m-0 mt-1 mb-1" [textContent]="stage.stageName"></h2>
--- a/bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.html
+++ b/bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.html
@@ -41,7 +41,7 @@
     >
       <ktb-selectable-tile
         *ngFor="let stage of project.getStages(parentStage)"
-        (click)="selectStage($event, stage)"
+        (click)="selectStage($event, stage, undefined)"
         [selected]="selectedStage === stage"
       >
         <h2 class="m-0 mt-1 mb-1" [textContent]="stage.stageName"></h2>


bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
```diff
@@ -52,6 +57,17 @@ export class KtbStageOverviewComponent implements OnDestroy, OnInit {
     });
   }
 
+  public ngAfterContentInit(): void {
+    let stageName = this.route.snapshot.queryParamMap.get('stage');
```
renepanzar, line 61: The query params should be the **single source of truth**. Now, we need to do the exact same steps twice in `ngAfterContentInit` and in `selectStage` (set the selected stage and emit the change).

We can subscribe to the `queryParamMap` here (instead of using the snapshot) and then do the rest of this method inside of the `subscribe` block. In the `selectStage` method we would only trigger the router navigation (and stop the propagation), and let the subscription do the magic.

renepanzar (changes requested):
Avoid doing the same thing twice when not necessary.

bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
TannerGilbert: That's actually exactly how I did it at first but then I couldn't propagate the `filterType` from the `selectStage` method.

bridge/cypress/support/pageobjects/DashboardPage.ts
```diff
@@ -39,6 +40,15 @@ class DashboardPage {
     cy.get('.dt-button-primary > span.dt-button-label').contains('Create a new project').click();
     return new NewProjectCreatePage();
   }
+
+  public clickStageTag(stage: string): EnvironmentPage {
+    cy.wait(500)
```
Kirdock, line 45: Instead of `wait(500)` I would wait for `@projects` before

bridge/cypress/support/pageobjects/DashboardPage.ts
```diff
@@ -39,6 +40,15 @@ class DashboardPage {
     cy.get('.dt-button-primary > span.dt-button-label').contains('Create a new project').click();
     return new NewProjectCreatePage();
   }
+
+  public clickStageTag(stage: string): EnvironmentPage {
```
Kirdock, line 44: This one may be flaky. We only have the stage here and a stage name may be in several projects.
I would introduce a new `uitestid` here for the `ktb-project-tile`. `ktb-project-${projectName}` or something.
Then we could write something like
```ts
cy.byTestId(`ktb-project-${projectName}`).find('dt-tag a').contains(stageName).click()
```

bridge/client/app/app.routing.ts
```diff
@@ -94,6 +94,8 @@ const routes: Routes = [
               { path: '', pathMatch: 'full', redirectTo: 'project' },
             ],
           },
+          { path: 'environment', component: KtbEnvironmentViewComponent },
```
Kirdock, line 97: I would also add a redirect here. `project/${projectName}` should be redirected to `project/${projectName}/environment`. @ermin-muratovic, @renepanzar  what's your opinion on that?

Kirdock (changes requested):

bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
```diff
@@ -52,6 +59,17 @@ export class KtbStageOverviewComponent implements OnDestroy, OnInit {
...
 
+  public ngAfterContentInit(): void {
+    const stageName = this.route.snapshot.paramMap.get('stageName');
+    if (stageName && this.project) {
+      const stage = this.project.getStage(stageName);
+      if (stage) {
+        this.selectedStage = stage;
+        this.selectedStageChange.emit({ stage: stage, filterType: undefined });
+      }
+    }
```
renepanzar, line 70: This is still kind of hacky... I'd rather go with you through my ideas. Will contact you.

renepanzar (changes requested):

bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
```diff
@@ -1,77 +1,103 @@
...
+
+  public params$ = combineLatest([this.selectedStageName$, this.paramFilterType$, this.project$]).pipe(
+    tap(([stageName, filterType, project]) => {
+      const stage = project!.getStage(stageName!);
+      if (stage) {
+        this.selectedStageChange.emit({ stage: stage!, filterType: (filterType as ServiceFilterType) ?? undefined });
+      }
+    }),
+    takeUntil(this.unsubscribe$),
+    shareReplay(1)
```
heinzburgstaller, line 69: Is this observable used anywhere outside of this component class? Otherwise it can be private. And I do not think the shareReplay is necessary. Furthermore I would not `tap` and then call a method, but `subscribe` and there call the function.

bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
```diff
@@ -1,77 +1,103 @@
...
       // services can be deleted or added; adjust filter
-      const services = this.globalFilter[this.project.projectName]?.services || [];
+      const services = this.globalFilter[project.projectName]?.services || [];
       newFilter = services.filter((service) =>
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        this.project!.getServices().some((pService) => pService.serviceName === service)
+        project!.getServices().some((pService) => pService.serviceName === service)
       );
     } else {
       newFilter = [];
```
heinzburgstaller, line 103: Can be simplified to:
```
const createFilter = (project): string[] => {
  // services can be deleted or added; adjust filter
  const services = this.globalFilter[project.projectName]?.services || [];
  return services.filter((service) => project!.getServices().some((pService) => pService.serviceName === service);
} 
const newFilter: string[] = project ? createFilter(project) : [];
```

bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
renepanzar: That is completely correct. We wanted to improve this further to avoid the actual subscription in the `ngAfterViewInit()`, but subscribe to it in the template. This will be a further investigation by Gilbert AFAIK, so there may be changes. Otherwise, in case we keep `subscribe()`, the observable needs to be adapted according to the comment you provided, @heinzburgstaller.

renepanzar, line 69: In case we don't subscribe in the template, remove the `params$` property and change the subscription in `ngAfterViewInit()`:

```suggestion
  ngAfterViewInit() {
    combineLatest([this.selectedStageName$, this.paramFilterType$, this.project$])
      .pipe(takeUntil(this.unsubscribe$))
      .subscribe(([stageName, filterType, project]) => {
        const stage = project!.getStage(stageName!);
        if (stage) {
          this.selectedStageChange.emit({ stage: stage!, filterType: (filterType as ServiceFilterType) ?? undefined });
        });
    }
```

bridge/client/app/_components/ktb-stage-overview/ktb-stage-overview.component.ts
renepanzar: We actually did not want to touch this method at all, but wanted to remove the `this.project` property. Should we still simplify the code?

bridge/cypress/support/pageobjects/DashboardPage.ts
```diff
@@ -19,7 +20,7 @@ class DashboardPage {
   }
 
   public clickProjectTile(projectName: string): EnvironmentPage {
-    cy.wait(500).get(this.PROJECT_TILE_LOCATOR.replace('proj_pattern', projectName)).click();
+    cy.wait('@projects').get(this.PROJECT_TILE_LOCATOR.replace('proj_pattern', projectName)).click();
```
renepanzar, line 23: Thanks for removing the random wait time with an actual response to wait for :+1:

bridge/client/app/app.routing.ts
renepanzar: @Kirdock , @ermin-muratovic - you're more experienced with the project, what are your thoughts?

Kirdock: Maybe it's a bit unclear.
- At the moment the environment screen is `project/${projectName}`.
- With the additional params we introduced `project/${projectName}/environment/...` in order to not interfere with other routes. `project/${projectName}/environment/` did not exist before, so there aren't any links to it.
- It is a weird behavior if you go to the environment screen (`project/${projectName}`), select a stage, and then `/environment/...` is added to the path

renepanzar: To be honest, I'm not sure if anyone is paying attention to the URL when using the app. With this in mind I wouldn't spend too much time thinking about this "weird behavior". If it works, it works :)

bridge/client/app/_components/ktb-trigger-sequence/ktb-trigger-sequence.component.ts
```diff
@@ -79,7 +79,7 @@ export class KtbTriggerSequenceComponent implements OnInit, OnDestroy, AfterView
   public customFormData: CustomSequenceFormData = {};
 
   @Input() public projectName!: string;
-  @Input() public stage: string | undefined;
+  @Input() public stage: string | undefined | null;
```
Kirdock, line 82: We actually don't need `undefined` here. With the async pipe it can either be null or string.
```ts
@Input() public stage: string | null = null
```

Kirdock (changes requested):
- [x] The number of the types (remediation, failed quality gate, approval) is now underlined (1).
- [x] The "no service" message now is blue on hover (2)
image

- [ ] If you set a filter type (remediation, approval, failed quality gate), the filter is not applied on refresh. We also need an UI test for that.
- [ ] The filter should not be set or clickable if the amount is 0. As you can see the filter on the right side (stage-details) is disabled if it is 0.

@renepanzar @ermin-muratovic @heinzburgstaller what do you say to redirects? Should we use `router.navigate` and trigger the observables again (route params, query params, etc.) or should we use `location.go()` as often as possible?

heinzburgstaller:
> @renepanzar @ermin-muratovic @heinzburgstaller what do you say to redirects? Should we use `router.navigate` and trigger the observables again (route params, query params, etc.) or should we use `location.go()` as often as possible?

We should use the Angular router as much as possible.

bridge/cypress/support/pageobjects/EnvironmentPage.ts
```diff
@@ -32,6 +34,16 @@ class EnvironmentPage {
     return this;
   }
 
+  public clickFilterType(stage: string, filterType: string): this {
```
Kirdock, line 37: Can be a bit improved. We only have the filter types 'problem', 'evaluation' and 'approval' so the parameter can be set to `filterType: 'problem' | 'evaluation' | 'approval'`

bridge/cypress/support/pageobjects/EnvironmentPage.ts
```diff
@@ -13,8 +13,10 @@ class EnvironmentPage {
     return this;
   }
 
-  public visit(project: string): this {
-    cy.visit(`/project/${project}`).wait('@metadata').wait('@project');
+  public visit(project: string, stage = '', filterType = ''): this {
+    cy.visit(stage ? `/project/${project}/environment/stage/${stage}?filterType=${filterType}` : `/project/${project}`)
```
Kirdock, line 17: I would still differentiate here and only add the query params if they are set.
```ts
public visit(project: string, stage?: string, filterType?: string): this {
    const query = filterType ? `?filterType=${filterType}` : '';
    cy.visit(stage ? `/project/${project}/environment/stage/${stage}${query}` : `/project/${project}`)
```

bridge/cypress/integration/environment.spec.ts
```diff
@@ -37,6 +37,33 @@ describe('Environment Screen default requests', () => {
...
+  });
+
+  it('stage-detail component should exist when navigating to /environment/stage url', () => {
+    environmentPage.visit(project, stage);
+    cy.get('ktb-stage-details h2').should('contain.text', stage);
+  });
+
+  it('filter should be set when navigating to /environment/stage?filterType=filter', () => {
+    environmentPage.visit(project, 'staging', 'evaluation');
+    cy.byTestId(`ktb-stage-details-evaluation-button`).should('not.be.disabled');
```
Kirdock, line 53: Could you move that into the page object?
Something like
```ts
public assertStageDetailsFilterEnabled(filterType: 'problem' | 'evaluation' | 'approval', enabled: boolean): this {
    cy.byTestId(`ktb-stag-details-${filterType}-button`).should(enabled ? 'be.enabled' : 'be.disabled');
    return this;
}

File: bridge/client/app/_views/ktb-environment-view/ktb-stage-overview/ktb-stage-overview.component.html [hunk 13]
```diff
@@ -1,29 +1,30 @@
...
-      <button dt-button class="mb-3" (click)="isTriggerSequenceOpen = true" uitestid="keptn-trigger-button-open">
-        <dt-icon name="flash"></dt-icon>Trigger a new sequence
-      </button>
-    </div>
-    <div class="mb-3 pr-1" *ngIf="isTriggerSequenceOpen">
+    <div *ngIf="isTriggerSequenceOpen; else triggerSequenceClosed">
       <ktb-trigger-sequence
         [projectName]="project.projectName"
-        [stage]="selectedStage?.stageName"
+        [stage]="selectedStageName$ | async"
```
Kirdock, line 16: You know that this observable is subscribed two times in the view right? But it might not be a problem when looking at the pipe (because it is executed twice then).

Kirdock (changes requested):
- [x] The filter for (evaluation, remediation and open approval) can actually be set through two ways.
image
(2) does not add/remove a query parameter, only (1).
- [ ] Currently, if we have a filter type applied (e.g. "evaluation") and change the service filter so that it does not have any services for that filter type anymore, we reset the selection. This is not considered by the query parameter. The query parameter is still here.

bridge/client/app/_views/ktb-environment-view/ktb-stage-overview/ktb-stage-overview.component.ts
```diff
@@ -17,65 +17,79 @@ import { ServiceFilterType } from '../ktb-stage-details/ktb-stage-details.compon
...
   private globalFilter: { [projectName: string]: { services: string[] } } = {};
   private unsubscribe$: Subject<void> = new Subject<void>();
 
+  public project$ = this.route.params.pipe(
+    map((params) => params.projectName),
+    filter((projectName): projectName is string => !!projectName),
+    distinctUntilChanged(),
+    switchMap((projectName) => this.dataService.getProject(projectName)),
+    filter((project): project is Project => !!project),
+    tap((project) => {
```
heinzburgstaller, line 34: Be aware `tap()` can have side effects: If more than one is subscribing to this observable, the tap is executed multiple times

bridge/client/app/_views/ktb-environment-view/ktb-stage-overview/ktb-stage-overview.component.ts
```diff
@@ -17,65 +17,79 @@ import { ServiceFilterType } from '../ktb-stage-details/ktb-stage-details.compon
...
+    this.isTriggerSequenceOpen = this.dataService.isTriggerSequenceOpen;
+    this.dataService.isTriggerSequenceOpen = false;
   }
 
-  private setFilter(projectChanged: boolean): void {
+  ngAfterContentInit(): void {
+    combineLatest([this.selectedStageName$, this.paramFilterType$, this.project$])
+      .pipe(takeUntil(this.unsubscribe$))
+      .subscribe(([stageName, filterType, project]) => {
+        const stage = project.stages.find((s) => s.stageName === stageName);
```
heinzburgstaller, line 65: Do
```ts
this.setFilter(project, true);
```
here instead of the `tap` above

bridge/client/app/_views/ktb-environment-view/ktb-stage-details/ktb-stage-details.component.ts
```diff
@@ -77,6 +69,11 @@ export class KtbStageDetailsComponent implements OnInit, OnDestroy {
   selectFilterEvent($event: DtToggleButtonChange<any>): void {
     if ($event.isUserInput) {
       this.filterEventType = $event.source.selected ? $event.value : null;
+
+      // Add filterType query parameter
+      this.router.navigate([], {
+        queryParams: { filterType: $event.source.selected ? $event.value : null },
```
heinzburgstaller, line 75: Better:
```ts
   queryParams: { filterType: this.filterEventType},
```
so you don't have code duplication

bridge/client/app/_views/ktb-environment-view/ktb-stage-overview/ktb-stage-overview.component.ts
```diff
@@ -63,6 +60,7 @@ export class KtbStageOverviewComponent implements AfterContentInit, OnDestroy {
       .pipe(takeUntil(this.unsubscribe$))
       .subscribe(([stageName, filterType, project]) => {
         const stage = project.stages.find((s) => s.stageName === stageName);
+        this.setFilter(project, true);
```
Kirdock, line 63: Actually, this should be called on project change and should not depend on the other observables. Currently, the service filter is not set when initially loaded. With this change, `setFilter` is also called if `selectedStageName$` or `paramFilterType$` changes

Kirdock (changes requested):

bridge/client/app/_views/ktb-environment-view/ktb-stage-overview/ktb-stage-overview.component.ts
TannerGilbert: Yes, that's true. Should I change it back to using `tap` on the `project$` observable or is there a better solution?

Kirdock: We could split it into two observables. We have two subscriptions for `project$` I guess. We have some approaches here:
1. The view (HTML-part) that subscribes to it shall contain the `tap` and the one in `combineLatest` shall not.
2. The view and the `combineLatest` takes the `project$` without the tap and we introduce another one, where we subscribe directly to it.

@heinzburgstaller what would you say?

heinzburgstaller: I would manually subscribe to `project$` in the component and only call `setFilter` there. Therefore you have to manually unsubscribe, but it looks cleaner.

Kirdock (approved):

heinzburgstaller (approved):

PR #7737, opened 2022-05-13
feat(bridge): Show automatic provisioning message

laneli:
If automatic provisioning is enabled (git upstream is not required) and the message is set via helm value it is shown in the form.

### DIFF
--- a/bridge/client/app/_components/ktb-project-settings/ktb-project-settings.component.ts
+++ b/bridge/client/app/_components/ktb-project-settings/ktb-project-settings.component.ts
@@ -69,7 +69,7 @@ export class KtbProjectSettingsComponent implements OnInit, OnDestroy, PendingCh
     filter((keptnInfo: KeptnInfo | undefined): keptnInfo is KeptnInfo => !!keptnInfo),
     map((keptnInfo: KeptnInfo) => ({
       gitUpstreamRequired: !keptnInfo.metadata.automaticprovisioning,
-      automaticProvisioningMessage: keptnInfo.bridgeInfo.automaticProvisioningMsg?.trim(),
+      automaticProvisioningMessage: keptnInfo.bridgeInfo.automaticProvisioningMsg,
     })),
     startWith({ gitUpstreamRequired: undefined, automaticProvisioningMessage: undefined })
   );
--- a/bridge/client/app/_components/ktb-project-settings/ktb-project-settings.component.ts
+++ b/bridge/client/app/_components/ktb-project-settings/ktb-project-settings.component.ts
@@ -25,6 +25,7 @@ type DialogState = null | 'unsaved';
 
 interface ProjectSettingsState {
   gitUpstreamRequired: boolean | undefined;
+  automaticProvisioningMessage: string | undefined;
 }
 
 @Component({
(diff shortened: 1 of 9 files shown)


installer/manifests/keptn/charts/control-plane/values.yaml
```diff
@@ -130,6 +130,7 @@ bridge:
     userIdentifier: ""
     mongoConnectionString: ""
   nodeSelector: {}
+  automaticProvisioningMsg: ""
```
mowies, line 133: maybe it would make sense to put this value next to `automaticProvisioningURL` under `features` for consistency?

installer/manifests/keptn/charts/control-plane/templates/core.yaml
```diff
@@ -201,6 +201,8 @@ spec:
               value: {{ .Values.mongo.auth.bridgeAuthDatabase | default "openid" }}
             - name: CONFIG_DIR
               value: "/config"
+            - name: AUTOMATIC_PROVISIONING_MSG
+              value: {{.Values.bridge.automaticProvisioningMsg}}
```
mowies, line 205:
```suggestion
              value: {{. Values.bridge.automaticProvisioningMsg }}
```

bridge/client/app/_components/ktb-project-settings/ktb-project-settings.component.ts
```diff
@@ -66,8 +67,11 @@ export class KtbProjectSettingsComponent implements OnInit, OnDestroy, PendingCh
 
   readonly state$: Observable<ProjectSettingsState> = this.dataService.keptnInfo.pipe(
     filter((keptnInfo: KeptnInfo | undefined): keptnInfo is KeptnInfo => !!keptnInfo),
-    map((keptnInfo: KeptnInfo) => ({ gitUpstreamRequired: !keptnInfo.metadata.automaticprovisioning })),
-    startWith({ gitUpstreamRequired: undefined })
+    map((keptnInfo: KeptnInfo) => ({
+      gitUpstreamRequired: !keptnInfo.metadata.automaticprovisioning,
+      automaticProvisioningMessage: keptnInfo.bridgeInfo.automaticProvisioningMsg?.trim(),
```
Kirdock, line 72: Should we move the `trim()` to the bridge server? Then it's one time on bridge server startup instead of every time the data is loaded in the frontend
```ts
const automaticProvisioningMsg = process.env.AUTOMATIC_PROVISIONING_MSG?.trim();
```

Kirdock (changes requested):

installer/manifests/keptn/charts/control-plane/templates/core.yaml
ermin-muratovic: @mowies I think you meant:

```suggestion
              value: {{ .Values.bridge.automaticProvisioningMsg }}
```

mowies: woops, yes exactly 😄

installer/manifests/keptn/charts/control-plane/values.yaml
thisthat: I think we should follow up with a separate enhancement ticket to merge under the same object `automaticProvisioning` the `serivceURL` and the `message`. Similarly to:
```

features:
  automaticProvisioning:
    serviceURL: ""
    message: ""

```

PR #7744, opened 2022-05-16
feat(cp-connector): Ensure mandatory CloudEvent attributes are set

bacherfl:
In addition to the event time, also an event ID and the CloudEvents spec version need to be set

### DIFF
--- a/cp-connector/pkg/nats/nats.go
+++ b/cp-connector/pkg/nats/nats.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/google/uuid"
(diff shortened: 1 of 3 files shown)


cp-connector/pkg/nats/nats.go
```diff
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	cloudevents "github.com/cloudevents/sdk-go/v2"
```
warber, line 7: i think we can completely avoid this dependency by creating just our own constant:
```
const CloudEventsVersionV1 = "1.0"
```

warber (changes requested):

cp-connector/pkg/nats/nats.go
bacherfl: agree, will replace that

warber (approved):

PR #7766, opened 2022-05-17
fix(bridge): Fixed triggering of validation on token reset

### LINKED ISSUE #7725: After setting/updating the GIT token for a project a hint appears that the token field cannot be empty
Just a small thing: When updating the git token for a project, it is confirmed that the change has been made, but at the same time a popup appears that is saying that the Token field must not be empty.

image

Kirdock:
Fixes #7725


ermin-muratovic (approved):
lgtm

renepanzar (approved):
Whaaat!? Anyway, LGTM :)

PR #7777, opened 2022-05-17
fix(mongodb-datastore): Change name of integration to name of service instead of pod name

bacherfl:
This ensures that multiple instances of the mongodb-datastore are registered with the same name, and therefore are part of the same nats queue group


RealAnna (approved):
LGTM

PR #7780, opened 2022-05-17
test(bridge): Removed dev dependency and use cypress/schematics instead

Kirdock:
Removed dev dependency `start-server-and-test`.


ermin-muratovic (approved):
lgtm

PR #7787, opened 2022-05-18
fix: Make sure that all events are being processed before shutting down lighthouse/approval service

bacherfl:
In the previous implementation of the `getGracefulContext()` function, we started a goroutine that waits for the waitGroup to be finished, but we did not wait for that goroutine - which lead to the pod being shut down prematurely even though there might still be some events being processed.

Also, during testing it seemed like addiing a small delay between cancelling the context (i.e. unsubscribing from nats) and calling `wg.Wait()` seemed to solve the problem where it was still the case that events were not fully processed before shutting down. My theory for that is that if `wg.Wait()` being called before an event handler has the chance to increase the wait group might lead to this behavior.

### DIFF
--- a/approval-service/main.go
+++ b/approval-service/main.go
@@ -79,11 +79,11 @@ func main() {
 	}
 
 	// this segment will be reached once the context has been cancelled - i.e. due to receiving the SIGTERM signal
-	logger.Info("Waiting for event handlers to finish")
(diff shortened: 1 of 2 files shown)


odubajDT (dismissed):
lgtm

odubajDT (approved):

PR #7793, opened 2022-05-18
fix(shipyard-controller): Dispatch new sequence directly only if no older sequence is waiting

### LINKED ISSUE #7767: Shipyard-controller skipped a sequence
**Environment**

* Keptn Version (`keptn version`): `0.15.0`

**Affected Component**

* [x] Shipyard-controller

**Describe the bug**

Keptn skipped a sequence even though it has an older timestamp: 

image

**To Reproduce**
* Trigger a delivery sequence multiple times. I triggered it for `carts-db` as well as `carts`:
```
keptn trigger delivery --project=sockshop --service=carts --image=docker.io/keptnexamples/carts:0.13.3
keptn trigger delivery --project=sockshop --service=carts-db --image=docker.io/mongo:4.2.2 --sequence=delivery-direct
keptn trigger delivery --project=sockshop --service=carts --image=docker.io/keptnexamples/carts:0.13.2
keptn trigger delivery --project=sockshop --service=carts --image=docker.io/keptnexamples/carts:0.13.1
```

**Expected behavior**
* The sequences are executed in the same order they were triggered

**Current behavior**
* For some reason, the sequence for `carts:0.13.1` was overtaken by `0.13.1`

odubajDT:

## This PR

- dispaches new sequences directly only if no older sequence is executed (in state triggered or started) in the same stage, service and project
- refactoring the Add() method
- adding unit tests

### Related Issues

Fixes #7767

### Integration tests
https://github.com/keptn/keptn/actions/runs/2369401694

### DIFF
--- a/shipyard-controller/db/mongodb_sequence_execution_repo.go
+++ b/shipyard-controller/db/mongodb_sequence_execution_repo.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"time"
+
 	apimodels "github.com/keptn/go-utils/pkg/api/models"
 	"github.com/keptn/keptn/shipyard-controller/db/models/sequence_execution"
 	v02 "github.com/keptn/keptn/shipyard-controller/db/models/sequence_execution/v1"
@@ -13,7 +15,6 @@ import (
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
-	"time"
 )
 
 const sequenceExecutionCollectionNameSuffix = "sequence-execution"
@@ -341,6 +342,8 @@ func (mdbrepo *MongoDBSequenceExecutionRepo) getCollection(collectionName string
 func (mdbrepo *MongoDBSequenceExecutionRepo) getSearchOptions(filter models.SequenceExecutionFilter) bson.M {
 	searchOptions := bson.M{}
 
+	log.Infof("filer je %+v", filter)
+
 	searchOptions = appendFilterAs(searchOptions, filter.Name, "sequence.name")
 	searchOptions = appendFilterAs(searchOptions, filter.Scope.TriggeredID, "scope.triggeredId")
 	searchOptions = appendFilterAs(searchOptions, filter.Scope.KeptnContext, "scope.keptnContext")
@@ -348,6 +351,11 @@ func (mdbrepo *MongoDBSequenceExecutionRepo) getSearchOptions(filter models.Sequ
 	searchOptions = appendFilterAs(searchOptions, filter.Scope.Stage, "scope.stage")
 	searchOptions = appendFilterAs(searchOptions, filter.Scope.Service, "scope.service")
 	searchOptions = appendFilterAs(searchOptions, filter.CurrentTriggeredID, "status.currentTask.triggeredID")
+	if !filter.TriggeredAt.IsZero() {
+		searchOptions["triggeredAt"] = bson.M{
+			"$lt": filter.TriggeredAt,
+		}
+	}
 
 	if filter.Status != nil && len(filter.Status) > 0 {
 		matchStates := []bson.M{}
--- a/shipyard-controller/db/mongodb_sequence_execution_repo.go
+++ b/shipyard-controller/db/mongodb_sequence_execution_repo.go
@@ -342,8 +342,6 @@ func (mdbrepo *MongoDBSequenceExecutionRepo) getCollection(collectionName string
 func (mdbrepo *MongoDBSequenceExecutionRepo) getSearchOptions(filter models.SequenceExecutionFilter) bson.M {
 	searchOptions := bson.M{}
 
-	log.Infof("filer je %+v", filter)
-
 	searchOptions = appendFilterAs(searchOptions, filter.Name, "sequence.name")
 	searchOptions = appendFilterAs(searchOptions, filter.Scope.TriggeredID, "scope.triggeredId")
 	searchOptions = appendFilterAs(searchOptions, filter.Scope.KeptnContext, "scope.keptnContext")
--- a/shipyard-controller/handler/sequencedispatcher.go
+++ b/shipyard-controller/handler/sequencedispatcher.go
@@ -61,8 +61,9 @@ func (sd *SequenceDispatcher) Add(queueItem models.QueueItem) error {
 	if sd.mode == common.SDModeRW {
 		//if there is only one shipyard we can both read and write,
 		//so we try to dispatch the sequence immediately if the sequence queue is empty
-		queueEmpty, err := sd.isQueueEmpty(queueItem)
-		if (err != nil && !errors.Is(err, db.ErrNoEventFound)) || !queueEmpty {
+		sequenceBlocked, err := sd.areSequencesInState(queueItem, []string{apimodels.SequenceTriggeredState, apimodels.SequenceStartedState})
+		if err != nil || sequenceBlocked {
+			log.Infof("!!!!!SHIt: sequence blocked")
 			return sd.addItemToQueue(queueItem)
 		}
 		if err := sd.dispatchSequence(queueItem); err != nil {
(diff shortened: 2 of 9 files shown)


shipyard-controller/handler/sequencedispatcher.go
```diff
@@ -57,30 +58,56 @@ func NewSequenceDispatcher(
...
+		return err2
+	}
+	return ErrSequenceBlockedWaiting
+}
+
+func (sd *SequenceDispatcher) isQueueEmpty(queueItem models.QueueItem) (bool, error) {
+	queuedSequences, err := sd.sequenceQueue.GetQueuedSequences()
+	if err != nil {
+		if !errors.Is(err, db.ErrNoEventFound) {
+			return false, sd.addItemToQueue(queueItem)
```
bacherfl, line 102: calling `addItemToQueue()` within the `isQueueEmpty` seems counterintuitive - can we move that outside?

shipyard-controller/handler/sequencedispatcher.go
```diff
@@ -57,30 +58,56 @@ func NewSequenceDispatcher(
...
-			} else if errors.Is(err, ErrSequenceBlockedWaiting) {
-				//if the sequence is currently blocked and should wait, insert it into the queue
-				if err2 := sd.add(queueItem); err2 != nil {
-					return err2
+		//so we try to dispatch the sequence immediately if the sequence queue is empty
+		queueEmpty, err := sd.isQueueEmpty(queueItem)
+		if err != nil {
+			return err
+		}
+		if queueEmpty {
```
bacherfl, line 68: this can be made a little easier to follow by restructuring it like so:

```
if !queueEmpty {
  return sd.addItemToQueue(queueItem)
}

if err := sd.dispatchSequence(queueItem); err != nil {
 if errors.Is(err, ErrSequenceBlocked) {
 	//if the sequence is currently blocked, insert it into the queue
 	return sd.add(queueItem)
 } else if errors.Is(err, ErrSequenceBlockedWaiting) {
 	//if the sequence is currently blocked and should wait, insert it into the queue
 	return sd.addItemToQueue(queueItem)
 } else {
 	return err
 }
}
```

this way we can reduce the nesting by one layer

shipyard-controller/handler/sequencedispatcher.go
```diff
@@ -80,7 +81,28 @@ func (sd *SequenceDispatcher) Add(queueItem models.QueueItem) error {
...
 
+func (sd *SequenceDispatcher) addItemToQueue(queueItem models.QueueItem) error {
+	if err2 := sd.add(queueItem); err2 != nil {
+		return err2
+	}
+	return ErrSequenceBlockedWaiting
+}
+
+func (sd *SequenceDispatcher) isQueueEmpty(queueItem models.QueueItem) (bool, error) {
+	queuedSequences, err := sd.sequenceQueue.GetQueuedSequences()
```
bacherfl, line 94: would we achieve the same functionality by checking if there are SequenceExecutionStates with status `triggered` (for the same project/stage/service)? `sequenceQueue.GetQueuedSequences()` will return the complete queue, and also this means that having an item form another project in the queue will affect if a sequence will be immediately dispatched or not

odubajDT: sure we can try this approach

odubajDT: applied

shipyard-controller/handler/sequencedispatcher.go
```diff
@@ -80,7 +81,13 @@ func (sd *SequenceDispatcher) Add(queueItem models.QueueItem) error {
 		//if there are multiple shipyard we should only write
 		return sd.add(queueItem)
 	}
+}
 
+func (sd *SequenceDispatcher) addItemToQueue(queueItem models.QueueItem) error {
+	if err2 := sd.add(queueItem); err2 != nil {
+		return err2
+	}
+	return ErrSequenceBlockedWaiting
```
bacherfl, line 90: returning `ErrSequenceBlockedWaiting` seems counterintuitive at this point, as adding an item to the queue usually is already a consequence of the sequence being blocked - can we move the return of the `ErrSequenceBlockedWaiting` to the caller of this method?

odubajDT: Sure, we can move it there, but that will mean one more `if err != nil { return err} return ErrSequenceBlockedWaiting` in each caller place, so than it makes no sense to have a function for this few lines and we can paste them on the caller place directly

shipyard-controller/handler/sequencedispatcher.go
```diff
@@ -147,6 +146,55 @@ func (sd *SequenceDispatcher) dispatchSequences() {
...
+				Service: queueItem.Scope.Service,
+			},
+		},
+		Status: []string{apimodels.SequenceStartedState},
+	})
+	if err != nil {
+		return true, err
+	}
+
+	if len(startedSequenceExecutions) > 0 {
```
bacherfl, line 165: can we also add a log here to print the keptnContext ID of the currently started sequence? Retrieving the ID of the currently running sequence has often been complicated, especially in support cases where we don't have access to the DB, so this would be really helpful here :)

odubajDT: sure no problem :)

shipyard-controller/handler/sequencedispatcher_test.go
```diff
@@ -357,6 +358,146 @@ func TestSequenceDispatcher_AddError(t *testing.T) {
 
 }
 
+func TestSequenceDispatcher_QueueIsNotEmpty(t *testing.T) {
```
bacherfl, line 361: Can we also add a test case where there are currently no sequences with state = `started`, but other queued sequences that preceed the current one?

odubajDT: such a test is there, pushed it few minutes earlier, we can have a call to explain if you want :)

shipyard-controller/db/mongodb_sequence_execution_repo.go
```diff
@@ -348,6 +349,11 @@ func (mdbrepo *MongoDBSequenceExecutionRepo) getSearchOptions(filter models.Sequ
 	searchOptions = appendFilterAs(searchOptions, filter.Scope.Stage, "scope.stage")
 	searchOptions = appendFilterAs(searchOptions, filter.Scope.Service, "scope.service")
 	searchOptions = appendFilterAs(searchOptions, filter.CurrentTriggeredID, "status.currentTask.triggeredID")
+	if !filter.TriggeredAt.IsZero() {
```
bacherfl, line 352: should we rather add a `From`/`To` field to the `SequenceExecutionFilter`? It might not be immediately clear that a `$lt` operator is applied if `TriggeredAt` is provided in the filter

odubajDT: we actually want to filter always sequences which are less than the actual timestamp, I don't think `from`/`to` makes much sense here. If the `TriggeredAt` is not set, the behaviour of the filter stays unchanged and we do not need to deal with any incompatibilities. Also `from`/`to` will introduce extra logic, parameters and also slows the request when applied every time

shipyard-controller/handler/sequencedispatcher.go
```diff
@@ -147,6 +146,61 @@ func (sd *SequenceDispatcher) dispatchSequences() {
...
+			return true, nil
+		}
+	}
+
+	if len(triggeredSequenceExecutions) > 1 {
+		log.Infof("Sequence with KeptnContext %s is blocked due to triggered sequences in stage %s", queueItem.Scope.KeptnContext, queueItem.Scope.Stage)
+		return true, nil
+	}
+
+	log.Infof("Sequence with KeptnContext %s is not blocked by any other sequence in stage %s", queueItem.Scope.KeptnContext, queueItem.Scope.Stage)
```
bacherfl, line 200: i think we can remove this log line, as this will otherwise make the logs very verbose

odubajDT: adapted

bacherfl (approved):

PR #7794, opened 2022-05-18
fix(bridge): Fix logout not being visible if metadata is not returned

### LINKED ISSUE #7781: Bridge: Fetching metadata information should not prevent the user from logging out
The bridge aggregates some metadata (bridge backend and API service) at the beginning. If the call to the API service fails, the whole request fails, and then the user is not able to log himself out.

In views where the metadata is required, an error is shown (e.g. create/edit project).
The user menu should show the keptn info and the logout even if the call to the API service fails

image

Kirdock:
Fixes #7781


ermin-muratovic (approved):
lgtm

ermin-muratovic (approved):
lgtm

PR #7823, opened 2022-05-23
chore: Update references to cp-common

RealAnna:
### Follow-up Tasks
update SDK ref in webhook and remediation and distributor ref


odubajDT:
We should probably completely remove this file https://github.com/keptn/keptn/blob/master/distributor/pkg/api/internal.go and use cp-common instead

odubajDT (approved):

PR #7775, opened 2022-05-17
fix(shipyard-controller): Allow parallel sequence execution if the service is different

RealAnna:
## This PR

-  dispatcher allows parallel sequences on the same project and stage if at least the service is different.
-  adds unit test
-  adds integration test

### Related Issues

Fixes #7753 
https://github.com/keptn/keptn/actions/runs/2344206480

PR #7825, opened 2022-05-23
fix(shipyard-controller): Allow parallel sequence execution if the service is different

RealAnna:
backport of  #7775


odubajDT (approved):
lgtm, but let's wait for the unit tests :)

PR #7826, opened 2022-05-23
feat: Enable Resource-Service by default

thisthat:

## This PR

makes the `resource-service` enabled by default.

As a follow-up, we should work on some clean-up tasks for the K8s service name and pipeline: #7827


RealAnna (commented):
do we also enable directory by default ?

mowies:
is this a breaking change?

RealAnna (approved):
LGTM

Issue #7662, opened 2022-05-05
Improve Zero Downtime tests


RealAnna:
## Action Item 

Improve the existing zero downtime test pipeline with the following refinements:

- [x] Add the following parameters (functionality) to be configurable (via env var), e.g.:
   - apiProbeInterval  
   - sequencesInterval
   - upgradeCyclesNumber
   - [GOMAXPROCS](https://pkg.go.dev/runtime#GOMAXPROCS) (tight to the max amount of tests we run in parallel)
- [x] Run in parallel the cypress tests (via ENV VAR enable) [e.g. CMD](https://github.com/keptn/keptn/blob/master/bridge/package.json#L21)

## DoD

- [x] Provide good defaults for the values
- [x] Document the ENV VARs that can change the configuration
- [x] Add a job scheduled for the workflow

PR #7829, opened 2022-05-23
ci: Added cypress tests to Zero Downtime pipeline

RealAnna:
## This PR

- adds cypress test to ZD pipeline
- example run : https://github.com/keptn/keptn/runs/6556725066?check_suite_focus=true
- tests can be enabled at pipeline selecting it as in here
image

- test are run in parallel to zd and reported under **format cypress logs output** e.g.
  image

- artifact of logs is generated as well
- scheduled a run for every Monday 

### Related Issues

#7662

### DIFF
--- a/.github/workflows/zero-downtime-tests.yml
+++ b/.github/workflows/zero-downtime-tests.yml
@@ -354,7 +354,7 @@ jobs:
           set -euo pipefail 
           
           cd ../test/zero-downtime && go test -json -v -run Test_ZeroDowntime -v -timeout 90m 2>&1 | tee /tmp/gotest.log &
-          if [[ "${{ github.event.inputs.testUI }}" == true ]]; then
+          if [[ "${{ github.event.inputs.testUI }}" == 'true' ]]; then
             while [[ $(jobs -pr) ]]; do
               npm run test:ui >> /tmp/uitest.log 2>&1 
             done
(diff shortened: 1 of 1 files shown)


.github/workflows/zero-downtime-tests.yml
```diff
@@ -325,11 +334,15 @@ jobs:
           # Optional: pass GITHUB_TOKEN to avoid rate limiting.
           token: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Install npm
+        id: install_npm
```
mowies, line 338: please rename this step to make it a little bit clearer. (you can also remove the ID if it's not used somewhere else in the pipeline)
```suggestion
      - name: Install Node dependencies
```

.github/workflows/zero-downtime-tests.yml
```diff
@@ -426,14 +450,23 @@ jobs:
           path: support-archive/*.zip
 
         # Upload the original go test log as an artifact for later review.
-      - name: Upload test log
+      - name: Upload go test log
         uses: actions/upload-artifact@v2
         if: always()
         with:
           name: test-log
```
mowies, line 457: maybe also rename this to `go-test-log`?

.github/workflows/zero-downtime-tests.yml
```diff
@@ -1,5 +1,9 @@
 name: ZeroDowntime Tests
 on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron: '0 1 * * 1' # run zd tests at 1 AM (UTC), monday(1)
```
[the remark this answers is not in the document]
RealAnna: then I will do it in a separate PR since I need to figure out how to select the second image to upgrade to :(

PR #7832, opened 2022-05-23
fix(webhook-service): Added missing webhook-config version check

odubajDT:

### Notes
When retrieving resources from git repository, the first step is decoding the content of webhook.yaml to a struct. Here we have two versions (alpha and beta) which are mostly normalized the same way except beta has a few additional checks. These checks were executed only if the version of the webhook-config was v1beta1, if it was different, additional checks were not executed. So actually the in the integration test, we had a valid webhook-config, so if the extra pre-check is or is not executed, it does not matter, the webhook is executed normally (and it's working), just the extra pre-checks were not executed. 

### Integration tests
https://github.com/keptn/keptn/actions/runs/2372897926

### DIFF
--- a/test/go-tests/test_customendpoints.go
+++ b/test/go-tests/test_customendpoints.go
@@ -3,17 +3,18 @@ package go_tests
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
 	"github.com/imroc/req"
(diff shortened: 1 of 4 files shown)


RealAnna (dismissed):
LGTM

RealAnna (approved):
LGTM

mowies (approved):

PR #7833, opened 2022-05-23
fix(bridge): Fixed D3 heatmap issues

### LINKED ISSUE #7824: Bridge: Improve D3 heatmap
- The D3 heatmap uses the wrong ordering internally: if the first just has a score then the score is placed on the bottom instead on the top.
- Input selected data point accepts string (identifier) instead of a data point.
- Zoom level should not effect width
- Use identifiers for ui-test-id for a data point

Kirdock:
Fixes #7824 


ermin-muratovic (approved):
lgtm

PR #7837, opened 2022-05-23
fix(installer): Redirect output of preStop hook command to /dev/null

bacherfl:
On GKE 1.22 the preStop hook was failing with the error message `wget: can't open pre-stop: Read only file system`.
Redirecting the output to /dev/null should solve this


RealAnna (approved):
LGTM

mowies (approved):
This will actually redirect the output to stdout and not /dev/null but still LGTM

PR #7838, opened 2022-05-23
fix(shipyard-controller): Set the sequence execution state back to `started` when approval task has been finished

### LINKED ISSUE #7828: Sequences are not blocked by running sequences with completed approval task
If a sequence that is currently in execution contains an approval task that is already completed it should again block other sequences from being started again. However, this is currently not the case, as the sequence state stays at `waitingForApproval` after an `approval` task has been finished. This needs to be adapted here:

https://github.com/keptn/keptn/blob/master/shipyard-controller/handler/shipyardcontroller.go#L801-L804

bacherfl:
Closes #7828


odubajDT (approved):
lgtm

PR #7840, opened 2022-05-23
fix(shipyard-controller): Set the sequence execution state back to `started` when approval task has been finished 

bacherfl:
Backport of #7838


odubajDT (approved):
lgtm

PR #7860, opened 2022-05-25
test: Added envconfig for setting up zero downtime test parameters

RealAnna:
### Related Issues

#7662

### How to test

Run locally without env parameters setup and default values will be used

Pipeline run here : https://github.com/keptn/keptn/runs/6722144468?check_suite_focus=true


test/zero-downtime/testsuite_zerodowntime_test.go
```diff
@@ -28,13 +27,14 @@ spec:
...
-const apiProbeInterval = 5 * time.Second
-const sequencesInterval = 15 * time.Second
-
 type ZeroDowntimeEnv struct {
-	quit         chan struct{}
-	NrOfUpgrades int
-	Wg           *sync.WaitGroup
+	quit              chan struct{}
+	NrOfUpgrades      int           `envconfig:"NUMBER_OF_UPGRADES" default:"2"`
+	EnvInstallVersion string        `envconfig:"INSTALL_HELM_CHART" default:"https://github.com/keptn/helm-charts-dev/raw/69eea439a26a99ecc163e296860dbb5d43e41600/packages/keptn-0.15.1-dev.tgz"`
```
thisthat, line 33: I don't know if a default here is a good approach. I would force the env var to exist

RealAnna: I thought a default value would make it easier to play with it locally, but I can add it to the code as a comment and remove the default...

installer/manifests/keptn/charts/control-plane/values.yaml
```diff
@@ -83,7 +83,7 @@ remediationService:
     tag: ""
```
thisthat, line 83: I don't think this file should be part of this PR

odubajDT (dismissed):

odubajDT:
looks good, but shouldn't we introduce a simple unit tests for this ? :)

odubajDT (approved):

PR #7812, opened 2022-05-20
feat(bridge): Removes projects polling on dashboard #7796

renepanzar:
## This PR
- Removes polling for projects
- Removes dependency on POLLING_INTERVAL_MILLIS
- Adapts spec

### Related Issues
Resolves #7796 

### Notes
No useful screenshots available.

### How to test

1. Go to dashboard and open DevTools/Networking
2. No additional calls to `/api/controlPlane/v1/project?disableUpstreamSync=true&pageSize=50`

PR #7863, opened 2022-05-25
test(bridge): Fix failing auth.spec UI test

Kirdock:
https://github.com/keptn/keptn/pull/7812 removed the polling in the dashboard and https://github.com/keptn/keptn/pull/7794 added two checks here. Now without the polling, there is only one check for the response/error.


ermin-muratovic (approved):
lgtm

PR #7870, opened 2022-05-25
chore(bridge): Remove sequence metadata polling

### LINKED ISSUE #7800: Bridge :: Sequence view - Split endpoint for filters and current deployments and remove polling
For the sequences view there is a sequences metadata endpoint that is polled on a regular basis. This endpoint contains filters (services, stages) as well as the current deployed versions for each stage and service that are only used for approvals.
The polling for this endpoint should be removed. In this case it would make sense to split the endpoint and fetch the data for filters once when the user navigates to the page (as sequences and stages should not change that often) and the latest deployment only when there is an approval task in the trace. It could also be an alternative to only fetch the latest deployment only when the user expands the tile for the approval.

## Acceptance criteria
- [x] Polling for sequences metadata is removed
- [ ] BFF API endpoints for **filters** and **current deployment** for service+stage have been introduced
- [x] Filters are fetched on page navigation
- [x] Current deployment is fetched only when an approval task is expanded
- [x] `/sequence/metadata` is removed from BFF

## DoD
- [ ] Acceptance criteria implemented
- [ ] Unit tests adapted
- [ ] Unit tests in BFF have been created for new endpoints
- [ ] BFF API docs are adapted

Kirdock:
Closes #7800 

Instead of adding a new endpoint in the BFF for fetching the deployed image for a service in a specific stage, I used the existing service endpoint to fetch the service. Correct me if I should introduce a new endpoint that only returns the deployed image instead.


bridge/client/app/_components/ktb-sequence-tasks-list/ktb-sequence-tasks-list.component.ts
```diff
@@ -30,52 +25,32 @@ export class KtbSequenceTasksListComponent implements OnInit, OnDestroy {
...
   }
-
   set focusedEventId(value: string | undefined) {
     if (this._focusedEventId !== value) {
       this._focusedEventId = value;
     }
+    if (!value) {
+      this.focusLastSequence();
+    }
+    this._changeDetectorRef.markForCheck();
```
renepanzar, line 43: Is this necessary? If the input changes (reference equality), then the change detection should trigger anyway.

bridge/client/app/_components/ktb-sequence-tasks-list/ktb-sequence-tasks-list.component.ts
```diff
@@ -30,52 +25,32 @@ export class KtbSequenceTasksListComponent implements OnInit, OnDestroy {
...
-  @Input()
-  get stage(): string | undefined {
-    return this._stage;
-  }
-
-  set stage(value: string | undefined) {
-    if (this._stage !== value) {
-      this._stage = value;
-      this.focusLastSequence();
+      this._changeDetectorRef.markForCheck();
```
renepanzar, line 28: Is this necessary? If the input changes (reference equality), then the change detection should trigger anyway.

bridge/client/app/_components/ktb-sequence-tasks-list/ktb-sequence-tasks-list.component.ts
```diff
@@ -30,52 +25,32 @@ export class KtbSequenceTasksListComponent implements OnInit, OnDestroy {
...
     }
   }
 
   @Input()
```
renepanzar, line 32: You might need to move the `@Input()` to the setter for the change detection `OnPush` to work correctly.

bridge/client/app/_components/ktb-sequence-tasks-list/ktb-sequence-tasks-list.component.ts
```diff
@@ -1,24 +1,19 @@
...
   @HostBinding('class') cls = 'ktb-sequence-tasks-list';
-  public _tasks: Trace[] = [];
-  public _stage?: string;
-  public _focusedEventId?: string;
-  private readonly unsubscribe$ = new Subject<void>();
+  private _tasks: Trace[] = [];
+  private _focusedEventId?: string;
   private currentScrollElement?: HTMLDivElement;
 
   @Input()
```
renepanzar, line 19: You might need to move the `@Input()` to the setter for the change detection `OnPush` to work correctly.

bridge/client/app/_components/ktb-sequence-tasks-list/ktb-sequence-tasks-list.component.ts
Kirdock: In this component you are right, it is not needed. But it's weird because in another component (`ktb-heatmap-tooltip.component.ts`) this was necessary.

renepanzar: Kudos for removing `ChangeDetectorRef` :+1:

renepanzar (approved):
LGTM!

PR #7892, opened 2022-05-30
fix(lighthouse-service): Ensure sloFileContent property is always a base64 encoded string

### LINKED ISSUE #7891: Lighthouse service does not base64 encode `sloFileContent` in error cases
In some error cases, e.g. when no target score for an evaluation is defied in an SLO file, the `data.evaluation.sliFileContent` property is not base64 encoded. The cause for that lies in the following line, where the sloFileContent is directly set as a string without encoding it first:

https://github.com/keptn/keptn/blob/17d42457703d805c6531e90abaa621e1dd0352a1/lighthouse-service/event_handler/common.go#L198

bacherfl:
Closes #7891

### DIFF
--- a/lighthouse-service/event_handler/common.go
+++ b/lighthouse-service/event_handler/common.go
@@ -2,6 +2,7 @@ package event_handler
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


RealAnna (approved):
LGTM

PR #7894, opened 2022-05-30
chore(installer): Adapted default values for preStop hook times and resource-service resource limits

bacherfl:
This PR adapts the default values for the preStop hook times of the following services:

resource-service
mongodb-datastore
secret-service

Additionally, the memory limit for the resource-service has been increased to 256Mi

### DIFF
--- a/installer/manifests/keptn/charts/control-plane/values.yaml
+++ b/installer/manifests/keptn/charts/control-plane/values.yaml
@@ -74,7 +74,7 @@ apiGatewayNginx:
     tag: 1.21.6-alpine
   nodeSelector: {}
   gracePeriod: 60
-  preStopHookTime: 5
+  preStopHookTime: 20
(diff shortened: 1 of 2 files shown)


installer/manifests/keptn/charts/control-plane/values.yaml
```diff
@@ -203,7 +203,7 @@ configurationService:
     GIT_KEPTN_EMAIL: "keptn@keptn.sh"
   nodeSelector: {}
   gracePeriod: 60
-  preStopHookTime: 5
+  preStopHookTime: 20
```
odubajDT, line 206: why configuration-service ?

bacherfl: if for some reason it is still used instead of the resource-service it will very likely also need more time

odubajDT (approved):

PR #7901, opened 2022-05-30
fix(shipyard-controller): Make sure result and status are set if sequence is timed out

### LINKED ISSUE #7898: No sequence result set if shipyard-controller times out the sequence
Keptn Version (`keptn version`): 0.15.1-dev

**Affected Component**

* [x] Shipyard-controller

**Describe the bug**

At sequence timeout the returned event from shipyard-controller does not contain a result

**To Reproduce**

Run a delivery without helm service installed or with a custom task not being executed, wait for a time out.

**Expected behavior**

Result failed 

**Current behavior**

image 
image

bacherfl:
Closes #7898 

This PR also adjusts some of the intervals at which assertions are checked in the Nats connection unit tests since these tests had a few flaky runs recently


RealAnna (approved):
LGTM

PR #7902, opened 2022-05-30
fix(cli): Remove unnecessary `--sequence` flag from `keptn trigger sequence`

### LINKED ISSUE #7884: keptn trigger sequence invalid helper text / unnecessary --sequence flag
@afzal442 @odubajDT FYI

The `keptn trigger sequence` help text is invalid. It leads me to believe that `keptn trigger sequence delivery --project fulltour --service helloservice --stage qa` is valid whereas in fact the command needs to be `keptn trigger sequence delivery --sequence delivery --project fulltour --service helloservice --stage qa`

Is the flag even required if we're passing the sequence name after `keptn trigger sequence foo` it seems like pointless duplication to have that flag.

Affects keptn 0.15.0 and higher.

odubajDT:

## This PR

- removed --sequence flag from command

### Related Issues

Fixes #7884 

#### Before
```Usage:
  keptn trigger sequence [flags]

Aliases:
  sequence, sequence

Examples:
keptn trigger sequence <sequence-name> --project=<project> --service=<service> --stage=<stage>

Flags:
      --project string    The project containing the service for which the new artifact will be triggered
      --sequence string   The sequence for which the new artifact will be triggered
      --service string    The service for which the new artifact will be triggered
      --stage string      The stage in which the new artifact will be triggered
```

#### After
```Usage:
  keptn trigger sequence [flags]

Aliases:
  sequence, sequence

Examples:
keptn trigger sequence <sequence-name> --project=<project> --service=<service> --stage=<stage>

Flags:
      --project string   The project containing the service for which the new artifact will be triggered
      --service string   The service for which the new artifact will be triggered
      --stage string     The stage in which the new artifact will be triggered
```

### DIFF
--- a/cli/cmd/trigger_sequence.go
+++ b/cli/cmd/trigger_sequence.go
@@ -63,6 +63,9 @@ The name of the sequence has to be provided as an argument to the command. The s
 	Example:      `keptn trigger sequence <sequence-name> --project=<project> --service=<service> --stage=<stage>`,
 	SilenceUsage: true,
 	Args: func(cmd *cobra.Command, args []string) error {
+		if len(args) != 1 {
+			return errors.New("required argument sequence-name not set")
+		}
 		return nil
 	},
 	RunE: func(cmd *cobra.Command, args []string) error {
--- a/cli/cmd/trigger_sequence.go
+++ b/cli/cmd/trigger_sequence.go
@@ -22,7 +22,6 @@ type sequenceStruct struct {
 	Project  *string `json:"project"`
 	Service  *string `json:"service"`
 	Stage    *string `json:"stage"`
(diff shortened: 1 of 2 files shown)


thisthat (changes requested):
~There could be multiple delivery sequences in a shipyard file, hence the flag should not be removed instead, it should be made optional.~

RealAnna (changes requested):
please fix integration tests

RealAnna (approved):

odubajDT:
Hi @agardnerIT, no it was not, we do consider this problem as low priority, so no backports were made. Do you have a specific reason why it should be backported to 0.15.x?

Thanks.

PR #7919, opened 2022-05-31
fix(mongodb-datastore): Return [] instead of nil from get methods

RealAnna:
### Related Issues

image

https://github.com/keptn/keptn/runs/6669629188?check_suite_focus=true


odubajDT:
Looks good, but shouldn't we introduce a simple unit tests for this ? :)

odubajDT (approved):

PR #7939, opened 2022-06-01
fix: Fixed wrong paths for `APIHandler` when used with internally via `InternalAPISet`

warber:
The `APIHandler` utility in `go-utils` is delegating requests to either the api of the shipyard-controller or the api of the api-service.
Hence it cannot be used with `InternalAPISet` because there we use a fixed address (e.g. shipyard-controller:8080) to make the requests. This PR fixes the issue, by providing a custom `APIHandler` implementation that is internally just using two instances of `APIHandler` from `go-utils` each configured with another base URL.


cp-common/api/internal.go
```diff
@@ -218,3 +225,40 @@ func getClientTransport(rt http.RoundTripper) http.RoundTripper {
...
+func (i *InternalAPIHandler) DeleteProject(project models.Project) (*models.DeleteProjectResponse, *models.Error) {
+	return i.shippyApiHandler.DeleteProject(project)
+}
+
+func (i *InternalAPIHandler) CreateService(project string, service models.CreateService) (string, *models.Error) {
+	return i.shippyApiHandler.CreateService(project, service)
+}
+
+func (i *InternalAPIHandler) DeleteService(project string, service string) (*models.DeleteServiceResponse, *models.Error) {
+	return i.DeleteService(project, service)
```
bacherfl, line 259: => `i.shippyApiHandler.DeleteService(project, service)`

bacherfl (approved):

PR #7941, opened 2022-06-01
test(bridge): Fix flaky UI test on first day of a month

Kirdock:
On the first day of the month, some UI tests were not successful because it was not able to select two different days

### DIFF
--- a/bridge/cypress/support/pageobjects/TriggerSequenceSubPage.ts
+++ b/bridge/cypress/support/pageobjects/TriggerSequenceSubPage.ts
@@ -67,12 +67,12 @@ export class TriggerSequenceSubPage {
     return this;
   }
 
-  public setStartDate(calElement: number, hours: string, minutes: string, seconds: string): this {
-    return this.clickStartTime().selectDateTime(calElement, hours, minutes, seconds);
+  public setStartDate(calElement: number, hours: string, minutes: string, seconds: string, goBackMonths = 0): this {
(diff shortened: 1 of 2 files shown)


renepanzar (approved):
LGTM!

bridge/cypress/support/pageobjects/TriggerSequenceSubPage.ts
```diff
@@ -85,7 +85,16 @@ export class TriggerSequenceSubPage {
...
+  public selectDateTime(
+    calElement: number,
+    hours: string,
+    minutes: string,
+    seconds: string,
+    goBackMonths: number
+  ): this {
+    for (let i = 0; i < goBackMonths; ++i) {
+      cy.get('.dt-calendar-header-button-prev-month').click();
+    }
```
heinzburgstaller, line 97: Suggestion: You could do
```
[...Array(goBackMonths).keys()].forEach(() => cy.get('.dt-calendar-header-button-prev-month').click())
```
To get rid of for loop

Kirdock: This is still a for-loop, even a more complicated one.
You create an array with the length of 5, get the keys via iterator, create another array out of it and then iterate over it.
Also `array.forEach()` is as the same as `for (const i of array)`

Kirdock: I completely disagree here and for me, the for-loop approach is much more read-friendly than the forEach

PR #7952, opened 2022-06-01
fix(cp-connector): Subscribe to integrations before creating a job

odubajDT:

### DIFF
--- a/cp-connector/pkg/controlplane/subscriptionsource.go
+++ b/cp-connector/pkg/controlplane/subscriptionsource.go
@@ -53,11 +53,9 @@ func NewUniformSubscriptionSource(uniformAPI api.UniformV1Interface, options ...
 
 // Start triggers the execution of the UniformSubscriptionSource
 func (s *UniformSubscriptionSource) Start(ctx context.Context, registrationData RegistrationData, subscriptionChannel chan []models.EventSubscription) error {
-	go func() {
-		s.ping(registrationData.ID, subscriptionChannel)
-	}()
 	ticker := s.clock.Ticker(s.fetchInterval)
 	go func() {
+		s.ping(registrationData.ID, subscriptionChannel)
 		for {
 			select {
 			case <-ctx.Done():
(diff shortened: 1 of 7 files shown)


cp-connector/pkg/controlplane/subscriptionsource.go
```diff
@@ -53,13 +53,23 @@ func NewUniformSubscriptionSource(uniformAPI api.UniformV1Interface, options ...
 
 // Start triggers the execution of the UniformSubscriptionSource
 func (s *UniformSubscriptionSource) Start(ctx context.Context, registrationData RegistrationData, subscriptionChannel chan []models.EventSubscription) error {
+	s.logger.Info("Pulling subscriptions for registrationID: ", registrationData.ID)
```
warber, line 56: Why are you doing the initial ping outside the go routing that's already being scheduled? Is there a specific reason for it?
Further, you could extract the functionality in an unexported method.
Suggestions:

```go
func (s *UniformSubscriptionSource) Start(ctx context.Context, registrationData RegistrationData, subscriptionChannel chan []models.EventSubscription) error {
	ticker := s.clock.Ticker(s.fetchInterval)
	go func() {
		s.ping(registrationData.ID, subscriptionChannel)
		for {
			select {
			case <-ctx.Done():
				return
			case <-ticker.C:
				s.ping(registrationData.ID, subscriptionChannel)
			}
		}
	}()
	return nil
}

func (s *UniformSubscriptionSource) ping(id string, subscriptionChannel chan []models.EventSubscription) {
	s.logger.Info("Pulling subscriptions for registrationID: ", id)
	updatedIntegrationData, err := s.uniformAPI.Ping(id)
	if err != nil {
		s.logger.Errorf("Unable to ping control plane: %v", err)
		return
	}
	subscriptionChannel <- updatedIntegrationData.Subscriptions
}

```
?

RealAnna (dismissed):

RealAnna (approved):

PR #7960, opened 2022-06-02
fix(shipyard-controller): Avoid lost writes to subscriptions due to concurrent writes

bacherfl:
Closes #7956

This PR changes the logic of the Uniform mongodb repo implementation to use an append-only approach for adding subscriptions to a registered integration

Also, deleting subscriptions will be done using $pull (also to avoid concurrency issues)

Integration test run: https://github.com/keptn/keptn/runs/6706093301?check_suite_focus=true


odubajDT (approved):
lgtm

RealAnna (approved):

odubajDT (approved):

PR #7964, opened 2022-06-02
fix: Ensure ZD pipeline fails if any test fails

RealAnna:
Because the test actions runs 2 parallel processes sometimes if only one fails the pipeline does not detect it and thus triggers clean up of the cluster.

### DIFF
--- a/.github/workflows/zero-downtime-tests.yml
+++ b/.github/workflows/zero-downtime-tests.yml
@@ -367,9 +367,9 @@ jobs:
         run: |
           set -euo pipefail 
           
-           go test -json -v -run Test_ZeroDowntime -v -timeout 90m 2>&1 | tee /tmp/gotest.log &
-           cd ../../bridge &&
-           if [[ "${{ github.event.inputs.testUI }}" == 'true' ]]; then   
+          go test -json -v -run Test_ZeroDowntime -v -timeout 90m 2>&1 | tee /tmp/gotest.log &
+          cd ../../bridge &&
+          if [[ "${{ github.event.inputs.testUI }}" == 'true' ]]; then   
             while [[ $(jobs -pr) ]]; do
               echo "New Run: "  >> /tmp/uitest.log 
               npm run test:ui >> /tmp/uitest.log 2>&1 
--- a/.github/workflows/zero-downtime-tests.yml
+++ b/.github/workflows/zero-downtime-tests.yml
@@ -358,7 +358,6 @@ jobs:
       #######################################################################
 
       - name: Zero Downtime Tests
-        id: test_zero_downtime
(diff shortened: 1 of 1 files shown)


.github/workflows/zero-downtime-tests.yml
```diff
@@ -358,26 +358,31 @@ jobs:
...
         env:
           GOMAXPROCS: 10
           KEPTN_ENDPOINT: ${{ steps.determine_keptn_endpoint.outputs.KEPTN_ENDPOINT }}
         run: |
           set -euo pipefail 
           
-           go test -json -v -run Test_ZeroDowntime -v -timeout 90m 2>&1 | tee /tmp/gotest.log &
-           cd ../../bridge &&
-           if [[ "${{ github.event.inputs.testUI }}" == 'true' ]]; then   
+          go test -json -v -run Test_ZeroDowntime -v -timeout 90m 2>&1 | tee /tmp/gotest.log & GT=$!
```
mowies, line 370: I would suggest using a more descriptive variable name instead of `GT` :)

.github/workflows/zero-downtime-tests.yml
```diff
@@ -358,26 +358,31 @@ jobs:
...
+
+          cd ../../bridge &&
+          if [[ "${{ github.event.inputs.testUI }}" == 'true' ]]; then
```
mowies, line 373:
```suggestion
          if [[ "${{ github.event.inputs.testUI }}" == 'true' ]]; then
```

.github/workflows/zero-downtime-tests.yml
```diff
@@ -358,26 +358,31 @@ jobs:
...
+          if [[ "${{ github.event.inputs.testUI }}" == 'true' ]]; then   
             while [[ $(jobs -pr) ]]; do
               echo "New Run: "  >> /tmp/uitest.log 
               npm run test:ui >> /tmp/uitest.log 2>&1 
             done
           else
-            echo "UI tests will not run"
-          fi 
-
+            echo "UI tests will not run"
```
mowies, line 379:
```suggestion
            echo "UI tests will not run"
```

.github/workflows/zero-downtime-tests.yml
```diff
@@ -358,26 +358,31 @@ jobs:
...
           else
-            echo "UI tests will not run"
-          fi 
-
+            echo "UI tests will not run" 
+          fi
+        
+          wait $GT 
+          if [[ $? -ne 0 ]]; then
+            exit 1
```
mowies, line 384: Maybe add an error message before the exit?

mowies (approved):

PR #7997, opened 2022-06-03
fix(bridge): Update projects if dashboard is visited

### LINKED ISSUE #7942: Bridge: Dashboard does not update projects
First of all, this is not related to polling. Polling should not be here but if you visit the dashboard the projects should be updated which is currently not the case.

Kirdock:
Fixes #7942
Also removes one additional projects call while creating a project

### DIFF
--- a/bridge/client/app/app.component.ts
+++ b/bridge/client/app/app.component.ts
@@ -1,7 +1,9 @@
 import { HttpClient, HttpHeaders } from '@angular/common/http';
-import { Component, OnInit } from '@angular/core';
+import { Component } from '@angular/core';
 import { filter, take } from 'rxjs/operators';
 import { DataService } from './_services/data.service';
+import { ActivatedRoute } from '@angular/router';
+import { combineLatest, of } from 'rxjs';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any,@typescript-eslint/naming-convention
 declare let dT_: any;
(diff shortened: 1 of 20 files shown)


bridge/client/app/app.component.ts
```diff
@@ -11,20 +13,20 @@ declare let dT_: any;
...
+
+    combineLatest([this.route.firstChild?.data ?? of({}), keptnInfo$])
       .pipe(take(1))
-      .subscribe(() => {
-        this.dataService.loadProjects();
+      .subscribe(([data]) => {
+        if (!data.projectsHandledByComponent) {
+          this.dataService.loadProjects().subscribe();
+        }
       });
```
heinzburgstaller, line 30: Suggestion: Get rid of double subscription
```
 combineLatest([this.route.firstChild?.data ?? of({}), keptnInfo$]).pipe(
      take(1),
      filter(([data]) => !data.projectsHandledByComponent),
      mergeMap(() => this.dataService.loadProjects()),
).subscribe();

```

bridge/client/app/dashboard/dashboard.component.ts
```diff
@@ -11,18 +10,23 @@ import { DataService } from '../_services/data.service';
...
-      this.isQualityGatesOnly = isQualityGatesOnly;
-    });
+    this.dataService.keptnInfo
+      .pipe(
+        filter((keptnInfo) => !!keptnInfo),
+        take(1)
+      )
+      .subscribe(() => {
+        this.loadProjects();
+      });
```
heinzburgstaller, line 29: See get rid of double subscription above

bridge/client/app/app.component.ts
Kirdock: The reason why I put the check inside the subscribe instead of the pipe was to only trigger it once, as we don't care for any changes here.
If the filter is moved into the pipe, there would be another projects call if you redirect from the dashboard to another view.

heinzburgstaller (approved):

PR #8020, opened 2022-06-08
feat: Adapt go-utils changes in git credentials models

### LINKED ISSUE #7963: Refactor API model for storing git credentials
Currently we have a model where everything is in root level:
```
 {
  "gitPemCertificate": "string", // optional 
  "gitPrivateKey": "string", // optional 
  "gitPrivateKeyPass": "string", // optional 
  "gitProxyInsecure": "true", // optional 
  "gitProxyPassword": "string", // optional 
  "gitProxyScheme": "string", // optional 
  "gitProxyUrl": "string", // optional 
  "gitProxyUser": "string", // optional 
  "gitRemoteURL": "string", 
  "gitToken": "string", 
  "gitUser": "string", // optional 
  "name": "string", 
  "shipyard": "string" 
} 
```

suggestion:

```
 {
  "remoteURL": "string", 
  "user": "string", // optional 
  "https": { // optional 
     "token": "string", 
     "certificate": "string", //optional
     "insecureSkipTLS": "boolean", // default: false 
     "proxy": { // optional 
          "scheme": "string", 
          "url": "string", // split it in host and port? 
          "user": "string", // optional 
          "password": "string", // optional 
      } 
   }, 
   "ssh": { // optional 

     "privateKey": "string", 

     "privateKeyPass": "string", // optional 
    }, 

   "name": "string", 

  "shipyard": "string" 

}
```

odubajDT:
## This PR

- adapts services to use the update git remote repository credentials models
- implements migrator to migrate from old model to new one (secrets and DB)

### Fixes
Fixes #7963 

BREAKING CHANGE: Git credentials for git authentication were moved to a separate sub-structure in go-utils package and split to either ssh or http sub-structures depending on the used authentication method. This leads to new models for creating, updating and retrieving the project information.

### Integration tests
https://github.com/keptn/keptn/actions/runs/2502239906
https://github.com/keptn/keptn/actions/runs/2575663901


RealAnna:
since shipyard and resource service model changed I think you need to regenerate the swagger docs...

resource-service/common/git.go
```diff
@@ -80,23 +80,23 @@ func getGitKeptnEmail() string {
 }
 
 func getAuthMethod(gitContext common_models.GitContext) (transport.AuthMethod, error) {
-	if gitContext.Credentials.GitPrivateKey != "" && strings.HasPrefix(gitContext.Credentials.RemoteURI, "ssh://") {
-		publicKey, err := ssh.NewPublicKeys("git", []byte(gitContext.Credentials.GitPrivateKey), gitContext.Credentials.GitPrivateKeyPass)
+	if gitContext.Credentials.SshAuth != nil {
```
thisthat, line 83: I would expect `*.[https|ssh].*` as a format without the `Auth` part

configuration-service/common_models/git_credentials.go
```diff
@@ -1,8 +1,12 @@
 package common_models
 
-// GitCredentials contains git credentials info
-type GitCredentials struct {
-	User      string `json:"user,omitempty"`
-	Token     string `json:"token,omitempty"`
-	RemoteURI string `json:"remoteURI,omitempty"`
-}
+// // GitCredentials contains git credentials info
```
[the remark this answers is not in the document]
odubajDT: good catch, nope

resource-service/common/credentials.go
```diff
@@ -45,7 +46,7 @@ func (kr K8sCredentialReader) GetCredentials(project string) (*common_models.Git
 		return nil, errors2.ErrMalformedCredentials
 	}
 	if err := credentials.Validate(); err != nil {
-		logger.Debug("Issue with credentials : ", err)
+		logger.Info("Issue with credentials : ", err)
```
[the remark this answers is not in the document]
odubajDT: adapted

resource-service/common/git.go
odubajDT: then we need to change the models in go-utils :)

thisthat: we can do that in a follow up

cli/cmd/create_project.go
```diff
@@ -127,9 +127,9 @@ keptn create project PROJECTNAME --shipyard=FILEPATH --git-user=GIT_USER --git-r
 				return errors.New("Access token or private key must be set")
 			}
 
-			project.GitUser = *createProjectParams.GitUser
-			project.GitToken = *createProjectParams.GitToken
-			project.GitRemoteURL = *createProjectParams.RemoteURL
+			if *createProjectParams.GitToken != "" && *createProjectParams.GitPrivateKey != "" {
```
bacherfl, line 130: to be safe I suggest to use the `isStringFlagSet(s *string)` method (available in the same package in the CLI), which also checks if the passed string pointer is `nil`.

cli/cmd/create_project.go
```diff
@@ -139,29 +139,48 @@ keptn create project PROJECTNAME --shipyard=FILEPATH --git-user=GIT_USER --git-r
...
 				content, err := ioutil.ReadFile(*createProjectParams.GitPrivateKey)
 				if err != nil {
 					return fmt.Errorf("unable to read privateKey file: %s\n", err.Error())
 				}
 
-				project.GitPrivateKey = string(base64.StdEncoding.EncodeToString(content))
-				project.GitPrivateKeyPass = *createProjectParams.GitPrivateKeyPass
-			}
+				sshCredentials := apimodels.SshGitAuth{
+					PrivateKey:     string(base64.StdEncoding.EncodeToString(content)),
```
bacherfl, line 154: the `string()` conversion can be removed since `EncodeToString` already returns a `string`

cli/cmd/update_project.go
```diff
@@ -118,29 +114,48 @@ keptn update project PROJECTNAME --git-user=GIT_USER --git-remote-url=GIT_REMOTE
 				return errors.New("Proxy cannot be set without scheme")
 			}
 
-			project.GitProxyURL = *updateProjectParams.GitProxyURL
-			project.GitProxyScheme = *updateProjectParams.GitProxyScheme
-			project.GitProxyUser = *updateProjectParams.GitProxyUser
-			project.GitProxyPassword = *updateProjectParams.GitProxyPassword
-			project.InsecureSkipTLS = *updateProjectParams.InsecureSkipTLS
+			project.GitCredentials = &apimodels.GitAuthCredentials{
```
bacherfl, line 117: there seems to be quite some code duplication when it comes to setting the git credentials config if we look ad `update_project` and `create_project` - can we extract this part into a separate function and use it in both commands?

resource-service/common_models/git.go
```diff
@@ -4,68 +4,58 @@ import (
...
 		}
-	} else if strings.HasPrefix(g.RemoteURI, "ssh://") {
-		if g.GitPrivateKey == "" {
+	} else if g.SshAuth != nil {
+		if g.SshAuth.PrivateKey == "" {
 			return kerrors.ErrCredentialsPrivateKeyMustNotBeEmpty
 		}
 	} else {
-		return kerrors.ErrCredentialsInvalidRemoteURI
+		return kerrors.ErrCredentialsInvalidRemoteURL
```
bacherfl, line 36: the remote URL should have been already verified at the beginning of the function - should we introduce a specific error type for this case?
Also, let's add a unit test case that reaches this line

resource-service/errors/errors.go
```diff
@@ -88,10 +88,11 @@ var ErrExactSHA1NotSupported = New("server does not support exact SHA1 refspec")
 
 var ErrCredentialsNotFound = New("could not find upstream repository credentials")
 var ErrMalformedCredentials = New("could not decode upstream repository credentials")
-var ErrCredentialsInvalidRemoteURI = New("invalid remote URI")
+var ErrCredentialsInvalidRemoteURL = New("invalid remote URL")
 var ErrCredentialsTokenMustNotBeEmpty = New("token must not be empty")
 var ErrCredentialsPrivateKeyMustNotBeEmpty = New("private key must not be empty")
 var ErrProxyInvalidScheme = New("proxy scheme must be http or https")
+var ErrInvalidRemoteURL = New("RemoteURL scheme must be http, https or ssh")
```
bacherfl, line 95: please start the error message with a lower case

shipyard-controller/db/migration/project_credentials_migration.go
```diff
@@ -0,0 +1,59 @@
...
+		return fmt.Errorf("could not transform git credentials to new format: %s", err.Error())
+	}
+	if err := s.UpdateSecrets(projects); err != nil {
+		return err
+	}
+	return s.UpdateProjects(projects)
+
+}
+
+func (s *ProjectCredentialsMigrator) UpdateProjects(projects []*models.ExpandedProjectOld) error {
```
bacherfl, line 35: lets make this function unexported

shipyard-controller/db/migration/project_credentials_migration.go
```diff
@@ -0,0 +1,59 @@
...
+	for _, project := range projects {
+		err := s.projectRepo.UpdateProject(project)
+		if err != nil {
+			return fmt.Errorf("could not transform git credentials for project %s: %s", project.ProjectName, err.Error())
+		}
+	}
+	return nil
+}
+
+func (s *ProjectCredentialsMigrator) UpdateSecrets(projects []*models.ExpandedProjectOld) error {
```
bacherfl, line 48: lets make this function unexported

shipyard-controller/db/mongodb_project_credentials_repo.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		fmt.Println("Error retrieving projects from mongoDB: " + err.Error())
+		return nil, err
+	}
+	defer cursor.Close(ctx)
+	for cursor.Next(ctx) {
+		projectResult := &models.ExpandedProjectOld{}
+		err := cursor.Decode(projectResult)
+		if err != nil {
+			fmt.Println("Could not cast to *models.Project")
```
bacherfl, line 50: please use the logger lib instead of fmt.Println here

shipyard-controller/db/mongodb_project_credentials_repo.go
```diff
@@ -0,0 +1,101 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	apimodels "github.com/keptn/go-utils/pkg/api/models"
+	"github.com/keptn/keptn/shipyard-controller/models"
+	"go.mongodb.org/mongo-driver/bson"
+)
+
+type MongoDBProjectCredentialsRepo interface {
```
bacherfl, line 14: please add a short explanation as a comment to describe why this struct is needed in addition to the previous projectsMVRepo - also, just to clarify, this will be removed in the next version when we don't need to migrate anymore, right?

shipyard-controller/models/expanded_project_old.go
```diff
@@ -0,0 +1,42 @@
+package models
+
+import apimodels "github.com/keptn/go-utils/pkg/api/models"
+
+type ExpandedProjectOld struct {
```
bacherfl, line 5: please add a short explanation of why we have the `Old` struct in addition to the updated model, and what the differences are - otherwise, this might confuse us in a couple of months when we rediscover it :)

shipyard-controller/db/migration/project_credentials_migration.go
```diff
@@ -0,0 +1,59 @@
...
+	return s.UpdateProjects(projects)
+
+}
+
+func (s *ProjectCredentialsMigrator) UpdateProjects(projects []*models.ExpandedProjectOld) error {
+	if projects == nil {
+		return nil
+	}
+	for _, project := range projects {
+		err := s.projectRepo.UpdateProject(project)
```
bacherfl, line 40: will this always do an update of every project whenever the pod is restarted, or do we have a check that prevents the project from being updated if it is already using the current model?

shipyard-controller/db/mongodb_secret_credentials_repo.go
```diff
@@ -0,0 +1,96 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	apimodels "github.com/keptn/go-utils/pkg/api/models"
+	"github.com/keptn/keptn/shipyard-controller/common"
+	"github.com/keptn/keptn/shipyard-controller/models"
+)
+
+type MongoDBSecretCredentialsRepo interface {
```
bacherfl, line 13: i think we should rename this, as this component does not interact with mongodb

shipyard-controller/db/mongodb_project_credentials_repo.go
bacherfl, line 14: let's rename this interface to `ProjectCredentialsRepo`, as the `MongoDBRepo` is a concrete implementation of that interface

shipyard-controller/db/migration/project_credentials_migration_test.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			repo := ProjectCredentialsMigrator{
+				secretRepo:  tt.secretRepo,
+				projectRepo: tt.projectRepo,
+			}
+			err := repo.Transform()
+			require.Equal(t, tt.want, err)
```
bacherfl, line 189: since migrating all projects always comes with some risk I would suggest to make the assertions more detailed to make sure no data is being lost or wrongly modified. Maybe you can use the sequence executions migrator as an orientation for that: https://github.com/keptn/keptn/blob/a9c1d2fc748198326af8981319d851a430c2c654/shipyard-controller/db/migration/sequence_execution_migration_test.go#L93 :)

resource-service/common_models/git.go
odubajDT: agree to add a different error, but we have a unit test for this on line 153 in git_test.go

shipyard-controller/db/migration/project_credentials_migration.go
odubajDT: we do not have a check, it's updated every time the pod is restarted, but it's made the same way for all migrations AFAIK

shipyard-controller/db/mongodb_project_credentials_repo.go
odubajDT: added and yes, exactly :)

shipyard-controller/db/migration/project_credentials_migration_test.go
odubajDT: tests for transformation of the data are present in `project_credentials_repo_test.go` and `secret_credentials_repo_test.go` with hopefully all possible corner cases, there is no need to test it in this place again. This test aims to be more of a top level test, where the the used repositories are mocked

bacherfl (approved):

bacherfl (approved):

PR #8032, opened 2022-06-08
fix(bridge): Fix missing evaluation score of sequence

### LINKED ISSUE #8031: Bridge: Missing evaluation score for sequence
In the environment screen, the sequence does not show a score if the latest `evaluation.finished` event is not from the lighthouse service.
image

Kirdock:
Fixes #8031


ermin-muratovic (dismissed):
lgtm

PR #8039, opened 2022-06-09
chore(bridge): Improve has-logs polling

### LINKED ISSUE #7802: Bridge :: Improve uniform log polling mechanism
For determining if there are unread logs for any integration and showing the red bubble next to the settings a lot of calls have to be made in the BFF. For every integration logs have to be loaded and checked to see if there is anything new.
To check this on a regular basis provides a lot of value to the user. But the polling can be optimized in order to reduce traffic to the API.

## Acceptance criteria
- [ ] Reduce the polling interval to 2 minutes
- [ ] If there are unread logs and the bubble is already displayed there is no need to continue polling. We can stop it at this point
- [ ] When the user is then navigating to the integrations page to determine where logs are produced and the status is cleared, we can continue polling

## DoD
- [ ] Acceptance criteria implemented
- [ ] Unit tests adapted (?)

Kirdock:
Closes #7802

### DIFF
--- a/bridge/client/app/project-board/project-board.component.ts
+++ b/bridge/client/app/project-board/project-board.component.ts
@@ -22,7 +22,7 @@ export class ProjectBoardComponent implements OnInit, OnDestroy {
   public error$: Observable<string | undefined> = this._errorSubject.asObservable();
   public isCreateMode$: Observable<boolean>;
   public hasUnreadLogs$: Observable<boolean>;
-  private readonly uniformLogPollingInterval = 2 * 60_000;
+  private static readonly uniformLogPollingInterval = 2 * 60_000;
 
   constructor(
     private router: Router,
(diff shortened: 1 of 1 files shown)


bridge/client/app/project-board/project-board.component.ts
```diff
@@ -22,32 +22,43 @@ export class ProjectBoardComponent implements OnInit, OnDestroy {
   public error$: Observable<string | undefined> = this._errorSubject.asObservable();
   public isCreateMode$: Observable<boolean>;
   public hasUnreadLogs$: Observable<boolean>;
+  private readonly uniformLogPollingInterval = 2 * 60_000;
```
renepanzar, line 25: Does this need to be a field? It seems unnecessary to expose this variable to the whole component. Could this also be a constant outside of the class (`defaultUniformLogPollingInterval`), that is being used if `initialDelayMillis` is greater than 0? Something like this:

```
const defaultUniformLogPollingInterval = 2 * 60_000;

...
export class ...
...
    const uniformLogPollingInterval = initialDelayMillis > 0 ? defaultUniformLogPollingInterval : 0;
    const uniformLogTimer$ = projectName$.pipe(
      switchMap(() => AppUtils.createTimer(0, uniformLogPollingInterval)),
      takeUntil(this.unsubscribe$)
    );
```

ermin-muratovic (approved):
lgtm

ermin-muratovic (changes requested):
lgtm,
please include the feedback from @renepanzar

PR #8040, opened 2022-06-09
fix(resource-service): Remove token enforcement

thisthat:
## This PR

Removes the necessity to enforce a token to set an upstream to a project.

Integration test run: https://github.com/keptn/keptn/runs/7067226222?check_suite_focus=true


odubajDT (approved):
lgtm

PR #8050, opened 2022-06-10
fix(bridge): Fix broken UI if connection was lost

### LINKED ISSUE #8049: Bridge: The bridge does not work anymore if the connection was lost in between
Currently, if the connection is lost, a lot of things break in the UI even if the next calls succeed. The website has to be reloaded in order to work again.
- The service screen does not show any service deployments
- The environment is permanently showing "loading"
- The sequence screen does not show anything

This may be related to #7995

Kirdock:
Fixes #8049 

### DIFF
--- a/bridge/client/app/_services/data.service.ts
+++ b/bridge/client/app/_services/data.service.ts
@@ -341,8 +341,6 @@ export class DataService {
               projects?.splice(projectIdx, 1);
               this._projects.next(projects);
             }
-          } else {
-            this._projects.error(err);
           }
         }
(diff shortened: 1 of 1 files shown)


heinzburgstaller (commented):
Should we add an UI test, to ensure the pages are still working when there is an uncatched error?

heinzburgstaller (approved):

ermin-muratovic (approved):
lgtm

heinzburgstaller (approved):

PR #8059, opened 2022-06-13
feat(go-sdk): Use APISet instead of resource handler

### LINKED ISSUE #7436: go-sdk: use `APISet` instead of handler directly
Currently, the `go-sdk` offers the service implementation to get a handle to the Resource handler in order to upload resources to keptn (`IKeptn::GetResourceHandler`). Right now, this has the limitation that it is only usable when running inside control plane since an unauthenticated resource handler is provided. 

Instead, it should be possible to get a handle to the [APISet](https://github.com/keptn/go-utils/blob/003ee3f85292e5ab6a049f8412bbb8fa28d4b6eb/pkg/api/utils/client.go#L27). A task handler implementation should be able to get a handle to the `APISet` via the `sdkIKeptn` interface:

``` go
...
func (g *GreetingsHandler) Execute(k sdk.IKeptn, event sdk.KeptnEvent) (interface{}, *sdk.Error) {
...
api := k.APIV1()
...
}
```

## Acceptance Criteria
- go-sdk provides easy way to access Keptn API via `APISet`
- old `GetResourceHandler` method is deleted
- current (internal) services using the SDK are adapted
- Add a config flag to select internal/external API call implementation

## DoD 

- [x] Unit tests

## Resources

- [go-utils PR](https://github.com/keptn/go-utils/pull/377)

RealAnna:
Signed-off-by: RealAnna <anna.reale@dynatrace.com>

## This PR

- go-sdk provides easy way to access Keptn API via APISet

### Related Issues

Closes #7436

### Follow-up Tasks
Make issue to remove later on ResourceHandler (when we have APIv2)


warber (changes requested):
Suggestion in order to reduce public api surface:
Move `client.go`, `initializer.go` as well as `config.go` as they are into packages living in the `internal` directory, e.g.:

```
+ internal/api
    + ---- client.go
    + ---- client_test.go
    + ---- initializer.go
    + ---- initializer_test.go
+ internal/config
    + ---- config.go
```

go-sdk/example/handler_test.go
```diff
@@ -11,7 +12,7 @@ import (
 )
 
 func Test_Handler(t *testing.T) {
-	fakeKeptn := sdk.NewFakeKeptn("test-greeting-svc")
+	fakeKeptn := sdk.NewFakeKeptn("test-greeting-svc", &api.APISet{})
```
warber, line 15: if no api set is required / used, just pass `nil` 
```suggestion
	fakeKeptn := sdk.NewFakeKeptn("test-greeting-svc", nil)
```

PR #8063, opened 2022-06-13
fix(cp-connector): Synchronized shutdown of cp-connector during cancelation

warber:
This PR contains changes for `cp-connector` lib to do a synchronized shutdown whenever the context is canceled.

### DIFF
--- a/cp-connector/pkg/controlplane/controlplane.go
+++ b/cp-connector/pkg/controlplane/controlplane.go
@@ -11,6 +11,7 @@ import (
 	"github.com/keptn/keptn/cp-connector/pkg/logger"
 	"github.com/keptn/keptn/cp-connector/pkg/subscriptionsource"
 	"github.com/keptn/keptn/cp-connector/pkg/types"
+	"sync"
 )
 
 type EventSender = types.EventSender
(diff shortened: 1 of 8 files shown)


cp-connector/pkg/controlplane/controlplane.go
```diff
@@ -82,13 +83,15 @@ func (cp *ControlPlane) Register(ctx context.Context, integration Integration) e
 	cp.logger.Debugf("Registered with integration ID %s", cp.integrationID)
 	registrationData.ID = cp.integrationID
 
+	wg := &sync.WaitGroup{}
+	wg.Add(2)
```
thisthat, line 87: I would add a small comment to say one is for eventsource and the other is for subscription source

cp-connector/pkg/subscriptionsource/subscriptionsource.go
```diff
@@ -112,7 +114,7 @@ func NewFixedSubscriptionSource(options ...func(source *FixedSubscriptionSource)
 	return fss
 }
 
-func (s FixedSubscriptionSource) Start(ctx context.Context, data types.RegistrationData, c chan []models.EventSubscription) error {
+func (s FixedSubscriptionSource) Start(ctx context.Context, data types.RegistrationData, c chan []models.EventSubscription, wg *sync.WaitGroup) error {
 	go func() { c <- s.fixedSubscriptions }()
```
thisthat, line 118: no wg.Done() here?

warber: whoops... good catch :-O

PR #8083, opened 2022-06-14
refactor(bridge): update sequence screen data model

### LINKED ISSUE #5101: Update sequence screen data model

Kirdock:
Closes #5101 
Fixes a flaky UI test in the sequence screen


bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.html
```diff
@@ -1,37 +1,36 @@
...
+    <ng-template #triggerSequenceButton>
+      <div>
+        <a (click)="navigateToTriggerSequence(state.projectName)">
+          <button dt-button uitestid="keptn-trigger-button-open">
+            <dt-icon name="flash"></dt-icon>
+            Trigger a new sequence
+          </button>
+        </a>
+      </div>
+    </ng-template>
```
ermin-muratovic, line 12: After clicking through the Sequence screen, I noticed that the list of sequences is not stretching vertically to 100% height. Previously it would use the full height. I don't see actually which change might have caused that, maybe it's because of moving the `*ngIf` to the outer container? Can you try to reproduce it and fix it?
image

ermin-muratovic (changes requested):

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.html
Kirdock: I removed `encapsulation: ViewEncapsulation.None,` and then the style in the related scss file wasn't applied anymore.
Added `:host::ng-deep` to it 👍

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.html
```diff
@@ -53,13 +52,13 @@
...
-              >
+              <ktb-loading-distractor *ngIf="loading" uitestid="keptn-loadingOldSequences">
+                Loading …
+              </ktb-loading-distractor>
               <button
                 dt-show-more
-                *ngIf="!loading && !project?.allSequencesLoaded"
-                (click)="loadOldSequences()"
+                *ngIf="!loading && !state.sequenceInfo?.allSequencesLoaded"
+                (click)="loadOldSequences(state.projectName)"
```
ermin-muratovic, line 61: "Show older sequences" is displayed, although no older sequences available.
image

ermin-muratovic (changes requested):

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.html
Kirdock: Fixed. FYI this problem only appeared after the polling happened (the previous value was not considered)

bridge/client/app/_services/data.service.ts
```diff
@@ -404,12 +409,20 @@ export class DataService {
         })
       )
       .subscribe(([sequences, totalCount]: [Sequence[], number]) => {
-        this.addNewSequences(project, sequences, !!beforeTime, oldSequence);
+        const previousSequences = this._sequences.getValue();
+        const sequenceData = previousSequences[projectName];
+        const newSequences = this.addNewSequences(sequenceData?.sequences ?? [], sequences, !!beforeTime, oldSequence);
+        const allSequencesLoaded =
+          previousSequences[projectName]?.allSequencesLoaded ||
+          this.allSequencesLoaded(sequenceData?.sequences.length ?? 0, totalCount, fromTime, beforeTime);
```
ermin-muratovic, line 417:
```suggestion
          this.allSequencesLoaded(newSequences?.length ?? 0, totalCount, fromTime, beforeTime);
```

ermin-muratovic (changes requested):

ermin-muratovic (approved):
lgtm

ermin-muratovic (approved):
lgtm

heinzburgstaller (approved):

PR #8095, opened 2022-06-15
test(bridge): Run UI tests in production mode and log failed requests

### LINKED ISSUE #7552: CI UI tests should run with production mode enabled

When running the UI tests in the CI environment they are run with development mode enabled. While this is not a big problem it is the cause for a few issues:

* **Flaky tests** - the Angular change detection cycle is being run two times to check for modifications (the second run might re-render a component that has already been fetched by `cy.get()`, thus failing the test)
* The **HTML sanitizer logs a warning** when it needs found something to sanitize (see https://github.com/angular/angular/blob/1d2f5c1101ccd182f5b528de52583a1b98dd6789/packages/core/src/sanitization/html_sanitizer.ts#L273) - this leads to a failing test, as we expect no logs during our UI tests

The `start:ci` NPM script in `bridge/package.json` uses the test environment by setting `--configuration=test`.

The production mode can be enabled by simply using `production=true` in `environment.test.ts`, which will eventually call `enableProdMode()`.

Alternatives are to use the `prod` configuration, but this it uses polling in the UI (which we actually want to avoid for testing).

**Additional context**

This is how the proposed solution would look like:

```
diff --git a/bridge/client/environments/environment.test.ts b/bridge/client/environments/environment.test.ts
--- a/bridge/client/environments/environment.test.ts	(revision 6b2d90fe9aae687980c25c772ae760cded04b982)
+++ b/bridge/client/environments/environment.test.ts	(date 1648728592975)
@@ -3,6 +3,7 @@
 class Environment extends DynamicEnvironment {
   constructor() {
     super();
+    this.production = true;
     this.pollingIntervalMillis = 0;
   }
 }
```

##  Interceptor

The interceptor should check for correct status codes and log failures. The logger should fail the test when it receives failures.

## Acceptance criteria

- [ ] Switch tests to production mode
- [ ] UI Tests: Add an http-interceptor for the status code and check that is between 2xx

## DoD

- [ ] Tests are run with production mode enabled

Kirdock:
Closes #7552

### DIFF
--- a/bridge/client/environments/environment.test.ts
+++ b/bridge/client/environments/environment.test.ts
@@ -4,6 +4,7 @@ class Environment extends DynamicEnvironment {
   constructor() {
     super();
     this.pollingIntervalMillis = 0;
+    this.production = true;
   }
 }
 
--- a/bridge/cypress/integration/deep-links.spec.ts
+++ b/bridge/cypress/integration/deep-links.spec.ts
@@ -47,6 +47,7 @@ describe('Test deep links', () => {
     environmentPage.visit(mockedProject);
     cy.location('pathname').should('eq', `/project/${mockedProject}`);
     basePage.clickMainHeaderKeptn();
+    dashboardPage.waitForProjects();
 
     cy.location('pathname').should('eq', '/dashboard');
   });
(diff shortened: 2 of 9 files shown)


renepanzar (approved):
LGTM!

renepanzar (approved):

PR #8122, opened 2022-06-20
chore(cp-connector): Fixed missing error in queuesubscribe for nats

RealAnna:


cp-connector/pkg/nats/nats.go
```diff
@@ -139,6 +139,12 @@ func (nc *NatsConnector) QueueSubscribeMultiple(subjects []string, queueGroup st
 		return ErrSubNilMessageProcessor
 	}
 
+	// Immediately verify if the Connection is valid, to avoid starting go routines for a
+	// faulty nats connection with no subjects
+	if _, err := nc.getOrCreateConnection(); err != nil {
```
warber, line 144: nitpick: let's just rename this one to `ensureConnection()` i think this is also how it's called in mongodb datastore / shipyard-controller

Issue #7149, opened 2022-03-14
Core: Shipyard-controller :: Offer extension points for provisioning of repository


thisthat:
## User Story

As a user, I would like to create a project without taking care of creating a repository first

## Acceptance Criteria

- [x] Add to the `shipyard-controller` the env var `AUTOMATIC_PROVISIONING_URL` which is read from the helm value `features.automaticProvisioningURL`. The value defaults to an empty string.
- [x] If (i) the project creation API is called and (ii) the value of `AUTOMATIC_PROVISIONING_URL` is not empty and (iii) the gitRemoteUrl is empty, then invoke the `POST ${AUTOMATIC_PROVISIONING_URL}/repository` endpoint before continuing with the normal processing of the project creation.
- [x] If (i) the project deletion API is called and (ii) the value of `AUTOMATIC_PROVISIONING_URL` is not empty, then the shipyard-controller shall invoke the `DELETE ${AUTOMATIC_PROVISIONING_URL}/repository` before continuing with the project deletion. 
- [x] If an API call to `${AUTOMATIC_PROVISIONING_URL}/repository` fails, then the `shipyard-controller` should also fail fast logging an error and return a 424 HTTP code with the message `Error provisioning a project instance`/`Error deleting a provisioned project instance`.  
- [x] The swagger definition for the shipyard-controller is updated with the return HTTP code 424.

## Techincal Details

The expected API contract for the repository provisioned are defined in the following swagger definition: [swagger.txt](https://github.com/keptn/keptn/files/8341909/swagger.txt)

## DoD

- Unit test
- Integration test with a simple mock server that simulates the implementation of the automatic provisioning REST APIs

## Resources

https://editor.swagger.io/

## Dependencies
- https://github.com/keptn/keptn/issues/7151

PR #8137, opened 2022-06-21
feat(api): Create import endpoint

pchila:
## This PR

- adds a new import endpoint in api-service for importing configuration and resources by uploading a zip archive 
- add an emptyDir volume as import scratch space to the api-service in the helm charts and service.yaml
- added a couple of env vars for configuration of control plane endpoint and max size of an uncompressed import package in the helm chart and service.yaml
- endpoint is marked as `WIP/beta` since the implementation is still ongoing, will be cleared once the implementation is done

This first implementation provides a simple POST /import handler along with some zip archive handling and sanity checks.

### Follow-up Tasks

Business logic processing the upload will be developed in follow-up PRs (this PR will handle the upload and local storage of the archive, validation and then throws it away)

### How to test

Code has been covered as much as possible by unit tests, however it is possible to test the upload using the swagger UI at /api/swagger-ui/#/Import/import  and providing a project name and a .zip file


christian-kreuzberger-dtx (dismissed):
Looks good from my point of view. 
Please wait before we merge this.
@thisthat let me know if we want an additional review from people that recently contributed to the api-service itself.

warber (commented):
Since there is already a `handlers` folder and because I think there is no reason to put this code inside `pkg`, can you move the new `import` folder into `handlers` ? Also naming the package after a language keyword and having the need to compensate with `_` feels a bit strange. Maybe there is another suitable name for it, like `importer` ?

api/pkg/import/zip_package.go
```diff
@@ -0,0 +1,149 @@
...
+	"io"
+	"os"
+	"path"
+	"strings"
+
+	logger "github.com/sirupsen/logrus"
+)
+
+// ErrorUncompressedSizeTooBig signals that a zip archive has an uncompressed size greater than maximum size allowed
+var /*const*/ ErrorUncompressedSizeTooBig = errors.New("uncompressed size of package exceeds configured maximum size")
```
thisthat, line 16:
```suggestion
const ErrorUncompressedSizeTooBig = errors.New("uncompressed size of package exceeds configured maximum size")
```

api/pkg/import/zip_package.go
```diff
@@ -0,0 +1,149 @@
...
+					)
+				}
+
+				extractedSize += zippedFile.UncompressedSize64
+				return nil
+			}()
+
+			if err != nil {
+				return err
+			}
```
thisthat, line 124: Since we don't do anything fancy with the returned error, I suggest to remove the extra wrapping of a function

api/pkg/import/zip_package_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+	p, err := NewPackage(tempZipFile.Name(), testArchiveSize20MB)
+	assert.ErrorIs(t, err, os.ErrNotExist)
+	assert.Nil(t, p)
+
+	// the extraction folder is cleaned up
+	expectedExtractedPath := strings.TrimSuffix(tempZipFile.Name(), defaultImportArchiveExtension)
+	assert.NoDirExists(t, expectedExtractedPath)
+}
+
```
thisthat, line 117: I would add also a small test to check that we fail if an invalid zip is supplied

api/swagger.yaml
```diff
@@ -59,6 +59,42 @@ paths:
           description: Error
           schema:
             $ref: "#/definitions/error"
+  /import:
+    post:
+      tags:
+        - Import
+      operationId: import
+      summary: "WIP/DO NOT USE: Import a zip package"
```
thisthat, line 67:
```suggestion
      summary: "Alpha: Import a zip package"
      description: "This API is still in alpha state and we do not recommend its usage outside of testing purposes"
```

thisthat (changes requested):

api/pkg/import/handler.go
```diff
@@ -0,0 +1,130 @@
...
+	projectExists, err := ih.checker.ProjectExists(params.Project)
+
+	if err != nil || !projectExists {
+		message := fmt.Sprintf("project %s does not exist", params.Project)
+		if err != nil {
+			message = fmt.Sprintf("error checking for project %s existence : %v", params.Project, err)
+		}
+
+		error := models.Error{
+			Code:    http.StatusNotFound,
```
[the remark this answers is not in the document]
pchila: @christian-kreuzberger-dtx any thoughts ?

api/pkg/import/handler.go
```diff
@@ -0,0 +1,130 @@
...
+			Message: &message,
+		}
+		return import_operations.NewImportNotFound().WithPayload(&error)
+	}
+
+	tempFileName, err := func() (string, error) {
+		file, err := ioutil.TempFile(ih.tempStorageDir, "importConfig*"+defaultImportArchiveExtension)
```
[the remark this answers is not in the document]
pchila: The asterisk is there to identify the unique part of the filename so that we obtain a file with a name like `importConfig<random part>.zip`:  https://pkg.go.dev/io/ioutil#TempFile

api/swagger.yaml
pchila: @christian-kreuzberger-dtx please check if this is acceptable

pchila:
@warber regarding the package, I felt it was a little too much code to put all in the handlers package (aside from the fact that the whole zip archive handling does not really fit in there).
Agreed on the poor naming, I would like to hear from @christian-kreuzberger-dtx  if something like `importer` or something equivalent is ok...
Regardless of the final name choice, I would still create a dedicated package for the zip handling and the (upcoming) business logic for the manifest processing if it's ok with you...

api/pkg/import/project_checker.go
```diff
@@ -0,0 +1,43 @@
+package _import
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/keptn/keptn/api/utils"
+)
+
+const projectEndpoint = "/v1/project/"
```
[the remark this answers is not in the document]
pchila: if I understand correctly, you are proposing to use [this](https://github.com/keptn/go-utils/blob/d5ad06c725cddaadb72d60cf0566421fad1f2c12/pkg/api/utils/projectUtils.go#L106) ?

api/pkg/import/zip_package.go
```diff
@@ -0,0 +1,149 @@
...
+
+	m.extractedDir = extractionDir
+	return extractZipArchive(zipReader, extractionDir, maxSize)
+}
+
+// This function could be part of the manifest interface to abstract the physical location of the files
+// func (m *ZippedPackage) GetResource(resourcePath string) (io.ReadCloser, error) {
+// 	// TODO
+// 	return nil, nil
+// }
```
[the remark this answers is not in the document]
pchila: it was a hint for the next tasks but sure it's a comment and can be removed

api/pkg/import/zip_package.go
pchila: this cannot be done as the right-hand-side expression is not a constant expression.
We have to keep the constant as a global variable in this case

api/pkg/import/zip_package.go
pchila: In this case it's necessary to trigger the defer() calls at the end of each iteration rather than at the end of the function.
The idea is to close  the files as soon as possible (each iteration of the for loop extracts a file) hence the function.
If it looks messy this can be moved to a named private function...

api/pkg/import/zip_package_test.go
pchila: There was a corresponding test for the handler before the refactoring, I will remove that and create a new one for the zip archive

api/pkg/import/project_checker.go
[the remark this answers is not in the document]
thisthat: Exactly :)

christian-kreuzberger-dtx:
@warber We wanted to introduce the `pkg` folder to follow best practices seen in other Keptn integrations and microservices. This way we don't overload the `handler` package. Also, if we ever need to move the implementation, it's way easier if it lives in its own package.

api/pkg/import/handler.go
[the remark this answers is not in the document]
christian-kreuzberger-dtx: @thisthat I've told @pchila to avoid that for internal API calls, as in the end it would mean having *multiple* HTTP Status 500 API Calls instead of just one.
For me, the end-result is what matters: We could not find the project, for whatever reason.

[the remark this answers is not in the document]
thisthat: I agree with your reasoning @christian-kreuzberger-dtx and we should avoid 500 errors here! 
I think we should instead return a 424 error as we have done for the automatic provisioning: https://github.com/keptn/keptn/issues/7149

warber:
@christian-kreuzberger-dtx as I said, no need to put everything into the handler package, you could as well just move the `import` folder beside the handler package. I just don't want to have the `pkg` introduced without a clear purpose.

api/pkg/import/project_checker.go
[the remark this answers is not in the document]
pchila: Done

api/restapi/configure_keptn.go
```diff
@@ -56,7 +62,7 @@ func configureAPI(api *operations.KeptnAPI) http.Handler {
 		os.Exit(1)
 	}
 
-	/// configure the api here
+	// / configure the api here
```
christian-kreuzberger-dtx, line 65: That looks like an accidental space :)

api/importer/project_checker_test.go
```diff
@@ -0,0 +1,119 @@
...
+    {
+      "services": [],
+      "stageName": "production"
+    }
+  ]
+}
+`
+
+type fakeProjectEndPoint struct {
+	f func(w http.ResponseWriter, r *http.Request)
```
Raffy23, line 41: I would prefer to have a function name like `handleRequest` instead of just `f`

api/restapi/configure_keptn.go
pchila: This is coming from auto-formatting of the code once I touched the file (some comments are not well formatted and somlines are a little long). Undid the format changes for the unaffected parts of the file and pushed a fixup

api/test/data/import/sample-package/manifest.yaml
```diff
@@ -0,0 +1,55 @@
...
+  # Create a subscription
+  - name: "Create webhook subscription"
+    type: api
+    id: create-subscription
+    payload: "api/create-subscription.json"
+    action: "keptn-api-v1-uniform-create-webhook-subscription"
+    # internally this figures out the current webhook integration id
+    context: # -> goes into "api/create-subscription.json"
+      event: "sh.keptn.event.evaluation.triggered"
+      project: "{{ .inputs.project }}"
```
Raffy23, line 38: I think we should use the same parentheses for the templating as in the other files.
```suggestion
      project: "[[ .inputs.project ]]"
```

christian-kreuzberger-dtx (dismissed):
LGTM
@warber @thisthat any objections?

api/test/data/import/sample-package/manifest.yaml
```diff
@@ -0,0 +1,55 @@
...
+  - name: "Webhook config"
+    type: resource
+    resource: "resources/webhook.yaml"    # where is the file stored in the package
+    resourceUri: "webhook.yaml"           # what should the file be called in the upstream repo
+    # option 1: specify stage
+    stage: "dev"
+    # option 2: specify no stage at all --> default branch (e.g., omit --stage=dev)
+    context:   # Placeholders for templating in resources/webhook.yaml
+      event: "{{ .tasks.create-subscription.context.event }}"
+      subscriptionId: "{{ .tasks.create-subscription.response.id }}"
```
Raffy23, line 52: Same here:
```suggestion
      event: "[[ .tasks.create-subscription.context.event ]]"
      subscriptionId: "[[ .tasks.create-subscription.response.id ]]"
```

Raffy23 (dismissed):
LGTM

thisthat (approved):
LGTM, thank you very much for the contribution! 🦾

christian-kreuzberger-dtx:
The PR is approved, but @pchila please squash and sign the commit. I will re-approve it

christian-kreuzberger-dtx (dismissed):
LGTM

christian-kreuzberger-dtx (approved):

christian-kreuzberger-dtx (approved):
LGTM

PR #8138, opened 2022-06-21
chore(installer): Added  API_PROXY_HTTP_TIMEOUT to distributor helm values

RealAnna:


installer/manifests/keptn/templates/_helpers.tpl
```diff
@@ -97,6 +97,10 @@ lifecycle:
 {{- else }}
   value: {{ .Chart.AppVersion }}
 {{- end }}
+- name: API_PROXY_HTTP_TIMEOUT
+  valueFrom:
+   fieldRef:
+      fieldPath: {{ ((.Values.distributor.config).proxy).httpTimeout | default 30 }}
```
thisthat, line 103: I think this should be directly a value

PR #8247, opened 2022-06-29
feat(bridge): Support new webhook.yaml version v1beta1

### LINKED ISSUE #7839: Bridge :: Make use of Webhook `v1beta1`
## Technical Details

In #7238, we introduced a new version of the webhook format `v1beta1`. A webhook file in the new format looks like the following:

```
apiVersion: webhookconfig.keptn.sh/v1beta1
kind: WebhookConfig
metadata:
  name: webhook-configuration
spec:
  webhooks:
    - type: "sh.keptn.event.mytask.triggered"
      subscriptionID: 0123456789
      sendFinished: true
      requests:
        - url: http://mykeptnendpoint.org/api/v1/project/{{.data.project}}/stage/{{.data.stage}}
          method: GET
    - type: "sh.keptn.event.mytask.triggered"
      subscriptionID: 0123456987
      sendFinished: true
      envFrom: 
        - name: "secretKey"
          secretRef:
            name: "my-webhook-k8s-secret"
            key: "my-key"
      requests:
        - url: http://mykeptnendpoint.org/api/v1/project/{{.data.project}}
          method: GET
          headers:
            - key: x-token
              value: "{{.env.secretKey}}"
        - url: http://mykeptnendpoint.org/api/v1/project/{{.data.project}}/stage/{{.data.stage}}
          method: GET`
```

## Acceptance Criteria

- [x] Bridge can read both `v1alpha1` and `v1beta1` webhook versions
- [x] Bridge saves webhooks only in `v1beta1` version using the `api/configuration-service/v1` REST API

## DoD

- [x] Unit tests

Kirdock:
Closes #7839


bridge/server/models/webhook-config.utils.ts
```diff
@@ -0,0 +1,191 @@
...
+    data = JSON.stringify(JSON.parse(data), null, 2);
+  } catch {}
+  return data;
+}
+
+export function stringifyPayload(payload: string): string {
+  try {
+    return JSON.stringify(JSON.parse(payload));
+  } catch {
+    return payload.replace(/\r\n|\n|\r/gm, '');
```
heinzburgstaller, line 86: What if payload is null?

bridge/server/utils/curl.utils.ts
```diff
@@ -1,28 +1,5 @@
...
-      stringify = stringify.replace(/\r\n|\n|\r/gm, '');
-    }
-    params += `--data '${stringify}' `;
-  }
-  return `curl ${params}${webhookConfig.url}`;
-}
-
-function parseCurl(curl: string): { [key: string]: string[] } {
-  const startCommand = 'curl ';
+function parseCurl(curl: string, ignoreStart = false): { [key: string]: string[] } {
```
heinzburgstaller, line 1: Can curl be null

bridge/server/interfaces/webhook-config.ts
```diff
@@ -0,0 +1,5 @@
+export interface IWebhookConfigFilter {
+  projects: string[];
+  stages: string[] | [undefined];
+  services: string[] | [undefined];
```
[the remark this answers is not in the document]
Kirdock: Yes and no. This is for iterating over stages, services, etc.. The use case is to not differentiate between empty array and not defined.
https://github.com/keptn/keptn/blob/a59d0a21d09ea172b5e26583a7a6fe555a540f02/bridge/server/services/data-service.ts#L787-L807
If you create a webhook.yaml without a filter, it is created once for a project. If you have multiple stages and don't have any services ( => `[undefined]`), the webhook.yaml file is created for every stage. The same if you don't have a stage but multiple services.

bridge/server/utils/curl.utils.ts
Kirdock: No. The interface/webhook-version that is used here provides `string[]` for parsing the curl commands. It's either an empty array or a string.

heinzburgstaller (approved):

PR #8269, opened 2022-07-01
fix(bridge): Save client secret in k8s secret

### LINKED ISSUE #8241: OAuth client secret should be set by reference to a k8s secret
## Details

Currently, the value of a secret parameter of OAuth is passed in plain text via the `bridge.oauth.clientSecret` Helm value: https://github.com/keptn/keptn/blob/master/installer/manifests/keptn/templates/core.yaml#L134.
All secret values should be, instead, passed through K8s secrets.

Kirdock:
Fixes #8241 
But most of the PR is refactoring.


heinzburgstaller (dismissed):

heinzburgstaller (approved):

mowies:
helm changes LGTM!

PR #8298, opened 2022-07-05
test(bridge): Remove not needed and flaky UI test

Kirdock:
A UI test failed that a certain HTML element could not be found, even though the screenshot showed that the element is here. We don't really need this test anyway so I removed it.


ermin-muratovic (approved):
lgtm

PR #8309, opened 2022-07-05
feat(api): Process import package manifest and execute API tasks

pchila:
## This PR

- Adds the import package manifest processing and execution. 
- Adds a generic implementation for executing one-shot API calls
- Basic `[[project]]` placeholder replacement in URL
At the moment only the following tasks are implemented:
- keptn api task : `keptn-api-v1-create-service`


api/importer/execute/keptn_endpoints.go
```diff
@@ -0,0 +1,7 @@
+package execute
+
+type StaticKeptnEndpointProvider struct{}
+
+func (_ StaticKeptnEndpointProvider) GetControlPlaneEndpoint() string {
+	return "http://shipyard-controller:8080"
```
Raffy23, line 6: Instead of using a static value for creating the URL, the provider should probably read the `CONTROLPLANE_URI` environment variable to like it was done when creating the `ImportHandlerFunc`: 
https://github.com/keptn/keptn/blob/1b140dcc828e70543e218b628f871d67d65d79e5/api/restapi/configure_keptn.go#L111-L118

pchila: I was thinking of having multiple implementations of the `KeptnEndpointProvider` (static one for simplicity when running side by side with keptn and other one that is configurablem maybe with smart defaults). This is the simplest implem of them all though...
I was also thinking of making sure that the ProjectChecker and the executors use the same Keptn control plane, so probably they will receive the same EndpointProvider (based on env or not :stuck_out_tongue_winking_eye:  )

api/handlers/zip_package.go
```diff
@@ -20,6 +20,16 @@ type ZippedPackage struct {
...
 
+func (m *ZippedPackage) GetResource(resourceName string) (io.ReadCloser, error) {
+	actualPath := path.Join(m.extractedDir, resourceName)
+	file, err := os.Open(actualPath)
+	if err != nil {
+		return nil, fmt.Errorf("error accessing resource %s: %w", resourceName, err)
+	}
+
+	return file, nil
+}
```
christian-kreuzberger-dtx, line 31: Not sure if it applies to this piece of code, but could one hypothetically abuse this by, let's say, calling `GetResource("../../../../../../etc/passwd")`?

In any case, we could try to make sure that we are not allowing to go above the base directory - not sure how to do that in Go though...

api/test/utils/reader.go
```diff
@@ -0,0 +1,46 @@
...
+	throwError   bool
+}
+
+func NewTestReader(data []byte, allowedLoops int, throwError bool) *TestReader {
+	return &TestReader{
+		data:         data,
+		allowedLoops: allowedLoops,
+		throwError:   throwError,
+	}
+}
```
christian-kreuzberger-dtx, line 22: Hope i'm not nit-picking here... But I think this could benefit from some documentation, it's not immediately clear to me what this does, until I look up how it is used and why it is used in the actual test.

api/test/data/import/.gitignore
```diff
@@ -0,0 +1 @@
+!api
```
christian-kreuzberger-dtx, line 1: Why :thinking: ?

api/handlers/zip_package.go
pchila: We can:
-  Clean() the path and then check if we are still inside the temp directory
- Forbid the `.` and `..` in the manifest.yaml references
- Not extract any file and process everything by reading the zip file directly

api/test/data/import/.gitignore
pchila: Because otherwise I could not easily add to git the `api` directory of the sample packages in the `test/data` directory.
It already happened in the previous PR but I realized it only today whan I saw the CI tests failing because some files were not committed to git.
For reference:  https://github.com/keptn/keptn/blob/master/api/.gitignore#L10

api/importer/execute/keptn_endpoints.go
christian-kreuzberger-dtx: Please follow a pattern similar to
https://github.com/keptn/keptn/blob/5aa0a78dfd668c86f7a2c056882586c63d975ca6/api/utils/event.go#L17-L28

(I don't mind where you put the code though, that's your call)

api/test/utils/reader.go
pchila: It's a simple io.Reader implementation that goes boom on command to test that the code can handle I/O errors. I will add some comments...

api/restapi/configure_keptn.go
```diff
@@ -102,11 +104,16 @@ func configureAPI(api *operations.KeptnAPI) http.Handler {
...
+		new(model.YAMLManifestUnMarshaler), execute.NewKeptnExecutor(execute.StaticKeptnEndpointProvider{}),
+	)
+
 	api.ImportOperationsImportHandler = import_operations.ImportHandlerFunc(
 		handlers.GetImportHandlerFunc(
 			env.ImportBasePath,
-			importer.NewControlPlaneProjectChecker(os.Getenv(controlPlaneServiceEnvVar)),
+			handlers.NewControlPlaneProjectChecker(os.Getenv(controlPlaneServiceEnvVar)),
 			env.MaxImportUncompressedSize,
+			importProcessor,
```
christian-kreuzberger-dtx, line 116: Please consolidate the two different ways of getitng the control plane URL here. Personally, I'd focus on the `os.Getenv` approach but within a function, so you don't repeat yourself.

api/importer/execute/keptn_endpoints.go
pchila: Are you referring to the `SanitizeURL` usage or accessing directly the Environment (makes testing a little harder than it needs to be)?

api/importer/execute/default_api.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	defer response.Body.Close()
+	responseBody := new(any)
+	if response.Header.Get("Content-Type") == "application/json" {
+		bytes, err := io.ReadAll(response.Body)
+		if err == nil {
+			json.Unmarshal(bytes, responseBody)
+		}
+	}
```
christian-kreuzberger-dtx, line 88: I'm always expecting JSON as a response from the calls we do. 
However, I'm wondering whether we should log something if it's not JSON, to make troubleshooting easier.

christian-kreuzberger-dtx: Or, alternatively, handle "responseBody" differently in the case of not being application/json?

christian-kreuzberger-dtx (commented):
Thanks for the PR.
Left some comments, I don't consider any of them as "blocking". Still, would like to get some of these addressed before we merge, if you can.

api/importer/execute/keptn_endpoints.go
christian-kreuzberger-dtx: both, though feel free to provide a default value if the environment variable is not set.

api/handlers/zip_package.go
christian-kreuzberger-dtx: > Not extract any file and process everything by reading the zip file directly
That would be my favorite, but let's not go that route for the Pull Request.

IMO cleaning the path, and then using `strings.HasPrefix` or a similar methodology is sufficient for now. Not sure if this is actually a problem, or if it's just me being paranoid.

api/test/utils/reader.go
christian-kreuzberger-dtx: Yep I got that too, after looking at the test were it was used. But when looking just at this single file, I didn't get what it can be used for (maybe that's the case for many test helpers...)

api/test/data/import/.gitignore
christian-kreuzberger-dtx: oh because we ignore the go binary in the main folder :D Thanks!

mowies:
@pchila Please try to provide a more descriptive PR title so that users will know what's going on when they see it in our changelog :)

api/importer/execute/default_api.go
pchila: I would add a log in case it's declared json from the content-type but we cannot unmarshal properly...
In case content-type is not jsonm  I am not sure how you would like to represent the body bytes... As a simple string ?

api/handlers/zip_package.go
pchila: @christian-kreuzberger-dtx  @Raffy23  Please have another look at the latest fixup

Raffy23: I think the current implementation is sufficient to prevent most path traversal attacks, but I haven't looked too much into it.

api/importer/execute/default_api.go
pchila: @Raffy23 @christian-kreuzberger-dtx  have a look at the latest fixup ;)

Raffy23 (dismissed):
LGTM

christian-kreuzberger-dtx (dismissed):
LGTM

Raffy23 (approved):

christian-kreuzberger-dtx (approved):
LGTM

christian-kreuzberger-dtx (approved):
LGTM

PR #8316, opened 2022-07-06
feat(installer): Introduce flags to enable / disable Keptn services

### LINKED ISSUE #8075: Enable / Disable all services by Helm chart
Ability to enable / disable any services in the control plane chart.

For example what if I don't want the statistics-service just for a temporary demo? I shouldn't have to use kustomise for this.

@mowies ref our discussion on #6042

warber:
closes #8075

### DIFF
--- a/installer/manifests/keptn/templates/lighthouse-service.yaml
+++ b/installer/manifests/keptn/templates/lighthouse-service.yaml
@@ -1,5 +1,4 @@
 {{- if .Values.approvalService.enabled }}
----
 # lighthouse-service
 apiVersion: apps/v1
 kind: Deployment
@@ -85,5 +84,4 @@ spec:
       protocol: TCP
   selector: {{- include "keptn.common.labels.selectorLabels" . | nindent 4 }}
     app.kubernetes.io/name: lighthouse-service
-
 {{- end}}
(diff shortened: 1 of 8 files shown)


installer/manifests/keptn/templates/lighthouse-service.yaml
```diff
@@ -1,4 +1,4 @@
----
+{{- if .Values.approvalService.enabled }}
```
mowies, line 1:
```suggestion
{{- if .Values.lighthouseService.enabled }}
```

installer/manifests/keptn/templates/remediation-service.yaml
```diff
@@ -77,4 +77,4 @@ spec:
       protocol: TCP
   selector: {{- include "keptn.common.labels.selectorLabels" . | nindent 4 }}
     app.kubernetes.io/name: remediation-service
-
+{{- end}}
```
mowies, line 80:
```suggestion
{{- end }}
```

installer/manifests/keptn/templates/rbac.yaml
```diff
@@ -1,4 +1,3 @@
----
```
mowies, line 1: Please also make sure that the corresponding lighthouse-svc service account and rolebinding is not created when the service is disabled.

thisthat (commented):
Maybe we can also skip the creation of the service accounts if the svc is not enabled

mowies (approved):
LGTM!

warber:
@thisthat  done as part of the review commits :+1:

PR #8321, opened 2022-07-06
ci: Security Scans

### LINKED ISSUE #7951: Add security tools to the pipeline
To enhance security in Keptn, we could introduce a few tools that check for security best practices on a pipeline level.

Ideas:
- introduce tools in pipelines that are run manually
- introduce tools in pipelines that are run on a weekly basis and generate github issues from the findings/reports
- introduce tools that prevent PR merges
- In addition to security tools, also kubernetes/yaml best practice tools could be introduced

### Security
- [kubescape](https://github.com/armosec/kubescape)
- [kics](https://github.com/Checkmarx/kics)

### Best practices
- [kubeconform](https://github.com/yannh/kubeconform)
- [kubeval](https://github.com/instrumenta/kubeval)

mowies:
### This PR
- Adds a new pipeline to run security scans with the following tools:
  - KICS
  - kubescape
  - kubeconform
- the pipeline runs scheduled on mondays and can also be triggered manually

Fixes #7951

### DIFF
--- a/.github/kics-config.yml
+++ b/.github/kics-config.yml
@@ -1,4 +1,5 @@
 exclude-queries:
+  # query IDs can be found here: https://docs.kics.io/latest/queries/all-queries/
   - b9c83569-459b-4110-8f79-6305aa33cb37 # Using Kubernetes Native Secret Management
   - 48a5beba-e4c0-4584-a2aa-e6894e4cf424 # Pod or Container Without ResourceQuota
   - 8b36775e-183d-4d46-b0f7-96a6f34a723f # Missing AppArmor Profile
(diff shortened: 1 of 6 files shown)


.github/kics-config.yml
```diff
@@ -0,0 +1,37 @@
+exclude-queries:
+  # query IDs can be found here: https://docs.kics.io/latest/queries/all-queries/
```
thisthat, line 2: I would suggest for future self when revisiting this pipeline to add in the comment also the reason why we exclude the specific rule

.github/workflows/security-scans.yml
```diff
@@ -0,0 +1,117 @@
...
+        run: |
+          VERSION=$(cat './VERSION.txt')
+          cd ./installer/manifests/keptn
+          helm repo add bitnami https://charts.bitnami.com/bitnami
+          helm repo add nats https://nats-io.github.io/k8s/helm/charts/
+          helm dependency update ./
+          helm dependency build ./
+          # replace "appVersion: latest" with "appVersion: $VERSION" in all Chart.yaml files
+          find . -name Chart.yaml -exec sed -i -- "s/appVersion: latest/appVersion: ${VERSION}/g" {} \;
+          helm template ./ -n keptn --name-template keptn --version 0.17.0 > ${{ env.RENDERED_CHART_FILENAME }}
```
thisthat, line 37: would it make sense to get the version from the latest semver tag?

.github/kics-config.yml
mowies: i will add a comment stating why those queries are excluded

.github/workflows/security-scans.yml
mowies: yes, i will fix that :)

thisthat (approved):

warber (approved):

PR #8326, opened 2022-07-07
fix(installer): Fix Nginx not starting when statistics service is disabled

warber:
PR to fix problems with having statistics service disabled via the helm chart

### DIFF
--- a/installer/manifests/keptn/values.yaml
+++ b/installer/manifests/keptn/values.yaml
@@ -365,7 +365,7 @@ lighthouseService:
       cpu: "200m"
 
 statisticsService:
-  enabled: false
+  enabled: true
   image:
     repository: docker.io/keptn/statistics-service
     tag: ""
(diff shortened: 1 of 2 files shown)


installer/manifests/keptn/values.yaml
```diff
@@ -365,7 +365,7 @@ lighthouseService:
       cpu: "200m"
 
 statisticsService:
-  enabled: false
+  enabled: true
```
christian-kreuzberger-dtx, line 368: Does it make sense to have it enabled by default? Or was this an accidental change :D

warber: Actually, having it set to fals was an accidental change, which caused us to discover the problem with nginx :D :dagger:

warber: but I agree, we should think about whether false should be the default, actually....
@thisthat ?

RealAnna (approved):

warber:
This will fix the master branch again.
@christian-kreuzberger-dtx we'll discuss the option of disabling statistics service per default and do it in a follow up. I think everyone kind of agrees to that proposal...

PR #8332, opened 2022-07-07
fix(bridge): Fix error on viewing service deployment

### LINKED ISSUE #8331: Bridge: Service screen does not show deployment
There might be the case that an `evaluation.finished` event is the parent/trigger of another event. If that's the case it is not possible to select an deployment, because an error is thrown.
image

image

Kirdock:
Fixes #8331


heinzburgstaller (dismissed):

heinzburgstaller:
@Kirdock should we fix the code smell with this PR?

Kirdock:
> @Kirdock should we fix the code smell with this PR?

yes, I'm already on it

PR #8343, opened 2022-07-07
chore:  Added new component test in remediation service

### LINKED ISSUE #8026: Core :: Replace selfhealing integration test with unit and component test
Remediation service uses SDK, we can get rid of [this integration test ](https://github.com/keptn/keptn/blob/master/test/go-tests/test_selfhealing.go) and substitute it with a component test

DoD

- [ ] Remediation service integration test removed
- [ ] All checks translated into component tests using fake keptn

RealAnna:
## This PR

- adds new test to remediation service
- deletes self healing integration test

### Related Issues

Fixes #8026


remediation-service/handler/get_action_event_handler_test.go
```diff
@@ -24,6 +25,35 @@ func newGetActionTriggeredEvent(filename string) models.KeptnContextExtendedCE {
 	return event
 }
 
+// This test reproduces integration test self-healing until line 176 :
+// it triggers a remediation that fails because no remediation.yaml is available yet
+// remediation-service finished with result fail
+// message is "Could not get remediation.yaml file ..."
+// require.Equal(t, keptnv2.StatusErrored, finishedEventData.Status)
+// require.Equal(t, keptnv2.ResultFailed, finishedEventData.Result)
```
warber, line 33:
```suggestion
// Test_Fail_Missing_YAML tests triggering a remediation that fails because no "remediation.yam" is available yet
```

warber: I think that's enough :D

remediation-service/handler/get_action_event_handler_test.go
```diff
@@ -24,6 +25,35 @@ func newGetActionTriggeredEvent(filename string) models.KeptnContextExtendedCE {
...
+	fakeKeptn.AssertSentEvent(t, 1, func(ce models.KeptnContextExtendedCE) bool {
+		getActionFinishedData := keptnv2.GetActionFinishedEventData{}
+		ce.DataAs(&getActionFinishedData)
+		return strings.Contains(getActionFinishedData.Message, "Could not get remediation.yaml")
+	})
+
+}
+
+// this test already checks for a passing path and for retrieving an action to set EnablePromotion to value off
+// component test should not care about action.triggered event since this depends on unleash service
```
bacherfl, line 56: i think we can remove the mentioning of the unleash-service - otherwise one might think that the remediation-service only works in combination with the unleash-service, while it should not care at all which service will then handle the action.triggered event

PR #8348, opened 2022-07-08
feat(api): Add create-secret api action to import endpoint

Raffy23:
## This PR

- adds the `create-secret` API action to the import endpoint

### DIFF
--- a/api/importer/execute/default_api_test.go
+++ b/api/importer/execute/default_api_test.go
@@ -663,6 +663,8 @@ func Test_noRenderRequestFactory_CreateRequest(t *testing.T) {
 					if tt.errorContains != "" {
 						assert.ErrorContains(t, err, tt.errorContains)
 					}
+				} else {
+					assert.NoError(t, err)
 				}
 			},
 		)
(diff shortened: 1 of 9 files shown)


api/importer/execute/default_api_test.go
```diff
@@ -568,3 +568,103 @@ func Test_defaultEndpointHandler_ExecuteAPI(t *testing.T) {
...
+					path:       tt.fields.path,
+				}
+				got, err := rf.CreateRequest(tt.args.tCtx, tt.args.host, tt.args.body)
+				assert.Equalf(t, tt.want, got, "CreateRequest(%v, %v, %v)", tt.args.tCtx, tt.args.host, tt.args.body)
+				if tt.wantError {
+					assert.Error(t, err)
+					if tt.errorContains != "" {
+						assert.ErrorContains(t, err, tt.errorContains)
+					}
+				}
```
pchila, line 666: please add an assertion in case we don't want errors:

```suggestion
				} else {
				    assert.NoError(t, err)
				}
```

PR #8366, opened 2022-07-09
test: use `T.Setenv` to set env vars in tests

Juneezee:
## This PR

This PR replaces `os.Setenv` with `t.Setenv`. Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

### Notes

Reference: https://pkg.go.dev/testing#T.Setenv

### DIFF
--- a/api/handlers/event_test.go
+++ b/api/handlers/event_test.go
@@ -2,6 +2,11 @@ package handlers
 
 import (
 	"errors"
+	"io"
+	"net/http"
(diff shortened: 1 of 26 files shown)


warber (approved):
:+1: nice, lgtm :)

PR #8368, opened 2022-07-11
test: Remove unneeded executions of full delivery sequences

### LINKED ISSUE #6830: Core: Clean up integration tests
Currently, integration tests retrieve or store some info in local folders (continuous delivery, graceful shutdown, and backup-restore). To improve **running the tests locally** it would be good to:

- **reduce or completely remove the usage of folders in the test path** (for instance favoring temp dirs and downloading from git repo) 
- cleanup test suite: it  would be nice to have each test at least visually separated based on type, most importantly we should **doublecheck that all tests are isolated** and do not interfere with each other

- since we are there we could also improve logging 

Timebox this to 2 day(s).

### LINKED ISSUE #8028: Core :: Replace podtato-service delivery with a simpler sequence in integration tests
We should substitute podato delivery sequence with a simpler sequence to shorter integration test times

- [ ] [backuprestore](https://github.com/keptn/keptn/blob/master/test/go-tests/test_backuprestore.go) ( Also make sure to remove PV from the resource-service part of the test)
- [ ] [sshPublickeyauth ](https://github.com/keptn/keptn/blob/master/test/go-tests/test_sshpublickeyauth.go)
- [ ] [proxyauth](https://github.com/keptn/keptn/blob/master/test/go-tests/test_proxyauth.go)

Make sure to extend the assertions to properly cover the "restore process" in the database

bacherfl:
This PR replaces unneeded executions of full delivery sequences with shorter sequences to save some time during the integration tests. Also, this PR makes use of the `ioutil.TempDir()` function for creating temporary directories that should be deleted after a test execution

Closes #8028
Closes #6830

Integration test run: https://github.com/keptn/keptn/runs/7315912461?check_suite_focus=true


RealAnna (approved):
LGTM

PR #8375, opened 2022-07-11
feat(api): Upload resources from import package

pchila:
## This PR

- Adds support for uploading resources to Keptn projects at stage/service level during package import. If no stage is specified, the resource will be uploaded on all stages

### How to test

This can be tested by uploading the attached import package using the import endpoint. As an outcome the service `test-service` will be created and a sample webhook config will be uploaded twice with different resource URI both at stage and service level
[sample-package.zip](https://github.com/keptn/keptn/files/9082090/sample-package.zip)


api/importer/execute/resource.go
```diff
@@ -0,0 +1,90 @@
...
+
+	request.Header.Set("Content-Type", "application/json")
+	response, err := p.doer.Do(request)
+
+	if err != nil {
+		return nil, fmt.Errorf("error performing add resource request: %w", err)
+	}
+
+	responseBody := new(any)
+	if response.Header.Get("Content-Type") == "application/json" {
```
Raffy23, line 61: Checking the header should be probably done by a `HasPrefix` in case the service adds `; charset=utf8` in the future

pchila: Reworked the code for parsing the response body (I also aligned it a bit more to the processing we do for API processing)

Raffy23 (approved):
LGTM

christian-kreuzberger-dtx (approved):
LGTM

PR #8379, opened 2022-07-11
chore(bridge): Generalization of showing a running sequence

### LINKED ISSUE #7854: Generalization of showing a running sequence
### Use case

As a user, I want to see a sequence that is currently running on my deployed artifact. 

For example: When a remediation sequence is running, we currently display it as follows: 
image

Instead of focusing on `remediation` only (which is very opinionated), it helps to show any sequence that is executed on an already deployed artifact. Therefore, a replacement of the **red icon** with a **loading icon** can solve this request:

image

### Definition of Done
- [x] Any kind of sequence that is happening after a deployment is finished is displayed on the service screen
- [x] Any kind of sequence that is currently running (and not a deployment) is visualized using a loading icon
- [x] Adapt tests

Kirdock:
Closes #7854 
image


ermin-muratovic (approved):
lgtm

PR #8400, opened 2022-07-12
feat(shipyard-controller): Introduce Debug-UI for shipyard-controller

asamonik:
## This PR
- adds a debug API with a simple debug-ui on port 9090


shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,200 @@
...
+	payload.NextPageKey = paginationInfo.NewNextPageKey
+	c.JSON(http.StatusOK, payload)
+}
+
+// GetAllSequencesForProject godoc
+// @Summary      Get all sequences for specific project
+// @Description  Get the all the sequences which are present in a sequence
+// @Tags         Sequence
+// @Param        project              path      string                    true "The name of the project"
+// @Success      200                  {object}  []models.SequenceState    "ok"
```
odubajDT, line 75: pls add 400 and 404 codes checks

shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,200 @@
...
+// @Failure      500                  {object}  models.Error              "Internal error"
+// @Router       /debug/project/{project} [get]
+func (dh *DebugHandler) GetAllSequencesForProject(c *gin.Context) {
+	params := &models.GetProjectParams{}
+	projectName := c.Param("project")
+	sequences, err := dh.DebugManager.GetAllSequencesForProject(projectName)
+
+	if err != nil {
```
odubajDT, line 83: if project == nil && errors.Is(err, ErrProjectNotFound) {
			SetNotFoundErrorResponse(c, fmt.Sprintf(ProjectNotFoundMsg, projectName))
			return
		}

odubajDT: an example how to check for this

shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,200 @@
...
+// @Success      200                  {object}  []models.KeptnContextExtendedCE    "ok"
+// @Failure      404                  {object}  models.Error                       "not found"
+// @Failure      500                  {object}  models.Error                       "Internal error"
+// @Router       /debug/project/{project}/shkeptncontext/{shkeptncontext}/event [get]
+func (dh *DebugHandler) GetAllEvents(c *gin.Context) {
+	params := &models.GetProjectParams{}
+	shkeptncontext := c.Param("shkeptncontext")
+	projectName := c.Param("project")
+
+	events, err := dh.DebugManager.GetAllEvents(projectName, shkeptncontext)
```
odubajDT, line 147: pls check for 404 and 400 codes and then fallback to 500

shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,200 @@
...
+	projectName := c.Param("project")
+
+	event, err := dh.DebugManager.GetEventByID(projectName, shkeptncontext, eventId)
+
+	if err != nil {
+		SetInternalServerErrorResponse(c, fmt.Sprintf(InvalidRequestFormatMsg, err.Error()))
+		return
+	}
+
+	if &event == nil {
```
odubajDT, line 194: pls check for other errors before fallback to 500

shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,200 @@
...
+// @Tags         Project
+// @Success      200                  {object}  []apimodels.ExpandedProject     "ok"
+// @Failure      500                  {object}  models.Error                    "Internal error"
+// @Router       /debug/project [get]
+func (dh *DebugHandler) GetAllProjects(c *gin.Context) {
+	params := &models.GetProjectParams{}
+	projects, err := dh.DebugManager.GetAllProjects()
+
+	if err != nil {
+		SetInternalServerErrorResponse(c, fmt.Sprintf(InvalidRequestFormatMsg, err.Error()))
```
odubajDT, line 44: pls don't use InvalidRequestFormatMsg when setting internal server error

odubajDT: https://github.com/keptn/keptn/blob/master/shipyard-controller/handler/error.go

shipyard-controller/controller/debugcontroller.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func (controller DebugController) Inject(apiGroup *gin.RouterGroup) {
+	apiGroup.Static("/ui", "./debug-ui")
+
+	apiGroup.GET("/", func(c *gin.Context) {
+		c.Redirect(302, "/ui")
+	})
+
+	seq := apiGroup.Group("/debug")
```
thisthat, line 23: Since we are retrieving sequence info, I would expect this name
```suggestion
	seq := apiGroup.Group("/sequence")
```

shipyard-controller/db/mongodb_event_repo.go
```diff
@@ -76,6 +76,24 @@ func (mdbrepo *MongoDBEventsRepo) GetEvents(project string, filter common.EventF
...
+	}
+
+	defer cancel()
+
+	cur := collection.FindOne(ctx, getSearchOptions(filter))
+
+	var event apimodels.KeptnContextExtendedCE
+	cur.Decode(&event)
+
+	return &event, cur.Err()
```
thisthat, line 94: NIT: I'd keep it consistent with the rest of the methods
```suggestion
	return event, cur.Err()
```

shipyard-controller/db/mongodb_event_repo.go
```diff
@@ -76,6 +76,24 @@ func (mdbrepo *MongoDBEventsRepo) GetEvents(project string, filter common.EventF
...
+
+	if err != nil {
+		return event, err
+	}
+
+	defer cancel()
+
+	cur := collection.FindOne(ctx, getSearchOptions(filter))
+
+	cur.Decode(&event)
```
bacherfl, line 92: please add error handling here

shipyard-controller/db/mongodb_event_repo.go
```diff
@@ -76,6 +76,24 @@ func (mdbrepo *MongoDBEventsRepo) GetEvents(project string, filter common.EventF
...
+	collection, ctx, cancel, err := mdbrepo.getEventsCollection(projectName, status...)
+	var event apimodels.KeptnContextExtendedCE
+
+	if err != nil {
+		return event, err
+	}
+
+	defer cancel()
+
+	cur := collection.FindOne(ctx, getSearchOptions(filter))
```
bacherfl, line 90: please check for `cur.Err()` right after calling the function and add error handling - note that `ErrNoDocuments` should be handled differently than other kinds of errors - For this one we can return `nil, ErrNoEventFound` - for all other errors `nil, fmt.Errorf("could not retrieve event: %w", err)

shipyard-controller/db/mongodb_state_repo.go
```diff
@@ -113,6 +114,27 @@ func (mdbrepo *MongoDBStateRepo) FindSequenceStates(filter models.StateFilter) (
...
+	err := mdbrepo.DBConnection.EnsureDBConnection()
+	if err != nil {
+		return nil, err
+	}
+	collection := mdbrepo.DBConnection.Client.Database(getDatabaseName()).Collection(filter.Project + taskSequenceStateCollectionSuffix)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	cur := collection.FindOne(ctx, mdbrepo.getSearchOptions(filter))
```
bacherfl, line 130: same comments with regards to error handling as in `mongodb_event_repo.go`

shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	payload.TotalCount = float64(totalCount)
+	payload.NextPageKey = paginationInfo.NewNextPageKey
+	c.JSON(http.StatusOK, payload)
+}
+
+// GetAllSequencesForProject godoc
+// @Summary      Get all sequences for specific project
+// @Description  Get the all the sequences which are present in a sequence
```
bacherfl, line 80: -> Get all the sequences which are present in a project

shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,255 @@
...
+// @Tags         Sequence
+// @Param        project              path      string                    true  "The name of the project"
+// @Param        shkeptncontext       path      string                    true  "The shkeptncontext"
+// @Success      200                  {object}  models.SequenceState      "ok"
+// @Failure      400                  {object}  models.Error              "Bad Request"
+// @Failure      404                  {object}  models.Error              "not found"
+// @Failure      500                  {object}  models.Error              "Internal error"
+// @Router       /debug/project/{project}/shkeptncontext/{shkeptncontext} [get]
+func (dh *DebugHandler) GetSequenceByID(c *gin.Context) {
+	params := &models.GetProjectParams{}
```
bacherfl, line 124: the params seems to not be needed here, as we only have path parameters, which are assigned directly via `c.Param()`

shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,255 @@
...
+
+	shkeptncontext := c.Param("shkeptncontext")
+	projectName := c.Param("project")
+	sequence, err := dh.DebugManager.GetSequenceByID(projectName, shkeptncontext)
+
+	if err := c.ShouldBindQuery(params); err != nil {
```
bacherfl, line 130: this can be removed, as we do not have any query parameters in this handler

shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,255 @@
...
+// @Param        project              path      string                    true "The name of the project"
+// @Success      200                  {object}  []models.SequenceState    "ok"
+// @Failure      400                  {object}  models.Error              "Bad Request"
+// @Failure      404                  {object}  models.Error              "not found"
+// @Failure      500                  {object}  models.Error              "Internal error"
+// @Router       /debug/project/{project} [get]
+func (dh *DebugHandler) GetAllSequencesForProject(c *gin.Context) {
+	params := &models.GetProjectParams{}
+
+	if err := c.ShouldBindQuery(params); err != nil {
```
bacherfl, line 91: this can be removed, as we do not have any query parameters in this handler

shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,255 @@
...
+// @Description  Gets a single event of a project with the given shkeptncontext and event_id
+// @Tags         Sequence
+// @Param        project              path      string                             true  "The name of the project"
+// @Param        shkeptncontext       path      string                             true  "The shkeptncontext"
+// @Param        event_id             path      string                             true  "The Id of the event"
+// @Success      200                  {object}  models.KeptnContextExtendedCE      "ok"
+// @Failure      400                  {object}  models.Error                       "Bad Request"
+// @Failure      404                  {object}  models.Error                       "not found"
+// @Failure      500                  {object}  models.Error                       "Internal error"
+// @Router       /debug/project/{project}/shkeptncontext/{shkeptncontext}/event/{event_id} [get]
```
bacherfl, line 219: please use camelCase for the `event_id` parameter

shipyard-controller/handler/debugmanager.go
```diff
@@ -0,0 +1,76 @@
...
+	apimodels "github.com/keptn/go-utils/pkg/api/models"
+	"github.com/keptn/keptn/shipyard-controller/common"
+	"github.com/keptn/keptn/shipyard-controller/db"
+)
+
+type IDebugManager interface {
+	GetAllProjects() ([]*models.ExpandedProject, error)
+	GetSequenceByID(projectName string, shkeptncontext string) (*models.SequenceState, error)
+	GetAllSequencesForProject(projectName string) (*models.SequenceStates, error)
+	GetAllEvents(projectName string, shkeptncontext string) ([]*models.KeptnContextExtendedCE, error)
```
bacherfl, line 14: why an array of pointers instead of `[]models.KeptnContextExtendedCE`?

shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,221 @@
...
+
+// GetAllProjects godoc
+// @Summary      Get all keptn projects
+// @Description  Get all keptn projects
+// @Tags         Project
+// @Success      200                  {object}  []apimodels.ExpandedProject     "ok"
+// @Failure      400                  {object}  models.Error                    "Bad Request"
+// @Failure      500                  {object}  models.Error                    "Internal error"
+// @Router       /debug/project [get]
+func (dh *DebugHandler) GetAllProjects(c *gin.Context) {
```
bacherfl, line 41: do we actually need this endpoint? this seems to be a copy of https://github.com/keptn/keptn/blob/a66ac61809eaa8167f9f78f8fa4f200eaf344a3e/shipyard-controller/handler/projecthandler.go#L197

shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,221 @@
...
+// GetAllSequencesForProject godoc
+// @Summary      Get all sequences for specific project
+// @Description  Get all the sequences which are present in a project
+// @Tags         Sequence
+// @Param        project              path      string                    true "The name of the project"
+// @Success      200                  {object}  []models.SequenceState    "ok"
+// @Failure      400                  {object}  models.Error              "Bad Request"
+// @Failure      404                  {object}  models.Error              "not found"
+// @Failure      500                  {object}  models.Error              "Internal error"
+// @Router       /debug/project/{project} [get]
```
bacherfl, line 82: is the path `debug/project/{project}` still up to date? If yes, the naming should be adapted to reflect that this is intended for retrieving sequences

shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,221 @@
...
+	projectName := c.Param("project")
+	payload, err := dh.DebugManager.GetAllSequencesForProject(projectName)
+
+	if err != nil {
+		if errors.Is(err, ErrProjectNotFound) {
+			SetNotFoundErrorResponse(c, fmt.Sprintf(ProjectNotFoundMsg, projectName))
+			return
+		}
+
+		SetInternalServerErrorResponse(c, fmt.Sprintf(InvalidRequestFormatMsg, err.Error()))
```
bacherfl, line 93: let's use a different error message here, as the `InvalidRequestFormatMsg` is intended for 400 - Bad Request responses

shipyard-controller/handler/debughandler.go
```diff
@@ -0,0 +1,221 @@
...
+			SetNotFoundErrorResponse(c, fmt.Sprintf(ProjectNotFoundMsg, projectName))
+			return
+		}
+
+		if errors.Is(err, ErrSequenceNotFound) {
+			SetNotFoundErrorResponse(c, fmt.Sprintf(UnableFindSequenceMsg, shkeptncontext))
+			return
+		}
+
+		SetInternalServerErrorResponse(c, fmt.Sprintf(InvalidRequestFormatMsg, err.Error()))
```
bacherfl, line 127: let's use a different error message here, as the InvalidRequestFormatMsg is intended for 400 - Bad Request responses

shipyard-controller/main.go
```diff
@@ -293,6 +301,8 @@ func _main(env config.EnvConfig, kubeAPI kubernetes.Interface) {
 		}
 	}()
 
+	debugEngine.Run("localhost:9090")
```
bacherfl, line 304: the call to the `Run()` method is blocking, so everything else after this will not be executed - please move this into a new goroutine

shipyard-controller/handler/debugmanager.go
```diff
@@ -0,0 +1,88 @@
...
+
+	sort.SliceStable(sequences, func(i, j int) bool {
+		return sequences[i].TriggeredAt.After(sequences[j].TriggeredAt)
+	})
+
+	return sequences, err
+}
+
+func (dm *DebugManager) GetAllEvents(projectName string, shkeptncontext string) ([]*apimodels.KeptnContextExtendedCE, error) {
+	events, err := dm.eventRepo.GetEvents(projectName, common.EventFilter{KeptnContext: &shkeptncontext})
```
bacherfl, line 67: please handle the error right after the call to this function

shipyard-controller/handler/debugmanager.go
```diff
@@ -0,0 +1,88 @@
...
+				Project:      projectName,
+				KeptnContext: shkeptncontext,
+			},
+		})
+
+	return sequence, err
+}
+
+func (dm *DebugManager) GetAllSequencesForProject(projectName string) ([]models.SequenceExecution, error) {
+	sequences, err := dm.sequenceExecutionRepo.Get(models.SequenceExecutionFilter{
```
bacherfl, line 51: please handle the error right after the call to this function

bacherfl (dismissed):

shipyard-controller/debug-ui/fetch_sequences.js
```diff
@@ -0,0 +1,87 @@
+function fetchSequences() {
+  const project = document.getElementById("projectSelector").value;
```
thisthat, line 2: I would pass the project as a parameter of the function. This way it is easier to test it and if you change the HTML id you don't need to touch this file.

shipyard-controller/debug-ui/fetch_sequences.js
```diff
@@ -0,0 +1,87 @@
...
+    .then((res) => {
+      return res.json();
+    })
+    .then((response) => {
+      let table = document.getElementById("prodTable");
+
+      const show_finished = document.getElementById("checkboxFinished").checked;
+      const show_triggered =
+        document.getElementById("checkboxTriggered").checked;
+      const show_timedOut = document.getElementById("checkboxTimedOut").checked;
```
thisthat, line 16: Also, these values should be passed as parameters from the outside.

shipyard-controller/debug-ui/fetch_sequences.js
```diff
@@ -0,0 +1,87 @@
...
+  )
+    .then((res) => {
+      return res.json();
+    })
+    .then((response) => {
+      out = "";
+      response.forEach((object) => {
+        out += object.scope.keptnContext + "\n";
+      });
+      alert(out);
```
thisthat, line 85: Please, don't use alerts 😓

thisthat: I'd suggest `console.log` for the moment as an alternative. It would be better to have a dedicated page in the future.

shipyard-controller/debug-ui/fetch_sequences.js
```diff
@@ -0,0 +1,87 @@
...
+
+          if (object.status.state == "triggered") {
+            td_blocking.innerHTML =
+              "<button onclick=\"getBlocking('" +
+              object.scope.keptnContext +
+              "', '" +
+              object.scope.project +
+              "')\">" +
+              "get blocking sequences" +
+              "</button>";
```
thisthat, line 38: Maybe this would be easier with backticks `

shipyard-controller/debug-ui/fetch_sequences.js
```diff
@@ -0,0 +1,87 @@
...
+            object.scope.stage +
+            "</td>" +
+            "<td>" +
+            object.scope.service +
+            "</td>" +
+            '<td><a href="viewevents.html?shkeptncontext=' +
+            object.scope.keptnContext +
+            "&projectname=" +
+            project +
+            '"><button>View Events</button></a></td>';
```
thisthat, line 62: ditto

shipyard-controller/debug-ui/fetch_events.js
```diff
@@ -0,0 +1,75 @@
+fetchEvents();
```
thisthat, line 1: please, align this across files. I would suggest to don't call functions directly from the js files.

shipyard-controller/debug-ui/fetch_events.js
```diff
@@ -0,0 +1,75 @@
...
+
+function fetchEvents() {
+  const urlParams = new URLSearchParams(window.location.search);
+  const context = urlParams.get("shkeptncontext");
+  const projectname = urlParams.get("projectname");
+
+  let list = document.getElementById("tree_container");
```
thisthat, line 8: please, pass these as a parameters

shipyard-controller/debug-ui/fetch_events.js
```diff
@@ -0,0 +1,75 @@
...
+  list.innerHTML = "";
+
+  fetch(
+    "../sequence/project/" + projectname + "/shkeptncontext/" + context + "/event",
```
thisthat, line 12: the `..` makes it harder to reuse the script. Maybe it would be better to introduce a basepath value🤔

shipyard-controller/debug-ui/fetch_events.js
```diff
@@ -0,0 +1,75 @@
...
+    {
+      method: "get",
+    }
+  )
+    .then((res) => {
+      return res.json();
+    })
+    .then((response) => {
+      let rootevent_list = document.getElementById("tree_container");
```
thisthat, line 21: please, pass the html object from outside

shipyard-controller/debug-ui/fetch_events.js
```diff
@@ -0,0 +1,75 @@
...
+      let rootevent_li, taskevent_ul, label;
+
+      response.forEach((object) => {
+        if (object.type.split(".").length == 6) {
```
thisthat, line 25: why the magic number `6`?

shipyard-controller/debug-ui/fetch_events.js
```diff
@@ -0,0 +1,75 @@
...
+          label2.className = "caret";
+          label2.innerHTML = object.type;
+          taskevent_li.appendChild(label2);
+
+          let detail_ul = document.createElement("ul");
+          detail_ul.className = "nested";
+
+          for (let key in object.data) {
+            let li = document.createElement("li");
+            li.innerHTML = key + ": " + object.data[key];
```
thisthat, line 50: please, check if it is only TS or is also available in JS 
```suggestion
            li.innerHTML = `${key} :  ${object.data[key]}`;
```

shipyard-controller/debug-ui/fetch_sequences.js
```diff
@@ -0,0 +1,77 @@
+function fetchSequences(
+  project_name,
+  show_triggered,
+  show_finished,
+  show_timedOut,
+  table
+) {
+  fetch("/sequence/project/" + project_name, {
```
thisthat, line 8: For consistency 
```suggestion
  fetch(`/sequence/project/${project_name}`, {
```

shipyard-controller/debug-ui/fetch_sequences.js
```diff
@@ -0,0 +1,77 @@
...
+    .then((response) => {
+      table.innerHTML =
+        "<tr><th>shkeptncontext</th><th>Service</th><th>Project</th><th>Stage</th><th>Service</th><th>view events</ht><th>getblocking</th></tr>";
+
+      response.sequenceExecutions.forEach((object) => {
+        if (
+          (object.status.state == "triggered" && show_triggered) ||
+          (object.status.state == "finished" && show_finished) ||
+          (object.status.state == "timedOut" && show_timedOut)
+        ) {
```
thisthat, line 23: pls use `===`

shipyard-controller/debug-ui/fetch_sequences.js
```diff
@@ -0,0 +1,77 @@
...
+          let tr = document.createElement("tr");
+          let td_blocking = document.createElement("td");
+
+          if (object.status.state == "triggered") {
```
thisthat, line 27:
```suggestion
          if (object.status.state === "triggered") {
```

shipyard-controller/debug-ui/fetch_sequences.js
```diff
@@ -0,0 +1,77 @@
...
+          tr.appendChild(td_blocking);
+          table.appendChild(tr);
+        }
+      });
+    });
+}
+
+function getBlocking(context, project) {
+  fetch(
+    "/sequence/project/" + project + "/shkeptncontext/" + context + "/blocking",
```
thisthat, line 64: Please use also here the ` format

shipyard-controller/debug-ui/fetch_sequences.js
```diff
@@ -0,0 +1,77 @@
...
+    {
+      method: "get",
+    }
+  )
+    .then((res) => {
+      return res.json();
+    })
+    .then((response) => {
+      response.forEach((blockingSequence) => {
+        console.log(blockingSequence.scope.keptnContext);
```
thisthat, line 74: is `keptnContext` or `scope` always available? Maybe it is worth having a null check here

thisthat (approved):

PR #8402, opened 2022-07-12
test(bridge): Fix flaky UI test in dashboard.spec

Kirdock:
There was a flaky UI test that the projects call was not triggered (`should trigger loadProjects once per dashboard visit` in `dashboard.spec.ts`).

Probably the cause: The navigation to the create-project page and then the navigation back. The navigation-back was triggered too early and therefore canceled the navigation to the create page.


ermin-muratovic (approved):
lgtm

PR #8408, opened 2022-07-13
fix: Zero Downtime test for the webhook-service

RealAnna:
Fix ZD test of webhook back to work

### DIFF
--- a/.github/workflows/zero-downtime-tests.yml
+++ b/.github/workflows/zero-downtime-tests.yml
@@ -12,7 +12,7 @@ on:
       upgradeTo:
         description: 'The helm upgrade command will run with this version (e.g. 0.15.0-dev-PR-7504 or a full address like in the default value )'
         required: true
-        default: 'https://github.com/keptn/helm-charts-dev/raw/1d63e5d3c3f409279a4157d5ad537413abade2e5/packages/keptn-0.16.0-dev.tgz'
+        default: 'https://github.com/keptn/helm-charts-dev/raw/gh-pages/packages/keptn-0.17.0-dev.tgz'
       testUI:
         type: boolean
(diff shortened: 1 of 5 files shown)


.github/workflows/zero-downtime-tests.yml
```diff
@@ -12,7 +12,7 @@ on:
       upgradeTo:
         description: 'The helm upgrade command will run with this version (e.g. 0.15.0-dev-PR-7504 or a full address like in the default value )'
         required: true
-        default: 'https://github.com/keptn/helm-charts-dev/raw/1d63e5d3c3f409279a4157d5ad537413abade2e5/packages/keptn-0.16.0-dev.tgz'
+        default: 'https://github.com/keptn/helm-charts-dev/raw/gh-pages/packages/keptn-0.17.0-dev.tgz'
```
bacherfl, line 15: should we use 0.18.0-dev here as a default instead?

test/go-tests/test_webhook.go
```diff
@@ -490,7 +490,7 @@ func getWebhookYamlWithSubscriptionIDs(t *testing.T, taskTypes []string, project
 	return webhookYamlWithSubscriptionIDs
 }
 
-func deleteFile(t *testing.T, shipyardFilePath string) {
+func DeleteFile(t *testing.T, shipyardFilePath string) {
```
bacherfl, line 493: let's give the `shipyardFilePath` a more generic name here, as the function can also be used for deleting other files.

.github/workflows/zero-downtime-tests.yml
RealAnna: I tought to keep it backwar compatible but sure we can

bacherfl (approved):

PR #8415, opened 2022-07-13
ci: Change stalebot configuration

thisthat:


christian-kreuzberger-dtx (approved):
I think the quotes are only needed in case there is a white space in between, but it should work out this way too!
Also, I think adding known-limitationand epic into exemptLabels makes perfect sense.
LGTM

johannes-b (approved):
lgtm

PR #8432, opened 2022-07-14
chore(bridge): Enable resource-service by default

thisthat:

## This PR

Set the FF true by default since resource-service is the new default.
The removal of it will be done in #8433


renepanzar (approved):
Looks promising, go ahead ;)

renepanzar (approved):
👍

PR #8456, opened 2022-07-18
feat(api): Support simple templating for resource and api tasks in import manifests

pchila:
Signed-off-by: Paolo Chila <paolo.chila@dynatrace.com>

## This PR

- Adds support for simple context definition as a map<string, string> defined at task level
- Support templating both API payloads and resource contents


christian-kreuzberger-dtx (approved):
LGTM
regarding reviewdog yaml.v3 issue, I've asked Keptn Devs for help, but most likely we can ignore it.

PR #8490, opened 2022-07-21
feat(shipyard-controller): Introduce RemoteURL denyList

### LINKED ISSUE #8422: Deny list for Git remote upstream URLs
**What and why?** 
As a Keptn operator, I would like to be able to specify what Git providers must not be used. For example, when adopting Keptn within a company, I would like to prevent e.g. that the Keptn upstream repos are hosted at GitHub, Gitlab, etc. 

## Acceptance Criteria

- [x] A ConfigMap `keptn-git-config` is created with a key `git-remote-url-denylist`
- [x] A Helm value `features.tbd` fills the `git-remote-url-denylist` values, by default is empty
- [x] Each line of the `git-remote-url-denylist` should be handled as a regex. 
- [x] This ConfigMap shall be mounted to the shipyard-controller as a file via Volume.
- [x] When an API call setting/changing the Git remote upstream is performed, shippy reads the ConfigMap via the file and performs a validation.
- [x] When the Git remote upstream matches one of the values from `git-remote-url-denylist`, the API should return an [HTTP 422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) error.

## Technical Details

In the case of the automatic provisioned, the Git URL received is empty. Hence, repositories provided by the Automatic Provisioning service are not validated.

## **Definition of Done**
- [x] Unit Tests
- As an operator of Keptn, I can specify a denylist for Git remote upstream URLs e.g. using a regex during installation time

odubajDT:

## This PR

- adds RemoteURL denyList which allows the user to specify a list of forbidden URLs

### Related Issues

Fixes #8422 

### Integration tests
https://github.com/keptn/keptn/actions/runs/2747358216

### Note:
The way of setting the content of `git-remote-url-denylist` will be possible via `--set=Values.features.gitRemoteURLDenyList=<string>` only if we are considering a single line string (single denied `RemoteURL`). If we need to add multiple lines (multiple denied `RemoteURL`s), we need to create a `values_patch.yaml` file and apply it via `--values` during the install command. More info here https://stackoverflow.com/questions/64682166/how-to-override-a-multi-line-string-with-set-in-helm-install-command

Example of the `values_patch.yaml` file:

```
git:
  remoteURLDenyList: |-
    some
    randomizer
    data
```


installer/manifests/keptn/templates/shipyard-controller.yaml
```diff
@@ -23,6 +23,9 @@ spec:
       serviceAccountName: keptn-shipyard-controller
       containers:
         - name: shipyard-controller
+          volumeMounts:
+            - name: keptn-git-config-volume
+              mountPath: /keptn-git-config
```
mowies, line 28: please add `readOnly: true` here if you don't need to write to this volume

shipyard-controller/common/filereader.go
```diff
@@ -0,0 +1,61 @@
...
+func NewFileReader() *fileReader {
+	return &fileReader{
+		FileSystem: fileOpener{},
+	}
+}
+
+func (d *fileReader) GetLines(filePath string) []string {
+	gitConfigFile, err := d.FileSystem.Open(filePath)
+	if err != nil {
+		fmt.Printf(err.Error())
```
thisthat, line 38: please, remove this line

shipyard-controller/common/filereader.go
```diff
@@ -0,0 +1,61 @@
...
+		logrus.Errorf("cannot open %s: %s", filePath, err.Error())
```
thisthat, line 39: Log messages should start with a capital letter, error messages with a lower case :)

shipyard-controller/common/filereader_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"testing"
+	"testing/fstest"
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_FileReader(t *testing.T) {
+	denyListTestFileName := "test-file"
+	tests := []struct {
+		provider FileReader
```
thisthat, line 13: could you please provide a `name string`? This way the test matrix can have a title for each entry which makes it easier to understand the purpose of the entry

shipyard-controller/common/filereader_test.go
```diff
@@ -0,0 +1,105 @@
...
+			},
+			result: []string{},
+		},
+		{
+			provider: &fileReader{
+				FileSystem: fstest.MapFS{},
+			},
+			result: []string{},
+		},
+	}
```
thisthat, line 62: would it be easy to add a test for testing a file error?

shipyard-controller/handler/projecthandler.go
```diff
@@ -303,6 +305,15 @@ func (ph *ProjectHandler) CreateProject(c *gin.Context) {
 			},
 			User: provisioningData.GitUser,
 		}
+	} else if automaticProvisioningURL == "" {
```
thisthat, line 308: why not just an `else`?

shipyard-controller/handler/projecthandler.go
```diff
@@ -303,6 +305,15 @@ func (ph *ProjectHandler) CreateProject(c *gin.Context) {
...
+		if params.GitCredentials == nil {
+			SetBadRequestErrorResponse(c, fmt.Sprintf(common.InvalidRequestFormatMsg, "cannot create project without git credentials"))
+			return
+		}
```
thisthat, line 312: I would not add this extra assumption. If the credentials are necessary, then the resource-svc will fail

shipyard-controller/handler/projecthandler.go
```diff
@@ -372,6 +383,18 @@ func (ph *ProjectHandler) UpdateProject(c *gin.Context) {
 		return
 	}
 
+	automaticProvisioningURL := ph.Env.AutomaticProvisioningURL
+	if automaticProvisioningURL == "" {
+		if params.GitCredentials == nil {
+			SetBadRequestErrorResponse(c, fmt.Sprintf(common.InvalidRequestFormatMsg, "cannot update project without git credentials"))
+			return
+		}
```
thisthat, line 391: Similar here. Plus, this is a behavior change

shipyard-controller/handler/projecthandler_test.go
```diff
@@ -66,6 +67,11 @@ func TestGetAllProjects(t *testing.T) {
 				EventSender:           &fake.IEventSenderMock{},
 				EnvConfig:             config.EnvConfig{ProjectNameMaxSize: 200},
 				RepositoryProvisioner: &fake.IRepositoryProvisionerMock{},
+				RemoteURLValidator: fake.RequestValidatorMock{
+					ValidateFunc: func(url string) error {
+						return nil
+					},
+				},
```
thisthat, line 74: since it is repeated over and over, would it make sense to have it defined only once?

shipyard-controller/handler/remoteurl_validator_test.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"testing"
+
+	"github.com/keptn/keptn/shipyard-controller/common"
+	common_mock "github.com/keptn/keptn/shipyard-controller/common/fake"
+)
+
+func Test_RemoteURLValidator(t *testing.T) {
+	tests := []struct {
+		url        string
```
thisthat, line 12: Also here I would suggest having a `name string` property to have the purpose of the test explicit

shipyard-controller/common/filereader_test.go
odubajDT: we are already testing it, it is the last test

shipyard-controller/handler/projecthandler_test.go
odubajDT: we can, but we define everything like this, it's not a deal, I will rewrite it

shipyard-controller/handler/projecthandler.go
odubajDT: for discussion for tomorrow

installer/manifests/keptn/values.yaml
```diff
@@ -59,6 +59,7 @@ features:
   oauth:
     enabled: false
     prefix: "keptn:"
+  gitRemoteURLDenyList: ""
```
bacherfl, line 62: should we rather add a `git` object here to group together other potential git related config values?

shipyard-controller/common/filereader.go
```diff
@@ -0,0 +1,59 @@
...
+		lines = append(lines, scanner.Text())
+	}
+
+	return removeEmptyStrings(lines)
+}
+
+func removeEmptyStrings(s []string) []string {
+	r := []string{}
+	for _, str := range s {
+		if str != "" {
```
bacherfl, line 54: what happens if we have a string with multiple whitespaces?

installer/manifests/keptn/values.yaml
odubajDT: applied, thx

shipyard-controller/handler/projecthandler.go
```diff
@@ -126,10 +131,14 @@ func (p ProjectValidator) validateUpdateProjectParams(updateProjectParams *model
 		}
 	}
 
-	if updateProjectParams.GitCredentials == nil {
+	if p.AutomaticProvisioningURL != "" && updateProjectParams.GitCredentials == nil {
 		return nil
 	}
 
+	if updateProjectParams.GitCredentials == nil {
+		return fmt.Errorf("gitCredentials cannot be empty")
```
bacherfl, line 139: here we can rather just use `errors.New()`

shipyard-controller/handler/projecthandler.go
```diff
@@ -68,10 +69,14 @@ func (p ProjectValidator) validateCreateProjectParams(createProjectParams *model
 		return fmt.Errorf("provided shipyard file is not valid: %s", err.Error())
 	}
 
-	if createProjectParams.GitCredentials == nil {
+	if p.AutomaticProvisioningURL != "" && createProjectParams.GitCredentials == nil {
 		return nil
 	}
 
+	if createProjectParams.GitCredentials == nil {
+		return fmt.Errorf("gitCredentials cannot be empty")
```
bacherfl, line 77: here we can rather just use errors.New()

shipyard-controller/common/filereader.go
odubajDT: fixed, thx

bacherfl (approved):

PR #8492, opened 2022-07-21
fix: log.Fatal will call os.Exit, use log.Println instead

Abirdcfly:
Signed-off-by: Abirdcfly <fp544037857@gmail.com>

## This PR

- use `log.Println` instead of `log.Fatal` to avoid immediate exit.
 

### Related Issues

Fixes #

### Notes

`log.Fatal` source code see https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/log/log.go;l=364-366;drc=2580d0e08d5e9f979b943758d3c49877fb2324cb

[Description](https://pkg.go.dev/log#Fatal) of `log.Fatal`: Fatal is equivalent to Print() followed by a call to os.Exit(1).

### DIFF
--- a/helm-service/main.go
+++ b/helm-service/main.go
@@ -3,22 +3,20 @@ package main
 import (
 	"context"
 	"fmt"
+	"log"
+	"net/url"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
 
 	"github.com/keptn/keptn/helm-service/controller"
 	"github.com/keptn/keptn/helm-service/pkg/common"
 	"github.com/keptn/keptn/helm-service/pkg/configurationchanger"
 	"github.com/keptn/keptn/helm-service/pkg/helm"
 
-	"net/url"
-	"os/signal"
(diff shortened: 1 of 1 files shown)


thisthat (approved):

helm-service/main.go
```diff
@@ -31,7 +29,7 @@ import (
 	keptncommon "github.com/keptn/go-utils/pkg/lib/keptn"
 	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
 
-	utils "github.com/keptn/go-utils/pkg/api/utils"
+	"github.com/keptn/go-utils/pkg/api/utils"
```
thisthat, line 32: please, revert this change

thisthat (dismissed):

helm-service/main.go
Abirdcfly: revert it.

odubajDT (approved):

Issue #6163, opened 2021-11-24
Replace highcharts line chart with d3


Kirdock:
As part of #6001 the highcharts line chart has to be replaced by a line chart made with d3.

## Acceptance Criteria:
- Possibility to show data as a line or block. The score should be shown as block and line in the chart, and all SLIs as a line.
- Tooltip on hover, showing the score and SLIs (display name and value)
- Filter out some SLIs: It should be possible to show/hide SLIs or the score

## Definition of Done
- [ ] Highcharts line chart is replaced by d3
- [ ] UI tests
- [ ] Use d3 rendering only when FF `sh.keptn.d3` is enabled

PR #8499, opened 2022-07-22
refactor(bridge): Rename D3 feature flag

heinzburgstaller:
## This PR
- Rename variable from 'D3_HEATMAP_ENABLED' to 'D3_ENABLED'
- Adopt usages

### Related Issues
Relates to #6163 

Signed-off-by: Heinz Burgstaller <heinz.burgstaller@dynatrace.com>

### DIFF
--- a/installer/manifests/keptn/templates/core.yaml
+++ b/installer/manifests/keptn/templates/core.yaml
@@ -170,7 +170,7 @@ spec:
             - name: AUTH_MSG
               value: {{ .Values.bridge.authMsg | default "" | quote }}
             - name: D3_ENABLED
-              value: {{ or .Values.bridge.d3heatmap.enabled ((.Values.bridge).d3).enabled | default false | quote }}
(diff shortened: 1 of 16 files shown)


installer/manifests/keptn/templates/core.yaml
```diff
@@ -169,8 +169,8 @@ spec:
               value: {{ ((.Values.features).automaticProvisioning).message | default "" | quote }}
             - name: AUTH_MSG
               value: {{ .Values.bridge.authMsg | default "" | quote }}
-            - name: D3_HEATMAP_ENABLED
-              value: {{ .Values.bridge.d3heatmap.enabled | default false | quote }}
+            - name: D3_ENABLED
+              value: {{ .Values.bridge.d3.enabled | default false | quote }}
```
thisthat, line 173: I would suggest trying to be backward compatible if possible. So if a user had it enabled in the past it still works :) 
```suggestion
              value: {{ or .Values.bridge.d3heatmap.enabled ((.Values.bridge).d3).enabled | default false | quote }}
```

ermin-muratovic (approved):
lgtm

PR #8505, opened 2022-07-25
ci: Add back helm repo add commands in pipelines

mowies:
### This PR
- adds back some `helm repo add` commands to make helm happy in all our builds
- adds some better logging in the build helm charts script for better readability
- adds a depenedncy in the CI build between docker build and helm chart upload so that only when there is a docker build, the helm chart will be uploaded

### DIFF
--- a/.github/workflows/security-scans.yml
+++ b/.github/workflows/security-scans.yml
@@ -28,6 +28,8 @@ jobs:
         run: |
           VERSION=$(cat './VERSION.txt')
           cd ./installer/manifests/keptn
+          helm repo add bitnami https://charts.bitnami.com/bitnami
+          helm repo add nats https://nats-io.github.io/k8s/helm/charts/
           helm dependency build ./
           # replace "appVersion: latest" with "appVersion: $VERSION" in all Chart.yaml files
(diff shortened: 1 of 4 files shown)


gh-actions-scripts/build-helm-charts.sh
```diff
@@ -36,6 +36,9 @@ find . -name values.yaml -exec sed -i -- "s/docker.io\/keptn\//docker.io\/${DOCK
 
 mkdir keptn-charts/
 
+helm repo add bitnami https://charts.bitnami.com/bitnami
```
odubajDT, line 39: why adding these repos if dependencies are already baked to charts?

.github/workflows/security-scans.yml
```diff
@@ -28,6 +28,8 @@ jobs:
         run: |
           VERSION=$(cat './VERSION.txt')
           cd ./installer/manifests/keptn
+          helm repo add bitnami https://charts.bitnami.com/bitnami
```
odubajDT, line 31: same here

gh-actions-scripts/build-helm-charts.sh
mowies: because we call `helm dependency build` later on. and that command fails if it can't pull the deps

odubajDT (approved):

PR #8508, opened 2022-07-25
feat(api): Importer manifest validation

TannerGabriel:
## This PR

- Adds validation to importer manifest

### How to test

Unit tests are included. However, a full test can be performed by creating a Keptn project and importing the `api/test/data/import/sample-package)` with the parameters you want to test.


api/importer/package_processor_test.go
```diff
@@ -1153,3 +1157,207 @@ func TestImportPackageProcessor_Process_FullManifestRendering(t *testing.T) {
...
+	tests := []struct {
+		name               string
+		manifest           model.ImportManifest
+		expectedErrMessage string
+		expectError        bool
+	}{
+		{
+			name:               "valid id",
+			manifest:           manifests[0],
+			expectedErrMessage: "<nil>",
```
pchila, line 1287: To avoid confusion when reading the expectedErrMessage can be left at its zero-value since it's irrelevant in this testcase (expectError is false)
```suggestion
```

api/importer/package_processor.go
```diff
@@ -101,6 +108,57 @@ func (ipp *ImportPackageProcessor) Process(project string, ip ImportPackage) err
...
+	for _, task := range manifest.Tasks {
+		// Check if ID is not empty and formatted properly
+		if task.ID == "" {
+			return fmt.Errorf("task id cannot be empty")
+		} else if !re.MatchString(task.ID) {
+			return fmt.Errorf("task id %s can only consist of alphnumeric characters and underscores", task.ID)
+		}
+
+		// Check if task definition is correct
+		switch task.Type {
```
Raffy23, line 124: I think `task.APITask` and `task.ResourceTask` shouldn't be allowed to be set in one action.

TannerGabriel: Are there any specific reasons for this? Currently, the `ResourceTask` will not be used when calling an API task and vice versa, but it is not restricted not to provide both.

api/importer/package_processor.go
```diff
@@ -262,3 +320,15 @@ func (ipp *ImportPackageProcessor) renderContext(
 	}
 	return renderedContext, nil
 }
+
+func (ipp *ImportPackageProcessor) resourceDoesExits(resource string, ip ImportPackage) error {
+	_, err := ip.GetResource(resource)
```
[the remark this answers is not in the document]
TannerGabriel: Moved the functionality into the ImportPackage. I also switched to using `os.Stat()` since it is sufficient to check if the file exists.

api/importer/package_processor.go
```diff
@@ -101,6 +108,57 @@ func (ipp *ImportPackageProcessor) Process(project string, ip ImportPackage) err
...
+		} else if !re.MatchString(task.ID) {
+			return fmt.Errorf("task id %s can only consist of alphnumeric characters and underscores", task.ID)
+		}
+
+		// Check if task definition is correct
+		switch task.Type {
+		case apiTaskType:
+			if task.APITask != nil {
+				// Check if the action type is supported
+				if !slices.Contains(model.AllActions, task.APITask.Action) {
```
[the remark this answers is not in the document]
TannerGabriel: We could move the constants from `execute` to the `executor` and access them over the injected executor, or do you have another idea for implementing this?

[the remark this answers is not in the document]
pchila: I would expose a method from the executor interface like

```
ActionSupported(actionName string) bool
```
so that this can be checked before manifest execution but also without having to know all the action at processor level.
Furthermore the fact that the constant exists does not guarantee that the implementation is available ;) so the implementor of the interface can also check that the action is actually supported

api/importer/package_processor.go
pchila: It's not restricted but it's a sign of a badly written manifest

api/handlers/zip_package.go
```diff
@@ -37,6 +37,20 @@ func (m *ZippedPackage) GetResource(resourceName string) (io.ReadCloser, error)
 	return file, nil
 }
 
+func (m *ZippedPackage) CheckIfResourceExists(resourceName string) error {
```
pchila, line 40: Is this tested anywhere?

api/handlers/zip_package.go
```diff
@@ -37,6 +37,20 @@ func (m *ZippedPackage) GetResource(resourceName string) (io.ReadCloser, error)
...
+	actualPath := path.Clean(path.Join(m.extractedDir, resourceName))
+	if !strings.HasPrefix(actualPath, m.extractedDir) {
+		return ErrorInvalidResourcePath
+	}
+
+	_, err := os.Stat(actualPath)
```
pchila, line 46: maybe we should check a few more things (like that it's not a directory for example)

api/importer/package_processor.go
```diff
@@ -21,6 +21,7 @@ type Renderer interface {
 type ImportPackage interface {
 	io.Closer
 	GetResource(resourceName string) (io.ReadCloser, error)
+	CheckIfResourceExists(resourceName string) error
```
pchila, line 24: we can try to shorten the signature and not use the absence of error as a value

```
ResourceExists(resourceName string) (bool, error)
```

api/importer/package_processor.go
TannerGabriel: Fine, I added a condition that both cannot be set in the same action.

api/handlers/zip_package.go
TannerGabriel: No, I just added some Unit tests.

api/importer/package_processor.go
TannerGabriel: Renamed it and added the second return value.

Raffy23 (approved):
LGTM

christian-kreuzberger-dtx (approved):
LGTM

PR #8523, opened 2022-07-27
fix(installer): Disable nats cluster due to unreliable behavior

### LINKED ISSUE #8522: Lost/Delayed events when using Nats Cluster with multiple replicas
When testing the draft release of Keptn 0.18, I observed that in some cases events did not get to their subscribers (e.g. the helm service). Restarting the pod of the subscriber (sometimes) fixed the problem, but also not in all cases. Another issue was that sometimes events were received with some delay (ranging from a couple of seconds to minutes). Looking at the logs of the `keptn-nats` StatefulSet I observed a lot of `nats Error trying to connect to route` error messages.

Therefore I suggest to set the `nats.cluster.enabled` option to `false` per default for the upcoming release of 0.18, and take the time to gather more experience with the enabled cluster while preparing the next release

bacherfl:
Closes #8522

This PR sets the `nats.cluster.enabled` option to `false`, due to the reasons described in #8522

### DIFF
--- a/installer/manifests/keptn/values.yaml
+++ b/installer/manifests/keptn/values.yaml
@@ -64,8 +64,8 @@ nats:
   nameOverride: keptn-nats
   fullnameOverride: keptn-nats
   cluster:
-    enabled: true
-    replicas: 3
+    enabled: false
+    replicas: 1
     name: nats
   securityContext:
     runAsNonRoot: true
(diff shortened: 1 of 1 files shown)


mowies (approved):
LGTM!
You could leave the replicas at 3 for later use when we enable this again. They only influence the cluster behaviour but as long as it's disabled, it's always gonna be just 1 replica

mowies (approved):

PR #8524, opened 2022-07-27
fix(bridge): Fix evaluation badge wrapping

### LINKED ISSUE #7013: Refactor evaluation-info badge
At the moment the `dt-tag` of the barista components is used/modified to show the evaluation result. This is more like a misuse of the component and we should write our own one, else we have some unexpected behaviors (e.g. "Show x more" shows up and hides some evaluations).

### LINKED ISSUE #8437: Bridge does a line break on evaluation history when running in quality gates only mode
**Environment**

* Keptn Version (`keptn version`): 0.16.1

**Affected Component**

* [ x ] Bridge

**Describe the bug**
* When using Keptn for quality gates only, a message is displayed on the environment screen: `Use [keptn trigger evaluation](https://keptn.sh/docs/0.16.x/reference/cli/commands/keptn_trigger_evaluation/) to trigger a quality gate evaluation.`

* Unfortunately, this message is causing line breaks on the evaluation result history: 

image

**Expected behaviour**
* The evaluation result history is always shown in a row.

Kirdock:
Fixes #8437 
Closes #7013

### DIFF
--- a/bridge/client/app/_components/ktb-evaluation-badge/ktb-evaluation-badge.component.scss
+++ b/bridge/client/app/_components/ktb-evaluation-badge/ktb-evaluation-badge.component.scss
@@ -1,0 +1,63 @@
+@import 'client/variables';
+@import 'node_modules/@dynatrace/barista-components/core/src/style/variables';
+
+.badge {
+  padding: 2px 8px;
+  min-width: 30px;
+  height: 24px;
+  border-radius: 12px;
+  font-size: 12px;
+  color: black;
+  background-color: $gray-130;
+  text-align: center;
+
+  &.fill {
+    &.error {
+      background-color: $error-color;
+      color: $white;
+    }
+
+    &.warning {
+      background-color: $warning-color;
+      color: $white;
+    }
+
+    &.success {
+      background-color: $cta-color-active;
+      color: $white;
+    }
+  }
+
+  &.highlight {
+    border-color: $approval-color;
+  }
+
+  &.border {
+    background-color: $gray-200;
+    color: black;
+
+    &.success {
+      border: 1px solid $cta-color;
+    }
+
+    &.warning {
+      border: 1px solid $warning-color;
+    }
+
+    &.error {
+      border: 1px solid $error-color;
+    }
+  }
+}
+
+ktb-loading-spinner {
+  height: 20px;
+  width: 20px;
+  vertical-align: bottom;
+}
+
+/* fix incorrect height and width of the overlay */
+::ng-deep .cdk-overlay-connected-position-bounding-box {
+  height: 100% !important;
+  width: 100% !important;
+}
--- a/bridge/client/app/_components/ktb-evaluation-badge/ktb-evaluation-badge.component.scss
+++ b/bridge/client/app/_components/ktb-evaluation-badge/ktb-evaluation-badge.component.scss
@@ -28,24 +28,20 @@
     }
   }
 
-  &.highlight {
-    border-color: $approval-color;
-  }
-
   &.border {
     background-color: $gray-200;
     color: black;
 
-    &.success {
-      border: 1px solid $cta-color;
+    &.error {
+      border: 1px solid $error-color;
     }
 
     &.warning {
       border: 1px solid $warning-color;
     }
 
-    &.error {
-      border: 1px solid $error-color;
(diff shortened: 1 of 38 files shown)


bridge/client/app/_components/ktb-evaluation-badge/ktb-evaluation-badge.component.scss
```diff
@@ -0,0 +1,59 @@
...
+}
+
+ktb-loading-spinner {
+  height: 20px;
+  width: 20px;
+  vertical-align: bottom;
+}
+
+/* fix incorrect height and width of the overlay */
+::ng-deep .cdk-overlay-connected-position-bounding-box {
```
renepanzar, line 56: You might want to prefix this with `:host ::ng-deep ...`, or else this will be a global style applied also to other overlays on the same page.

bridge/client/app/_components/ktb-evaluation-badge/ktb-evaluation-badge.component.ts
```diff
@@ -0,0 +1,65 @@
...
+      this.updateOverlayPositionSubscription.unsubscribe();
+      this.overlayRef = undefined;
+    }
+  }
+
+  public ngOnDestroy(): void {
+    this.unsubscribe$.next();
+    this.unsubscribe$.complete();
+  }
+}
```
renepanzar, line 65: Can we only use one unsubscription method? If we keep the subscription anyway we might as well do it like this:

```suggestion
  public ngOnDestroy(): void {
    this.updateOverlayPositionSubscription.unsubscribe();
  }
```

No `unsubscribe$` observable needed, no `.pipe(takeUntil(this.unsubscribe$))` necessary.

Alternatively you could also do the following (and not use the subscription at all):

```
  public hideEvaluationOverlay(): void {
    if (this.overlayRef) {
      this._dtOverlay.dismiss();
      this.unsubscribe$.next(); // <--- 
      this.overlayRef = undefined;
    }
  }
```

bridge/client/app/_views/ktb-environment-view/ktb-stage-overview/ktb-services-list/ktb-services-list.component.html
```diff
@@ -83,7 +83,9 @@
           <ng-container *ngIf="row | toType: ServiceClass as service">
             <div *ngIf="service.latestSequence">
               <ktb-evaluation-info
-                [evaluation]="service.latestSequence.getEvaluationTrace(service.stage)"
+                *ngIf="{ value: service.latestSequence.getEvaluationTrace(service.stage) } as evaluation"
```
renepanzar, line 86: This will always be true? I suppose this is only used to wrap the value in an object, which in turn can also be undefined or null?

bridge/cypress.config.ts
```diff
@@ -15,7 +15,7 @@ export default defineConfig({
   chromeWebSecurity: false,
   e2e: {
     baseUrl: 'http://localhost:5000', // workaround until https://github.com/cypress-io/cypress/issues/21555 is fixed
-    specPattern: 'cypress/integration/**.spec.{js,jsx,ts,tsx}',
+    specPattern: 'cypress/integration/**/**.spec.{js,jsx,ts,tsx}',
```
renepanzar, line 18: Nice change to allow nesting specs in directories :+1:

renepanzar (changes requested):
A few adaptions are required, otherwise it looks good!

bridge/client/app/_components/ktb-evaluation-badge/ktb-evaluation-badge.component.scss
Kirdock: It has to be the global style because the overlay is not inside the child but on root level of the body

renepanzar: As discussed offline, this can be left in.

bridge/client/app/_components/ktb-evaluation-badge/ktb-evaluation-badge.component.ts
Kirdock: done

renepanzar (approved):
LGTM now!

PR #8525, opened 2022-07-27
test(bridge): Fix ignored UI tests in trigger-sequence.spec

Kirdock:
Only one test in `trigger-sequence.spec.ts` was executed because of `it.only`.
Removed the `only`, fixed two UI tests and fixed a bug in the environment screen regarding automatically opening the trigger-sequence form.

## Additional changes:
- Added a retry strategy for UI tests. A failed test will now run up to 2 additional times if it fails.


renepanzar (approved):
LGTM!

ermin-muratovic (approved):
lgtm

PR #8531, opened 2022-07-28
fix(bridge): Fix detection of pending changes when automatic provisioning active

### LINKED ISSUE #8493: Bridge 0.17.0: Incorrect detection of pending changes when accessing project details of a project with auto-provisioned project
Using Keptn 0.17.0 and the [keptn-gitea-provisioner-service](https://github.com/keptn-sandbox/keptn-gitea-provisioner-service), I created a new project, e.g.:
```bash
keptn create project jes4 --shipyard=shipyard.yaml
```

This will auto create a repo and asign it to the project, e.g.:

# Current Behaviour

When editing the project in Bridge, something weird is happening and this seems to cause Bridge to think that I have changed some values:
image

# Expected Behaviour

As I did not change any values, I am not expecting the popup regarding unsaved changes

ermin-muratovic:
Signed-off-by: ermin.muratovic <ermin.muratovic@dynatrace.com>

## This PR
- fixes the detection of pending changes when automatic provisioning is active
- shows the configured data correctly in the settings page

### Related Issues
Fixes #8493 

image


Kirdock (changes requested):
Could you add a Unit and a UI test for the right radio-box selection here?
e.g. `should not select NO_UPSTREAM if automatic provisioning is active and createMode is false`

Kirdock (approved):

PR #8535, opened 2022-07-28
fix(installer): RoleBinding is not installed if not needed for shippy leader election

### LINKED ISSUE #8534: Core :: Installation creates resources when not needed
Keptn provides the helm value `shipyardController.config.disableLeaderElection`: https://github.com/keptn/keptn/blob/master/installer/manifests/keptn/values.yaml#L292

However, in our RBAC.yaml file, we ask for the `leases` permission no matter the settings of the flag: 
https://github.com/keptn/keptn/blob/master/installer/manifests/keptn/templates/rbac.yaml#L183
https://github.com/keptn/keptn/blob/master/installer/manifests/keptn/templates/rbac.yaml#L338

mowies:
### This PR
Fixes an issue, where the role binding for shippy to acquire the lease for leader election was always installed. Now, it's only installed if leader election is actually enabled.
This PR also adjusts the helm value slightly to follow our own standards for helm values. (`disableLeaderElection` -> `leaderElection.enabled`)

Fixes #8534

### DIFF
--- a/installer/manifests/keptn/templates/rbac.yaml
+++ b/installer/manifests/keptn/templates/rbac.yaml
@@ -335,6 +335,7 @@ subjects:
     name: keptn-secret-service
 
 ---
+{{- if and (ge .Capabilities.KubeVersion.Minor "14") (.Values.shipyardController.config.leaderElection.enabled) }}
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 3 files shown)


thisthat (approved):
LGTM

bacherfl (approved):

PR #8536, opened 2022-07-28
test: Migrate QualityGates integration tests to component tests

### LINKED ISSUE #8030: Core :: Clean up qualitygates  integration test
Quality gates integration test should be only running a single full sequence, the rest of the checks can be moved as component test.

current stats for the test on gke19 are over 7 minutes:
Test_GKE/Test_QualityGates (182.02s) - keep it
Test_GKE/Test_QualityGates_SLIWrongFinishedPayloadSend (85.66s) - move as component test
Test_GKE/Test_QualityGates_AbortedFinishedPayloadSend (84.90s) - move as component test
Test_GKE/Test_QualityGates_ErroredFinishedPayloadSend (86.78s) - move as component test

DoD

- [x] One full qualitygates integration test around 100-200s long
- [x] All other test cases migrated to component/unit tests in lighthouse-service

odubajDT:
## This PR

- migrates quality gates integration tests to component tests
- refactoring of lighthouse-service to make it testable

### Related Issues

Fixes #8030

### Integration tests
https://github.com/keptn/keptn/actions/runs/2787887829


lighthouse-service/event_handler/common.go
```diff
@@ -211,16 +210,31 @@ type SLIProviderConfig interface {
...
+type K8sSLIProviderOption func(c *K8sSLIProviderConfig)
+
+func WithKubeAPIClient(k8sClient kubernetes.Interface) K8sSLIProviderOption {
+	return func(c *K8sSLIProviderConfig) {
+		c.KubeAPI = k8sClient
 	}
+}
 
-	configMap, err := kubeAPI.CoreV1().ConfigMaps(namespace).Get(context.TODO(), "lighthouse-config", v1.GetOptions{})
+func NewSLIProviderConfig(opts ...K8sSLIProviderOption) K8sSLIProviderConfig {
```
bacherfl, line 225: we can pass the k8s interface directly as an argument since this component does not function without any k8s client injected into it

lighthouse-service/main_test.go
```diff
@@ -0,0 +1,599 @@
...
+		),
+	)
+
+	logHandler := &gofake.LogAPIMock{}
+	logForwarder := logforwarder.New(logHandler)
+
+	controlPlane := controlplane.New(subscriptionSource, eventSource, logForwarder, controlplane.WithLogger(log))
+
+	os.Setenv("RESOURCE_SERVICE", configurationService.URL)
+	defer os.Unsetenv("RESOURCE_SERVICE")
```
warber, line 105:
```suggestion
     test.Setenv("RESOURCE_SERVICE", configurationService.URL)
```

lighthouse-service/main_test.go
```diff
@@ -0,0 +1,599 @@
...
+  warning: "65%"`
+
+const natsTestPort = 8370
+
+var keptnContext = "context"
+var projectName = "quality-gates-invalid-finish"
+var serviceName = "my-service"
+var stageName = "dev"
+var configurationService *httptest.Server
+var EventChan chan types.EventUpdate
```
warber, line 75: remove

lighthouse-service/main.go
```diff
@@ -65,15 +76,21 @@ func main() {
 	logForwarder := logforwarder.New(api.LogsV1(), logforwarder.WithLogger(log))
 
 	controlPlane := controlplane.New(subscriptionSource, eventSource, logForwarder, controlplane.WithLogger(log))
+	lighthouseService.env = env
+
+	_main(controlPlane, log, lighthouseService)
```
warber, line 81:
```suggestion
	_main(controlPlane, log, LighthouseService{KubeAPI: kubeAPI, env: env})
```

lighthouse-service/main.go
```diff
@@ -38,7 +41,15 @@ type envConfig struct {
...
 func main() {
+	kubeAPI, err := createKubeAPI()
+	if err != nil {
+		log.Fatalf("could not create kubernetes client: %s", err.Error())
+	}
+
 	var env envConfig
+	lighthouseService := LighthouseService{
+		KubeAPI: kubeAPI,
+	}
```
warber, line 52:
```suggestion
	lighthouseService := LighthouseService{
		KubeAPI: kubeAPI,
	}
```
```suggestion
```
you can delete this from here and pass it dirctly as arg to _main below

lighthouse-service/event_handler/handler.go
```diff
@@ -33,6 +35,8 @@ func NewEventHandler(ctx context.Context, event cloudevents.Event) (EvaluationEv
 		return nil, err
 	}
 
+	eventStore := processEventStore(es, keptnHandler)
```
warber, line 38: what is this doing ? why?

odubajDT: due to testability, if the es is nil (we don't pass the mocked EventStore), we should use the one from keptnHandler, if it's not nil (we passed the mocked implementation), we should use the mocked implementation

warber: If it's just about testing, this should not be part of the business logic (it is like introducing a `if mock {}` logic). Would be better to just pass the correct implementation of EventStore as a dependency. Also, if i delete this line, all the tests are still green

odubajDT: I could not find a solution how to inject this nicely, as the EventStore is a part of keptnHandler which is created in the NewEventHandler(), which is called in OnEvent() and depends on context. If I  try to inject it from OnEvent(), I cannot get the appropiate data from context, so I get an error of `could not get eventSender from context`. Can you maybe propose some solution?

warber: Instead of passing in an `EventStore` value which the constructor logic needs to evaluate whether it is nil or not, you can pass in a EventStore _Provider_ that returns the correct event store. The provider is a simple function that looks like this:
```
type EventStoreProvider func(k Keptn) EventStore
``` 
now, in the test you pass a provider that just ignores the `Keptn` parameter and returns the mocked event store. In the real implementation you pass a provider provider that returns `k.EventHandler`

odubajDT: resolved

lighthouse-service/event_handler/handler_test.go
```diff
@@ -97,13 +97,10 @@ func TestNewEventHandler(t *testing.T) {
...
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			GetConfig().GetKubeAPI = func() (kubernetes.Interface, error) {
-				return fake.NewSimpleClientset(), nil
-			}
 			tt.args.event.SetType(tt.eventType)
 			t.Setenv("RESOURCE_SERVICE", configurationServiceURL)
 
-			got, err := NewEventHandler(ctx, tt.args.event)
+			got, err := NewEventHandler(ctx, tt.args.event, fake.NewSimpleClientset(), func(k *keptnv2.Keptn) EventStore { return keptnHandler.EventHandler })
```
warber, line 103:
```suggestion
			got, err := NewEventHandler(ctx, tt.args.event, fake.NewSimpleClientset(), func(k *keptnv2.Keptn) EventStore { return k.EventHandler })
```

warber (approved):

PR #8567, opened 2022-08-03
feat(bridge): link to current running sequence

### LINKED ISSUE #7635: Link to sequence that is currently running
### Use case

As a user, I'm triggering multiple sequences at the same time. As a consequence, the first one is blocking the others from getting launched. 

This can result in the following picture, where one sequence is running (R) and two sequences are queued (Q): 

image

### Improvement

* In a queued sequence, provide a link to the currently running sequence:

image

* As a user, I can then directly jump to the sequence that is blocking the current execution. This is helpful in order to find stuck sequences that are blocking others. 

## Acceptance criteria

- [ ] When the user clicks the link, Bridge retrieves the right `shkeptncontext` of the running sequence (that blocks the currently selected sequence) and navigates to that sequence. 
- [ ] Validate the UX (the proposed link might not be the best) w/ @johannes-b
- [ ] Handle the corner case: the user open a blocked sequence and left the bridge open long enough for the current sequence to change its state from blocked to finished.

### Definition of Done
- [ ] UI & Unit tests
- [ ] As a user, I can directly jump to the sequence that is blocking the current sequence from getting executed. 

## Resources

- API PR: https://github.com/keptn/keptn/pull/8430

TannerGilbert:
## This PR

- adds link to current running sequence if the selected sequence is blocked

### Related Issues

Resolves #7635

### Notes

link-to-blocking-sequence

link-to-blocking-sequence-II

[link-to-blocking-sequence.webm](https://user-images.githubusercontent.com/36239763/182592610-6038f406-7bc5-401c-8a1f-6e3468bcbb7c.webm)


bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.ts
```diff
@@ -406,6 +406,30 @@ export class KtbSequenceViewComponent implements OnDestroy {
     this.router.navigate(['/project/' + projectName]);
   }
 
+  public navigateToBlockingSequence(currentSequence: SequenceState, state?: ISequenceViewState): void {
```
Kirdock, line 409: Could you add some unit tests for this function?

Kirdock (changes requested):

bridge/cypress/integration/sequences.spec.ts
```diff
@@ -97,6 +97,46 @@ describe('Sequences', () => {
       .assertIsSelectedSequenceWaiting(true);
   });
 
+  it.only('should navigate to blocking sequence', () => {
```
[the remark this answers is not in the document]
TannerGilbert: Tested this behavior using unit tests.

bridge/cypress/support/pageobjects/SequencesPage.ts
```diff
@@ -12,6 +12,9 @@ type EFilterGroup = 'Service' | 'Stage' | 'Sequence' | 'Status';
 export class SequencesPage {
   private readonly sequenceWaitingMessage = ' Sequence is waiting for previous sequences to finish. ';
```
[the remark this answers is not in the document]
TannerGilbert: This message is still used inside the `ktb-root-events-list.component`

bridge/shared/interfaces/sequence-execution-result.ts
```diff
@@ -56,15 +51,12 @@ export interface SequenceExecutionStatus {
...
-      result: string;
-      status: string;
+      properties: Record<string, unknown>;
+      result: ResultTypes;
+      status: StatusTypes;
       triggeredID: string;
     }
   ];
-  state: string;
+  state: 'triggered' | 'waiting' | 'suspended' | 'paused' | 'finished' | 'cancelled' | 'timedOut';
```
Kirdock, line 60: Could you also add 'started' here? And maybe move it into a separate type variable and use that for the status property of the sequence-execution API call in the apiService.

Kirdock (changes requested):
I guess we need a fallback for the stage
In this case
image
we don't have any stages.

We should discuss how we should handle that.

TannerGilbert:
I added code to sort the stages when calling `${this._baseUrl}/project/${projectName}/sequences/filter` to allow us to use the `metadata` object in the `ktb-sequence-view.component.ts` to get the available sequences. The sorting however doesn't help us navigate to the correct next stage if there are multiple parent stages. For this, we would need to update the [getSequencesFilter method](https://github.com/keptn/keptn/blob/29c9e7d6d250cde8247558df1303a72c900d4540/bridge/server/services/data-service.ts#L1242) to also return the parentStages of each individual stage.

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.ts
```diff
@@ -406,6 +408,45 @@ export class KtbSequenceViewComponent implements OnDestroy {
...
+        pageSize: 1,
+      })
+      .pipe(
+        finalize(() => {
+          this.navigationToRunningSequenceLoading = false;
+        })
+      )
+      .subscribe((sequenceExecutionResult) => {
+        if (sequenceExecutionResult.sequenceExecutions.length === 0) {
+          this.navigationToRunningSequenceLoading = false;
```
Kirdock, line 432: This is not necessary because it is already set in `finalize`

bridge/cypress/support/pageobjects/SequencesPage.ts
```diff
@@ -243,10 +243,10 @@ export class SequencesPage {
     return this;
   }
 
-  public assertIsSelectedSequenceWaiting(status: boolean): this {
+  public assertDtAlert(status: boolean): this {
```
Kirdock, line 246: Could you make it more clear in the function name what it does?
```suggestion
  public assertDtAlertExists(status: boolean): this {
```

Kirdock (dismissed):

heinzburgstaller (approved):
Next time: Changing a naming will be in a separate PR please ;)

PR #8578, opened 2022-08-04
feat(api): Better inbound event validation

### LINKED ISSUE #5544: Validation of inbound events
### User Story

> As a user,
> When I post an event to Keptn
> Then I want to be clear about whether the event payload is valid or not

#### Example:
User wants to post a `task.started` event using the `POST /event` endpoint of Keptn but uses the wrong event type (e.g. `sh.keptn.event.dev.verification.started` instead of `sh.keptn.event.verification.started`) 

### Acceptance Criteria
 - [x] check for missing mandatory fields (e.g., trigger id based on the event type, project, stage, service)
 - [x] type must have correct syntax and started events must not have a stage
 - [x] time format is correct
 - [x] if it is a well-defined event (from our spec), validate that it contains the mandatory fields
 - [x] `result` and `status` have a correct value set
 - [x] size of the [CloudEvent is respected](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#size-limits) (check Distributor for a reference implementation)

## DoD 
- [x] The user posts events that are **invalid** for the type of action he/she wants to trigger and gets a reason why in the logs.
- [ ] updated/new http endpoints are well documented in swagger doc
- [ ] integration test verifies that invalid events are marked as warnings by Keptn

warber:
closes #5544 

This PR adds following Event validation to the API-service:
#### task started events (i.e. `sh.keptn.event.<task>.started`)
* general validity of event data
* mandatory project stage and service as part of event data
* mandatory triggered id

#### task finished events (i.e. `sh.keptn.event.<task>.finished`)
* general validity of event data
* mandatory project, stage and service as part of event data
* mandatory result and status as part of event data
* mandatory triggered id

#### sequence triggered events (i.e. `sh.keptn.event.<stage>.<sequence>.triggered`)
* general validity of event data
* mandatory project, stage and service as part of event data
* stage name in event data must match stage mentioned in event type

#### error log events (i.e., `sh.keptn.log.error` )
* no specific validation(s) yet

#### invalidated events (sent by bridge, i.e. `sh.keptn.event.<task>.invalidated`)
* no specific validation(s) yet

#### problem events (i.e. `sh.keptn.events.problem`)
* no specific validation(s) yet

Events of different types are rejected.
Validation errors are returned with HTTP 400 Bad Request response codes.

Note: Event validation can be turned on/off using the `EVENTE_VALIDATION_ENABLED` environment variable

Additionally, there is a max size validation for Events entering the `/event` endpoint, which is configurable using the `MAX_EVENT_BODY_SIZE_KB` environment variable and, if not set, defaults to `64` if set to `0` this check is disabled.

Latest Integration test run: https://github.com/keptn/keptn/actions/runs/2833058931


api/handlers/eventvalidation.go
```diff
@@ -0,0 +1,149 @@
...
+		v0_2_0.EventData
+		Type string `json:"type,omitempty"`
+	}
+
+	var eventData data
+	if err := v0_2_0.Decode(e.Data, &eventData); err != nil {
+		return &EventValidationError{Msg: "could not parse common event data"}
+	}
+
+	if eventData.Project == "" || eventData.Service == "" || eventData.Type == "" {
```
bacherfl, line 144: i believe only the project name is mandatory for the monitoring.configure event, so we should remove the check for the service property

bacherfl (approved):

PR #8605, opened 2022-08-08
refactor(bridge): Split environment-details component

### LINKED ISSUE #7010: Split evaluation details component
At the moment the evaluation details component should be split into two components.
The reason why: We also use this component for the overlay of some evaluation results (like in the environment screen) and for the overlay we don't show and therefore don't need any previous evaluations. Therefore we don't need to fetch any previous evaluations and we don't need to aggregate the data for the charts.

Kirdock:
Closes #7010 

## Additional Changes
- The SLI breakdown now has a loading indicator if it needs to fetch compared events for calculating the compared value, if the compared value property of an indicator result does not exist.
- Fixed a bug in the SLI breakdown that `-100%` was shown if the value stayed the same (e.g. compared value: 0 and current value: 0)


heinzburgstaller (commented):
This PR is way to big. Is there a chance to split this PR into three or more?

1. Splitting the code into smaller components
2. Add unit tests improve code
3. Add UI tests

heinzburgstaller (approved):

PR #8612, opened 2022-08-09
fix(shipyard-controller): Fail sequence when receiving invalid status

### LINKED ISSUE #7367: Core: default behavior for result and status
## Technical details

Currently, the shipyard-controller considers a failed sequence only if one of the following is true:
- `result = fail`
- `status = error` 

Instead, the default behavior should be to fail and stop a running sequence if the events contains a non valid value for `result` or `status`. 

## Acceptance criteria

- [x] stop a running sequence will be the default behavior if a wrong value for `result` or `status` is set

## DoD

- [x] Unit tests
- [x] Add a component test for this in the shippy
- [x] Adapt existing integration tests

## Reference

- https://github.com/keptn/spec/blob/master/cloudevents.md#data

odubajDT:
## This PR

- sequence should fail when receiving unknown status or result

### Related Issues

Fixes #7367 

### Integration tests
https://github.com/keptn/keptn/actions/runs/2911236457

### DIFF
--- a/shipyard-controller/models/sequence_execution.go
+++ b/shipyard-controller/models/sequence_execution.go
@@ -242,7 +242,6 @@ func (e *TaskExecutionState) IsFailed() bool {
 	for _, event := range e.Events {
 		if keptnv2.IsFinishedEventType(event.EventType) {
 			if event.Result != keptnv2.ResultPass && event.Result != keptnv2.ResultWarning {
-				//if event.Result == keptnv2.ResultFailed {
 				return true
 			}
 		}
--- a/shipyard-controller/models/sequence_execution.go
+++ b/shipyard-controller/models/sequence_execution.go
@@ -1,8 +1,9 @@
 package models
 
 import (
-	"github.com/keptn/keptn/shipyard-controller/internal/common"
 	"time"
+
+	"github.com/keptn/keptn/shipyard-controller/internal/common"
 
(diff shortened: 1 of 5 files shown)


shipyard-controller/models/sequence_execution.go
```diff
@@ -235,7 +241,8 @@ func (e *TaskExecutionState) IsFinished() bool {
 func (e *TaskExecutionState) IsFailed() bool {
 	for _, event := range e.Events {
 		if keptnv2.IsFinishedEventType(event.EventType) {
-			if event.Result == keptnv2.ResultFailed {
+			if event.Result != keptnv2.ResultPass && event.Result != keptnv2.ResultWarning {
+				//if event.Result == keptnv2.ResultFailed {
```
bacherfl, line 245: please remove the commented line

odubajDT: sure

shipyard-controller/models/sequence_execution_test.go
```diff
@@ -570,6 +571,120 @@ func TestSequenceExecution_CompleteCurrentTask(t *testing.T) {
 				},
 			},
 		},
+		{
```
bacherfl, line 574: let's add some additional test cases to keep the coverage of `sequence_execution.go` at 100%

bacherfl (approved):

PR #8613, opened 2022-08-10
chore(bridge): Separate run of client and server (dev)

### LINKED ISSUE #8558: Bridge: Separate run of client and server (dev)
## Technical Details

Currently, for development, there are multiple pains when running/building the angular application with `ng build` instead of `ng serve`. Examples are
- The page is not automatically re-rendered if there is a change
- Caching. Refreshing the page is often not enough. There are also some cached lazy junks.

The client would need to run on a different port. Client should also have a proxy configuration to configure the bridge-server endpoint. 

## Acceptance Criteria

- [x] Add a new npm entry cmd to run both Client and Server in dev mode
- [x] Add new entry cmds to run Client and Server individually in dev mode
- [x] In dev mode, angular files are served via `ng serve`

## DoD

- The client runs on its own with `ng serve` and has a proxy configuration that points at the bridge server (the client has a different port than the server).
- There is one single command to run both, client and server, and two commands to run them separately.
- Manual tests

Kirdock:
Closes #8558 

### DIFF
--- a/bridge/README.md
+++ b/bridge/README.md
@@ -162,7 +162,7 @@ kubectl delete -f deploy/service.yaml
    set API_URL=http://keptn.127.0.0.1.nip.io/api
    set API_TOKEN=1234-exam-ple
    ```
-4. Run `yarn start:dev` from bridge root level to start the express server and the Angular app.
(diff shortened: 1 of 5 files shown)


heinzburgstaller (approved):
Live reloading works really nicely now. LGTM

TannerGilbert (approved):

PR #8615, opened 2022-08-10
chore(bridge): Upgrade to Angular 13 (WIP)

Kirdock:
## Blockers
Dynatrace barista components: https://github.com/dynatrace-oss/barista
At the moment v10 of barista does not support Angular v13.
There is a v11 release candidate and it works if `dompurify` is added to `package.json` but it hasn't been updated for months.

## Additional changes
Adjusted some classes and interfaces because the Angular HTML strict check is really weird if it comes to generics. e.g. `myFunction<T extends Trace>(): T` does not really work. If the class that is using this function is in the ts-file instead it works without any problems.


mowies:
@Kirdock is this still relevant?

Issue #8571, opened 2022-08-04
Bridge :: Streamline development documentation


TannerGilbert:
Currently the development documentation for the Keptn Bridge is split into two documents [docs/developer/bridge.md](https://github.com/keptn/keptn/blob/master/docs/developer/bridge.md) and [bridge/README.md](https://github.com/keptn/keptn/blob/master/bridge/README.md). This should be streamlined to make it more intuitive for new contributors.

Discussion with the team:
The [bridge/README.md](https://github.com/keptn/keptn/blob/master/bridge/README.md) should contain the current information about developing with the Bridge. The [docs/developer/bridge.md](https://github.com/keptn/keptn/blob/master/docs/developer/bridge.md) links to the [bridge/README.md](https://github.com/keptn/keptn/blob/master/bridge/README.md).

# DoD

- [ ] [docs/developer/bridge.md](https://github.com/keptn/keptn/blob/master/docs/developer/bridge.md) links to the [bridge/README.md](https://github.com/keptn/keptn/blob/master/bridge/README.md)
- [ ] [bridge/README.md](https://github.com/keptn/keptn/blob/master/bridge/README.md) is updated with the newest information

PR #8635, opened 2022-08-11
chore(bridge): Remove resource service feature flag

### LINKED ISSUE #8433: Bridge :: Remove FF `RESOURCE_SERVICE_ENABLED`
## Technical details

Keptn has only one service by now which is the resource-service. 
The FF to know if configuration- or resource-service is not needed anymore and its logic should be removed.
Remove also all code branches that have special logics for when the FF is false

## DoD 

- [ ] Bridge considers only resource-service

Kirdock:
Closes #8433


heinzburgstaller (commented):
In the file `keptn/bridge/deploy/service.yaml` is the flag 'RESOURCE_SERVICE_ENABLED' still there. Can also be deleted?

Kirdock:
@heinzburgstaller  This file is deprecated and does not even work. With #8571 we wanted to delete it anyway

heinzburgstaller (approved):

PR #8719, opened 2022-08-22
feat(cli): Use sh.keptn.event.configure-monitoring.triggered instead of sh.keptn.event.monitoring.configure for keptn configure monitoring CLI command

### LINKED ISSUE #8718: keptn configure monitoring sends inconsistent Cloud Event types, leading to problems with go-sdk
We (@TannerGabriel and myself) are currently trying to change [prometheus-service](https://github.com/keptn-contrib/prometheus-service) to use the go-sdk.

prometheus-service has two use-cases that requires communication with Cloud Events:

* Configure Monitoring (triggered by using the CLI `keptn configure monitoring prometheus`),
* Quality Gate (triggered via `get-sli.triggered` Cloud Event)

This would be the usage with the SDK:
```go
const serviceName = "prometheus-service"
const envVarLogLevel = "LOG_LEVEL"
const monitoringTriggeredEvent = "sh.keptn.event.configure-monitoring.triggered" // "sh.keptn.event.monitoring.configure"
const getSliTriggeredEvent = "sh.keptn.event.get-sli.triggered"

...

	log.Fatal(sdk.NewKeptn(
		serviceName,
		sdk.WithTaskHandler(
			monitoringTriggeredEvent,
			eventhandling.NewConfigureMonitoringEventHandler(),
			prometheusTypeFilter),
		sdk.WithTaskHandler(
			getSliTriggeredEvent,
			eventhandling.NewGetSliEventHandler(*kubeClient),
			prometheusSLIProviderFilter),
		sdk.WithLogger(logrus.New()),
	).Start())
```

Unfortunately, we identified a huge problem with configure monitoring.

**keptn configure monitoring prometheus**  will send the following CloudEvent:
https://github.com/keptn/keptn/blob/3253bf522644f19ed44bbfdec53ab678f4dc86ce/cli/cmd/configure_monitoring.go#L80-L87

Whereas `keptn.ConfigureMonitoringEventType` is set to `sh.keptn.event.monitoring.configure` (see https://github.com/keptn/go-utils/blob/764b633614cb1cdc672d525d1dd3f9fab2514640/pkg/lib/events.go#L11-L12 for reference).

**However, the CLI then waits for a `sh.keptn.event.configure-monitoring.finished` Cloud Event**:
https://github.com/keptn/keptn/blob/3253bf522644f19ed44bbfdec53ab678f4dc86ce/cli/cmd/configure_monitoring.go#L117-L121

First of all, this Cloud Event Type is not specified in [keptn/spec](https://github.com/keptn/spec/blob/master/cloudevents.md).
Second, this is heavily inconsistent (CLI sends `sh.keptn.event.monitoring.configure` but expects `sh.keptn.event.configure-monitoring.finished`).
Third, and that's the biggest problem: While we could use [`WithAutomaticResponse(false)`](https://github.com/keptn/go-utils/blob/5ff2d4010e671b0e3fc14e55dd8214bd5643bede/pkg/sdk/keptn.go#L113-L119) from the SDK, this would disable the automatically generated response also for the `get-sli.triggered` handler, and therefore create a huge mess in our implementation.

# Affected Components / Services

* SLI providers (prometheus-service, dynatrace-service, datadog-service, sumologic-service)
* lighthouse service?
* CLI
* spec
* docs?

# Proposed Solutions

Either one of those could probably solve the problem.

- [ ] Rework the CLI to send a Cloud Event of type `configure-monitoring.triggered`, and wait for a `configure-monitoring.finished` (already there)
- [ ] Rework the CLI, keep sending a `monitoring.configure`, but wait for a `monitoring.configure.finished` Cloud Event
- [ ] Support both Cloud Event types for sending (waiting is still on `configure-monitoring.finished`) / adapt the SDK to handle this corner case properly
- [ ] Adapt the SDK to include a way to disable automatic response within a single task handler (not the whole `newKeptn`)

## Decision 

- Adapt the go SDK to provide a `WithAutomaticResponse(false)` per task handler and not globally 

## DoD

- [ ] Doc is updated
- [ ] Go SDK is adapted
- [ ] Unit tests

christian-kreuzberger-dtx:
This breaking change also affects
* prometheus-service
* dynatrace-service
* datadog-service
* sumologic-service

Fixes #8718

### DIFF
--- a/cli/cmd/configure_monitoring.go
+++ b/cli/cmd/configure_monitoring.go
@@ -81,7 +81,7 @@ keptn configure monitoring datadog --project=PROJECTNAME --service=SERVICENAME
 
 		sdkEvent := cloudevents.NewEvent()
 		sdkEvent.SetID(uuid.New().String())
-		sdkEvent.SetType(keptn.ConfigureMonitoringEventType)
+		sdkEvent.SetType(keptnv2.GetTriggeredEventType("configure-monitoring"))
 		sdkEvent.SetSource(source.String())
 		sdkEvent.SetDataContentType(cloudevents.ApplicationJSON)
 		sdkEvent.SetData(cloudevents.ApplicationJSON, configureMonitoringEventData)


odubajDT:
pls fix your commit msg

thisthat (changes requested):
This would break a lot of SLI-services and the lighthouse-service. 
I think we need to support the `configure` event in a native way from the go-SDK

christian-kreuzberger-dtx:
Agreed. I will close this PR, as my fix does not work as expected.

PR #8730, opened 2022-08-24
feat: Introduce signed Keptn Helm charts

### LINKED ISSUE #6610: Create provenance files to be able to verify keptn helm charts integrity
Helm has a feature called provenance files to check charts for their integrity. We should make use of that feature in the keptn project.
The only thing that is needed is a PGP key pair and some small adaptations to the helm chart upload pipeline steps to also push the created provenance files.
In addition the keptn charts could be verified nightly like the docker images (#6598).

`helm package --sign` provides the command to generate such a provenance file for our charts. (more info [here](https://helm.sh/docs/topics/provenance/))

Tasks:
- find out how PGP key pairs are created and if they need some certificate authority to be valid (probably not)
- create a PGP key pair, store it securely and save the private in the keptn GH secrets
- adjust the (pre)release pipeline to use the `--sign` argument during helm chart packaging so that it will use the PGP private key to create a provenance file for the chart to be released
- adjust the (pre)release pipeline to also commit the provenance file to the keptn/helm-charts git repo

Create a small GH actions pipeline that runs on a schedule to verify the latest helm chart of keptn similar to the [docker image digest checker pipeline](https://github.com/keptn/keptn/blob/master/.github/workflows/check-docker-image-digests.yml).
It should just download the latest chart from the keptn/helm-charts repo for keptn, helm-service and jmeter-service and then use `helm verify` to verify the chart against the provenance file that should already be in the chart repo.

References:
https://helm.sh/docs/topics/provenance/
[Example of a provenance file](https://github.com/grafana/helm-charts/blob/gh-pages/packages/fluent-bit-0.0.5.tgz.prov)

mowies:
### This PR
- adds functionality to sign helm charts with a private/public keypair
- signing only happens for releases and pre-releases
- the public key is added to the repo and to every new release together with the Helm provenance `*.tgz.prov` files
- adds a small section to the readme to point users into the right direction if they want to verify the keptn charts

#### How to test
Check the helm chart artifacts in this run for the provenance files: https://github.com/keptn/keptn/actions/runs/2919304598
They can be verified by downloading the public key under `assets/pubring.gpg` and importing it into gpg.
Use the following steps:
- Download the helm chart and provenance file from the workflow run, as well as the public key. Then run below commands.
- `gpg --import pubring.gpg`
- `helm verify keptn-0.19.0-dev-PR-8730.tgz`

(you will maybe need to export your gpg public key ring since Helm cannot handle `*.kbx` keyrings, only `*.gpg` keyrings, more info on this can be found [here](https://helm.sh/docs/topics/provenance/#:~:text=Warning%3A%20the,gnupg/secring.gpg))

Fixes #6610


.github/workflows/pre-release.yml
```diff
@@ -426,3 +429,10 @@ jobs:
           cat ../dist/image-digests/digest-*.txt >> $DIGEST_FILE
 
           gh release upload "$RELEASE_TAG" "$DIGEST_FILE"
+
+      - name: Attach helm provenance public key to release
```
thisthat, line 433:
```suggestion
      - name: Attach Helm provenance public key to release
```

gh-actions-scripts/build-helm-charts.sh
```diff
@@ -26,7 +27,28 @@ if [ -z "$DOCKER_ORG" ]; then
...
 fi
 
+if [ -z "$SIGN_CHART" ]; then
+  echo "No flag set for signing charts, defaulting to no signing"
+  SIGN_CHART=false
+fi
+
+if [[ "$SIGN_CHART" == 'true' ]]; then
+  if [ -z "$SIGNING_KEY_BASE64" ] || [ -z "$SIGNING_KEY_PASSPHRASE_BASE64" ] || [ -z "$SIGNING_KEY_NAME" ]; then
+    echo 'The following variable need to be set to enable Helm chart signing: SIGNING_KEY_BASE64 SIGNING_KEY_PASSPHRASE_BASE64 SIGNING_KEY_NAME'
```
thisthat, line 37:
```suggestion
    echo 'The following env variables need to be set to enable Helm chart signing: SIGNING_KEY_BASE64 SIGNING_KEY_PASSPHRASE_BASE64 SIGNING_KEY_NAME'
```

.github/workflows/release.yml
```diff
@@ -423,3 +426,10 @@ jobs:
           cat ../dist/image-digests/digest-*.txt >> $DIGEST_FILE
 
           gh release upload "$RELEASE_TAG" "$DIGEST_FILE"
+
+      - name: Attach helm provenance public key to release
```
thisthat, line 430:
```suggestion
      - name: Attach Helm provenance public key to release
```

thisthat (approved):

thisthat (approved):

PR #8736, opened 2022-08-24
feat(shipyard-controller): debugUI documentation & feature Flag

odubajDT:
## This PR

-  adds documentation to the DebugUI
- adds a feature flag to the DebugUI to enable it


installer/manifests/keptn/templates/shipyard-controller.yaml
```diff
@@ -49,6 +49,8 @@ spec:
           env:
             - name: RESOURCE_SERVICE
               value: "http://resource-service:8080"
+            - name: DEBUG_UI_ENABLED
+              value: {{ ((.Values.features).debugUI).enabled | default "false" | quote }}
```
bacherfl, line 53: not 100% sure, but this might result in `value: ""false""` in the rendered manifest, if the value is not present in the values file. Can you please double check if it should rather be `default false | quote`?

odubajDT: you are right, fixing

bacherfl (approved):

PR #8775, opened 2022-08-30
feat(installer): Add initcontainers to prevent installation errors

### LINKED ISSUE #7580: Understand Installation Dependencies
During initial installtion, Keptn can take some time to "become usable" by this I mean having all pods in a `Running` state.

Frequently pods will `CrashLoopBackOff` and `Error` until "something" becomes available and / or "settles down".

Main concerns:

1. A new user will attempt a `keptn install` see pods erroring out and think Keptn is broken - leave and never progress
2. A user wishes to rollout Keptn quickly and these delays cause startup delays
3. Any monitoring of Keptn will be throwing noisy alerts during an install and thus special "no alert" windows will have to be used. Increasing overhead to working with Keptn.

All 3 are risks to the adoption or use of Keptn.

If the dependencies can be understood then (via separate tickets) solutions can be sought.

agardnerIT:
## This PR
- Adds initContainers to prevent `Error`s and `CrashLoopBackOff`s during a Keptninstallation.

### Related Issues
Fixes #7580

### Notes
This is my "best guess" as to the order of the dependencies and probably needs adjusting.

### Follow-up Tasks
This is a temporary fix and should be solved with other methods (like improved readiness probes and health checks).

### How to test
Install a version without this PR and you'll see the pods fall into `Error` and `CrashLoopBackOff` states until everything settles down. This is due to dependencies between the Keptn pods.

Install this version and initcontainers should prevent `Error`s and `CrashLoopBackOff`s.

Green integration test run: https://github.com/keptn/keptn/actions/runs/3264768834

### DIFF
--- a/installer/manifests/keptn/templates/_helpers.tpl
+++ b/installer/manifests/keptn/templates/_helpers.tpl
@@ -463,3 +463,66 @@ Usage:
   {{- end }}
 {{- end -}}
 
+{{- define "keptn.initContainers.wait-for-nats" -}}
+- name: "wait-for-nats"
+  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.version }}"
+  env:
+  - name: "ENDPOINT"
+    value: "{{ .Values.nats.nameOverride }}"
+  command: ['sh', '-c', 'until nslookup $ENDPOINT; do echo waiting for $ENDPOINT; sleep 2; done;']
+{{- end -}}
+
+{{- define "keptn.initContainers.wait-for-keptn-mongo" -}}
+- name: "wait-for-keptn-mongo"
+  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.version }}"
+  env:
+  - name: "ENDPOINT"
+    value: "keptn-mongo"
+  command: ['sh', '-c', 'until nslookup $ENDPOINT; do echo waiting for $ENDPOINT; sleep 2; done;']
+{{- end -}}
+
+{{- define "keptn.initContainers.wait-for-mongodb-datastore" -}}
+- name: "wait-for-mongodb-datastore"
+  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.version }}"
+  env:
+  - name: "ENDPOINT"
+    value: "mongodb-datastore"
+  command: ['sh', '-c', 'until nslookup $ENDPOINT; do echo waiting for $ENDPOINT; sleep 2; done;']
+{{- end -}}
+
+{{- define "keptn.initContainers.wait-for-shipyard-controller" -}}
+- name: "wait-for-shipyard-controller"
+  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.version }}"
+  env:
+  - name: "ENDPOINT"
+    value: "shipyard-controller"
+  command: ['sh', '-c', 'until nslookup $ENDPOINT; do echo waiting for $ENDPOINT; sleep 2; done;']
+{{- end -}}
+
+{{- define "keptn.initContainers.wait-for-secret-service" -}}
+- name: "wait-for-secret-service"
+  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.version }}"
+  env:
+  - name: "ENDPOINT"
+    value: "secret-service"
+  command: ['sh', '-c', 'until nslookup $ENDPOINT; do echo waiting for $ENDPOINT; sleep 2; done;']
+{{- end -}}
+
+{{- define "keptn.initContainers.wait-for-api-service" -}}
+- name: "wait-for-secret-service"
+  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.version }}"
+  env:
+  - name: "ENDPOINT"
+    value: "api-service"
+  command: ['sh', '-c', 'until nslookup $ENDPOINT; do echo waiting for $ENDPOINT; sleep 2; done;']
+{{- end -}}
+
+{{- define "keptn.initContainers.wait-for-resource-service" -}}
+- name: "wait-for-resource-service"
+  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.version }}"
+  env:
+  - name: "ENDPOINT"
+    value: "resource-service"
+  command: ['sh', '-c', 'until nslookup $ENDPOINT; do echo waiting for $ENDPOINT; sleep 2; done;']
+{{- end -}}
+
--- a/installer/manifests/keptn/templates/_helpers.tpl
+++ b/installer/manifests/keptn/templates/_helpers.tpl
@@ -465,7 +465,7 @@ Usage:
 
 {{- define "keptn.initContainers.wait-for-nats" -}}
 - name: "wait-for-nats"
-  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.version }}"
+  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.tag }}"
   env:
   - name: "ENDPOINT"
     value: "{{ .Values.nats.nameOverride }}"
@@ -474,7 +474,7 @@ Usage:
 
 {{- define "keptn.initContainers.wait-for-keptn-mongo" -}}
 - name: "wait-for-keptn-mongo"
-  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.version }}"
+  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.tag }}"
   env:
   - name: "ENDPOINT"
     value: "keptn-mongo"
@@ -483,7 +483,7 @@ Usage:
 
 {{- define "keptn.initContainers.wait-for-mongodb-datastore" -}}
 - name: "wait-for-mongodb-datastore"
-  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.version }}"
+  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.tag }}"
   env:
   - name: "ENDPOINT"
     value: "mongodb-datastore"
@@ -492,7 +492,7 @@ Usage:
 
 {{- define "keptn.initContainers.wait-for-shipyard-controller" -}}
 - name: "wait-for-shipyard-controller"
-  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.version }}"
+  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.tag }}"
   env:
   - name: "ENDPOINT"
     value: "shipyard-controller"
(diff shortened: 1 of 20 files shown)


installer/manifests/keptn/templates/api-service.yaml
```diff
@@ -17,6 +17,37 @@ spec:
...
       terminationGracePeriodSeconds: {{ .Values.apiService.gracePeriod | default 60 }}
+      initContainers:
+        - name: wait-for-nats
+          image: busybox:1.35.0
+          env:
+          - name: "ENDPOINT"
+            value: "keptn-nats"
+          command: ['sh', '-c', 'until nslookup $ENDPOINT; do echo waiting for $ENDPOINT; sleep 2; done;']
+        - name: wait-for-keptn-mongo
+          image: busybox:1.35.0
```
thschue, line 28: Please make the busybux image version and the endpoint configurable as values. Furthermore, we should fail after some time ...

thschue (changes requested):
I think the addition makes sense, but we should fail nevertheless after a certain time. I would propose adding this as a common function (similar to the security context) and including it in the respective pods.

installer/manifests/keptn/templates/api-service.yaml
agardnerIT: Can you help with this? I tried adding this to `values.yaml`:

```
initContainers:
  image: "busybox"
  version: "1.35.0"
```
and setting line 28 (above) as `image: "{{ .Values.initContainers.image }}:{{ .Values.initContainers.version }}"` and I get a warning during `helm install`:

```
coalesce.go:160: warning: skipped value for initContainers: Not a table
```

No idea how to build in the initContainer timeout.

mowies (changes requested):
Since the init containers are repeated in all the template files, it would be great to extract them into helper functions in`_helpers.tpl`

mowies:
@agardnerIT did you try a solution with just readiness probes instead of init containers?

agardnerIT:
> @agardnerIT did you try a solution with just readiness probes instead of init containers?

No. I'm at the limit of my knowledge here so any assistance or better solutions to move this PR towards being ready for a merge is appreciated.

mowies:
I will look into it soon, sorry for the long wait, vacation, other prios in Keptn...

installer/manifests/keptn/templates/api-gateway-nginx.yaml
```diff
@@ -383,6 +383,11 @@ spec:
     spec:
       {{- include "keptn.apiGatewayNginx.pod-security-context" . | nindent 6 }}
       terminationGracePeriodSeconds: {{ .Values.apiGatewayNginx.gracePeriod | default 60 }}
+      initContainers:
+        {{- include "keptn.initContainers.wait-for-nats" . | nindent 8 }}
+        {{- include "keptn.initContainers.wait-for-keptn-mongo" . | nindent 8 }}
+        {{- include "keptn.initContainers.wait-for-mongodb-datastore" . | nindent 8 }}
+        {{- include "keptn.initContainers.wait-for-shipyard-controller" . | nindent 8 }}
```
thisthat, line 390: I don't think this is necessary for nginx since it relies already on the K8s services to be ready

installer/manifests/keptn/values.yaml
```diff
@@ -5,6 +5,12 @@ global:
     registry: docker.io/keptn
     ## @param global.keptn.tag The tag of Keptn that should be used for all images
     tag: ""
+  initContainers:
+    ## @param global.initContainers.image Init container image to enable staggered rollout of Keptn
+    image: "busybox"
```
thisthat, line 10: is there a smaller image that provides `nslookup`?

installer/manifests/keptn/templates/_helpers.tpl
```diff
@@ -463,3 +463,66 @@ Usage:
...
+  command: ['sh', '-c', 'until nslookup $ENDPOINT; do echo waiting for $ENDPOINT; sleep 2; done;']
+{{- end -}}
+
+{{- define "keptn.initContainers.wait-for-keptn-mongo" -}}
+- name: "wait-for-keptn-mongo"
+  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.tag }}"
+  env:
+  - name: "ENDPOINT"
+    value: "keptn-mongo"
+  command: ['sh', '-c', 'until nslookup $ENDPOINT; do echo waiting for $ENDPOINT; sleep 2; done;']
```
thisthat, line 481: here we might have a problem if a user has an external mongo installation. I would suggest to do the lookup only if there is no external connection set via `mongo.external.connectionString`

installer/manifests/keptn/templates/api-gateway-nginx.yaml
```diff
@@ -383,6 +383,11 @@ spec:
     spec:
       {{- include "keptn.apiGatewayNginx.pod-security-context" . | nindent 6 }}
       terminationGracePeriodSeconds: {{ .Values.apiGatewayNginx.gracePeriod | default 60 }}
+      initContainers:
+        {{- include "keptn.initContainers.wait-for-nats" . | nindent 8 }}
+        {{- include "keptn.initContainers.wait-for-keptn-mongo" . | nindent 8 }}
```
thisthat, line 388: Generally speaking, only `shipyard-controller` and `mongo-datastore` connects directly to the DB. So I would have it only for these two services the wait for `keptn-mongo`

installer/manifests/keptn/templates/mongodb-datastore.yaml
```diff
@@ -18,6 +18,8 @@ spec:
       {{- include "keptn.common.pod-security-context" . | nindent 6 }}
       serviceAccountName: keptn-default
       terminationGracePeriodSeconds: {{ .Values.mongodbDatastore.gracePeriod | default 60 }}
+      initContainers:
+        {{- include "keptn.initContainers.wait-for-nats" . | nindent 8 }}
```
thisthat, line 22: Here I would also add the wait for MongoDB

thisthat (commented):
I left some comments but I think we have a good solution here 👍
Thanks for your contribution @agardnerIT

installer/manifests/keptn/values.yaml
mowies: the busybox image has 1.24MB, that's pretty hard to beat...

installer/manifests/keptn/templates/api-gateway-nginx.yaml
mowies: done

installer/manifests/keptn/templates/_helpers.tpl
mowies: this is fixed in the templates for shippy and mongodb-datastore. the waiting is only done if no external connection string is provided in the helm values

installer/manifests/keptn/templates/_helpers.tpl
```diff
@@ -463,3 +463,66 @@ Usage:
...
+    value: "http://shipyard-controller:8080/health"
+  command: ['sh', '-c', 'until curl -s $ENDPOINT; do echo waiting for $ENDPOINT; sleep 2; done;']
+{{- end -}}
+
+{{- define "keptn.initContainers.wait-for-secret-service" -}}
+- name: "wait-for-secret-service"
+  image: "{{ .Values.global.initContainers.image }}:{{ .Values.global.initContainers.tag }}"
+  env:
+  - name: "ENDPOINT"
+    value: "http://secret-service:8080/v1/secret"
```
thisthat, line 507: is there a special reason why we don't use the health endpoint here?

installer/manifests/keptn/templates/core.yaml
```diff
@@ -96,6 +96,10 @@ spec:
         app.kubernetes.io/name: bridge
     spec:
       {{- include "keptn.bridge.pod-security-context" . | nindent 6 }}
+      initContainers:
+        {{- include "keptn.initContainers.wait-for-nats" . | nindent 8 }}
```
thisthat, line 100: Bridge does not need NATS, but the `resource-service`

installer/manifests/keptn/templates/remediation-service.yaml
```diff
@@ -18,6 +18,10 @@ spec:
         app.kubernetes.io/name: remediation-service
     spec:
       {{- include "keptn.common.pod-security-context" . | nindent 6 }}
+      initContainers:
+        {{- include "keptn.initContainers.wait-for-nats" . | nindent 8 }}
+        {{- include "keptn.initContainers.wait-for-mongodb-datastore" . | nindent 8 }}
+        {{- include "keptn.initContainers.wait-for-shipyard-controller" . | nindent 8 }}
```
thisthat, line 24: I think that this service uses the API of the resource-service

installer/manifests/keptn/templates/resource-service.yaml
```diff
@@ -38,6 +38,10 @@ spec:
           type: RuntimeDefault
         {{- end }}
       terminationGracePeriodSeconds: {{ .Values.resourceService.gracePeriod | default 60 }}
+      initContainers:
+        {{- include "keptn.initContainers.wait-for-nats" . | nindent 8 }}
+        {{- include "keptn.initContainers.wait-for-mongodb-datastore" . | nindent 8 }}
+        {{- include "keptn.initContainers.wait-for-shipyard-controller" . | nindent 8 }}
```
thisthat, line 44: There should not be any dependency here, am I right @bacherfl ?

installer/manifests/keptn/templates/secret-service.yaml
```diff
@@ -48,6 +48,10 @@ spec:
       {{- include "keptn.common.pod-security-context" . | nindent 6 }}
       serviceAccountName: keptn-secret-service
       terminationGracePeriodSeconds: {{ .Values.secretService.gracePeriod | default 60 }}
+      initContainers:
+        {{- include "keptn.initContainers.wait-for-nats" . | nindent 8 }}
+        {{- include "keptn.initContainers.wait-for-mongodb-datastore" . | nindent 8 }}
+        {{- include "keptn.initContainers.wait-for-shipyard-controller" . | nindent 8 }}
```
thisthat, line 54: Also the secret-service has not dependency

installer/manifests/keptn/templates/_helpers.tpl
mowies: I just used the same endpoint that the readiness probe also uses, but I can change that as well

installer/manifests/keptn/templates/remediation-service.yaml
mowies: added the resource service

installer/manifests/keptn/templates/resource-service.yaml
mowies: I removed the init containers here for now

installer/manifests/keptn/templates/secret-service.yaml
mowies: removed

installer/manifests/keptn/templates/_helpers.tpl
thisthat: Thanks 👍 that's a good explanation, we can use it

thisthat (approved):
LGTM

mowies (approved):

thschue (commented):
lgtm

thschue (approved):
lgtm

agardnerIT:
All good for me. I can verify this works. Awesome work @mowies .

PR #8782, opened 2022-08-31
feat(webhook-service): Added unmarshalling of curl responses 

### LINKED ISSUE #8256: webhook-service: Add response as JSON object if content-type is application/json
**Status-quo**
Currently, the webhook-service embeds the response as a plain string in the event.

**What?**
However, especially if the response is of content-type `application/json`, it would be good to embed the JSON object in the event (not as string) to simplify further processing.

## DoD

- [x] Unit tests
- [x] Component Tests

RealAnna:
For each response, webhook attempts to convert it into an object and appends it to the finished event
image

closes #8256

### DIFF
--- a/webhook-service/handler/handler_test.go
+++ b/webhook-service/handler/handler_test.go
@@ -698,6 +698,67 @@ func Test_HandleIncomingFinishedEventWithResultingError(t *testing.T) {
 
 }
 
+func Test_HandleIncomingFinishedEventWithJSON(t *testing.T) {
+
+	t.Run("Test_HandleIncomingFinishedEventWithJSON - BETA", func(t *testing.T) {
+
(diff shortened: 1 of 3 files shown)


webhook-service/handler/handler_test.go
```diff
@@ -698,6 +698,67 @@ func Test_HandleIncomingFinishedEventWithResultingError(t *testing.T) {
 
 }
 
+func Test_HandleIncomingFinishedEventWithJSON(t *testing.T) {
```
odubajDT, line 701: is it possible to add a test case, where the string is not unmarshable, so the string is returned instead of JSON object ?

RealAnna: I added this in the unit test, do you think it is needed in the component test too?

odubajDT (approved):

PR #8792, opened 2022-09-01
fix(installer): Normalize 401 responses

### LINKED ISSUE #7524: Core: provide a consistent 401 error
## Technical details

Keptn uses Nginx to redirect each API request toward `POST /auth` (example https://github.com/keptn/keptn/blob/master/installer/manifests/keptn/charts/control-plane/templates/api-gateway-nginx.yaml#L140). 
In the case of an unauthenticated request, this causes two different error responses based if the call is to the api-service or another service. 
The api-service directly handles the auth request and answers with JSON. The other services API calls, instead, are redirected via Nginx and return the default 401 html page. 

## Acceptance Criteria

- [ ] When an unauthorized request is performed, the error message is in JSON format.

## DoD 

- [ ] Acceptance Criteria
- [ ] Integration tests

odubajDT:

## This PR

- adds this new feature

### Related Issues

Fixes #7524


installer/manifests/keptn/templates/api-gateway-nginx.yaml
```diff
@@ -135,7 +140,8 @@ data:
       # auth via backend (if the subrequest returns a 2xx response code, the access is allowed. If it returns 401 or 403,
       # the access is denied) before we store the file
       # see http://nginx.org/en/docs/http/ngx_http_auth_request_module.html
-      auth_request               {{ .Values.prefixPath }}/api/v1/auth;
+      auth_request               /api/v1/auth;
```
thisthat, line 143: please, retain the `prefixPath`

installer/manifests/keptn/templates/api-gateway-nginx.yaml
```diff
@@ -90,6 +90,10 @@ data:
       proxy_busy_buffers_size 256k;
     }
 
+    location @error401 {
+      return 302 /api/v1/auth;
```
thisthat, line 94: I don't think this would work since you create a loop: user request -> (auth endpoint -> failure)+
Besides, why 302?

thisthat (changes requested):
I would instead go with the solution of @mowies proposed in the ticket. This way, we don't need to adapt any endpoint but only override the default 401 page of ngninx

installer/manifests/keptn/templates/api-gateway-nginx.yaml
odubajDT: works, but will try the solution

installer/manifests/keptn/templates/api-gateway-nginx.yaml
```diff
@@ -90,6 +90,10 @@ data:
       proxy_busy_buffers_size 256k;
     }
 
+    location @error401 {
+      return 401 '{"code":401,"message":"unauthenticated for invalid credentials"}';
```
mowies, line 94: i think you meant:
```suggestion
      return 401 '{"code":401,"message":"unauthenticated or invalid credentials"}';
```

mowies (approved):

PR #8820, opened 2022-09-07
ci: Remove schedule for ZD test

RealAnna:
Removing the ZD scheduled test, since we know we do have a regression due to MongoDB changes
Workflow can still be manually triggered https://github.com/keptn/keptn/actions/runs/3005279968


philipp-hinteregger (approved):
LGTM

PR #8822, opened 2022-09-07
feat(cli): Support adding resources to all stages without defining a service + refactoring

### LINKED ISSUE #8818: Core:: CLI add-resource --all-stages without specifying service
When calling keptn add resource with the flag  --all-stages without specifying the service I get an error: "--service and --project need to be supplied when using --all-stages"

I would expect --project and --all-stages to store the resource in each stage folder or branch and only if the service is specified to actually store in all stages for the specified service.

odubajDT:

Fixes #8818

### DIFF
--- a/cli/cmd/add_resource.go
+++ b/cli/cmd/add_resource.go
@@ -45,9 +45,11 @@ From a technical perspective, the file provided via the *--resource* flag is sto
 `,
 	Example: `keptn add-resource --project=musicshop --stage=hardening --service=catalogue --resource=slo.yaml
 keptn add-resource --project=musicshop --stage=hardening --service=catalogue --resource=slo-quality-gates.yaml --resourceUri=slo.yaml
-keptn add-resource --project=sockshop --stage=dev --service=carts --resource=./jmeter.jmx --resourceUri=jmeter/functional.jmx
-keptn add-resource --project=rockshop --stage=production --service=shop --resource=./basiccheck.jmx --resourceUri=jmeter/basiccheck.jmx
-keptn add-resource --project=keptn --service=keptn-control-plane --all-stages --resource=0.7.3_keptn-installer.tgz --resourceUri=helm/keptn-control-plane.tgz`,
+keptn add-resource --project=sockshop --resource=./jmeter.jmx --resourceUri=jmeter/functional.jmx
+keptn add-resource --project=rockshop --stage=production --resource=./basiccheck.jmx --resourceUri=jmeter/basiccheck.jmx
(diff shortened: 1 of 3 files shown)


cli/cmd/add_resource.go
```diff
@@ -83,27 +85,47 @@ keptn add-resource --project=keptn --service=keptn-control-plane --all-stages --
...
+				// project resource
+				logging.PrintLog("Adding resource "+*addResourceCmdParams.Resource+" as a project resource to project "+*addResourceCmdParams.Project, logging.InfoLevel)
+				_, errorObj := api.ResourcesV1().CreateResources(*addResourceCmdParams.Project, "", "", resources)
+				if errorObj != nil {
+					return errors.New("Resource " + *addResourceCmdParams.Resource + " could not be uploaded as a project resource: " + *errorObj.Message)
+				}
+			} else if !isStringFlagSet(addResourceCmdParams.Service) && isBoolFlagSet(addResourceCmdParams.AllStages) {
+				// stage resource to all stages
+				logging.PrintLog("Adding resource "+*addResourceCmdParams.Resource+" as a stage resource to all stages for project "+*addResourceCmdParams.Project, logging.InfoLevel)
+				api, err := internal.APIProvider(endPoint.String(), apiToken)
```
bacherfl, line 99: i belive the initialization of `api, err` can be removed since that already happened some lines above

cli/cmd/add_resource.go
```diff
@@ -83,27 +85,47 @@ keptn add-resource --project=keptn --service=keptn-control-plane --all-stages --
...
+				stages, err := api.StagesV1().GetAllStages(*addResourceCmdParams.Project)
+				if err != nil {
+					return fmt.Errorf("Failed to retrieve stages for project %s: %v", *addResourceCmdParams.Project, err)
+				}
+
+				if len(stages) == 0 {
+					return fmt.Errorf("No stages found")
+				}
+
+				for _, stage := range stages {
```
bacherfl, line 113: should we include a new log message whenever a resource is uploaded to the next stage? Sometimes the command can seem a bit unresponsive if there are many stages

cli/cmd/add_resource.go
odubajDT: fixed

cli/cmd/add_resource.go
odubajDT: ok, we can add that

odubajDT: fixed

bacherfl (dismissed):

bacherfl (approved):

PR #8824, opened 2022-09-07
feat(resource-service): Compute git auth method once per API request

### LINKED ISSUE #7034: Research: Prevent computing publicKey on every git action
## Technical Details 

When using resource-service with SSH and private/public keys for authentication against remote git repository, the public key structure is computed for every git action (clone/fetch/pull/push), which consumes time and resources.

Function `getAuthMethod()` is called on each git action to retrieve the auth credentials for authentication, where we can see, `publicKey` structure is computed from the privateKey, which is slowing the process down. 
https://github.com/keptn/keptn/blob/master/resource-service/common/git.go#L77

For example, when creating a project with multiple stages, the structure is computed ~8 times (depends on the number of stages etc.) because of multiple fetches/pulls/pushes. This needs to be avoided.

## Expected outcome 

We should compute the `publicKey` structure only once per action (for example calling API endpoint for creating a project). We should store the `publicKey` structure to the current context of the API endpoint and not compute it for each git action.

### Suggestions 

- Compute it once per API request and pass it through the context
- Make sure that from one API call to the other, the key is recomputed

odubajDT:

## This PR

- Compute git auth method once per API request

### Related Issues

Fixes #7034 

### Integration tests
https://github.com/keptn/keptn/actions/runs/3007829806
https://github.com/keptn/keptn/actions/runs/3034862862

### DIFF
--- a/resource-service/errors/errors.go
+++ b/resource-service/errors/errors.go
@@ -99,6 +99,7 @@ var ErrInvalidCredentials = New("credentials need to have ssh or http auth metho
 // Error messages
 
 const ErrMsgCouldNotRetrieveCredentials = "could not read credentials for project %s: %w"
+const ErrMsgCouldEstablishAuthMethod = "could not establish auth method for project %s: %w"
 const ErrMsgInvalidRequestFormat = "Invalid request format"
 const ErrMsgCouldNotSetUser = "could not set git user: %w"
 const ErrMsgCouldNotCreatePath = "could not create path %s: %w"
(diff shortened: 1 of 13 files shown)


resource-service/errors/errors.go
```diff
@@ -99,6 +99,7 @@ var ErrInvalidCredentials = New("credentials need to have ssh or http auth metho
 // Error messages
 
 const ErrMsgCouldNotRetrieveCredentials = "could not read credentials for project %s: %w"
+const ErrMsgCouldEstablishAuthMethod = "could not establish auth method for project %s: %w"
```
bacherfl, line 102: `ErrMsgCouldEstablishAuthMethod` -> `ErrMsgCouldNotEstablishAuthMethod`

odubajDT: fixed, thx

bacherfl (approved):

PR #8844, opened 2022-09-12
feat: Add a multi-integration and multi-stage test

### LINKED ISSUE #8695: Core :: Create a good integration test to substitute continuous delivery
Since we removed helm service and continuous delivery, we currently miss an integration test that runs across multiple stages. 
To test the number of logs in shippy and their usefulness @warber used a test scenario that involves multiple stages and simulates longer times to run. Trigger multiple sequences for the same stage and service using echo service and make it so that sequences overlap and queue each other. He added delays using https://deelay.me/. 

This seems like a nice test candidate to have saved somewhere either as an integration test or as a script. 

- [x] Create an integration test using a shipyard that is configured to let multiple keptn integrations react to a *.triggered event.
- [x] Make sure to have at least 2 stages
- [x] Have multiple sequences

RealAnna:
## This PR

- adds test across 2 stages with two integrations answering to the same event
- verifies all finished events for both stages 
- verifies no pending stage

### Related Issues

Closes #8695 

image
image

### How to test

integration test run : https://github.com/keptn/keptn/actions/runs/3043833675/jobs/4903547005 and https://github.com/keptn/keptn/actions/runs/3234427330

### DIFF
--- a/test/go-tests/test_multiple_integrations.go
+++ b/test/go-tests/test_multiple_integrations.go
@@ -4,13 +4,11 @@ import (
 	"fmt"
 	"github.com/keptn/go-utils/pkg/api/models"
 	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
+	"github.com/stretchr/testify/require"
 	"os"
 	"strings"
 	"testing"
 	"time"
-
-	//models "github.com/keptn/keptn/shipyard-controller/models"
-	"github.com/stretchr/testify/require"
 )
 
 const sequenceMultipleShipyard = `--- 
--- a/test/go-tests/test_multiple_integrations.go
+++ b/test/go-tests/test_multiple_integrations.go
@@ -5,6 +5,7 @@ import (
 	"github.com/keptn/go-utils/pkg/api/models"
 	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
 	"os"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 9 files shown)


test/go-tests/test_webhook.go
```diff
@@ -470,7 +470,7 @@ func Test_ExecutingWebhookTargetingClusterInternalAddressesFails(t *testing.T) {
 
 }
 
-func getWebhookYamlWithSubscriptionIDs(t *testing.T, taskTypes []string, projectName string, webhookYamlWithSubscriptionIDs string) string {
+func GetWebhookYamlWithSubscriptionIDs(t *testing.T, taskTypes []string, projectName string, webhookYamlWithSubscriptionIDs string) string {
```
odubajDT, line 473: maybe we can move this method to test_utils.go ?

test/go-tests/testsuite_gke_test.go
```diff
@@ -8,6 +8,8 @@ func Test_GKE(t *testing.T) {
 	// Shut DownTests
 	t.Run("Test_BackupRestore", Test_BackupRestore)
 	// Common Tests
+	t.Run("Test_multipleIntegrations", Test_multipleIntegrations)
+
```
odubajDT, line 12: don't we want to execute this test also on other platforms?

test/go-tests/test_multiple_integrations.go
```diff
@@ -0,0 +1,331 @@
...
+import (
+	"fmt"
+	"github.com/keptn/go-utils/pkg/api/models"
+	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	//models "github.com/keptn/keptn/shipyard-controller/models"
```
odubajDT, line 12: pls remove if not needed

PR #8983, opened 2022-10-10
fix(bridge): Fix missing update in project settings on project change

### LINKED ISSUE #8981: Bridge :: Changing project from selectbar doesn't fetch the new data in project setting page
## How to reproduce

Create two projects with two different upstreams.
Navigate to the project setting page of the first project.
From the top navigation bar, select the second project. 
The data in the page is not updated with the info of the second project.

Kirdock:
Fixes #8981 

Root cause: The child component that consumes the project data only set the data in `ngOnInit`. A change will not be detected anymore.

Solution: On project change we set the state to loading/init again. We need a loading state because we are fetching data here. This would cause a re-render for the child components, because they are not shown for a short duration, and the data will be correctly set. => No change for child components needed

### DIFF
--- a/bridge/client/app/_views/ktb-settings-view/ktb-project-settings/ktb-project-settings.component.ts
+++ b/bridge/client/app/_views/ktb-settings-view/ktb-project-settings/ktb-project-settings.component.ts
@@ -64,9 +64,8 @@ export class KtbProjectSettingsComponent implements OnInit, OnDestroy, PendingCh
 
   public projectName$: Observable<string | null> = this.route.paramMap.pipe(map((params) => params.get('projectName')));
 
-  public gitInputDataExtended$: Observable<IGitDataExtended | undefined> = this.projectName$.pipe(
-    mergeMap((projectName) => (projectName ? this.dataService.loadPlainProject(projectName) : of(undefined))),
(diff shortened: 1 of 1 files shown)


bridge/client/app/_views/ktb-settings-view/ktb-project-settings/ktb-project-settings.component.ts
```diff
@@ -80,11 +79,20 @@ export class KtbProjectSettingsComponent implements OnInit, OnDestroy, PendingCh
...
-    this.gitInputDataExtended$,
+    this.project$,
     this.projectNames$,
     this.projectCreated$,
   ]).pipe(
-    map(([keptnInfo, metadata, projectName, gitInputDataExtended, projectNames, projectCreated]) => {
+    map(([keptnInfo, metadata, projectName, project, projectNames, projectCreated]) => {
+      let state: ProjectSettingsStatus = ProjectSettingsStatus.LOADED;
+
+      // if project is changed, and it is loading, show loading state
```
heinzburgstaller, line 89: Comment is kind of misleading?

heinzburgstaller (approved):

PR #9007, opened 2022-10-13
feat(installer): Make bridge username configurable from helm values

### LINKED ISSUE #8541: Make bridge username configurable on install
The bridge username appears hardcoded to `keptn`. This issue is to make it configurable at install via helm.

I believe the correct place is here: https://github.com/keptn/keptn/blob/55db322e351d0b94ae0e63a767191f83f820736f/installer/manifests/keptn/templates/core.yaml#L55

@StackScribe please tag as a `good first issue`

afzal442:
Signed-off-by: afzal442 <afzal442@gmail.com>

## This PR

- allows user to configure username on helm install

### Related Issues

Fixes #8541 

### Notes

To test:
```
export USERNAME=mykeptn
helm install --dry-run --set username=$USERNAME --debug [chart name] [chart path]
```


installer/manifests/keptn/values.yaml
```diff
@@ -1227,6 +1226,9 @@ logLevel: "info"
 ## @param prefixPath URL prefix for all Keptn URLs
 prefixPath: ""
 
+## @param keptnUsername for Keptn Bridge login
+keptnUsername: ""
```
mowies, line 1230: Please put this under the `bridge` values

installer/manifests/keptn/values.yaml
```diff
@@ -68,7 +68,6 @@ mongo:
       cpu: 1000m
       memory: 500Mi
 
-## @section Keptn Features
```
mowies, line 71: Please leave this in :)

installer/manifests/keptn/templates/core.yaml
```diff
@@ -108,6 +108,11 @@ spec:
                 secretKeyRef:
                   name: {{ default "keptn-api-token" .Values.apiService.tokenSecretName }}
                   key: keptn-api-token
+            - name: "USERNAME"
+              valueFrom:
+                secretKeyRef:
+                  key:  keptnUsername
+                  name: bridge-credentials
```
mowies, line 115: I don't think you need this here, since the username is used by the bridge directly from the secret.

mowies (changes requested):
Hi @afzal442 , thanks for the contibution! :)
I left some comments for you

installer/manifests/keptn/values.yaml
afzal442: Thanks much @mowies for your quick reply! I loved that.

installer/manifests/keptn/templates/core.yaml
```diff
@@ -52,7 +52,7 @@ metadata:
     app.kubernetes.io/name: bridge
 type: Opaque
 data:
-  BASIC_AUTH_USERNAME: {{ "keptn" | b64enc | quote }}
+  BASIC_AUTH_USERNAME: {{ default "keptn" .Values.bridge.keptnUsername | b64enc | quote }}
```
mowies, line 55: I think we should also add a block above that fetches. the username from the secret if it already exists, like it's done for the bridge password. So something like this:
```
{{- $bridgePassword := (randAlphaNum 20) | b64enc | quote }}
{{- $bridgeUsername := default "keptn" .Values.bridge.keptnUsername | b64enc | quote }}
{{- $bridgesecret := (lookup "v1" "Secret" .Release.Namespace "bridge-credentials") }}
{{- if $bridgesecret }}
  {{- $bridgeUsername = index $bridgesecret.data "BASIC_AUTH_USERNAME" }}
  {{- $bridgePassword = index $bridgesecret.data "BASIC_AUTH_PASSWORD" }}
{{- end -}}
```

And maybe with that it also makes sense to rename the value to `basicAuthUsername` to make a clear distinction between OAuth and HTTP basic auth

installer/manifests/keptn/values.yaml
```diff
@@ -512,6 +512,8 @@ bridge:
     mongoConnectionString: ""
   ## @param bridge.authMsg Overrides Keptn auth command displayed inside the Bridge user menu
   authMsg: ""
+  ## @param bridge.basicAuthUsername Keptn auth username for Bridge user login
+  basicAuthUsername: ""
```
mowies, line 516: Please add the default value here too:
```suggestion
  basicAuthUsername: "keptn"
```
(don't forget to re-generate the helm value readme :) )

mowies (changes requested):

installer/manifests/keptn/values.yaml
afzal442: Thanks again! I made it. 🛩️

Kirdock (approved):

PR #9041, opened 2022-10-17
ci: Create scheduled workflow for govulncheck

### LINKED ISSUE #8937: Implement a workflow to scan Go modules with Govulncheck
This is a follow-up ticket of #8899 
The workflow should be schedule run once per week.
If the initial checks pass, the workflow should be configured to automatically generate an issue if a vulnerability is found.

## DoD
- [ ] The workflow is implemented and scheduled to run once a week
- [ ] The workflow automatically creates an issue if a vulnerability is found

philipp-hinteregger:
Signed-off-by: Philipp Hinteregger <philipp.hinteregger@dynatrace.com>

## This PR

- adds a scheduled workflow to execute govulncheck for every go module

### Related Issues

Fixes #8937

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -1,0 +1,138 @@
+name: Vulnerability checks
+on:
+  schedule:
+    - cron: '0 3 * * 1' # run tests at 1 AM (UTC), every monday (1)
+  workflow_dispatch:
+
+env:
+  GO_VERSION: "~1.18"
+  BRIDGE_ARTIFACT_PREFIX: "BRIDGE"
+  API_ARTIFACT_PREFIX: "API"
+  APPROVAL_SVC_ARTIFACT_PREFIX: "APPROVAL_SVC"
+  DISTRIBUTOR_ARTIFACT_PREFIX: "DISTRIBUTOR"
+  SHIPYARD_CONTROLLER_ARTIFACT_PREFIX: "SHIPYARD_CONTROLLER"
+  SECRET_SVC_ARTIFACT_PREFIX: "SECRET_SVC"
+  RESOURCE_SVC_ARTIFACT_PREFIX: "RESOURCE_SVC"
+  REMEDIATION_SVC_ARTIFACT_PREFIX: "REMEDIATION_SVC"
+  LIGHTHOUSE_SVC_ARTIFACT_PREFIX: "LIGHTHOUSE_SVC"
+  MONGODB_DS_ARTIFACT_PREFIX: "MONGODB_DS"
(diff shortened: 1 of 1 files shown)


.github/workflows/govulncheck.yml
```diff
@@ -0,0 +1,150 @@
...
+
+      - name: Check for vulnerabilities
+        if: matrix.config.artifact != 'bridge2'
+        working-directory: ./${{ matrix.config.working-dir }}
+        run: govulncheck ./...
+
+  create_issue:
+    name: Create GitHub Issue
+    runs-on: ubuntu-22.04
+    needs: security-scans
```
thisthat, line 124: I guess
```suggestion
    needs: vulnerability-check
```

.github/workflows/govulncheck.yml
```diff
@@ -0,0 +1,150 @@
...
+        echo "* Link to run: $GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID" >> vulnerability-found.md
+        if [[ $GITHUB_EVENT_NAME == 'schedule' ]]; then
+          echo "* Triggered by: Scheduled build" >> vulnerability-found.md
+        else
+          echo "* Triggered by: @$GITHUB_ACTOR" >> vulnerability-found.md
+        fi
+        echo "" >> vulnerability-found.md
+        echo "Note: This issue was auto-generated from [govulncheck.yml](.github/workflows/govulncheck.yml)" >> vulnerability-found.md
+
+    - name: Create issue if versions differ
```
thisthat, line 145:
```suggestion
    - name: Create issue
```

thisthat (changes requested):
The PR LGTM - I would, however, put in on hold. There are few imports that need to be adjusted from the example run and if we merge this PR, we would get an issue daily 😅  Hence, I would suggest to merge this after the fixes are in :)

.github/workflows/govulncheck.yml
```diff
@@ -0,0 +1,150 @@
...
+      BUILD_MATRIX: ${{ steps.check_modified_files.outputs.BUILD_MATRIX }}
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v3
+
+    - name: Prepare artifact build matrix
+      id: check_modified_files
+      env:
+        BUILD_EVERYTHING: true
+      run: ./gh-actions-scripts/prepare_changed_artifacts.sh ""
```
mowies, line 92: You could use a much simpler matrix that doesn't even need to be dynamically generated here...

mowies (approved):

PR #9047, opened 2022-10-18
fix(shipyard-controller): Handle HTTP error response range when provisioning GIT repository

### LINKED ISSUE #9046: ShipyardController: Call to GIT provisioning API does not handle errors corrrectly
The logic responsible for calling the GIT provisioning API only handles 409 Error and assumes that it receives a well defined response payload otherwise. This results a de-serialization error shadowing the root cause if something goes wrong.

warber:
fixes #9046

### DIFF
--- a/shipyard-controller/internal/provisioner/repository_provisioner.go
+++ b/shipyard-controller/internal/provisioner/repository_provisioner.go
@@ -53,8 +53,8 @@ func (rp *RepositoryProvisioner) ProvideRepository(projectName, namespace string
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode == http.StatusConflict {
-		return nil, fmt.Errorf(common.UnableProvisionInstance, http.StatusText(http.StatusConflict))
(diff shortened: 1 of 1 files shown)


shipyard-controller/internal/provisioner/repository_provisioner.go
```diff
@@ -53,8 +53,8 @@ func (rp *RepositoryProvisioner) ProvideRepository(projectName, namespace string
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode == http.StatusConflict {
-		return nil, fmt.Errorf(common.UnableProvisionInstance, http.StatusText(http.StatusConflict))
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
```
thisthat, line 56: Maybe is worth to log the response data. I am thinking about a 302, this way an operator can follow through the redirect. 
WDYT?

thisthat (approved):

PR #9055, opened 2022-10-19
fix(bridge): Show correct border color for task items in sequence view

### LINKED ISSUE #9054: Bridge :: Showing sequences in red but the sequence is passing
## Problem

Bridge shows incorrectly the status of the sequence. See the screenshot.
If an SLI step fails, Bridge colors the sequence in red although the overall status of the sequence is passed.
This is not a problem in the sequence list as it is correctly colored green.

## How to reproduce

1. Create a project with a service and install an SLI provider
2. Add an SLI.yaml file for your service which has misconfigured SLI definitions
3. Do NOT add any SLO.yaml file
4. Trigger an evaluation

TannerGilbert:
## This PR

- Fixes problem with incorrect color being displayed if one subtask fails, but the task passes.

### Related Issues

Resolves #9054 

## Notes

Screenshot from 2022-11-02 14-46-47

Signed-off-by: gilbert.tanner <gilbert.tanner@dynatrace.com>

### DIFF
--- a/bridge/shared/models/trace.ts
+++ b/bridge/shared/models/trace.ts
@@ -227,10 +227,8 @@ export class Trace {
 
   public isFaulty(stageName?: string): boolean {
     let result = false;
-    if (this.data) {
-      if (this.isFailed()) {
-        result = stageName ? this.data.stage === stageName : true;
-      }
+    if (this.data && this.isFailed()) {
+      result = stageName ? this.data.stage === stageName : true;
     }
     return result;
   }
--- a/bridge/shared/models/trace.ts
+++ b/bridge/shared/models/trace.ts
@@ -254,7 +254,7 @@ export class Trace {
   }
 
   public getFinishedEvent<T extends Trace>(this: T): T | undefined {
-    return (this.isFinishedEvent() ? this : this.traces.find((t) => t.isFinishedEvent())) as T;
+    return (this.isFinishedEvent() ? this : this.traces.find((t) => t.isFinishedEvent(this.getShortType()))) as T;
   }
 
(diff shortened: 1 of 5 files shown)


bridge/shared/models/trace.ts
```diff
@@ -228,7 +228,7 @@ export class Trace {
   public isFaulty(stageName?: string): boolean {
     let result = false;
     if (this.data) {
-      if (this.isFailed() || this.traces.some((t) => t.isFaulty())) {
+      if (this.isFailed()) {
```
Kirdock, line 231: This may not work with the webhook.
Example
```
evaluationSequence.triggered
-> webhook.triggered
-> webhook.started
-> trigger delivery
---->start, finished
-> trigger evaluation
---->start, finished
-> webhook.finished
-> evaluationSequence.finished
```
for `isFailed`, it takes the result of the first found finished event.
If the webhook succeeds, the overall sequence is success but it may be failed.

Correct me if I'm wrong here

Kirdock (changes requested):

bridge/shared/models/trace.ts
TannerGilbert: You're right this doesn't work with webhook events. With help of @ermin-muratovic this should be fixed now. There is also a UI test that checks for it.

heinzburgstaller, line 231: please refactor to single line if

heinzburgstaller:
please add screenshot to PR's description

heinzburgstaller (changes requested):

heinzburgstaller (approved):

PR #9077, opened 2022-10-25
ci: Use new action to get changed files

mowies:
### This PR
- exchanges the `jitterbit/get-changed-files@v1` action for `tj-actions/changed-files@v33` since the old action doesn't seem to be maintained anymore and also has some issues with github actions


philipp-hinteregger (approved):
LGTM!

PR #9107, opened 2022-11-03
fix(resource-service): Force checkout event if unstaged changes are present

odubajDT:

### DIFF
--- a/resource-service/common/git.go
+++ b/resource-service/common/git.go
@@ -457,6 +457,7 @@ func (g *Git) CheckoutBranch(gitContext common_models.GitContext, branch string)
 
 	err := g.checkoutBranch(gitContext, &git.CheckoutOptions{
 		Branch: b,
+		Force:  true,
 	})
 	if err != nil {
 		if errors.Is(err, plumbing.ErrReferenceNotFound) {
--- a/resource-service/common/git.go
+++ b/resource-service/common/git.go
@@ -243,12 +243,20 @@ func (g Git) StageAndCommitAll(gitContext common_models.GitContext, message stri
 
 	id, err := g.commitAll(gitContext, message)
 	if err != nil {
+		err := g.ResetHard(gitContext, "HEAD~0")
+		if err != nil {
+			logger.WithError(err).Warn("could not reset after")
+		} else {
+			logger.Warn("untracked changes were removed")
+		}
 		return "", fmt.Errorf(kerrors.ErrMsgCouldNotCommit, gitContext.Project, mapError(err))
 	}
 	rollbackFunc := func() {
 		err := g.ResetHard(gitContext, "HEAD~1")
 		if err != nil {
 			logger.WithError(err).Warn("could not reset")
+		} else {
+			logger.Warn("commited changes were removed")
 		}
 	}
 	err = g.Pull(gitContext)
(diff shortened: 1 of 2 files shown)


resource-service/common/git.go
```diff
@@ -457,6 +457,7 @@ func (g *Git) CheckoutBranch(gitContext common_models.GitContext, branch string)
 
 	err := g.checkoutBranch(gitContext, &git.CheckoutOptions{
 		Branch: b,
+		Force:  true,
```
agrimmer, line 460: Do we know what changes would be lost if we do a force checkout?

agrimmer: Only the resource-service can produce unstaged changes on our local branch.

thisthat: I don't have much insights, but I would expect that a failed operation (e.g., due to ctx cancellation) can trigger unstashed data

thisthat (approved):
LGTM - Given that the upstream contains always the truth, I think it makes sense to force the operation.

agrimmer (changes requested):
But what data are we then losing if a write operation of resource-serivce fails?

thisthat:
@agrimmer does it make sense what data we lose? The truth is in the upstream. If there was an op that didn't succeed, it should be re-executed in a clean state, which is what this PR is aiming to introduce.

agrimmer:
With this change, we could lose data that a Keptn service (e.g. dynatrace-service) tried to write. 
Do we have appropriate re-try logic in place?

resource-service/common/git.go
```diff
@@ -243,12 +243,20 @@ func (g Git) StageAndCommitAll(gitContext common_models.GitContext, message stri
 
 	id, err := g.commitAll(gitContext, message)
 	if err != nil {
+		err := g.ResetHard(gitContext, "HEAD~0")
+		if err != nil {
+			logger.WithError(err).Warn("could not reset after")
```
thisthat, line 248:
```suggestion
			logger.WithError(err).Warn("could not reset after commitAll")
```

resource-service/common/git.go
```diff
@@ -243,12 +243,20 @@ func (g Git) StageAndCommitAll(gitContext common_models.GitContext, message stri
 
 	id, err := g.commitAll(gitContext, message)
 	if err != nil {
+		err := g.ResetHard(gitContext, "HEAD~0")
```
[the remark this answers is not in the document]
odubajDT: sure, will look at it

thisthat (approved):

resource-service/common/git.go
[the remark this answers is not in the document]
odubajDT: Due to the structure of the code it is basically impossible to create a proper test to verify the fix. The fix is therefore verified via testing the `ResetHard` method directly in multiple scenarios, where it is quite clear, that when `commitAll` method in `StageAndCommitAll` method fails due to error when adding or commiting changes, methos `ResetHard` should take care of the cleanup

agrimmer (approved):
LGTM

odubajDT:
In the current state, it is basically not possible to test `StageAndCommitAll` function directly, so we tested the scenario, we are aiming for, which is: create changes + reseting the git + checking if the changes are stashed.

The reason why it is not possible to create a direct test is, that in function `StageAndCommitAll` we need to mock the `commitAll` function. In this case, both functions are part of the same structure/interface, so it is not possible to mock only one of them. Either both (which is useless for testing) of none (in this case, `commitAll` is never able to return an error). If we want to test it, we need a big refactoring, which is not in plan in the current state.

From this reason we used only testing the `ReseHard` function itself, as it is quite clear, that when `commitAll` returns an error, `ResetHArd` is executed and should take care of the cleanup. This should cover testing the changes we made in an indirect way.

PR #9108, opened 2022-11-03
fix(resource-service): Force checkout event if unstaged changes are present

odubajDT:


odubajDT:
close in favor of https://github.com/keptn/keptn/pull/9113

PR #9112, opened 2022-11-07
feat(resource-service): Use libgit2 as a fallback when cloning repo

### LINKED ISSUE #9110: Use git2go as a fallback for "unexpected client" errors during the clone of an Azure repo
On Azure hosted upstream repositories, we experienced an error where the resource service could not authenticate itself on an Azure hosted upstream repository, due to a two way handshake protocol used by Azure, which is not implemented by the go-git library used in the resource service. A way to solve this is to use the [libgit2go](https://github.com/libgit2/git2go) library, which supports this protocol to do the clone of a repository, if an `unexpected client` error occurs.

A rough PoC is implemented in #9109 

Note: In this PR, all git functions that interact with the remote repository are implemented, however in fact only the Clone command is needed to establish a successful authentication to an Azure upstream. After that, all operations can be done using the go-git library again.

Note: that with these changes in place, the resource service will only compile if `libgit2` is installed on the machine it is running on. Also, version handling is more complicated with this library, because the latest libgit2 version (1.5.0) is not available on alpine 3.16 - the latest version here is 1.4.3 (see https://pkgs.alpinelinux.org/packages?name=libgit2-dev&branch=v3.16&repo=&arch=&maintainer=). However, there is no [libgit2go](https://github.com/libgit2/git2go) release that supports this version. As a result, I had to downgrade the alpine image to 3.15, where libgit2 1.3.0, which is also supported by [libgit2go](https://github.com/libgit2/git2go) is available

bacherfl:
Closes #9110 

Note that with these changes in place, the resource service will only compile if `libgit2` is installed on the machine it is running on. Also, version handling is more complicated with this library, because the latest libgit2 version (1.5.0) is not available on alpine 3.16 - the latest version here is 1.4.3 (see https://pkgs.alpinelinux.org/packages?name=libgit2-dev&branch=v3.16&repo=&arch=&maintainer=). However, there is no [libgit2go](https://github.com/libgit2/git2go) release that supports this version. As a result, I had to downgrade the alpine image to 3.15, where libgit2 1.3.0, which is also supported by [libgit2go](https://github.com/libgit2/git2go) is available

Integration test run: https://github.com/keptn/keptn/actions/runs/3418103500/jobs/5689997683


resource-service/Dockerfile
```diff
@@ -43,7 +45,7 @@ RUN sed -i "s/version: develop/version: ${version}/g" docs/swagger.yaml
 RUN GOOS=linux go build -ldflags '-linkmode=external' -gcflags="${SKAFFOLD_GO_GCFLAGS}" -v main.go
 
 # Use a Docker multi-stage build to create a lean production image.
-FROM alpine:3.16 as production
+FROM alpine:3.15 as production
```
thisthat, line 48: is there a way to use 3.16 with the lib? We would introduce some open vulnerabilities with 3.15 :/

resource-service/common/git.go
```diff
@@ -164,6 +169,12 @@ func (g Git) rewriteDefaultBranch(path string, env envconfig.EnvConfig) error {
 
 func (g Git) init(gitContext common_models.GitContext, projectPath string) (*git.Repository, error) {
 	init, err := g.git.PlainInit(projectPath, false)
+	if err != nil && !errors.Is(err, git.ErrRepositoryAlreadyExists) {
+		return nil, err
+	}
```
thisthat, line 174: Given the if err != nil at line 178, I guess this can be removed

resource-service/common/go-git.go
```diff
@@ -17,40 +26,55 @@ func (t GogitReal) PlainOpen(path string) (*git.Repository, error) {
...
-	Head() (*plumbing.Reference, error)
-	Reference(name plumbing.ReferenceName, resolved bool) (*plumbing.Reference, error)
-	References() (storer.ReferenceIter, error)
-	Worktree() (*git.Worktree, error)
-	ResolveRevision(rev plumbing.Revision) (*plumbing.Hash, error)
-}*/
+type Git2Go struct{}
+
+func (g Git2Go) PlainClone(gitContext common_models.GitContext, path string, isBare bool, o *git.CloneOptions) (*git.Repository, error) {
+	_, err := git2go.Clone(o.URL, path, &git2go.CloneOptions{
```
thisthat, line 50: what variable we are ignoring and for which purpose?

resource-service/config/config.go
```diff
@@ -7,10 +7,14 @@ import (
...
 var Global EnvConfig
 
+const GitImplementationLibGit2 = "libgit2"
+const GitImplementationGoGit = "go-git"
+
 type EnvConfig struct {
 	LogLevel                         string `envconfig:"LOG_LEVEL" default:"info"`
 	DirectoryStageStructure          bool   `envconfig:"DIRECTORY_STAGE_STRUCTURE" default:"false"`
 	DefaultRemoteGitRepositoryBranch string `envconfig:"DEFAULT_REMOTE_GIT_BRANCH" default:"master"`
+	GitImplementation                string `envconfig:"GIT_IMPLEMENTATION" default:"go-git"`
```
thisthat, line 17: should we add an env variable in the helm deployment? 🤔

resource-service/Dockerfile
bacherfl: by using the repo of 3.15 only for installing libgit2 it was possible to keep using 3.16 again

resource-service/common/go-git.go
```diff
@@ -17,40 +26,55 @@ func (t GogitReal) PlainOpen(path string) (*git.Repository, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	repo, err := git.PlainOpen(path)
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = repo.Head()
```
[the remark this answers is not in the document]
bacherfl: added some comments

resource-service/common/go-git.go
bacherfl: added some comments

RealAnna (approved):

resource-service/common/git.go
```diff
@@ -164,6 +169,14 @@ func (g Git) rewriteDefaultBranch(path string, env envconfig.EnvConfig) error {
...
+		if errors.Is(err, git.ErrRepositoryAlreadyExists) {
+			init, err = g.git.PlainOpen(projectPath)
+		} else {
+			return nil, err
+		}
+	}
+
 	if err != nil {
 		return nil, err
 	}
```
thisthat, line 182: I would move this lines at line 174. Here they are confusing since there is 2 if err != nil in a row

thisthat (approved):

PR #9114, opened 2022-11-07
fix(bridge): Fix relative weight calculation (#9099)

Kirdock:
Backport of #9099


thisthat:
Would it make sense also to add a test that covers the usage of `getRelativeSliWeight`?

Kirdock:
> Would it make sense also to add a test that covers the usage of `getRelativeSliWeight`?

Yes, but then we would need anoher backport to the main branch. This change is something that should have been considered in the original PR.

Kirdock:
> Would it make sense also to add a test that covers the usage of `getRelativeSliWeight`?

Yes, but then we would need anoher backport to the main branch. This change is something that should have been considered in the original PR.

thisthat:
Fair point, this would violate the cherry-picking. Therefore, I would suggest that it is not needed for this PR

thisthat (approved):

PR #9115, opened 2022-11-07
ci: Remove cosign image signing for now

mowies:


mattmoor:
I am curious why you disabled signing?  Did you run into some issue?

I happed to come across this compiling a list of open source projects signing with cosign, and figured I'd ask.  i see it's still setup on `main`, but there wasn't a lot of explanation here as context.  thanks!

mowies:
Yes, we had some troubles during our last release where Fulcio was returning an invalid key all of a sudden. This is the first time we had any issues with it though.
Here's a screenshot of the error from our release pipeline run (this happened with all our released container images):
Screenshot 2022-11-16 at 07 46 44

PR #9116, opened 2022-11-07
fix(shipyard-controller): Return 4xx error responses for upstream repository problems

### LINKED ISSUE #9038: Core :: Better error codes for project creation
## Technical details

When a project is created with an already initialized repository, `shipyard-controller` returns a 500 error instead of a more descriptive code.

## Acceptance Criteria

- [x] If `resource-service` returns 409 Conflict`, then also `shipyard-controller` shall return `409 Conflict`.

## References

- https://github.com/keptn/keptn/blob/0.19.x/shipyard-controller/internal/handler/projecthandler.go#L353

odubajDT:

## This PR

- unified HTTP error response to 404 when upstream repository is not found
- return 409 HTTP error response when the upstream repository is already initialized when creating or updating project

BREAKING CHANGE: Keptn returns 404 instead of 400 HTTP error code when creation of the project fails when upstream repository is not found. Also returns 409 instead of 500 HTTP error code when creation or update of project fails due to upstream repository is already initialized.

### Related Issues

Fixes #9038 

Screenshot from 2022-11-07 16-32-15

### DIFF
--- a/shipyard-controller/internal/handler/projecthandler.go
+++ b/shipyard-controller/internal/handler/projecthandler.go
@@ -346,7 +346,7 @@ func (ph *ProjectHandler) CreateProject(c *gin.Context) {
 			SetConflictErrorResponse(c, err.Error())
 			return
 		}
-		if errors.Is(err, common.ErrAlreadyInitializedRepository) {
+		if err.Error() == common.AlreadyInitializedRepositoryMsg {
 			SetConflictErrorResponse(c, err.Error())
 			return
 		}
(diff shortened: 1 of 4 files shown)


shipyard-controller/internal/handler/projecthandler.go
```diff
@@ -418,6 +422,10 @@ func (ph *ProjectHandler) UpdateProject(c *gin.Context) {
 			SetBadRequestErrorResponse(c, err.Error())
 			return
 		}
+		if err.Error() == common.AlreadyInitializedRepositoryMsg {
```
bacherfl, line 425: this seems to be a duplicate

shipyard-controller/internal/handler/projecthandler.go
```diff
@@ -346,6 +346,10 @@ func (ph *ProjectHandler) CreateProject(c *gin.Context) {
 			SetConflictErrorResponse(c, err.Error())
 			return
 		}
+		if err.Error() == common.AlreadyInitializedRepositoryMsg {
```
bacherfl: here we should use `strings.Contains()`, as an exact string comparison will not work for wrapped errors

odubajDT: fixed

bacherfl (approved):

PR #9118, opened 2022-11-08
ci: Remove dynatrace github actions

RealAnna:
## This PR

- removes dynatrace action: it uses old nodejs dependances, was updated last time an year ago and we do not need these metrics anyhow
-  updates echo action

### Integration tests
https://github.com/keptn/keptn/actions/runs/3418713575


odubajDT (dismissed):

test/go-tests/test_utils.go
```diff
@@ -179,11 +179,17 @@ func GetInternalKeptnAPI(ctx context.Context, internalService, localPort string,
 	if err != nil {
 		return nil, err
 	}
-	keptnInternalAPI, err := NewAPICallerWithBaseURL("http://127.0.0.1:" + localPort)
+	var internalKeptnAPI *APICaller
+
+	err = retry.Retry(func() error {
+		internalKeptnAPI, err = NewAPICallerWithBaseURL("http://127.0.0.1:" + localPort)
+		return err
+	}, retry.NumberOfRetries(5), retry.DelayBetweenRetries(1*time.Minute))
```
christian-kreuzberger-dtx, line 187: 1 Minute? That's quite a big delay between retries. Wouldn't it make sense to lower that in the range of several seconds?

christian-kreuzberger-dtx:
@mowies this PR completely gets rid of pipeline observability, is that intended? If so, the PR title is misleading and/or the removal of pipeline observability should be moved into another PR.

@RealAnna 
> removes deprecated dynatrace action
Where is this coming from? I'm not seeing a deprecation notice for https://github.com/dynatrace-oss/dynatrace-github-action 

Not going to block the merge of the PR, just raising some questions as it seems appropriate.

mowies:
@christian-kreuzberger-dtx thanks for bringing attention to this. The removal was discussed internally, since the action has really old dependencies and spams our pipelines with warnings. Also, with less feature work and more maintenance going on in keptn/keptn and additionally some great pipeline optimization work that happened, we don't really need the monitoring of timings anymore.
You're definitely right about the PR title though, I updated it :D

test/go-tests/test_utils.go
RealAnna: tested multiple iterations of this, in all cluster it works without even retry logic but k3d 24 fails all the time even with 40sec :(

mowies (approved):

mowies:
@RealAnna Could you please also backport this to the 0.19.x branch? Then we can also get rid of the access token

PR #9166, opened 2022-11-17
feat(bridge): Unify logging

### LINKED ISSUE #9164: Bridge :: Uniform logging
## Details 

Currently the Bridge server produces logs in two format:

```
GET /api/controlPlane/v1/sequence/performance-test-27?pageSize=5 - - ms - -
[Keptn] 2022-11-17T06:49:06.058Z error   [API] [object Object]
```

For an easier processing, they should be unified as in the second line

## DoD

- Unique format of log lines

thisthat:

## This PR

Fixes #9164 

```
[Keptn] 2022-11-17T07:44:56.153Z info    [Express] GET /api/controlPlane/v1/project?disableUpstreamSync=true&pageSize=50 200 
[Keptn] 2022-11-17T07:44:56.248Z info    [Express] GET /api/controlPlane/v1/sequence/test?pageSize=5 200
[Keptn] 2022-11-17T07:45:03.554Z info    [API] get http://api-gateway-nginx/api/controlPlane/v1/project/test 200
```

### DIFF
--- a/bridge/server/app.ts
+++ b/bridge/server/app.ts
@@ -6,7 +6,6 @@ import * as http from 'http';
 import { contentSecurityPolicy, frameguard, noSniff, permittedCrossDomainPolicies, xssFilter } from 'helmet';
 import express, { Express, NextFunction, Request, Response } from 'express';
 import { fileURLToPath, URL } from 'url';
-import logger from 'morgan';
 import { ComponentLogger } from './utils/logger';
 import cookieParser from 'cookie-parser';
 import AdmZip from 'adm-zip';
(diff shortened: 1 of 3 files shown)


bridge/server/app.ts
```diff
@@ -74,7 +74,24 @@ async function init(configuration: BridgeConfiguration): Promise<Express> {
   );
 
   // add some middlewares
-  app.use(logger('dev'));
+  const logExpress = new ComponentLogger('Express');
+  app.use(
```
Kirdock, line 78: Do we even need morgan then?

Couldn't we write something like this?
```ts
app.use((req: Request, response: Response, next: NextFunction) =>
{
  logExpress.log(...);
  next();
});
```

bridge/server/package.json
```diff
@@ -24,7 +24,6 @@
     "express-session": "1.17.3",
     "helmet": "5.1.1",
     "mongodb": "4.10.0",
-    "morgan": "1.10.0",
```
Kirdock, line 27: It's not removed in the lock file right?
Could you remove these dependencies via `yarn remove morgan @types/morgan`? Then yarn will take care that the package.json and the lock file is corrected

Kirdock (changes requested):

Kirdock (approved):

Issue #9111, opened 2022-11-04
Core :: Improve logging levels in Keptn core services


thisthat:
## Action Item

At the moment, the majority of logs is either at info or error level. The goal of this ticket is to re-evaluate the log levels used of each service.

## DoD

The following services use verbose/info/debug/warning/error log levels correctly:

- [x] shipyard: https://github.com/keptn/keptn/pull/9167
- [x] bridge: https://github.com/keptn/keptn/pull/9183
- [x] api-service: https://github.com/keptn/keptn/pull/9168
- [x] approval-service: https://github.com/keptn/keptn/pull/9171
- [x] lighthouse-service: https://github.com/keptn/keptn/pull/9169
- [x] mongo-datastore: https://github.com/keptn/keptn/pull/9170
- [x] remediation-service: No changes required
- [x] resource-service: https://github.com/keptn/keptn/pull/9172
- [x] secret-service: https://github.com/keptn/keptn/pull/9175
- [x] webhook-service: https://github.com/keptn/keptn/pull/9176

thisthat:
All services have improved their logging

PR #9167, opened 2022-11-17
chore(shipyard-controller): Clean up logs/errors

bacherfl:
Part of #9111

### DIFF
--- a/shipyard-controller/internal/controller/shipyardcontroller.go
+++ b/shipyard-controller/internal/controller/shipyardcontroller.go
@@ -2,7 +2,6 @@ package controller
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"github.com/keptn/keptn/shipyard-controller/internal/common"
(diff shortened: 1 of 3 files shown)


shipyard-controller/internal/controller/shipyardcontroller.go
```diff
@@ -239,7 +238,8 @@ func (sc *ShipyardController) handleSequenceTriggered(event apimodels.KeptnConte
 	sequence, err := GetTaskSequenceInStage(eventScope.Stage, taskSequenceName, shipyard)
 	if err != nil {
 		msg := fmt.Sprintf("Unable to start sequence %s: %v", taskSequenceName, err)
-		log.Error(msg)
+		// use 'warn' level here since a sequence not being present is not an error in the service
+		log.Warn(msg)
```
odubajDT, line 242:
```suggestion
		log.Warnf("Unable to start sequence %s: %v", taskSequenceName, err)
```

bacherfl: The `msg` var is then also passed to the function that emits the finished event containing that message, so I would prefer to keep it this way

odubajDT (approved):

PR #9168, opened 2022-11-17
chore(api): Clean up logs/errors

bacherfl:
Part of #9111

### DIFF
--- a/api/handlers/metadata.go
+++ b/api/handlers/metadata.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -61,7 +62,7 @@ type metadataHandler struct {
 }
 
 func (h *metadataHandler) getMetadata() middleware.Responder {
-	logger.Info("API received a GET metadata event")
+	logger.Debug("API received a GET metadata event")
 
 	namespace := os.Getenv("POD_NAMESPACE")
 	automaticProvisioningURL := os.Getenv("AUTOMATIC_PROVISIONING_URL")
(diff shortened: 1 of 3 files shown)


api/handlers/metadata.go
```diff
@@ -93,19 +94,23 @@ func (h *metadataHandler) getMetadata() middleware.Responder {
...
+		return "", fmt.Errorf("no K8s client available")
 	}
 	bridgeDeployment, err := h.k8sClient.AppsV1().Deployments(namespace).Get(context.TODO(), "bridge", metav1.GetOptions{})
 	if err != nil {
-		return "", fmt.Errorf("unable to get bridge version %w", err)
+		return "", fmt.Errorf("unable to get bridge version: %w", err)
+	}
+
+	if len(bridgeDeployment.Spec.Template.Spec.Containers) == 0 {
+		return "", errors.New("bridge deployment does not contain any image")
```
odubajDT, line 105:
```suggestion
		return "", fmt.Errorf("bridge deployment does not contain any image")
```
let's try to be consistent

api/handlers/metadata.go
```diff
@@ -93,19 +94,23 @@ func (h *metadataHandler) getMetadata() middleware.Responder {
...
 	}
 
 	v := strings.Split(bridgeDeployment.Spec.Template.Spec.Containers[0].Image, ":")
 	if len(v) >= 2 {
 		return v[1], nil
 	}
 
-	return "", fmt.Errorf("unable to get bridge version")
+	return "", errors.New("could not determine bridge version from deployment")
```
odubajDT, line 113:
```suggestion
	return "", fmt.Errorf("could not determine bridge version from deployment")
```

api/handlers/metadata.go
bacherfl: if we don't pass any parameters it does not make sense to use `fmt.Errorf`

odubajDT (approved):

PR #9183, opened 2022-11-21
chore(bridge): Improve Bridge logging levels and messages

thisthat:

Part of #9111

### DIFF
--- a/bridge/server/app.ts
+++ b/bridge/server/app.ts
@@ -75,7 +75,7 @@ async function init(configuration: BridgeConfiguration): Promise<Express> {
   // add some middlewares
   const logExpress = new ComponentLogger('Express');
   app.use((req: Request, res: Response, next: NextFunction) => {
-    logExpress.info(`${req.method} ${req.url} ${res.statusCode} :: ${logExpress.prettyPrint(req.rawHeaders)}`);
+    logExpress.info(`${req.method} ${req.url} ${res.statusCode} :: ${logExpress.prettyPrint(req.headers)}`);
     next();
   });
   app.use(express.json());
--- a/bridge/server/app.ts
+++ b/bridge/server/app.ts
@@ -145,7 +145,7 @@ async function setBasicAUTH(app: Express, auth: AuthConfig): Promise<void> {
     ) {
       updateBucket(!!(login || password), auth, userIP);
 
-      log.error(`Access denied for IP: ${userIP}`);
+      log.error(`Access denied`);
       res.set('WWW-Authenticate', 'Basic realm="Keptn"');
       next({ response: { status: 401 } });
       return;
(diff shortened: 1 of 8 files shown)


heinzburgstaller (approved):

bridge/server/app.ts
```diff
@@ -309,9 +312,25 @@ function handleError(err: any, req: Request, res: Response, authType: AuthType):
...
+    delete filteredHeaders[item];
+  }
+  // remove additional sec headers
+  Object.keys(headers).forEach((key) => {
+    if (key.startsWith('sec')) {
+      delete filteredHeaders[key];
+    }
+  });
+  return filteredHeaders;
+}
```
heinzburgstaller, line 334: You can try to be more functional here. Something like:
```ts
const record = {...headers};
const filtered = Object.entries(record)
  .filter(entry => !(entry[0].startsWith('sec') || ['cookies', 'user-agent'].includes(entry[0])))
  .reduce((prev, current) => ({...prev, [current[0]]: current[1]}), {});
return filtered;
```

bridge/server/app.ts
```diff
@@ -134,7 +137,7 @@ async function setBasicAUTH(app: Express, auth: AuthConfig): Promise<void> {
     userIP = userIP instanceof Array ? userIP[0] : userIP;
 
     if (userIP && isIPThrottled(userIP, auth)) {
-      log.error('Request limit reached');
+      log.error(`Request limit reached for IP ${userIP}`);
```
heinzburgstaller, line 140: Is IP logging allowed? GDPR wise?

bridge/server/app.ts
```diff
@@ -309,9 +312,25 @@ function handleError(err: any, req: Request, res: Response, authType: AuthType):
   }
 
   printError(err);
-  log.info(`Response status: ${err.response?.status}`);
+  log.error(`Response status ${err.response?.status} for ${req.method} ${req.url}`);
 
   return err.response?.status || 500;
 }
 
+function filterHeaders(headers: IncomingHttpHeaders): IncomingHttpHeaders {
```
heinzburgstaller, line 320: A unit test would be nice

bridge/server/app.ts
thisthat: I think it makes very hard to follow the logic and purpose of the code.

bridge/server/app.ts
thisthat: I know there is an exception for data collection if it is for legitimate interest, which I guess this case fits since someone is trying to misuse the product. However, better be safe than sorry :) I will remove it to be on the safe side

bridge/server/app.ts
heinzburgstaller: Then just add named intermediate steps:
```ts
const record = {...headers};
const removeSecurityRelatedHeaders = (entry) =>  !(entry[0].startsWith('sec') || ['cookies', 'user-agent'].includes(entry[0]));
const toRecord = (prev, current) => ({...prev, [current[0]]: current[1]});
const filtered = Object.entries(record)
  .filter(removeSecurityRelatedHeaders)
  .reduce(toRecord, {});
return filtered;
```

thisthat: I agree that this is more readable. However, the whole code base is built without a functional and I would propose to follow similar code pattern that are already there.

heinzburgstaller (approved):

PR #9200, opened 2022-11-29
fix(bridge): Fix navigation to blocking sequence

### LINKED ISSUE #9199: Bridge: Navigate to blocking sequence does not work
If the selected waiting sequence does not have any stages (sequence-state `stages` is empty), the "go to blocking sequence"-call may fail because the blocking stage is not the first available one.

In this case the blocking stage was prod but the request was filtered for staging, because it is the first one in the shipyard file.

image

Kirdock:
Fixes #9199
Also fixes another bug that the "go to blocking sequence" button did not work immediately (two clicks needed) if the blocking sequence was not initially loaded.

### DIFF
--- a/bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.ts
+++ b/bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.ts
@@ -159,7 +159,7 @@ export class KtbSequenceViewComponent implements OnDestroy {
       filter((state) => state.sequenceInfo?.state === SequencesState.LOAD_UNTIL_ROOT)
     );
     combineLatest([this.route.paramMap, loadUntilRootTriggered$])
-      .pipe(takeUntil(this.unsubscribe$))
+      .pipe(takeUntil(this.unsubscribe$), take(1))
       .subscribe(([params, state]) => {
         this.setParams(params, state);
       });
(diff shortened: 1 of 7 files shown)


bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.ts
```diff
@@ -423,12 +422,30 @@ export class KtbSequenceViewComponent implements OnDestroy {
...
+    if (lastEventStage) {
+      const lastIndex = stages.indexOf(lastEventStage);
+      // if the stage is finished it would be in sequence.stages therefore we don't need to check if the stage is actually not finished
+      if (lastIndex !== -1) {
+        return lastIndex;
+      }
+      return lastIndex + 1;
+    }
+
+    return 0;
```
heinzburgstaller, line 441: Please improve by "early return", no nested IFs

bridge/client/app/_views/ktb-sequence-view/ktb-sequence-view.component.ts
```diff
@@ -456,7 +473,15 @@ export class KtbSequenceViewComponent implements OnDestroy {
...
-          this.dataService.loadUntilRoot(sequenceExecution.scope.project, sequenceExecution.scope.keptnContext);
+          this.dataService
+            .loadUntilRoot(sequenceExecution.scope.project, sequenceExecution.scope.keptnContext)
+            .subscribe((sequences) => {
+              const newSequence = sequences.find((seq) => seq.shkeptncontext === sequenceExecution.scope.keptnContext);
+              if (newSequence) {
+                this.selectSequence({ sequence: newSequence });
+                this.loadTraces(newSequence, undefined, sequenceExecution.scope.stage);
+              }
+            });
```
heinzburgstaller, line 484: There should be a way to get rid of the double subscribe? By chaining with RXJS operators...

heinzburgstaller (changes requested):

heinzburgstaller (approved):

PR #9202, opened 2022-11-29
fix(bridge): Fix incorrect score-bar representation in evaluation

### LINKED ISSUE #9201: Bridge: Incorrect progress bar for score
There may be the case that an evaluation has a score of 100 but failed because there aren't any indicator results.
If you click on some evaluation which fills the bar and then go back to the one that does not fill it, the bar is filled instead of being empty.

initial evaluation:
image

select a valid one:
image

go back to previous one:
image

As you can see, the progress bar is incorrectly filled

Kirdock:
Fixes #9201


heinzburgstaller (approved):
Maybe a (unit) test is needed?

PR #9334, opened 2022-12-19
fix(shipyard-controller): Add new stage to sequence state immediately after receiving triggered event

bacherfl:
This will prevent sequences that have been finished in one stage from being displayed as waiting when they are actually waiting in the next stage


shipyard-controller/main_test.go
```diff
@@ -1027,7 +1027,7 @@ func Test__main_SequenceStateParallelStages(t *testing.T) {
 			return false
 		}
 		return true
-	}, 3*time.Second, 100*time.Millisecond)
+	}, 10*time.Second, 100*time.Millisecond)
```
bacherfl, line 1030: technically not related to this PR, but this should improvve the reliability of the component tests which have been flaky lately

PR #9391, opened 2023-01-01
feat: Publish SBOM for Keptn artifacts/images

### LINKED ISSUE #7163: Publish SBOM for Keptn artifacts/images
It would be really nice to publish SBOMs for all Keptn deliverables, preferably in the SPDX format so that it can be consumed by the Linux Foundation tooling ecosystem and downstream projects.

Scope:
- [ ] Keptn service
- [ ] Keptn integration services
- [ ] Keptn CLI

riteshsonawane1372:
## This PR

Added Sboms for Keptn services and CLI using the anchore SBOM action 

### Related Issues

Fixes #7163


mowies (commented):
This is a good start! But there's some stuff missing still.
First of all, you could use the matrix we already have in place in the `docker_build` job.
Something like this:
```
build_sboms:
  name: Build SBOMs
  needs: [release-please, prepare]
  strategy:
    matrix: ${{ fromJson(needs.prepare.outputs.BUILD_MATRIX) }}
  steps:
  - name: SBOM ${{ matrix.config.artifact }}
    uses:  anchore/sbom-action@v0.13.1
    with:
      format: spdx
      path: ./${{ matrix.config.working-dir }}
      artifact-name: sboms
```

This uses the matrix with all artifacts and then it should pack them all into the same artifact.
After this, we need a another step to attach that `sboms` artifact to the release though.
You can essentially  do something like [this](https://github.com/keptn/keptn/blob/master/.github/workflows/release.yml#L290)

PR #9411, opened 2023-01-09
chore: fixed false positive security errors

odubajDT:

https://github.com/keptn/keptn/actions/runs/3881855545

### DIFF
--- a/.github/.kubescape/exceptions.json
+++ b/.github/.kubescape/exceptions.json
@@ -14,5 +14,25 @@
         }
       }
     ]
+  },
+  {
+    "name": "application-credentials-in-configuration-files",
+    "policyType": "postureExceptionPolicy",
+    "actions": [
+      "alertOnly"
+    ],
+    "resources": [
+      {
+        "designatorType": "Attributes",
+        "attributes": {
+          "kind": "Deployment"
+        }
+      }
+    ],
+    "posturePolicies": [
+        {
+            "controlID": "C-0012"
+        }
+    ]
   }
 ]
--- a/.github/.kubescape/exceptions.json
+++ b/.github/.kubescape/exceptions.json
@@ -34,5 +34,147 @@
             "controlID": "C-0012"
         }
     ]
+  },
+  {
+    "name": "images-allowed-registry",
+    "policyType": "postureExceptionPolicy",
+    "actions": [
+      "alertOnly"
+    ],
+    "resources": [
+      {
+        "designatorType": "Attributes",
+        "attributes": {
+          "kind": "Deployment"
+        }
+      }
+    ],
+    "posturePolicies": [
+        {
+            "controlID": "C-0078"
+        }
+    ]
+  },
+  {
+    "name": "auto-map-service-account-nats",
+    "policyType": "postureExceptionPolicy",
+    "actions": [
+      "alertOnly"
+    ],
+    "resources": [
+      {
+        "designatorType": "Attributes",
+        "attributes": {
+          "kind": "ServiceAccount",
+          "name": "keptn-nats"
+        }
+      }
+    ],
+    "posturePolicies": [
+        {
+            "controlID": "C-0034"
+        }
+    ]
+  },
+  {
+    "name": "auto-map-service-account-mongo",
+    "policyType": "postureExceptionPolicy",
+    "actions": [
+      "alertOnly"
+    ],
+    "resources": [
+      {
+        "designatorType": "Attributes",
+        "attributes": {
+          "kind": "Deployment",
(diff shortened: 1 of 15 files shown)


.github/.kubescape/exceptions.json
```diff
@@ -14,5 +14,167 @@
...
+      }
+    ],
+    "posturePolicies": [
+        {
+            "controlID": "C-0012"
+        }
+    ]
+  },
+  {
+    "name": "images-allowed-registry",
```
mowies, line 39: instead of disabling this rule, you could add the missing registry to the allowed list. see [link](https://hub.armosec.io/docs/c-0078)

.github/.kubescape/exceptions.json
```diff
@@ -14,5 +14,167 @@
...
+          "kind": "StatefulSet"
+        }
+      }
+    ],
+    "posturePolicies": [
+        {
+            "controlID": "C-0030"
+        }
+    ]
+  },
```
mowies, line 139: you can probably combine those 2 rules with 1 resource each into 1 rule with 2 resources

```suggestion
  {
    "name": "ingress-egress-blocked",
    "policyType": "postureExceptionPolicy",
    "actions": [
      "alertOnly"
    ],
    "resources": [
      {
        "designatorType": "Attributes",
        "attributes": {
          "kind": "Deployment"
        }
      },
      {
        "designatorType": "Attributes",
        "attributes": {
          "kind": "StatefulSet"
        }
      }
    ],
    "posturePolicies": [
        {
            "controlID": "C-0030"
        }
    ]
  },
```

.github/.kubescape/exceptions.json
```diff
@@ -14,5 +14,167 @@
...
+  {
+    "name": "default-namespace",
```
mowies, line 141: IMHO this is something that should actually be fixed in the helm chart instead of disabling the rule...

.github/.kubescape/exceptions.json
```diff
@@ -14,5 +14,167 @@
...
+      }
+    ],
+    "posturePolicies": [
+        {
+            "controlID": "C-0061"
+        }
+    ]
+  },
+  {
+    "name": "immutable-container-filesystem",
```
mowies, line 161: please disable this only for the specific resources where it came up. this will ensure that we remove false positives but actual positives in other resources will still be caught

.github/.kubescape/exceptions.json
```diff
@@ -14,5 +14,167 @@
...
+          "name": "keptn-mongo"
+        }
+      }
+    ],
+    "posturePolicies": [
+        {
+            "controlID": "C-0034"
+        }
+    ]
+  },
```
mowies, line 99: this can be combined into one:
```suggestion
  {
    "name": "auto-map-service-account-nats",
    "policyType": "postureExceptionPolicy",
    "actions": [
      "alertOnly"
    ],
    "resources": [
      {
        "designatorType": "Attributes",
        "attributes": {
          "kind": "ServiceAccount",
          "name": "keptn-nats"
        }
      },
      {
        "designatorType": "Attributes",
        "attributes": {
          "kind": "Deployment",
          "name": "keptn-mongo"
        }
      }
    ],
    "posturePolicies": [
        {
            "controlID": "C-0034"
        }
    ]
  },
```

.github/.kubescape/exceptions.json
```diff
@@ -14,5 +14,167 @@
         }
       }
     ]
+  },
+  {
+    "name": "application-credentials-in-configuration-files",
```
mowies, line 19: pls disable this rule only where it comes up now so that we can catch new actual positives in the future for other resources

installer/manifests/keptn/templates/_helpers.tpl
```diff
@@ -465,6 +465,7 @@ Usage:
 
 {{- define "keptn.initContainers.wait-for-nats" -}}
 - name: "wait-for-nats"
+  namespace: "{{ .Release.Namespace }}"
```
mowies, line 468: this is not needed here since the init container is filled directly into the pod.spec.initConainers field

installer/manifests/keptn/templates/_helpers.tpl
```diff
@@ -488,6 +489,7 @@ Usage:
 
 {{- define "keptn.initContainers.wait-for-keptn-mongo" -}}
 - name: "wait-for-keptn-mongo"
+  namespace: "{{ .Release.Namespace }}"
```
mowies, line 492: this is not needed here since the init container is filled directly into the pod.spec.initConainers field

mowies (approved):

PR #9420, opened 2023-01-10
ci: add concurrency group to release pipeline to avoid race conditions

afzal442:
Signed-off-by: afzal442 <afzal442@gmail.com>

## This PR

The release job is in the "release" group and will run on the latest version of Ubuntu, and the concurrency option is set to 1, so only 1 release job will run at a time, just to avoid race condition.


mowies (approved):

mowies:
Thanks for the contribution @afzal442 !

PR #9495, opened 2023-02-06
fix(lighthouse-service): inconsistent behavior when SLI provider returns warning

### LINKED ISSUE #9494: Lighthouse :: Inconsistent behavior
## Goal

The lighthouse should provide consistent handling of SLI events. 

## Technical Description

Lighthose awaits SLI providers for data used to make decisions on evaluations. The handling of the get SLI event needs to be more consistent. Here we list how Lighthouse shall handle the SLI provider's different answers.

We have three types of answers for the SLI provider: 

1. Provider returns `succeeded` -> Lighthouse evaluates the SLO definition based on the SLI values
2. Provider returns `warning`
3. Provider returns `failed` -> Lighthouse, in turn, fails

In the second case, Lighthouse has to look into each SLO definition and there could be two cases:
1. The SLO has **NO** pass/warning criteria: this is an informative SLO and Lighthouse will ignore this entry and will NOT contribute to the overall score, but the SLI value is reported. This way, it gets into the Heatmap in the Bridge.
2. The SLO has pass/warning criteria. The eval of this entry will contribute to the score:
	* if the SLI indicator has success=false the SLO entry is considered failed
	* otherwise, the SLO definition is normally evaluated

odubajDT:

## This PR

- fixes inconsistent behaviour when SLI provider returns warning

### Related Issues

Fixes #9494 

https://github.com/keptn/keptn/actions/runs/4121475886

### DIFF
--- a/lighthouse-service/event_handler/evaluate_sli_handler.go
+++ b/lighthouse-service/event_handler/evaluate_sli_handler.go
@@ -149,6 +149,13 @@ func (eh *EvaluateSLIHandler) processGetSliFinishedEvent(ctx context.Context, sh
 		return sendEvent(shkeptncontext, triggeredID, keptnv2.GetFinishedEventType(keptnv2.EvaluationTaskName), commitID, eh.KeptnHandler, &evalResult)
 	}
 
+	if e.Result == keptnv2.ResultFailed {
+		evalResult.EventData.Result = e.Result
+		evalResult.EventData.Status = keptnv2.StatusSucceeded
+		evalResult.Message = fmt.Sprintf("lighthouse failed because SLI failed with message %s", e.Message)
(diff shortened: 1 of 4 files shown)


lighthouse-service/event_handler/evaluate_sli_handler.go
```diff
@@ -42,6 +42,12 @@ type criteriaObject struct {
 	CheckIncrease   bool
 }
 
+type KeySLI struct {
```
bacherfl, line 45: do we need this struct somewhere else or would it make sense to make it unexported?

odubajDT: it can be unexported, good catch!

bacherfl (approved):
lgtm

PR #9497, opened 2023-02-06
fix(bridge): Fixed prefix path

### LINKED ISSUE #9496: Bridge: Prefix-path does not work with the bridge client
If keptn is installed with a custom prefixPath, the bridge client still uses `/bridge` as a base-href within the app instead of the right one.

Kirdock:
Fixes #9496


heinzburgstaller (approved):

thisthat (approved):
Thanks @Kirdock

PR #9535, opened 2023-02-27
chore: update go.mod and pipeline to use the same Go version of Dockerfile

thisthat:
## This PR

Align Go version to what is used in the Dockerfile

Integration test run: https://github.com/keptn/keptn/actions/runs/4283773729

### DIFF
--- a/webhook-service/lib/template_engine_test.go
+++ b/webhook-service/lib/template_engine_test.go
@@ -45,7 +45,7 @@ func TestTemplateEngine_ParseTemplate(t1 *testing.T) {
 			},
 			want:    "",
 			wantErr: true,
-			errMsg:  "unexpected",
+			errMsg:  "template: :1: bad character U+007D '}'",
 		},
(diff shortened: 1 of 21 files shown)


webhook-service/lib/template_engine_test.go
```diff
@@ -45,7 +45,7 @@ func TestTemplateEngine_ParseTemplate(t1 *testing.T) {
 			},
 			want:    "",
 			wantErr: true,
-			errMsg:  "unexpected",
+			errMsg:  "template: :1: bad character U+007D '}'",
```
mowies, line 48: why is this change included?

thisthat: It looks like that with go 1.19 the std library to handle templating has better error messages

mowies: ok cool

mowies (approved):

PR #9564, opened 2023-03-14
ci(bridge): bump nodejs version

thisthat:
## This PR

- bump nodejs version to use the latest LTS

### DIFF
--- a/bridge/Dockerfile
+++ b/bridge/Dockerfile
@@ -26,7 +26,7 @@ RUN yarn build && \
 
 # Use a Docker multi-stage build to create a lean production image.
 # https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds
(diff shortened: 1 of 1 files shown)


bridge/Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN yarn build && \
 
 # Use a Docker multi-stage build to create a lean production image.
 # https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds
-FROM node:14-alpine3.15 as production
+FROM node:lts-alpine3.17 as production
```
mowies, line 29: i would prefer to use a pinned node version here. this could have influence on our builds even if we don't change any code since the `lts` tag is a moving one

thisthat: fair point, also for reproducibility of our images

mowies (approved):

PR #9576, opened 2023-03-15
chore: move away from docker

mowies:
### This PR
- removes docker from most occurrences by replacing it with GHCR or removing it completely, depending on the circumstances
- removes the usage of the keptndev dockerhub registry and instead replaces it with ghcr.io/keptn

Related to keptn/community#250


thisthat (dismissed):
LGTM and thanks also for improving small things here and there.

gh-actions-scripts/cleanup_docker_images.sh
RealAnna, line 1: cleanup_images would be a better name for the script now

mowies: i reverted the script since it does actually only work with docker

gh-actions-scripts/cleanup_docker_images.sh
```diff
@@ -5,7 +5,7 @@
 # Required secrets/params:                                                                          #
 # - REGISTRY_USER: A dockerhub user.                                                                #
 # - REGISTRY_PASSWORD: The corresponding password. CAUTION: Personal access tokens don't work here! #
-# - DOCKER_ORG: The organization on dockerhub where the images are located                          #
```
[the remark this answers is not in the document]
mowies: i reverted the script since it does actually only work with docker.
maybe we can re-write the script to work with GHCR in the future

thisthat (approved):

PR #9792, opened 2023-10-19
fix: minishift instalation script

odubajDT:


odubajDT:
Keptn V1 has reached EOL on December 22nd, 2023, so this PR has been closed.

PR #9808, opened 2023-12-13
build(deps): update helm release nats to v1

bacherfl:
This PR is used to ensure everything works after upgrading the nats chart dependency


bacherfl:
Closing this as the 1.x version of the nats helm chart introduced a number of significant changes that would need to be adapted and carefully evaluated in our helm chart - see https://github.com/nats-io/k8s/tree/main/helm/charts/nats#upgrade-nodes

